PR #623, opened 2011-10-14
This allows ActiveAdmin::ResourceController::Sorting to sort by a column in another table

Zequez:
The previous behavior always appended the current table name at the beginning of the column.
With this patch it only appends the current table name if the column was given without any specific table.

So now when you can call something like this:

column :name, sortable: 'people.name'

It works as expected. Though when loading the model it has to be joined to the other table, of course.

### DIFF
--- a/lib/active_admin/resource_controller/collection.rb
+++ b/lib/active_admin/resource_controller/collection.rb
@@ -46,9 +46,13 @@ module ActiveAdmin
 
         def sort_order(chain)
           params[:order] ||= active_admin_config.sort_order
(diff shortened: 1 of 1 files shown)


AlexVentura:
@romeuhcf It worked for me, but can you explain or elaborate your solution a little bit more to clarify the advantages for using this instead of the most common `end_of_association_chain.includes(:related_model)`?

backwardm:
Thank you @rabidpraxis for your note! That helped me be able to sort a column while in different scopes.

PR #866, opened 2011-12-21
Add config option to render flash messages as raw -- this enables you to include a link in the flash message, such as undo and redo actions

jmonster:


luctus:
It's not working for me. (Rails 4.2.1)

PR #1439, opened 2012-06-28
Allow CSS class for rows in attributes_table

### LINKED ISSUE #1381: how do I add a new css class to the row?
``` ruby
ActiveAdmin.register Ad do
  show do |ad|
    attributes_table do
      row :title
      row :image do
        image_tag(ad.image.url)
      end
    end
    active_admin_comments
  end
end
```

how do I add a new css class to the row?

seanlinsley:
Code to fix #1381

This way you can include a custom CSS class to identify rows. The default has been set to 'row'.

``` ruby
attributes_table do
  row("Total", :class => :total) {  }
end
```

I'd add unit tests to `attributes_table_spec.rb`, but I'm not entirely sure what all is happening there.

### DIFF
--- a/lib/active_admin/views/components/attributes_table.rb
+++ b/lib/active_admin/views/components/attributes_table.rb
@@ -17,13 +17,16 @@ module ActiveAdmin
         attrs.each {|attr| row(attr) }
       end
 
-      def row(attr, &block)
-        @table << tr do
+      def row(*args, &block)
+        title   = args[0]
+        options = args.extract_options!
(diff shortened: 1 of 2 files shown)


IonutPerfectWeb:
@abhimediratta I know it's been 2 years, but I am posting this here if it helps anyone else :) 

you can add class to index by doing:

```
 index do
    column :value, class: 'your-class'
end
```

johnflux:
Can you add a class to just the td inside the row?

Issue #909, opened 2012-01-16
Let pass multiple columns to sortable


behrangsa:
Hi,

I have a column that displays values of a virtual attribute `full_name` that is composed of multiple fields `first_name` and `last_name`. Right now there's no way to sort ascending or descending using both columns. If I specify:

``` ruby
:sortable => "timesheet_user.first_name, timesheet_user.last_name"
```

The generated link for sorting would become:

```
order=user.first_name%2C+user.last_name_desc
```

But it should become something like

```
order=user.first_name_desc%2C+user.last_name_desc
```

It would be nice if we can pass multiple columns to sortable like this:

``` ruby
:sortable => ["user.first_name", "user.last_name"]
```

In order to be able to sort based on multiple columns.

PR #1766, opened 2012-11-13
#909 - multiple columns to sortable

awinabi:
Adding in this feature for #909. 
This pull request will let pass multiple columns to sortable in the following formats

```
:sortable => [:first_name, :last_name]
:sortable => ["user.first_name", "user.last_name"]
:sortable => "first_name, last_name"
```

Added in a couple of specs as well


brightbytes-dude:
LOL - ended up back here 8 years later randomly enough; the solution is to simply override `def apply_sorting(chain)`.

PR #1815, opened 2013-02-08
Batch Action forms

seanlinsley:
FYI: this is an Issue that was [converted into a Pull Request](http://opensoul.org/blog/archives/2012/11/09/convert-a-github-issue-into-a-pull-request/)

I want to add optional forms to Batch Actions; an example use-case being that you want the user to add in a text box the reason why they're taking this action. Since we're already including jQuery UI and it has [a similar UI for pop-over forms](http://jqueryui.com/dialog/#modal-form), a bit of the work has already been done.

~~What I'm curious about is what you think a good DSL syntax would be for this--should I include a `form do |f|` block **inside** a `batch_action` declaration, or should they be defined much like the new/edit pages are, with something like the below syntax?~~

``` ruby
batch_form :for => :an_action do |f|
  #...
end
```
#### UPDATES
- The syntax is now:

``` ruby
    batch_action :foo, confirm: 'You sure?', form: {bar: :text} do |ids, inputs|
      inputs[:bar] # access user input from this hash
      # do your stuff!
    end
```
- I changed the behaviour of the `:confirm` option. Passing a string still works, but now if you pass `true`, it will use a default confirm message.
- The UI now looks like this:
  Modal styling

### DIFF
--- a/config/locales/bg.yml
+++ b/config/locales/bg.yml
@@ -53,7 +53,8 @@ bg:
       link: "Създаване"
     batch_actions:
       button_label: "Масови действия"
-      delete_confirmation: "Сигурни ли сте, че искате да изтриете тези %{plural_model}? Това действие не може да бъде върнато назад."
+      default_confirmation: "Наистина ли искате да направите това?"
+      delete_confirmation: "Сигурни ли сте, че искате да изтриете тези %{plural_model}?"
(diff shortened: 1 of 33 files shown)


modsaid:
This optional :form attribute would be very useful in page_action or even action_item too

Issue #1930, opened 2013-02-20
How to create customized redirect_to after creation of new resource


yangchenyun:
Saying, after submitting the form post, instead of being redirected to the show page I want to go to the index page, how could I archieve that?

PR #2001, opened 2013-03-18
[1930] Allow IR shorthand for redirection

seanlinsley:
For #1930; supersedes #1944

This removes the Inherited Resources overrides we had. Now the shorthand syntax works:

``` ruby
controller do
  # Redirects to index page instead of rendering updated resource
  def update
    update!{ collection_path }
  end
end
```


dchersey:
Just saw this ... 7 years later.  Solved a nagging and persistent problem with all of my AA sites.

It would be AWESOME to see this in the documentation on http://activeadmin.info !

deivid-rodriguez:
We're very happy to take a PR documenting this :)

PR #2021, opened 2013-03-26
remove unused Batch Actions code

seanlinsley:
This removes a set of Batch Actions code that isn't actually being used. If someone besides myself could download this branch and test it out, I'd appreciate it.

This spawned from #2016, where I noticed that there was duplicate Ruby and JS code just laying about.

/cc @macfanatic


javierjulio:
@timoschilling looks like this wasn't ever merged in but this should verify that the popover related code can be safely removed as @seanlinsley had done it but stopped due to the `switch-to-bootstrap` effort which now has seemed to have died out. If you're good with me removing the popover code I'll do it in my CSS WIP branch unless you prefer I do it separately (although if it was deemed unused so long ago can't see how it would be an issue to just remove now from my branch).

timoschilling:
@javierjulio remove it, but pleas do it in a separately PR

timoschilling:
@javierjulio pleas ref this PR in your one

Issue #301, opened 2011-07-18
Don't register ActiveAdmin::Comment as Comments


gregbell:
It can be confusing to not be able to register your own model named Comment because of the default comments that come with Active Admin. We should register ActiveAdmin::Comments as something other than Comments.

abrambailey:
I know this is very old, but apparently I missed the ability to change the registered names for comments using [config.comments_registration_name = 'AdminComment'](https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/docs/1-general-configuration.md#comments) (posting this comment in case someone else is digging around looking for the solution).

edwincarbajal:
@abrambailey +1 thank you! Saved me a lot of time. Cheers!

shootingfly:
How stupid it is!
Why not make `config.comments_registration_name = 'AdminComment' ` as default?

sarmaddevc:
> config.comments_registration_name = 'AdminComment'
in which file i past it

zulays:
> 
 you don't put it anywhere, you uncomment it in the config folder. It's in this [config repo ](https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/docs/1-general-configuration.md#comments)

PR #2113, opened 2013-04-16
Refactor Comments UI

seanlinsley:
While working on #1979, I noticed that we happen to register an index page for Comments, but it's hidden by default. This PR updates that UI so it's worthy of being shown by default.
##### Changes
- new configuration options:
  - menu item visibility (defaults to true)
  - `:as` registration name. For #301; replaces #2060 (defaults to Comment)
- index page:
  - added a scope for every AA namespace, with the current namespace as the default
  - updated `resource_type` and `author_type` filters to be select fields (for Rails >= 3.2)
  - updated table to show every attribute
- show page:
  - instead of redirecting to the associated resource, we now have a proper show page

### DIFF
--- a/features/comments/commenting.feature
+++ b/features/comments/commenting.feature
@@ -34,7 +34,7 @@ Feature: Commenting
         config.comments = false
       end
     """
-    Then I should not see "Comments"
+    Then I should not see the element "div.comments.panel"
 
   Scenario: View a resource in a namespace that doesn't have comments
     Given a configuration of:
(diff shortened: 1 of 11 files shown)


abrambailey:
I know this is very old, and by the way hello @coreyward, but apparently I missed the ability to change the registered names for comments using [config.comments_registration_name = 'AdminComment'](https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/docs/1-general-configuration.md#comments)

PR #2163, opened 2013-05-03
Advanced routing constraints

macfanatic:
Advanced routing configuration per namespace, or globally, with a new setting.  Inspired when loading engines into my parent application with ActiveAdmin, client wanted each engine to be available at it's own URL.

The old URL "http://www.mygreatapp.com/some_engine" is now served up at "http://engine.mygreatapp.com/" with the following config change:

``` ruby
ActiveAdmin.setup do |config|
  config.namespace :some_engine do |e|
    e.routing_constraint = { subdomain: "engine" }
  end
end
```

The new setting with accept a Proc, Hash, Class or Class instance - any object that responds to `#matches?`.

### DIFF
--- a/docs/1-general-configuration.md
+++ b/docs/1-general-configuration.md
@@ -78,6 +78,47 @@ individually.
 Each setting available in the Active Admin setup block is configurable on a per
 namespace basis.
 
+### Namespace Routing Constraints
+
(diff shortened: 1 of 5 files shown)


javierjulio:
Closing out since its been stale for a long time now.

PR #2421, opened 2013-08-18
File import

seanlinsley:
I've noticed that there are a lot of gems that provide import functionality to Active Admin ([[1]](https://github.com/krhorst/active_admin_importable), [[2]](https://github.com/Fivell/active_admin_import), and the _days old_ [[3]](https://github.com/Papercloud/active_admin_csv_import)), but no one has ever tried adding the feature directly to AA. Maybe that's because people think that it'll never get merged, or they don't feel comfortable with AA internals. At any rate, there's no need for such a core feature to be isolated into yet another gem. This PR is my attempt to get file import into AA proper.

The goal is to support every filetype that we currently allow for export (CSV, JSON, XML), plus any others that don't require too much added complexity.

I also want it to be built in such a way that it can fit every scenario. So if you want to disable validations in order to do a quick-and-dirty import, it'll work, and if you want to add an extra filetype or use a specific library like [activerecord-import](https://github.com/zdennis/activerecord-import), AA will seamlessly see that you have the gem installed and so will use that when saving.

The actual implementation so far was based on one of the options given in the related [railscast](http://railscasts.com/episodes/396-importing-csv-and-excel).

For added context: http://stackoverflow.com/questions/7754896

The features so far:
- you get an "Import" action_item button on the index page
- when visiting the import page, you get a list of all attributes that can be set, plus their data type
- if you import an existing record and include the ID, this will update the existing one instead of creating a duplicate
- when validations fail, you get a descriptive error and the row it came from

Concerns:
- how/when to allow for external gem behavior for saving, validations, and filetype support
- how to update this for Rails 4
- how to properly translate the page, plus the data type definitions
- it might be a better idea to create only one "import" page, instead of creating one for every model

To try and foster a conversation: CC @tspacek, @Fivell, @krhorst, @scottshea

Here's a snapshot:
screen shot 2013-08-18 at 5 30 40 pm

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,8 @@ else
   raise "Rails #{rails_version} is not supported yet"
 end
 
+gem 'roo'
+
 group :development do
   gem 'better_errors',     '~> 0.9.0' # Web UI to debug exceptions. Go to /__better_errors to access the latest one
   gem 'binding_of_caller', '~> 0.7.1' # Retrieve the binding of a method's caller in MRI Ruby >= 1.9.2
--- a/app/views/active_admin/resource/import.html.arb
+++ b/app/views/active_admin/resource/import.html.arb
@@ -1,0 +1,1 @@
+insert_tag renderer_for(:import)
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -74,6 +74,8 @@ en:
       title_content: "Comments (%{count})"
       errors:
         empty_text: "Comment wasn't saved, text was empty."
+    import:
+      title: "Import"
     devise:
       login:
         title: "Login"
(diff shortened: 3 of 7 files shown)


seanlinsley:
For reference, this is how GitHub deals with CSV errors on Gists:

screen shot 2016-06-17 at 10 12 11 pm


From: https://gist.github.com/strand/12330cdecec469bb678c7e4b57b76a7d

Fivell:
@pcreux , I noticed you is good citizen in AA community and there is   a gem https://github.com/pcreux/csv-importer , did you ever try to use it with AA ?

varyonic:
FWIW I think large file import is non-trivial, and I am happy to see it remain in separate plugin(s), especially if it adds more gem dependencies.

Fivell:
@varyonic , I don't see big benefits from implementing this in this, think we might close this discussion/issue.

timoschilling:
I agree with @varyonic and @Fivell

PR #2432, opened 2013-08-21
making sort_order accept multiple fields

dedsm:
now config.sort_order will accept sorting by multiple fields,
the format is "field1_(asc|desc), field2_(asc|desc)" preserved backwards
compatibility

Issue https://github.com/gregbell/active_admin/issues/659

### DIFF
--- a/lib/active_admin/resource_controller/data_access.rb
+++ b/lib/active_admin/resource_controller/data_access.rb
@@ -217,17 +217,24 @@ module ActiveAdmin
 
       def apply_sorting(chain)
         params[:order] ||= active_admin_config.sort_order
-        if params[:order] && params[:order] =~ /^([\w\_\.]+)_(desc|asc)$/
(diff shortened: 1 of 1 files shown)


RobinDaugherty:
This does not seem to be working in 1.0.0+. To get around this, you can use:
```rb
  config.sort_order = 'customer_name_asc'

  order_by(:customer_name) do |order_clause|
    %w[sites.customer_name sites.customer_store_number sites.customer_site_name].map { |column|
      "#{column} #{order_clause.order}"
    }.join(", ")
  end
```

This is somewhat documented in [the table index documentation](https://activeadmin.info/3-index-pages/index-as-table.html).

sunloverz:
@RobinDaugherty does it work in rails 4?

RobinDaugherty:
@sunloverz not sure, I believe this was with Rails 5.1.

PR #2546, opened 2013-10-08
Belongs to patch

paultyng:
Per @edestecd, see gregbell/active_admin#2082

This is the patch as suggested in the issue.  The tests seem to still be passing locally.

I am attempting to write some tests, not sure what model would make the most sense to add to the test site under posts.  I was originally thinking comments but that gets confusing with the built in active admin comments.  Any other suggestions?


seanlinsley:
It's been nearly two years since this PR has seen activity, so I'm going to close it.

PR #1994, opened 2013-03-15
Multiple sortable columns and sorting on associated resources

vjt:
Hello,

this pull is a superset of #1766, that uses `meta_search`'s [`meta_sort`](https://github.com/ernie/meta_search/#sorting-columns) to implement sorting on multiple columns (#909), and also sorting on associated models' columns.

The only thing I don't like about this implementation discriminates between [virtual and concrete columns](https://github.com/ifad/active_admin/commit/a7bdb84b38c326b901c447f68e5cce388d298836#L0R221), by looking at the `.select_values`.

Anyway, specs & features pass (this is my first experience with cucumber - sorry if they look bad), and there's also an unrelated improvement to the tests set up, in which instead of using Rails 3.2.0 by default, the latest version is [fetched from the `.travis.yml`](https://github.com/ifad/active_admin/commit/b8978ecbfdb970bda3db9c3b2ed3bb960d96e7d1) configuration file.

Enjoy!

cc: @npj - we'll soon be able to remove that monkeypatch :beers:
cc: @Daxter - here we go! :dancers: 
cc: @zzeni - I've taken your suggestion into account, thanks! :+1:

PR #2608, opened 2013-10-23
use Ransack to sort collections

seanlinsley:
This is a continuation of where @vjt left off in #1994

Added functionality:
- you can sort by multiple columns

Problems:
- sorting on a virtual column doesn't work
- when sorting on multiple columns, how do you tell which column (on an
  index table) is currently being sorted?

### DIFF
--- a/docs/3-index-pages.md
+++ b/docs/3-index-pages.md
@@ -59,7 +59,7 @@ index as: ActiveAdmin::Views::IndexAsTable do
 end
 ```
 
-## Index Filters
+## Filters
 
 By default the index screen includes a "Filters" sidebar on the right hand side
 with a filter for each attribute of the registered model. You can customize the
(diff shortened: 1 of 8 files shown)


hirokihokari:
@seanlinsley Seems like this got all removed on master now?

senid231:
@seanlinsley what if we just extend AA with some sort of custom sorting option
something like:
```ruby
ActiveAdmin.register Post do
  custom_sort full_name: ->(records, direction) { records.order([{last_name: direction, first_name: :asc}]) },
                       comments_count: :order_by_comments_count 
                       # shortcut for ->(records, direction) { records.order_by_comments_count(direction) }

  index do
    column :id # sorted by posts.id asc/desc, query param id_asc/id_desc
    column :full_name # sorted by lambda provided above, query param full_name_asc/full_name_desc
    column :reversed_name, sortable: :full_name do |r| # sorted same as :full_name
      r.full_name.reverse
    end
    column :comments, :comments_tag, sortable: :comments_count 
    # sorted by `order_by_comments_count` scope or class method, 
    # query param comments_count_asc/comments_count_desc
  end
end
```

Issue #2621, opened 2013-10-26
The icon API doesn't work as advertised


seanlinsley:
Our [built-in icons](http://activeadmin-icons.herokuapp.com/admin) are SVG, so you should be able to [resize them](https://github.com/gregbell/active_admin/blob/0f663e3ad728e1791be138d42e35f9f52c93d93b/lib/active_admin/iconic.rb#L32-L35). Unfortunately, it doesn't work:

``` ruby
icon :calendar, height: '50px', width: '50px'
```

Issue #880, opened 2011-12-23
Let's add a sortable option for the has_many formbuilder


michaek:
This would add a jQuery UI sortable to the items in a has_many, with the :sortable parameter accepting the field the sortable should update. That field would need to be included in the fields for the associated model - for example:

```
f.has_many :images, :sortable => :sort do |i|
  i.input :image
  i.input :sort, :as => :hidden
end
```

The feature will only update the value of the hidden field, which would only be persisted to the database upon submitting the form.

PR #2656, opened 2013-11-04
sortable has_many

igorbernstein:
This is a revival of #2614, #881, #2117, #880
Its a still a work in progress, but I wanted to post this early on to make sure I'm on the right path and this doesn't slip through the cracks.

TODO:
- ~~force initial ordering~~
- ~~add an indicator icon to show that the fieldsets can be dragged~~
  - http://activeadmin-icons.herokuapp.com/admin
  - http://rdoc.info/github/gregbell/active_admin/master/ActiveAdmin/Iconic
  - https://groups.google.com/forum/#!topic/activeadmin/SpdnDLf4i5I
  - #2621
- ~~add css to change the cursor to indicate that the fieldset is draggable~~
- add tests
  - http://www.dixis.com/?p=626
  - http://forum.jquery.com/topic/fire-events-to-simulate-drag-n-drop
- ~~add docs~~
- ~~deal with nested sortable has_many fieldsets~~

### DIFF
--- a/app/assets/javascripts/active_admin/base.js.coffee
+++ b/app/assets/javascripts/active_admin/base.js.coffee
@@ -1,6 +1,7 @@
 #= require jquery
 #= require jquery.ui.widget
 #= require jquery.ui.datepicker
+#= require jquery.ui.sortable
 #= require jquery_ujs
 #
 #= require_self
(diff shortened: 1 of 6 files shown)


Babbz75:
@kirel Thank you! 

For acts_as_list people wondering why the list is rotating you need to start the sort at 1. i.e. sortable_start: 1 otherwise it bumps the zero to one and then with two positions at 1 it will bump the first one to the bottom. At least I believe that's what was happening to me

PR #2674, opened 2013-11-08
rename popover to aaPopover to avoid naming conflict

alexagranov:
saw that aaDropdownMenu had been "namespaced" to AA so thought I'd change popover as well as its causing me problems with Foundation/JQuery


javierjulio:
Can we remove Popover? It seems to have been replaced by Dropdown Menu.

PR #2770, opened 2013-12-07
Add the belongs_to functionality to Page resources

zorab47:
Allow Page resources to use the same belongs_to functionality as Resource.

``` ruby
ActiveAdmin.register_page "Post Report" do
  belongs_to :post

  content do
    # ...
  end
end
```

### DIFF
--- a/features/registering_pages.feature
+++ b/features/registering_pages.feature
@@ -146,3 +146,19 @@ Feature: Registering Pages
     And I follow "Check"
     Then I should see the content "Chocolate lØves You Too!"
     And I should see the Active Admin layout
+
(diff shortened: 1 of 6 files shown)


Fivell:
@zorab47 , this is very nice feature,  maybe you could remember why there is a conflict with arbe ```parent``` method when using page, and no conflict when using resource?

zorab47:
It has to do with the Arbre context and variable resolution. My thought is
Arbre is intercepting the call for 'parent' though its use of
method_missing and returns the parent HTML node.

Fivell:
@zorab47 , let's give this second chance in #4759

PR #2771, opened 2013-12-07
Change CSVBuilder to get column name from I18n

mateusg:
These changes will make the `CSVBuilder` try to get the column name through `Model.human_attribute_name`.
- If the attribute name is not defined on the locale, the name returned will no longer be "titleized", it will be "humanized".
- If the value specified to the `column` method is not a symbol, the builder will print it humanized.
- Very useful when using locales. :)

Instead of doing this:

``` ruby
csv do
  column("Nome") { |r| r.name }
  column("Idade") { |r| r.age }
  column("Altura") { |r| r.height }
end
```

I'll do this:

``` ruby
csv do
  column(:name)
  column(:age)
  column(:height)
end
```

And have my translations defined on the locale file:

``` yaml
pt-BR:
  activerecord:
    attributes:
      person:
        name: Nome
        age: Idade
        height: Altura
```

### DIFF
--- a/spec/unit/csv_builder_spec.rb
+++ b/spec/unit/csv_builder_spec.rb
@@ -5,16 +5,16 @@ describe ActiveAdmin::CSVBuilder do
   describe '.default_for_resource using Post' do
     let(:csv_builder) { ActiveAdmin::CSVBuilder.default_for_resource(Post) }
 
-    it "should return a default csv_builder for Post" do
+    it 'returns a default csv_builder for Post' do
       expect(csv_builder).to be_a(ActiveAdmin::CSVBuilder)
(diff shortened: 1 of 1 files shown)


Onumis:
I know this is an old one, but is there anyway to skip `humanize`?

I need a column like this: `column 'Input Power (W)' do ... end`

And this is being converted into: `Input power (w)`, which is bad because Wattage should be expressed with capital 'W'.

Onumis:
Ah! It's right there on the comments: https://github.com/activeadmin/activeadmin/blob/v1.2.1/lib/active_admin/csv_builder.rb#L9

:D

PR #2814, opened 2013-12-23
Resource action_item makeup fix for dropdown_menu

### LINKED ISSUE #2813: Resource action_item makeup fix for dropdown_menu
Here is what was:

And here is what should be:

I'll send pull request in a while

sandrew:
fixes #2813


adisos:
@sandrew 
Could you please explain, how to use `action_item` with dropdown feature?

PR #2827, opened 2013-12-25
Make has_many nested records' Remove button uniform

davidstosik:
Hello,
At the moment, has_many nested record form offer two ways to remove a record, depending on wether it's new (Remove button that also removes the record form from the page) or existing (simple destroy checkbox).
I propose in this pull request to render this remove feature uniform, by replacing the destroy checkbox by a Remove button similar to the one used on new records.
The destroy checkbox becomes a hidden input, which is set to true when the user clicks on Remove. The record's form is also hidden in the process.
Sortable still works.
I also updated rspec tests.

Thanks in advance for your review,

David


sonar0007:
I know that a lot of time passed but is that going to be reviewed/merged any time?

seanlinsley:
@sonar0007 if you would like to branch off of this PR to continue the work, I'd be happy to continue reviewing it.

timoschilling:
Maybe some of this is already happen in #3907

sonar0007:
I will take a look at that

timoschilling:
This PR should wait until #3862 is merged

Fivell:
@javierjulio ,did you see this?

javierjulio:
@Fivell yes but I don't think we should introduce this. I'd like to update all the JS for v2 and this introduces a lot more. To be honest I'm not sure I'm a fan of the button approach. I think the checkbox for marking something deleted is fine. Has many forms are just.. blah. 🤢 I think its best to have a good core with our CSS and JS before retackling this. Its not important since you are able to delete has many records currently. I'll close out because its stale and with many conflicts. We can definitely consider an updated approach in a new PR!

PR #2857, opened 2014-01-12
Pundit authorization adapter

romansklenar:
I've found that [Pundit gem](https://github.com/elabs/pundit) is nice alternative to CanCan and it would be nice to support it out of the box. Thanks to authorization adapter implementation shouldn't be so hard. Some resources about Pundit:
- [Simple authorization in Ruby on Rails apps](http://www.elabs.se/blog/52-simple-authorization-in-ruby-on-rails-apps)
- [Keeping it Simple: Migrating to Pundit from CanCan](http://blog.carbonfive.com/2013/10/21/migrating-to-pundit-from-cancan/)

I'd like to contribute to AA and give back something to comunity but I'll need your feedback and code review. I've made basic tests implementation same as CanCan adapter tests but there are 2 things:
- Any tips how can I improve specs and specially mocking of testing classes?
- I'm unable now to get `authorization_pundit.feature` test passing since I'm not experienced enough with cucumber ([info about failing feature tests](https://gist.github.com/romansklenar/cee8a83afef0d8942b1c)).


jessesanford:
I get a Pundit::AuthorizationNotPerformedError in Admin::DashboardController#index error when accessing the dashboard. Any ideas?

jessesanford:
To anyone else who gets here via google and has trouble with AuthorizationNotPerformedErrors I posted an explination and solution over here on SO: http://stackoverflow.com/questions/34664645/active-admin-devise-and-pundit-punditpolicyscopingnotperformederror

PR #2878, opened 2014-01-19
I added function of soft-delete.

alfa-jpn:
I added function of soft-delete.
So, please consider merging from my repository if possible.

This code support general gems of soft-delete.
- paranoia
- acts_as_paranoid

example

``` ruby
ActiveAdmin.register Nanika do
  actions :all
  soft_delete_actions
end
```

nanika

added action and batch of soft-delete and restore and hard-delete.

problem
- I created config/locales/en.yml and ja.yml, because I can't translate locale file without them.

I'm poor at English.
If you can't understand what I mean, please let me know.
Sorry for any inconvinience I may cause you.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -8,6 +8,10 @@ en:
     edit: "Edit"
     delete: "Delete"
     delete_confirmation: "Are you sure you want to delete this?"
+    soft_delete: "SoftDelete"
+    soft_delete_confirmation: "Are you sure you want to soft-delete this?"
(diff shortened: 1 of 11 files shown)


Fivell:
@timoschilling , maybe this is candidate for 3d party gem ?

varyonic:
I see [active_admin_paranoia](https://github.com/raihan2006i/active_admin_paranoia) was written and released during 2015.

timoschilling:
closed while released as plugin gem

PR #2891, opened 2014-01-23
Allow index page_title to be set in controller

quainjn:
Can set the index page title with the @page_title instance variable in the controller, so you can avoid duplication if you want the page_titles to be the same for that action and other actions. 

Example:

```
ActiveAdmin.register Post do
  controller do
    before_filter { @page_title = "#{current_user} Posts" }
  end
end
```


seanfcarroll:
This worked for me in Rails 5.2

```ruby
 before_action { @page_title = AmldConference.find(params[:id]).try(:conference) 
```

PR #2906, opened 2014-01-28
Move controllers and concerns under app/controllers and app/concerns

### LINKED ISSUE #697: ApplicationController and Helpers are not being reloaded in Dev Env
why is the application_controller cached?
and how can i disable this?

cheers

a-chernykh:
In order to get files automatically reloaded in Rails, these files should be under `ActiveSupport::Dependencies.autoload_paths` directory. `Rails::Engine` adds `app/controllers` path automatically to `autoload_paths`.

Reference: https://github.com/rails/rails/issues/12195#issuecomment-24432724

Closes #697


javierjulio:
As maintainer I would be open to changes like these to improve the development experience. I’ve had much success with extracting view partials for a v4 release of ActiveAdmin (using TailwindCSS) that now reload on local development of the library. I’d love to have this for the host app. If someone wants to take this on that would be most helpful.

PR #3038, opened 2014-04-02
Implements streaming CSVs.

craigmcnamara:
Allows unlimited downloads of CSV data using ActiveRecord batch APIs and ActionController response streaming.

Works on Rails 3.2+

### DIFF
--- a/lib/active_admin/resource_controller/streaming.rb
+++ b/lib/active_admin/resource_controller/streaming.rb
@@ -31,6 +31,7 @@ module ActiveAdmin
           options = default.merge active_admin_config.csv_builder.options
           columns = active_admin_config.csv_builder.render_columns(self)
 
+          headers['X-Accel-Buffering'] = 'no'
           headers['Cache-Control'] = 'no-cache'
 
           self.response_body = Enumerator.new do |csv|


jhblacklock:
Loving this commit!  It works great locally; cool idea.  Starts streaming the file instead of serving the csv from the web server.

For production on heroku I get H12 errors trying to download around 10K records with children.

Are there any reasons this would not work on Heroku currently?  I'll be happy to do any research and provide any other info about my implementation. 

activeadmin (1.0.0.pre)
Rails 4.1.0

joallard:
@jhblacklock Sounds like the timeout value in Unicorn. That, or stop having children download your CSVs and have it done by an adult!

jhblacklock:
@joallard That is definitely a problem of mine (children) but they don't seem to be the issue I'm having with streaming CSVs ^_^.

My unicorn timeout value is hardcoded at 600.  I also run unicorn local and do not experience any issues.

The initial web request hits in 10ms but the server doesn't initiate the download before the server times out in 30 seconds.

```
2015-08-05T22:07:00.792698+00:00 app[web.1]: Started GET "/admin/newcomers.csv" for 97.77.148.58 at 2015-08-05 22:07:00 +0000
2015-08-05T22:07:00.852062+00:00 app[web.1]: Completed 200 OK in 10ms (ActiveRecord: 4.1ms)
2015-08-05T22:07:00.837900+00:00 app[web.1]: Processing by Admin::NewcomersController#index as CSV
2015-08-05T22:07:30.774222+00:00 heroku[router]: at=error code=H12 desc="Request timeout" method=GET path="/admin/newcomers.csv" host=admin-ttr-challenge.herokuapp.com request_id=cf4050f3-f162-4717-b64a-205f0be0fb07 fwd="97.77.148.58" dyno=web.1 connect=1ms service=30000ms status=503 bytes=0

```

My csv block is lengthy but I wouldn't think the timeout would occur on the first iteration.

```
csv do
    [:first_name, :last_name, :num_of_children, :email].each do |col|
      column col
    end
    column(:event_name) { |newcomer| newcomer.event.name if newcomer.event }
    column(:event_date) { |newcomer| newcomer.event.starts_time if newcomer.event }
    column(:business_name) { |newcomer| newcomer.event.user.business_name if newcomer.event && newcomer.event.user }
    column(:host_first_name) { |newcomer| newcomer.event.user.first_name if newcomer.event && newcomer.event.user }
    column(:host_last_name) { |newcomer| newcomer.event.user.last_name if newcomer.event && newcomer.event.user }
    column(:newcomer_address_name) { |newcomer| newcomer.address.name if newcomer.address }
    column(:newcomer_address_one) { |newcomer| newcomer.address.address_one if newcomer.address }
    column(:newcomer_address_two) { |newcomer| newcomer.address.address_two if newcomer.address }
    column(:newcomer_city) { |newcomer| newcomer.address.city if newcomer.address }
    column(:newcomer_state) { |newcomer| newcomer.address.state if newcomer.address }
    column(:newcomer_zip) { |newcomer| newcomer.address.zip if newcomer.address }
    column(:children) { |newcomer| newcomer.children.pluck(:name, :gender, :age).reject(&:blank?).join(", ") if newcomer.children }
  end
```

jhblacklock:
Yar. I feel sheepish. :blush:

My Procfile was not registering with the heroku deploy so it was defaulting to webrick.  One process = not working.

Thanks for everything!

joallard:
Methinks we should document this issue seeing how frequent it is

seanlinsley:
It's also important to mention that if an exception occurs during the stream, the stack trace will be appended to the resulting file, but the request will only end after your web server times out. That's not necessarily a problem in production (and it might be very hard to fix), but in development it can make debugging much more difficult. Which is why I created #3535

BTW @jhblacklock you can use syntax highlighting to make your code easier to read: https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown

seanlinsley:
Actually, it is a problem in production, because the exception doesn't bubble up. So the only way you'll find out is by the users of your app reporting that their CSV exports don't include all the records they should.

Fivell:
@seanlinsley , I faced with this problem, have no logs to find out why csv was not downloaded.

haraldkrischner:
This PR might be old, but we are running into the same issue as @seanlinsley , so is there any approch to it?

luiszacheu:
I downloaded active admin 1.0.0.pre4, but would like to know how to implement the export CSV streaming? Tks

DaniilVas1:
@seanlinsley  Thanks for your implementing. but I have some questions. 
I had built the Activeadmin with rails and I have 20k records. I hope to download this with CSV. But this happens the timeout error. I deployed this on Heroku. How can I fix this?
Please help me.
Thanks.

PR #3053, opened 2014-04-08
split Resource::ActionItems#add_default_action_items into separate methods

timoschilling:


timoschilling:
I reopen this while it's a solution for @scarver2 problem in https://github.com/activeadmin/activeadmin/pull/3463#issuecomment-87300983.
> Is there a way to sort the actions when combining with custom actions?

The solution could look like:
```ruby
actions defaults: false do
  item :custom_action_a
  default_new_action_item
  item :custom_action_b
end
```
`default_new_action_item` is currently named `add_default_new_action_item` in this PR, I will rename them before merge.

achiinto:
any news with this?

scarver2:
@timoschilling I like where this development is going. Givig the developer control to include all the default actions or cherry-picking specific actions is quite beneficial. I like that the default actions continue to respect the authorization and inclusion settings dynamically.

Will the new syntax in IndexAsTable be this:

To insert all default action items between two custom action items:
```ruby
actions defaults: false do
  item :custom_action_a
  default_action_items
  item :custom_action_b
end
```

To insert individual default action items intermixed with custom action items:
```ruby
actions defaults: false do
  item : custom_action_a
  default_new_action_item
  item :custom_action_b
  default_edit_action_item
  item :custom_action_b
  default_detete_action_item
end
```

1. I would prefer the "default_" prefix was not needed. Is that possible?
2. Any concern that the actions are not alphabetically sorted similar to the other ActiveAdmin DSLs like menus? If we follow those conventions, the organization of the action items would default to alphabetical and then the priority attribute would be used to control placement. I have no personal preference so long as performance is not affected.

PR #3159, opened 2014-05-22
2514 generator should detect require tree

scicco:
Hi,

here you have my PR.

Note: Still missing the proper spec for all the options under install_spec.rb

Any advice how to simulate an user choose of an option? How avoid generating three different rails installation for each case?


javierjulio:
Closing out as its gone stale. I'm afraid we are overthinking this. At most what we should do is show a warning on install if we detect that require_tree is in either file that it has to be removed for the admin styles to not be included in other parts of their app.

Issue #1027, opened 2012-02-16
Twitter bootstrap usage instead


akashkamboj:
It would be so great if you use Twitter Bootstrap instead. There are a few issues that are coming when working with IE i.e. IE8 or below doesn't show Topbar Dropdown Menu. And if you use Twitter bootstrap instead these problems will be gone as well as we will get some extra features to customize the look and feel using the javascripts that bootstrap offers.

I think I can give you extra hands if you want to achieve this.

UPDATE: I want to start on this. Can you suggest shall i create styles according to current Arbre layouts? or do I modify the Arbre layouts according to Bootstrap?

PR #3163, opened 2014-05-24
Responsive design

ball-hayden:
This is by no means a perfect and complete solution, but more a step towards responsive design for ActiveAdmin. There are probably quite a few things that I have missed.

There has been some discussion about this in the past (mostly #1027 - which seems to me like overkill). I'm quite keen to see it done - not because ActiveAdmin is primarily used on mobile devices, but because it's useful to be able to occasionally change something when not at a desk...

I've added a `responsive` mixin, which defines styles for viewports with a width of less than `$active-admin-responsive-width`, and can be disabled completely by setting `$active-admin-responsive: false`.

Tables are made responsive using an idea stolen from http://elvery.net/demo/responsive-tables/#flip-scroll. As this could potentially upset a few people, I've also added `$active-admin-flip-table`, which can be set to `false` to switch this off.

Since these are mostly style changes... have a few screenshots!

screenshot from 2014-05-24 10 26 50
screenshot from 2014-05-24 11 12 17
screenshot from 2014-05-24 11 15 08
screenshot from 2014-05-24 11 16 00

### DIFF
--- a/app/assets/stylesheets/active_admin/_forms.css.scss
+++ b/app/assets/stylesheets/active_admin/_forms.css.scss
@@ -26,6 +26,11 @@ form {
         font-size: 1.0em;
         font-weight: bold;
         color: $form-label-color;
+
+        @include responsive {
+          width: 100%;
+        }
+
         abbr {
           border: none;
           color:  $required-field-marker-color;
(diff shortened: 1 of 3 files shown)


noctivityinc:
I wonder why there wasn't more thought given to this amazing gem.  The fact that AA is not responsive in 2020 is insane to me.

PR #3170, opened 2014-05-26
Router rewrite and support of singleton belongs to resources

dmitry:
It's WIP branch, which will be squashed and completed, after you give me a green light on it.

- [x] Rewrite of the router definitions
- [x] Add`s ability to define singleton resource

#### Related

- Support multiple belongs_to (https://github.com/gregbell/active_admin/issues/221)
- Support nested belongs_to (#3280)

#### Singleton resource

```ruby
ActiveAdmin.register Place
ActiveAdmin.register Location do
  belongs_to :place, singleton: true
end
```

Generates correct route with `resource` DSL command, instead of `resources`, so it's possible to get access location using the url: `/admin/users/1/location` instead of `/admin/users/1/locations`.

### DIFF
--- a/lib/active_admin/resource/belongs_to.rb
+++ b/lib/active_admin/resource/belongs_to.rb
@@ -28,6 +28,10 @@ module ActiveAdmin
         @options[:optional]
       end
 
+      def singleton?
+        @options[:singleton]
+      end
+
       def required?
         !optional?
       end
--- a/lib/active_admin/router.rb
+++ b/lib/active_admin/router.rb
@@ -1,5 +1,11 @@
 module ActiveAdmin
   class Router
+    extend ActiveSupport::Autoload
+
+    autoload :ResourceRoutes
+
+    attr_reader :application
+
     def initialize(application)
       @application = application
     end
@@ -13,12 +19,14 @@ module ActiveAdmin
     #   end
     #
     def apply(router)
-      define_root_routes router
-      define_resource_routes router
+      define_root_routes(router)
+      define_resource_routes(router)
     end
 
+    private
+
     def define_root_routes(router)
-      router.instance_exec @application.namespaces.values do |namespaces|
+      router.instance_exec application.namespaces.values do |namespaces|
         namespaces.each do |namespace|
           if namespace.root?
             root to: namespace.root_to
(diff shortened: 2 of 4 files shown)


bamorim:
I'll try to handle this soon if I finish my week tasks on my job.

However, what refactoring you think should be applied?

I'm thinking about an elegant way to do this and #3280. I'll first finish this and then I'd copy my tests and rewrite it and update my PR.
Suggestions are welcome.

timoschilling:
The plan of @dmitry was to split [`define_routes`](https://github.com/activeadmin/activeadmin/pull/3170/files#diff-dd9fd6675d3b22ad75de1071cb9f17b1R26) into 2 different ResourceRouter's, one for each kind.

timoschilling:
And we need a rebase

timoschilling:
@bamorim the best think should be you create a new branch based on `origin/pr/3170`, continue the work and create a new PR with refs this one.

bamorim:
@timoschilling wow, I was going to suggest it anyway. hahaha
So it's fine. I'll see if I can do this in the weekend.

Another thing, maybe we shouldn't have another type of resource? just like Resouce and Page?
Something like register_singleton.
I mean, this have completely diferent use cases than a normal resource. Also, removing the singleton: true from the belongs_to we will be able to do root singletons.

But this is discussion for another PR, I think.

Actually, I think we have to rethink the architecture. Just passing belongs_to to inherited resources is not going to help. We are strongly dependent on them, and valim just stopped supporting it.

Probably we should have some kind of adapter for the Inherited Resources. I don't know.
IMO we really consider dropping inherited resources or still using it through that adapter. This will allow us to move quickly.

timoschilling:
@bamorim I agree with you, but as you say it's all for new PR's / Issues.

@rafaelfranca is supporting `inherited_resources` until Rails 5.0 for us. But after that we need to replace it, I think with a own solution. Maybe @bamorim and @rafaelfranca can work on that. There is already a Issue for that #3604.

timoschilling:
@bamorim any updates?

nileshtrivedi:
What's the status of this PR?

seanlinsley:
Given the limited time we have to review PRs and that we'll have to remove Inherited Resources soon (#3604), I don't think we can afford to accept this PR.

I'm really sorry it turned out this way :disappointed:

PR #3193, opened 2014-06-03
Fix compatibility with Inherited Resources 1.5.x

petergoldstein:
Specs (and app behavior) were failing for the belongs_to association in a Controller.  This was failing because Inherited Resources was no longer inferring the parent_class from the parameters passed in (as it was in 1.4.1).  So when find was called on the parent_class, this was resulting in a NPE.

I fixed this by doing this inference ourselves, classifying and then constantizing the belongs_to :target.  Unfortunately this means that if the target name is not a class name, then the parent_class will need to be explicitly set.

I'm not 100% sure this is the right fix - I'd love to get some other opinions.

### DIFF
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'bourbon'
   s.add_dependency 'coffee-rails'
   s.add_dependency 'formtastic',          '~> 3.0'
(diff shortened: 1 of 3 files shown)


preston:
@rafaelfranca Thanks! There are also a ton of projects using inherited_resources that aren't using activeadmin. Any general guidance for migrating off prior to Rails 5 would be appreciated!

PR #3207, opened 2014-06-15
Restyling with Bootstrap, Fontawesome, & More

krainboltgreene:
This pull request represents a unit of work that attempts to root out the previous homebrew style for a more "modular" and addon-rich environment of Bootstrap. It also replaces the previous SVG icon library with the font-based Fontawesome.


Jamevid:
+1 :)

Another reason (not related to design) could be to just get faster and smoother page loads with ajax (eg.Turbolinks etc)  - when going from AA to the front-end or rails-admin or rails_db etc or other parts that most likely is based on Bootstrap where new CSS now has to be loaded.. which is specially slow for mobile networks. 

Ok, you could place new CSS in the body but it wont be as smooth and tweaky. 
Http request with new CSS can really take forever to load on mobile networks.. So it could be a great improvement for speedier apps.  

Maybe this has already been mentioned. Please have mercy in that case :)

seanlinsley:
Feel free to open a PR to add Turbolinks support for Active Admin.

seanlinsley:
Chrome's developer tools let you simlulate a slow network connection, so it would be interesting to see how much Turbolinks speeds up page navigation.

Jamevid:
Thanks for the answer and the suggestion to add Turbolinks :)

**Turbolinks PR idea below**
- add something like [this](https://gist.github.com/Jamevid/9121460c0f6d41b282a06e47ae901b0b) to base_controller.rb
- in form.rb: add "remote: true" in default_form_options method
- ....same repond_to and remote true for devise and filters when action code is final
- (install Turbolinks gem and add JS)
- add a config.turbolinks in ActiveAdmin.setup and add proper conditions if true on all above code. 

The code should work. What do you think? 


**What I meant about Bootstrap**
For example ..an app hosted in US and browsed on _mobile_ networks in South East Asia - Post requests with redirect_to AND new css can take 5-7 seconds to render in the browser :P.  So Turbolinks/ajax solves this to about 1 second page load which is fine If same CSS/JS throughout the app was used for both fronted and backend. So its not about Turbolinks. Its the same principal without Turbolinks because of the instability of the mobile networks (as I understand it). Its just connected to the best solution. 

Hope this bring something new to the CSS discussion thats more technical  - not so much about taste or personal preferences. Bootstrap as an option would be great too  https://github.com/activeadmin/activeadmin/issues/4266 :)

**Active Bootstrap skin**
Now I found below bootstrap skin. It was just created, will try it out :)
https://github.com/vinhnglx/active_bootstrap_skin

Issue #2232, opened 2013-05-29
Add jquery-ui-rails as a Dependency


dsilver829:
Active Admin should add `jquery-ui-rails` as a dependency.

Active Admin currently depends only on `jquery-rails`. But `jquery-ui-rails` is the gem that the `jquery-rails` maintainers themselves recommend for jQueryUI.

I spent a while today [drilling into a bug](https://github.com/rails/jquery-rails/issues/126) that occurred because Active Admin calls `//= require jquery-ui`, which pulls up the (discouraged) jQueryUI file from within `jquery-rails`. This, in turn, clobbers the `jquery-ui-rails` asset `//= require jquery.ui.all`.

Since `jquery-rails` is planning to [remove jQueryUI support soon](https://github.com/rails/jquery-rails/issues/46#issuecomment-18590453), it makes sense for Active Admin to add  `jquery-ui-rails` as a dependency.

PR #3251, opened 2014-07-04
Fix jquery ui asset

hendrauzia:
_Problem:_
I stumbled upon an issue where active admin throws error `couldn't find file 'jquery.ui.datepicker'`, I have looked at issue #2232, and checked the `jquery-ui-rails` gem version in `Gemfile.lock`, it already had the correct version, unfortunately in my case it still throws errors.

_Solution:_
I checked on jquery-ui-rails (https://github.com/joliss/jquery-ui-rails), and noticed that it uses a different way of adding asset, which uses `jquery-ui/datepicker` instead of the one used in `base.js.coffee` which uses `jquery.ui.datepicker`, hopefully this satisfies other users that had the same issue.

_Tested On:_
Mac OSX 10.9.3
Ruby 2.1.2

### DIFF
--- a/app/assets/javascripts/active_admin/base.js.coffee
+++ b/app/assets/javascripts/active_admin/base.js.coffee
@@ -1,8 +1,8 @@
 #= require jquery
-#= require jquery.ui.datepicker
-#= require jquery.ui.dialog
-#= require jquery.ui.sortable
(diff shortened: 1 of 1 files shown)


iamxsx:
It works !!!

PR #3280, opened 2014-07-15
Adding nested belongs_to

### LINKED ISSUE #3274: Nested Belongs_to does not work as expected
It would be nice if active_admin support InheritedResources nested belongs_to

I'm putting it here but I'll try to implement that feature.

My suggestion is that ```belongs_to``` method instead of setting ```@belongs_to``` it would push a configuration to an array.

Then, the router could run through the belongs_tos setting the routes.

bamorim:
Fixes #3274
I initially created a pull request with non descriptive branch name on #3276

Now it's okay.
Another PR that attempted to solve this was #2546 but I think it's a little bit out of date.

Hope you enjoy.


halilim:
Ping...

varyonic:
After some confusion, this has now been merged as #4618

PR #3287, opened 2014-07-18
Fix cases where events bound to document were 'rebounding'

rodrigoddalmeida:
This PR will fix cases dealing with Turbolinks or anything messing with the ready event. Events bound to the 'document' object were being triggered twice (or more).

Since this method of event binding is independent of the target object existing in the DOM at the moment of its runtime, it is actually better to leave them out of the ready function.


timoschilling:
replaced by #3753

PR #3361, opened 2014-08-26
introduce JS: ActiveAdmin.flash

timoschilling:
with this we can show flash messages via js at runtime, for example a message after a ajax call or remote link click


adisos:
@timoschilling 

Hi, are you still using it? Because now `<div class="flashes"></div>` is hidden by default.

seanlinsley:
@adisos, @timoschilling's example is still working for me

PR #3388, opened 2014-09-01
Added tabs dsl for organizing content on active admin pages

yagudaev:
Added tabs to active admin to enable users to better organize content on different page (e.g. show pages).

This is the result:
screen shot 2014-09-01 at 12 14 20 am

DSL:

```
tabs do
  tab :overview do
    span 'tab 1'
  end

  tab :details do
    span 'tab 2'
  end
end
```

I figured I will have it reviewed before starting to write documentation for it. Hope we can get this merged quickly. It is a relatively small change, but really useful when a model has too much data on the screen. You can even be cool like github and display a preview for your html/markdown in a tab just like github does :).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,9 +17,12 @@ gem 'pundit'
 gem 'rake', require: false
 gem 'parallel_tests'
 
+group :development, :test, :cucumber do
+  gem 'pry'                # Easily debug from your console with `binding.pry`
+end
+
 group :development do
   # Debugging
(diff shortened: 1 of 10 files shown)


Fivell:
@yagudaev, thanks for tabs! would be very nice to see also jQuery accordion as well  :+1:

yagudaev:
:). About the accordion, you should be able to do it by using panels and just using the right css selector in your jquery to instantiate it easily.

PR #3435, opened 2014-09-14
replace ActiveAdmin::Event with ActiveSupport::Notifications

timoschilling:
This is the implementation for the discussion in #3364.

`ActiveAdmin::Event` has a MessageQueue implementation that supports the basics, but is not thread safe for example. `ActiveSupport::Notification` provides already a MessageQueue, so why should ActiveAdmin implement his own?

This pull request replace the usage of `ActiveAdmin::Event` with `ActiveSupport::Notifications` and deprecate all `ActiveAdmin::Event` methods, so it can be removed in a future version.

```ruby
ActiveSupport::Notifications.subscribe ActiveAdmin::Application::BeforeLoadEvent do |event, *args|
  # some code
end

ActiveSupport::Notifications.publish ActiveAdmin::Application::BeforeLoadEvent, "some data"
```

### DIFF
--- a/lib/active_admin/views/components/sidebar_section.rb
+++ b/lib/active_admin/views/components/sidebar_section.rb
@@ -13,6 +13,11 @@ module ActiveAdmin
         build_sidebar_content
       end
 
+      # Renders attributes_table_for current resource
+      def attributes_table(*args, &block)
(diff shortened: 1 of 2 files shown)


Fivell:
:+1:

PR #3463, opened 2014-10-02
use same dsl for actions and actions with dropdown (fix #2625)

### LINKED ISSUE #2625: index actions block doesn't properly buffer Arbre
Given this:

``` ruby
index do
  # ...
  actions defaults: false do
    a 'Details'
    a 'Update'
  end
end
```

You get this:
screen shot 2013-10-27 at 5 10 00 pm

timoschilling:
## Before

Before this PR the index action DSL is quite different and partly very dirty.
#### 1. actions _with dropdown_ and custom actions has a different mark up than actions _without dropdown_ and custom actions

``` ruby
actions dropdown: true do |admin_user|
  item 'Grant Admin', grant_admin_admin_user_path(admin_user)
  # here you need to use `item`
end

actions do |admin_user|
  link_to 'Grant Admin', grant_admin_admin_user_path(admin_user)
  # here you need to use a string retuning method
end
```
#### 2. If you want to show two custom actions (without dropdown), you need to do a very dirty thing:

``` ruby
actions do |resource|
  (link_to "Foo", foo_admin_resource_path(resource)) + " " +
  (link_to "Bar", bar_admin_resource_path(resource))
end
```
## After

After this PR both actions types use the same DSL and you don't need to concat strings to display 2 custom actions

``` ruby
actions do |resource|
  item "Foo", foo_admin_resource_path(resource)
  item "Bar", bar_admin_resource_path(resource)
end

actions dropdown: true do |resource|
  item "Foo", foo_admin_resource_path(resource)
  item "Bar", bar_admin_resource_path(resource)
end

# You can use the arbre too:
actions dropdown: true do |resource|
  a "Foo", href: foo_admin_resource_path(resource)
  a "Bar", href: bar_admin_resource_path(resource)
end
```

NOTE: the old way still works, but should be depreciated

This PR fix #2625

### DIFF
--- a/docs/3-index-pages/index-as-table.md
+++ b/docs/3-index-pages/index-as-table.md
@@ -72,7 +72,7 @@ index do
   selectable_column
   column :title
   actions do |post|
-    link_to "Preview", admin_preview_post_path(post), class: "member_link"
+    item "Preview", admin_preview_post_path(post), class: "member_link"
   end
 end
 ```
@@ -83,7 +83,18 @@ Or forego the default links entirely:
 index do
   column :title
   actions defaults: false do |post|
-    link_to "View", admin_post_path(post)
+    item "View", admin_post_path(post)
+  end
+end
+```
+
+Or append custom action with custom html via arbre:
+
+```ruby
+index do
+  column :title
+  actions do |post|
+    a link_to "View", admin_post_path(post)
   end
 end
 ```
--- a/lib/active_admin/views/index_as_table.rb
+++ b/lib/active_admin/views/index_as_table.rb
@@ -69,7 +69,7 @@ module ActiveAdmin
     #   selectable_column
     #   column :title
     #   actions do |post|
-    #     link_to "Preview", admin_preview_post_path(post), class: "member_link"
+    #     item "Preview", admin_preview_post_path(post), class: "member_link"
     #   end
     # end
     # ```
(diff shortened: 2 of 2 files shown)


scarver2:
@timoschilling Thank you for documenting the custom actions DSL and the dropdown feature.

Is there a way to sort the actions when combining with custom actions?

timoschilling:
@scarver2 It's documented in [index-as-table.md](https://github.com/activeadmin/activeadmin/blob/master/docs/3-index-pages/index-as-table.md#defining-columns) too or do you miss something there? But we should separate actions from columns in the docs.

There is no way to sort the actions at the moment. But I reopen #3053 as a solution for that.

timoschilling:
@scarver2 now there is a way, take a look on #3053

AlexVPopov:
@timoschilling The new way doesn't leave a space between the links. If you have:

```ruby
item "Foo", foo_admin_resource_path(resource)
item "Bar", bar_admin_resource_path(resource)
```

You get [Foo](#)[Bar](#), instead of [Foo](#) [Bar](#).

timoschilling:
@AlexVPopov Thats needs to be solved by css or you make this:
```ruby
item "Foo", foo_admin_resource_path(resource)
text_node " " # or "&nbsp;"
item "Bar", bar_admin_resource_path(resource)
```

AlexVPopov:
And what if you want to pass additional options to the link, such as data attributes like `method` and `confirm`?

timoschilling:
If you take a look into the implementation of `item` https://github.com/activeadmin/activeadmin/pull/3463/commits/f52c8f789fff2493b63b524d229516a401f4a2ac#diff-fbc2a61dc2383a7fac8bafee1b8a3634R341 you will see that `item *args` is just a `text_node link_to *args`, that means any arg you pass into `item` will be passed to `link_to`.

PR #2138, opened 2014-10-22
Better defaults for the Show page

timoschilling:
I'll turn this into a PR once there's code to go along with the idea.

This replaces the stale #1077.

``` ruby
# Renders the default table
show do
  attributes_table
end

# Removes specific attributes
show do
  attributes_table except: [:foo, :bar]
end

# Adds extra (virtual) attributes
show do
  attributes_table with: :baz
end
```

### DIFF
--- a/lib/active_admin/views/components/attributes_table.rb
+++ b/lib/active_admin/views/components/attributes_table.rb
@@ -7,12 +7,12 @@ module ActiveAdmin
       def build(obj, *attrs)
         @collection     = is_array?(obj) ? obj : [obj]
         @resource_class = @collection.first.class
-        options = { }
+        options = attrs.extract_options!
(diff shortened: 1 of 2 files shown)


seanlinsley:
It looks like this PR removes the ability to pass in a Hash, but as I [said above](https://github.com/activeadmin/activeadmin/pull/2138#issuecomment-60093227), it's not necessary to drop that feature.

varyonic:
I propose to close this as incomplete (insufficient tests) and abandoned.  Sean thought it could be salvaged but Timo was not convinced.  I looked at finishing this myself but the benefit is minor and I don't think it is worth it.

Fivell:
@varyonic , agree

PR #3511, opened 2014-10-24
Remove metasearch backports

timoschilling:
In following to https://github.com/activeadmin/activeadmin/pull/3486#issuecomment-60332348 from @seanlinsley

> The next release is going to be 1.0.0, and we'll be following semantic versioning. I really don't want that cruft in the code base until 2.0.0

We should remove the `metasearch` backports.


pboling:
Would it be possible to also get active_admin to loosen up the dependency on ransack as well?  Ransack is already past a 1.6.x release, but the dependency definition in the gemspec here requires a 1.3.x version.  We use ransack in our app outside of active_admin, and would like to be able to upgrade it independently (within reason).  You are using the `between` concept for the rails dependency with [`s.add_dependency 'rails',               '>= 3.2', '< 5.0'`](https://github.com/activeadmin/activeadmin/blob/master/activeadmin.gemspec#L25), so perhaps ransack could be:
```
s.add_dependency 'ransack',             '>= 1.3', '< 2'
```

timoschilling:
@pboling `~> 1.3` requires a version `1.x` not a `1.3.x`, so `~>1.3` is the same as your `'>= 1.3', '< 2.0'`

pboling:
Oh, right.  I knew that at one point.  Thanks.

seanlinsley:
Before my nearly year-long hiatus, I had been working on AA for around two full years, during which I made a ton of backwards-incompatible changes without actually making a release. Now that I'm back, I'm very hesitant to make any more breaking changes unless they're necessary.

I still prefer verbs like `contains` versus `cont`, but more importantly people other than myself have been relying on this for almost two years now.

timoschilling:
@seanlinsley I give you right about the "`contains` versus `cont`", but ransack name it `con` so we should not rename it. For more reasons read my previous comment https://github.com/activeadmin/activeadmin/pull/3511#issuecomment-60452831. Why do we need backports at this place?

seanlinsley:
> but ransack name it `con` so we should not rename it.

I assume this was a typo, and you meant to type `cont`

As I said, other than my preferring the more human `contains`, the other reason to keep it is those that ported their AA apps from Metasearch to Ransack two years ago were at least able to keep those verb names the same. Now you're asking them to change their code again, when we haven't even made a stable release since the Metasearch -> Ransack migration was completed two years ago.

And besides this being a breaking change, it doesn't affect the readability or performance of our code. So this change should at the very least wait until version 2.0.0.

I'm sorry, but we're wasting our time here arguing about this. There are much more important issues to resolve.

I'm going to close this PR. We can come back to it once we start considering what breaking changes to include in 2.0.0.

timoschilling:
I reopen this PR, while the backports are buggy, see #4078 for example

javierjulio:
@timoschilling since you had this for v2, are you still interested in getting this out? Would @Fivell be able to help out here? I've noticed a lot of his work with these predicates and such.

javierjulio:
@deivid-rodriguez do you think we could still get this in? Maybe easier to create a new PR and redo the changes rather than resolve each merge conflict?

javierjulio:
This will be addressed in #8010 and part of our next major release around Ransack changes.

PR #3535, opened 2014-10-29
Don't stream CSV in development

seanlinsley:
This enables any exceptions to bubble up, instead of waiting until the request times out before sending back any data. This also lets you use in-browser debuggers, like better_errors.

TODO: add tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ env:
   matrix:
     - RAILS=3.2.22
     - RAILS=4.1.13
-    - RAILS=4.2.4
     - RAILS=master
+    - RAILS=4.2.4
   global:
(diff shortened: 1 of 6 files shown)


timoschilling:
Maybe we should name it `disable_streaming_in` to make sure that it is a array property.
Or we should allow `false` as a value for compelled disabling.

timoschilling:
@seanlinsley can you finish this PR?

seanlinsley:
Okay I just revisited this PR. Oddly, in development my CSV exports consist of a single character:

screen shot 2015-11-19 at 7 37 58 pm

seanlinsley:
Ah nevermind, I was missing a line after rebasing.

seanlinsley:
This is ready for review.

Issue #3539, opened 2014-10-30
Default render of XML, CSV... is highly unsecure (even with download_links = false)


micred:
In standard activeadmin configuration all formats renderer are enabled by default (even if download_links = false).
CSV, XML... render by default all database attributes.
I think it's important to explicitly enable that feature, otherwise sensible data not shown in activeadmin show or index will leak out.

How can I disable output different from HTML?

juliogarciag:
I don't know if this issue is still active but i found that constraining the format in a `constraint` block to just html when mounting activeadmin invalidates the `xml` and `json` routes.

```ruby
constraints :format => :html do
  ActiveAdmin.routes(self)
end
```

seanlinsley:
I don't see this as a serious security issue, since Active Admin is meant for true administrators of your website to use (people who already have your trust). While it is a security issue for certain uses of Active Admin, fixing it isn't a priority.

However that doesn't mean a PR wouldn't be accepted.

timoschilling:
@seanlinsley I'm with you but we should leave this open

seanlinsley:
:+1:

vraravam:
hi - I am looking for this exact solution: just hiding the download links does not work, I need to be able to un-authorize a user if their role does not allow to actually download CSVs, etc. Could this be merged soon?

PR #3542, opened 2014-10-30
dynamic respond_to in ResourceController

timoschilling:
this is a fix for #3539

befor this change, AA respond to html, xml and json on each action, that's a securty problem while the user can view a record as xml and see the raw data.

after this change AA respond only to html on each action and respond to xml, csv, json only it the format is configured for the download links.

Now a user can only see that data he has access to

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ index download_links: ->{ can?(:view_all_download_links) || [:pdf] }
 ### Security Fixes
 
 * Prevents potential DOS attack via Ruby symbols [#1926][] by [@seanlinsley][]
+* Prevents access to formats that the user not permitted to see [#3542][] by [@timoschilling][]
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


seanlinsley:
This PR is nearly a year old, and is still missing tests. While it would be nice to limit download links based on the user's permissions, this isn't the right approach. Instead it should be limited by `ActiveAdmin::AuthorizationAdapter`.

@timoschilling if you still plan to work on this PR, feel free to reopen.

timoschilling:
We should keep this open, while it provides a solution. The age isn't necessary, we have PR's that are older.

vraravam:
hi - I am looking for this exact solution: just hiding the download links does not work, I need to be able to un-authorize a user if their role does not allow to actually download CSVs, etc. Could this be merged soon?

dmitry:
@vraravam once it will be updated with tests and auth changes - it will be merged.

Fivell:
@timoschilling , I think this PR is worth rebasing

timoschilling:
@Fivell feel free to continue the work on this

Issue #2714, opened 2013-11-20
Mongoid support


seanlinsley:
Inspired from [a conversation in IRC](https://botbot.me/freenode/activeadmin/msg/7993192/) with @machonemedia, I actually read the source code of @elia's [activeadmin-mongoid gem](https://github.com/elia/activeadmin-mongoid) and was surprised by how little code was actually there. I've never used Mongoid, but it's about time that Active Admin supported a different ORM.

Now in this ticket I'm not looking for +1s; #26 has already made the desire for this clear. Instead I want to identify the hurdles in the way for Active Admin to natively support Mongoid, and what we can do to get from here to there.

There are obviously changes (like #1713) to be made to Active Admin, but I'm more concerned about our dependencies:
#### Kaminari

Do things like #2638 apply to huge MongoDB data sets?
#### Ransack

Ransack [doesn't currently support](https://github.com/activerecord-hackery/ransack/issues/120) Mongoid, and it [may never happen](https://github.com/activerecord-hackery/ransack/issues/120#issuecomment-16276809). There's an [in-progress attempt](https://github.com/kristianmandrup/ransack-mongoid) at adding that functionality, but I don't know how far that's going to go.

I've set the 1.0.0 milestone to February, so the question is can we get a compatibility layer for Ransack 90% done in that time? If that's a dead end, what other options do we have?

CC @johnnyshields, @kristianmandrup

maran:
It's been a while. Anybody did any work on this lately?

pencilcheck:
It's sad but @Zhomart version still has a lot of holes with the current master of activeadmin, it just doesn't work out of box, and in the errors are methods that should've been overwritten a long time ago.

Zhomart:
activeadmin is improved a lot in the last month. Now I got latest master changes and started adding mongoid as @timoschilling suggested earlier. It doesn't work yet, after finishing basic mongoid support and comments I'll make pull request. I'm working on https://github.com/Zhomart/active_admin/tree/mongoid.

timoschilling:
@Zhomart what do you think about activerecord-hackery/ransack#469, how should we deal with it?

Zhomart:
@timoschilling we need to improve it too.

timoschilling:
@Zhomart If you work on the adapter interface, please create a issue for the discussion about the API.

Zhomart:
[...]

PR #3551, opened 2014-11-01
rename orm folder to object_mappers

timoschilling:
Mongoid is a ODM (Object Document Mapper) and not a ORM (Object Relational Mapper)
ActiveRecord is a ORM

Object Mapper covers both kinds.


timoschilling:
closed, while covered by the upcoming PR resulting from #2714

Issue #3451, opened 2014-09-24
don't use !important in SVG color mixin


yoyos:
Hello, this commit made a color bug in my ActiveAdmin UI: https://github.com/activeadmin/activeadmin/commit/930d34cd224272687613dcf1db4cf5b1347f1098

Are you sure !important wasn't important for something ^^ ?
I didn't put any custom css, here what I have:

![ActiveAdmin UI](https://cloud.githubusercontent.com/assets/1993859/4389901/212a866a-43f9-11e4-8f68-01f5e36f583d.png)

timoschilling:
closed by #3553

Issue #3399, opened 2014-09-03
switch from svg icons to Font Awesome


timoschilling:
I would prefer a switch from the ~140 [svg icons](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/iconic/icons.rb) to the ~550 [Font-Awesome](http://fortawesome.github.io/Font-Awesome/cheatsheet/) icons.

One other reason is that svg is not colorable from css, Font-Awesome is that!
For example I work on a Icon in Flash messages, but coloring for svg smells:
```ruby
        def build_flash_messages
          div class: 'flashes' do
            flash.each do |type, message|
              div class: "flash flash_#{type}" do
                color = case type
                when :error
                  "#b33c33"
                when :notice
                  "#416347"
                end
                span Iconic.icon :x_alt, color
                span message
              end
            end
          end
        end
```

Font coloring works out of the box

sepastian:
+1 for font awesome

timoschilling:
closed by #3553

Fivell:
@timoschilling , sorry but not very clean for me,  how to use new icons? should I add  gem font-awesome-rails to Gemfile or active_admin  has it as dependency ?

yoyos:
Same question.

Something like that ? How to implement it ?

```
actions defaults: false do |p|
      links = '<span class="default_actions"> '.html_safe
      links += link_to '<i class="fa fa-camera-retro fa-lg"></i>'.html_safe, edit_admin_user_path(p), :title =>  'my title'      
      links += '</span>'.html_safe
      links
    end
```

timoschilling:
We haven't switched to Font Awesome, but we have removed the Icon support, thats why I closed this issue. // cc @Fivell

timoschilling:
I have a idea for a new gem `font-awesome-i18n`, where you can do things like:
```yaml
en:
  edit: "{{edit}}"
```
Which is rendered as:
```
<i class="fa fa-edit"></i>
```

yoyos:
Good idea :)

terry90:
+1

PR #3553, opened 2014-11-02
Remove iconic

timoschilling:
ref: #3399, #3451

Changelog and backport how to will follow


javierjulio:
Great that we are dropping this but I'm not sure font icons are the best way to go. I think it was good to ditch Iconic for another solution but I get the feeling SVG will be the better way going forward. Although when I did a search on this recently, seems hard to find a library with SVG icons (at least one that is free/open source).

timoschilling:
I think in the future ActiveAdmin will not support Icons natively, the old way was creepy and works only on a few places and I thing we should not build in a new way.

I have described a possible new way in https://github.com/activeadmin/activeadmin/issues/3399#issuecomment-91053676.

seanlinsley:
I don't think this should be in 1.0.0. It's a breaking change that causes unnecessary pain.

timoschilling:
@seanlinsley it's gone since over halve a year. A 1.0.0 means that there is a breaking change! You can put them into a separate gem.

seanlinsley:
It's a breaking change with no replacement. I think this should wait until we have a suitable replacement. Until then, you're causing unnecessary pain for people who try upgrading.

seanlinsley:
It's fine if this stays on master, but I think the official 1.0.0 release on Rubygems shouldn't include it.

seanlinsley:
I'd have a different opinion if we had a replacement ready for the 1.0.0 release (which I'd like to get out in the next month or so).

timoschilling:
I think we should not have a replacement, if some want Icons he can use fontawesome & co

javierjulio:
Yes I'm with @timoschilling on this. Until we can get a theme into the app that is easily configurable in a lot of areas this should be best left out and let users add them as they want it. Or a solution that can be easily turned off and not included if someone wants to use their own library. We should worry less about including heavy defaults and let users add their own without having bulk.

PR #3591, opened 2014-11-06
change view context for csv builder

### LINKED ISSUE #3197: Helpers unavailable in CSV export
Title was: "display_name is unusable in CSV export"

I recently updated an application of mine from AA ced7d35703bf22a967b0ce66ec9dd2f4bb5ac881 to 8b4c1b5ef65ebdf77413bcaa955c0f7badb26ffb. Turns out that this makes `display_name` cause the CSV file to not have any content. I don't get any stack trace either, so the solution isn't clear.

timoschilling:
this change is proposed by @seanlinsley in #3197 (close #3197)

### DIFF
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -39,10 +39,10 @@ module ActiveAdmin
       @columns << Column.new(name, @resource, column_transitive_options.merge(options), block)
     end
 
-    def build(view_context, receiver)
-      @collection = view_context.send(:collection)
+    def build(controller, receiver)
+      @collection = controller.send(:collection)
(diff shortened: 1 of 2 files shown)


varyonic:
I am baffled why this breaks my Capybara chromedriver spec, but it does.

seanlinsley:
@varyonic specifically this commit? So the commit before it (1a3ce35b29b6a87b9ce9bfbf0c8477bc5829b26e) works fine?

varyonic:
Yes.  The test was a little hokey as it unregistered the CSV mime type and tested the contents of the rendered page.  I rewrote the test to wait on the file to be downloaded and then test the content of the file.  I'm not going to pursue this any further.

Issue #1126, opened 2012-03-12
Isolated kaminari views & styles


stgeneral:
acitve_admin should have and use own isolated views and/or styles and do not override kaminari's defaults.

Now, if you use active_admin then you are forced to specify theme for your front-end.

PR #3653, opened 2014-11-25
remove kaminari theme

timoschilling:
ActiveAdmin use the default Kaminari theme, thats not nessary to have this theme inside of ActiveAdmin we can use it from the gem.


killthekitten:
@timoschilling it was necessary because of https://github.com/activeadmin/activeadmin/issues/1126 this issue. Now my app's theme breaks ActiveAdmin's paginated views.

PR #3698, opened 2014-12-16
Fix issue with model attribute matching metasearch method name

azach:
Not sure if you're still accepting PRs fixing 0.6 only issues. This is a fix for the case when an attribute on a model matches a method on the metasearch search object. In my case, it was a string attribute called `size`. I was expecting it to create a standard string filter, but got something like this instead:

screen shot 2014-12-16 at 3 26 03 pm

(757 is the actual count of records of this model).

After this fix, it looks like you'd expect:

screen shot 2014-12-16 at 3 40 08 pm

I checked metasearch's docs to see if there were a better method for determining what methods are actually search methods but didn't see anything, so this seemed like the simplest fix.

### DIFF
--- a/lib/active_admin/inputs/filter_string_input.rb
+++ b/lib/active_admin/inputs/filter_string_input.rb
@@ -9,7 +9,7 @@ module ActiveAdmin
       # If the filter method includes a search condition, build a normal string search field.
       # Else, build a search field with a companion dropdown to choose a search condition from.
       def to_html
-        if @object.respond_to? method
+        if @object.methods.exclude?(method) && @object.respond_to?(method)
           input_wrapping do
             label_html <<
             builder.text_field(method, input_html_options)


timoschilling:
Can you rebase this on the top of master to get the tests passing

azach:
yep, done

timoschilling:
Thanks, but I have the following problem: I know this kind of bug from 1.0.0.pre and ransack, but I'm not sure about that this is the right solution. I don't have a 0.6 app to test it and I don't know about that this brakes other things.

azach:
I agree it's not the ideal solution. However, these kind of filters will always work incorrectly due to how the Formtastic builder determines the value for the input field. This means it's impossible to use any methods that exist on an object (like size) otherwise, since there is no real workaround you can otherwise use.

Open to suggestions though.

timoschilling:
Let's give this a try

Issue #362, opened 2011-08-12
Right to left language support


aalsaif:
Could you please add support for right to left languages? I will try to help with Arabic translations, too.

amit0001:
Hi, is there another solution to this, or has this been included in some other way?

javierjulio:
@amit0001 no there isn't but after I complete my work in #3862 which will make up v2 I'd love to have help for this. I'm thinking it would be a good candidate for a 2.1 release perhaps.

PR #3716, opened 2014-12-29
362 right to left layout

eshaiju:
Added right to left layout option by adding

```
 <html lang="<%=I18n.locale%>" >
```

 Also added rtl.scss for switching styles.
Now we need to add documention which instruct user to add 

```
html[lang=ar] {
  @import "active_admin/rtl.scss";
}
```

### DIFF
--- a/app/assets/stylesheets/active_admin/mixins/_all.scss
+++ b/app/assets/stylesheets/active_admin/mixins/_all.scss
@@ -8,4 +8,5 @@
 @import "active_admin/mixins/sections";
 @import "active_admin/mixins/utilities";
 @import "active_admin/mixins/typography";
+@import "active_admin/mixins/rtl";
 @import "bourbon";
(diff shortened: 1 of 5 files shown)


timoschilling:
There are some layout problems in the RTL state:

1. The breadcrumbs are on the right side, but still in LTR order
2. The index table is not resigning in RTL, but is resigning in LTR

eshaiju:
Hi @timoschilling 

Didint understood the 2 nd problem

timoschilling:
The EN images are ok, AR are wrong

With a big browser window:



With a small window:

eshaiju:
Hi @timoschilling  

I will rework on all RTL CSS with my designer friend @imran04u and push with in a week

timoschilling:
any updates?

maysam:
any updates?

amit0001:
Any updates?

javierjulio:
Going to close out as stale. We can track working on a new updated solution in the issue #362. I think it would be great to support RTL languages in AA and with v2 it will hopefully be easier with the CSS overhaul I'm doing in #3862. I'm thinking RTL support could be a great candidate for a v2.1 release perhaps.

PR #3723, opened 2014-12-31
Avoid CSS conflict

lboix:
Between application style and Active Admin one


lboix:
Thanks for merging it into the right new section!

PR #3732, opened 2015-01-04
Set form multipart option automatically.

### LINKED ISSUE #3577: Form builder no longer adds multipart attribute automatically
We recently updated ActiveAdmin from `05778f4b7c8e821d584c7b4b08bcd79c02a0db76` to `0becbef0918a38b91ad120aba7c2f6641e9b5440` and uploading files via CarrierWave is broken.

After digging for a while, I found that the main cause is missing multipart attribute for form tag, which used to be added automatically. 

For those also encounter this issue, there's a simple workaround:

```ruby
form html: { multipart: true } do |f|  
  # Your codes here.
end
```

Possibly related: #3559

varyonic:
Rails 4 sets multipart automatically if a file field is present, but to build an Arbre proxy the form tag has already been rendered before the file field.  Fixes #3577

### DIFF
--- a/lib/active_admin/views/components/active_admin_form.rb
+++ b/lib/active_admin/views/components/active_admin_form.rb
@@ -33,6 +33,12 @@ module ActiveAdmin
 
         @opening_tag, @closing_tag = split_string_on(form_string, "</form>")
         instance_eval(&block) if block_given?
+
+        # Rails 4 sets multipart automatically if a file field is present,
+        # but the form tag has already been rendered before the block eval.
+        if multipart? && @opening_tag !~ /multipart/
+          @opening_tag.sub!(/<form/, '<form enctype="multipart/form-data"')
+        end
       end
 
       def inputs(*args, &block)
@@ -70,6 +76,10 @@ module ActiveAdmin
 
       def has_many(*args, &block)
         insert_tag(HasManyProxy, form_builder, *args, &block)
+      end
+
+      def multipart?
+        form_builder && form_builder.multipart?
       end
 
       def object
--- a/spec/unit/form_builder_spec.rb
+++ b/spec/unit/form_builder_spec.rb
@@ -124,6 +124,20 @@ describe ActiveAdmin::FormBuilder do
     end
   end
 
+  if Rails::VERSION::MAJOR > 3
+    context "file input present" do
+      let :body do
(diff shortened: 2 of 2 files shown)


spec/unit/form_builder_spec.rb
```diff
@@ -124,6 +124,19 @@ def build_form(options = {}, form_object = Post.new, &block)
...
   end
 
+  context "file input present" do
+    let :body do
+      build_form do |f|
+        f.input :body, as: :file
+      end
+    end
+    it "adds multipart attribute automatically" do
+      if Rails::VERSION::MAJOR > 3
```
timoschilling: The if should wrap the context not only the expectation
And pleas leaf a empty line between the let and it block

BTw: nice to see you back on board!

varyonic:
This change breaks form_helper_spec .active_admin_for.  I suggest the spec is redundant and propose to remove it.

timoschilling:
With wich spec is it redundant?

varyonic:
The form_helper_spec is mocking the call to semantic_form_for so @form_builder is never initialized and attempting to render the view fails because @form_builder in the code I added is nil.  I could make this pass by testing for nil in my patch but that should never be true in production.  I believe all of form_builder_spec is exercising the same path covered by form_helper_spec.  If you have a way to fix the spec I’m open to suggestions.

timoschilling:
Do it and we will see

timoschilling:
I agree with your last comment, but I prefer to fix the code or the test.

```ruby
if @form_builder && @form_builder.multipart? && @opening_tag !~ /multipart/
```

lib/active_admin/views/components/active_admin_form.rb
```diff
@@ -33,6 +33,12 @@ def build(resource, options = {}, &block)
 
         @opening_tag, @closing_tag = split_string_on(form_string, "</form>")
         instance_eval(&block) if block_given?
+
+        # Rails 4 sets multipart automatically if a file field is present,
+        # but the form tag has already been rendered before the block eval.
+        if multipart? && @opening_tag !~ /multipart/
+          @opening_tag.sub!(/<form/,'<form enctype="multipart/form-data"')
```
timoschilling: "<form" don't need to be a regex

PR #3748, opened 2015-01-17
Allow pagination to be controller with dynamic per_page

dziemian007:
First of all, I know there were issues and feature requests to support per_page configuration. In one of my project I needed this functionality: allow user to change number of results on one page. And I succeed, but it looked kinda hackish. I know you can do before_filter to set `@per_page` based on some parameter. But: you need to amend each resource, inject some extra html, and ... Maybe it's good idea to deliver that feature out of the box? 

If that sounds like a good candidate to merge, here's how it could word:
* global config from initializer `config.default_per_page` could be set to ether integer or array.
* it could be overwritten in resource controller register `config.per_page`, just like it was before, but now accepting also arrays
* when `per_page` is an array, and pagination on, resource#index adds select with array of options next to pagination links
* selecting some option preserves current query string, manipulating only per_page param

Here's how it would look like:
screenshot from 2015-01-17 21 39 20

What do you guys think?

### DIFF
--- a/lib/active_admin/resource_controller/data_access.rb
+++ b/lib/active_admin/resource_controller/data_access.rb
@@ -276,9 +276,21 @@ module ActiveAdmin
 
       def per_page
         if active_admin_config.paginate
-          @per_page || active_admin_config.per_page
+          dynamic_per_page || configured_per_page
         else
           max_per_page
+        end
+      end
+
+      def dynamic_per_page
+        params[:per_page] || @per_page
+      end
+
+      def configured_per_page
+        if active_admin_config.per_page.is_a?(Array)
+          active_admin_config.per_page[0]
+        else
+          active_admin_config.per_page
         end
       end
 
--- a/spec/unit/views/components/paginated_collection_spec.rb
+++ b/spec/unit/views/components/paginated_collection_spec.rb
@@ -216,5 +216,21 @@ describe ActiveAdmin::Views::PaginatedCollection do
       end
     end
 
+    context "when specifying per_page: array option" do
+      let(:collection) do
+        posts = 10.times.map { Post.new }
+        Kaminari.paginate_array(posts).page(1).per(5)
+      end
+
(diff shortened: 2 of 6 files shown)


lib/active_admin/resource_controller/data_access.rb
```diff
@@ -276,7 +276,11 @@ def apply_pagination(chain)
 
       def per_page
         if active_admin_config.paginate
-          @per_page || active_admin_config.per_page
+          params[:per_page] || @per_page || if active_admin_config.per_page.is_a?(Array)
+            active_admin_config.per_page[0]
+          else
+            active_admin_config.per_page
+          end
```
timoschilling: You need to rewrite this.

A || B || if ... is a bad style.

spec/unit/views/components/paginated_collection_spec.rb
```diff
@@ -216,5 +216,20 @@ def paginated_collection(*args)
...
+        posts = 10.times.map{ Post.new }
+        Kaminari.paginate_array(posts).page(1).per(5)
+      end
+
+      let(:pagination) { paginated_collection(collection, per_page: [1,2,3]) }
+      let(:pagination_html) { pagination.find_by_class('pagination_per_page').first }
+
+      it "should render per_page select tag" do
+        expect(pagination_html.content).to match(/Per page:/)
+        expect(Capybara.string(pagination_html.to_s)).to have_css('select option', count: 3)
```
timoschilling: Move `Capybara.string(pagination_html.to_s)` in a separate line

timoschilling:
Good work, but please change, @houndci and my notes.

dziemian007:
Thanks @timoschilling for such quick feedback! 
I must admit, this @houndci is pretty nasty :smile:

Fivell:
@dziemian007 , I suggest after changing  per_page value, page param should be removed and ui should be forced to first page. otherway after clicking on last page  and increasing per_page I  see "no records found"

timoschilling:
@Fivell you are right, create a PR

PR #3503, opened 2014-10-20
change exec context for batch_action form option

timoschilling:
fix for a https://github.com/activeadmin/activeadmin/pull/1815#issuecomment-59659670

PR #3751, opened 2015-01-19
use render_in_context for eval form block

Fivell:
https://github.com/activeadmin/activeadmin/pull/3503

### DIFF
--- a/lib/active_admin/batch_actions/controller.rb
+++ b/lib/active_admin/batch_actions/controller.rb
@@ -5,9 +5,10 @@ module ActiveAdmin
       # Controller action that is called when submitting the batch action form
       def batch_action
         if action_present?
-          selection =            params[:collection_selection] ||  []
-          inputs    = JSON.parse params[:batch_action_inputs]  || '{}'
-          inputs    = inputs.with_indifferent_access.slice *instance_exec(&current_batch_action.inputs).keys
+          selection  =            params[:collection_selection] ||  []
+          inputs     = JSON.parse params[:batch_action_inputs]  || '{}'
+          valid_keys = render_in_context(self, current_batch_action.inputs).try(:keys)
+          inputs     = inputs.with_indifferent_access.slice *valid_keys
           instance_exec selection, inputs, &current_batch_action.block
         else
           raise "Couldn't find batch action \"#{params[:batch_action]}\""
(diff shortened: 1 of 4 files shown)


Fivell:
@timoschilling , can you review logic before I start playing with houndci  ? Thanks

lib/active_admin/batch_actions/controller.rb
```diff
@@ -7,7 +7,8 @@ def batch_action
         if action_present?
           selection =            params[:collection_selection] ||  []
           inputs    = JSON.parse params[:batch_action_inputs]  || '{}'
-          inputs    = inputs.with_indifferent_access.slice *instance_exec(&current_batch_action.inputs).keys
+          valid_keys = render_in_context(self, current_batch_action.inputs).try(:keys)
+          inputs    = inputs.with_indifferent_access.slice *valid_keys
```
timoschilling: bring all line into one column

timoschilling:
look's good, don't worry about the line length comments from @houndci

Fivell:
>>bring all line into one column 
@timoschilling , I hope I understood you.

lib/active_admin/batch_actions/controller.rb
```diff
@@ -5,9 +5,10 @@ module Controller
       # Controller action that is called when submitting the batch action form
       def batch_action
         if action_present?
-          selection =            params[:collection_selection] ||  []
-          inputs    = JSON.parse params[:batch_action_inputs]  || '{}'
-          inputs    = inputs.with_indifferent_access.slice *instance_exec(&current_batch_action.inputs).keys
+          selection  =            params[:collection_selection] ||  []
+          inputs     = JSON.parse params[:batch_action_inputs]  || '{}'
```
Fivell: @timoschilling ,  maybe we should use single_quotes style ? 
like https://gist.github.com/salbertson/05e579933807493cf1c5

timoschilling: I know there is a minimal performance difference between single and double quotes. But I prefer option B from the https://github.com/bbatsov/ruby-style-guide#consistent-string-literals, while this style can used for each case, with single quotes some times you need to switch to double quotes

timoschilling:
:+1:

Issue #3744, opened 2015-01-15
turbolinks and batch select 


nextbyte:
If I enable turbolinks selecting items for batch actions is not working anymore. When I click the checkbox I get the following error in console:

Uncaught Error: cannot call methods on aaDropdownMenu prior to initialization; attempted to call method 'enable'

 Any idea were to look to fix it?

![aa-turbolinks-error](https://cloud.githubusercontent.com/assets/88761/5753468/a49b2e68-9c92-11e4-9126-bcf810a2db44.png)

nextbyte:
There are some javascript blocks that need to be ran on document:ready/page:load events I'll fix them

timoschilling:
Can you test this PR #3287?

nextbyte:
That is a different issue, doesn't help with this issue

timoschilling:
Is my understanding right? It is the same kind of problem, but with other JS code?

nextbyte:
In that PR the problem was that some events were triggered twice. I think he might had jquery-turbolinks gem also because turbolinks gem only doesn't trigger anything on document ready just on page:load. 

In my case the checkbox in the header that should select all items loaded in the page is not working properly because drop menu was not initialized yet. For example in dropdown-menu.js.cofee I had to change 

$ ->
  $('.dropdown_menu').aaDropdownMenu()

to 

$(document).on 'ready page:load', ->
  $('.dropdown_menu').aaDropdownMenu()

timoschilling:
Would you like to provide a pull request for that. Does we have more places with this problem in our JS code?

nextbyte:
Yes, I am working on it. I think I fixed all from the lib folder I'll check if there are some in other places also.

nextbyte:
https://github.com/activeadmin/activeadmin/pull/3753

timoschilling:
closed by #3753

PR #3753, opened 2015-01-21
Fixed some JS initializers to run on turbolinks page:load also

nextbyte:


nextbyte:
https://github.com/activeadmin/activeadmin/issues/3744

timoschilling:
You are sure that the 3 files are all related files? @rodrigoddalmeida has reported the problem for the `has_many.js` too.

timoschilling:
But :+1: good work

nextbyte:
I tested has many and it works fine. It might be needed in the newly added per_page.js.coffee also, I didn't see it was added before I submitted the PR. I'll check.

rodrigoddalmeida:
I'll just comment that, even if adding the handlers to respond to _page:load_ works, it may not be a reasonable way to do it since it's not a 'default' event name. Solving it like this might cause more maintenance of said code (i.e Turbolinks changing its event names).

Since activeadmin is already highly reliant on jQuery, I do recommend using [kossnocorp/jquery.turbolinks](https://github.com/kossnocorp/jquery.turbolinks) when using Turbolinks, anyways.

Also, there are callbacks being bound to the _document_ and being delegated to the wanted selector. Those cases should be out of the _ready_ event function to not be prone to rebound (called more than once).

nextbyte:
I don't know if it is okay to force users to use jquery.turbolinks in order to make turbolinks work with AA, jquery.turbolinks is provided as an workaround and rails >= 4 only includes the turbolinks gem by default.

Yes, I know there are callbacks that shouldn't be set more then once. From my tests I didn't see any issue like that with the changes I've made.

timoschilling:
I share @nextbyte's point of view

rodrigoddalmeida:
To be fair, it's not about forcing the use of the jquery.turbolinks. The thing is: being neutral and sticking to the default (JS and jQuery) pattern is easier and it'll still work. The code in question is not optimal, with or without Turbolinks.

It could be also solved by adding the jquery.turbolinks asset to the project the same way jQuery is, as a dependency (which I think is not the case at all).

So, I think, it all boils down to AA actively supporting Turbolinks or not, and why should it be supported instead of other options.

nextbyte:
When you create an new rails 4 application turbolinks is added and enabled by default. I know it is not enabled by default in active_admin.js now also but I think users would expect to get it with AA also since it's part of default rails now.

Also with the changes suggested it still works fine even if you don't enable turbolinks so I don't see what is the problem if we support it. I think is should be supported because it is used by default in rails now and I don't know other similar options that we could support.

timoschilling:
@nextbyte :+1:

PR #3765, opened 2015-01-28
Fix default delete prompt on Rails 3.0.x

craigmcnamara:
Comes through in HTML with a data=“{:ruby=>’hash’}”


dmitry:
For me it looks ok. 

@craigmcnamara  BTW why still rails 3.0.x?

craigmcnamara:
It's a long upgrade path for an old app.

PR #3771, opened 2015-01-28
allow tabs to be used in sidebar

Fivell:


Fivell:
before 
before
after  
 after

Issue #3764, opened 2015-01-27
batch actions on a model with default_scope


Vizvamitra:
In my app I have a model with default scope. In some cases admin user can through activeadmin create records of this model that are outside the default scope. The problem is, in this case when admin tries to perform a default batch action :destroy on these records, activeadmin can't find them.

Here is a string in [:destroy batch action sources](https://github.com/activeadmin/activeadmin/blob/abe59fdc6cd73803bd94e6b8403b9df9ec11d947/lib/active_admin/batch_actions/resource_extension.rb#L70) that performs find and destroy:

     active_admin_config.resource_class.find(selected_ids).each { |r| r.destroy }

where `active_admin_config.resource_class` is my model. So it performs MyModel.find(ids), and records with those ids are not in the default scope, and I get RecordNotFound error.

It there anything I can do with that without writing 15 lines of code that redefines :destroy batch action just to change that one line (`active_admin_config.resource_class` => `MyModel.unscoped`)

timoschilling:
There are 3 ways:
1. `remove_batch_action : destroy` and rewrite them
2. redefine `find` on your Models
```ruby
ActiveRecord::Base.class_eval do
  def find(*args)
    unscoped.find(*args)
  end
end
```
3. write a PR to change this :wink: 

You point me to a other problem, it looks like the destroy batch action doesn't respect the constraints form cancan.

Fivell:
@timoschilling, how do you think  should default batch actions to use ```find_resource``` DSL method and/or  ```method_for_find ```  instead of direct `find` ? I mean this implementation of resource retrieving   https://github.com/activeadmin/activeadmin/blob/802ccd2ff2a9d3051122034afe733b794af8c54d/lib/active_admin/resource.rb#L146

timoschilling:
@Fivell yes that's a solution for @Vizvamitra problem. But it doesn't fix the cancan problem. And maybe `find_resource` has the same issue

timoschilling:
And it solves @Vizvamitra problem only if he writes a finder method wich calls unscoped. I think we should fix this by using the `end_of_assioisation_chain` method and using `find_resource`

Fivell:
@timoschilling , maybe ```DataAccess``` module can be used,  there is one more ```find_resource``` in data_access which call collections, which is already authorised.
[...]

Issue #2375, opened 2013-07-26
UI for batch actions for all filtered results


Fivell:
There are some problems with active_admin trying to update filtered results. For instance I there was about 100 pages where I  should update every record (disable/enable, change attribute, etc) after  I had filtered data with sidebar form ....
To do this I must do the same action 100 times. 
Examples: 
1)I need to update all prices (increase on 10%, decrease on 5%, etc) in some category which has more than 100 000 items. 
2) I need to disable all items which has price >= 100 USD ... (more than 1k records)

Such actions can be done with 1 sql query, but there is no way to do this with current UI.

Are there possible ways to improve this?   Any thoughts? Thanks.

Fivell:
@timoschilling , there is a gem that has all the features for now 
https://github.com/activeadmin-plugins/active_admin_scoped_collection_actions

timoschilling:
@Fivell way not as PR?

Fivell:
@timoschilling, because early alpha, lack of time,  no specs, no localization, not  DRY solution for merging.

mmustala:
I would think integrating DataTables into ActiveAdmin would solve this issue. DataTables is able to filter the data in many ways allowing the user to get exactly the wanted records into the table. Then just use the current batch action methods. And possibly merge my pull request for ability to define a custom batch action dialog. https://github.com/activeadmin/activeadmin/pull/3965

jasper-lyons:
I figured out a fairly light weight solution, by providing a wrapper function around ```batch_action``` in ```ActiveAdmin::DSL``` I could identify a lot of useful information about the ```batch_action``` call that allowed me to apply filters, select all records or just do the default behavior.

Heres the implementation:
```ruby
# ./config/initializers/active_admin.rb
module ActiveAdmin
  class DSL
    def filtered_batch_action(title, options, &block)
      self.batch_action title, options do |ids, options|
        if params[:collection_selection_toggle_all] != "on"
          instance_exec(ids, options, &block)
        else
          # pluck is ActiveAdmin specific, probably needs abstracting
          if params[:q].nil?
            instance_exec(resource_class.pluck(:id), options, &block)
          else
            instance_exec(resource_class.ransack(params[:q]).pluck(:id), options, &block)
          end
        end
      end
    end
  end
end 
```
Which can be used like this:
```ruby
ActiveAdmin.register Post do
  filtered_batch_action :delete do |ids, options|
    # do things as you would in normal batch actions
  end
end
```
[...]

PR #3775, opened 2015-01-29
WIP: DataAccess module impovements

Fivell:
Goal of this PR is to provide flexible access to collection in different scopes of request livecycle.
DataAccess#find_collection  becomes more generic and can return different result depends on ```options``` argument.
Can be used in batch actions (no sorting,pagination needed), csv exports(no pagination needed), etc

### DIFF
--- a/docs/9-batch-actions.md
+++ b/docs/9-batch-actions.md
@@ -21,7 +21,7 @@ your desired batch action on all of them:
 ```ruby
 ActiveAdmin.register Post do
   batch_action :flag do |ids|
-    Post.find(ids).each do |post|
+    batch_action_collection.find(ids).each do |post|
       post.flag! :hot
     end
     redirect_to collection_path, alert: "The posts have been flagged."
--- a/lib/active_admin/batch_actions/controller.rb
+++ b/lib/active_admin/batch_actions/controller.rb
@@ -25,8 +25,16 @@ module ActiveAdmin
         active_admin_config.batch_actions.detect{ |action| action.sym.to_s == params[:batch_action] }
       end
 
+      COLLECTION_APPLIES = [
+        :authorization_scope,
+        :collection_decorator,
+        :filtering,
+        :scoping,
+        :includes
+      ].freeze
+
       def batch_action_collection
-        find_collection(only: [:authorization_scope, :collection_decorator, :filtering, :scoping])
+        find_collection(only: COLLECTION_APPLIES)
       end
     end
   end
--- a/lib/active_admin/batch_actions/controller.rb
+++ b/lib/active_admin/batch_actions/controller.rb
@@ -25,6 +25,9 @@ module ActiveAdmin
         active_admin_config.batch_actions.detect{ |action| action.sym.to_s == params[:batch_action] }
       end
 
+      def batch_action_collection
+        find_collection(only: [:authorization_scope, :collection_decorator, :filtering, :scoping])
+      end
     end
   end
 end
--- a/lib/active_admin/batch_actions/controller.rb
+++ b/lib/active_admin/batch_actions/controller.rb
@@ -33,8 +33,8 @@ module ActiveAdmin
         :includes
       ].freeze
 
-      def batch_action_collection
-        find_collection(only: COLLECTION_APPLIES)
+      def batch_action_collection(only = COLLECTION_APPLIES)
+        find_collection(only: only)
       end
     end
   end
--- a/lib/active_admin/batch_actions/resource_extension.rb
+++ b/lib/active_admin/batch_actions/resource_extension.rb
@@ -67,7 +67,10 @@ module ActiveAdmin
         }
 
         add_batch_action :destroy, proc { I18n.t('active_admin.delete') }, destroy_options do |selected_ids|
-          active_admin_config.resource_class.find(selected_ids).each { |r| r.destroy }
+          batch_action_collection.find(selected_ids).each do |record|
+            authorize! ActiveAdmin::Auth::DESTROY, record
+            destroy_resource(record)
+          end
 
           redirect_to active_admin_config.route_collection_path(params),
                       notice: I18n.t("active_admin.batch_actions.succesfully_destroyed",
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -40,7 +40,7 @@ module ActiveAdmin
     end
 
     def build(controller, receiver)
-      @collection = controller.send(:collection)
+      @collection = controller.send(:find_collection, except: :pagination)
       options = ActiveAdmin.application.csv_options.merge self.options
       columns = exec_columns controller.view_context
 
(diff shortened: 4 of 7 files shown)


Fivell:
travis-ci failed  with cucumber 
> Failing Scenarios:
cucumber features/index/filters.feature:109 # Scenario: Clearing filter preserves custom parameters

trace

>ReferenceError: Can't find variable: options
ReferenceError: Can't find variable: options
    at http://127.0.0.1:56079/assets/active_admin.js:20289 in CheckboxToggler
    at http://127.0.0.1:56079/assets/active_admin.js:20779 in TableCheckboxToggler

Fivell:
@timoschilling , can you help to understand why build failed ? 
I can't understand how batch actions can be related to form specs ...
>> ActiveAdmin::FormBuilder with has many inputs with allow destroy with an existing post should include a boolean field for _destroy

timoschilling:
I think build fails while a dependency has changed. See #3781.
Problem is that I don't have the time to find the bug, at the moment.

timoschilling:
please rebase this commit, #3781 is closed

Fivell:
@timoschilling , feel like few comments and specs should be added, but I will have free time only next week, so let's wait a bit.

Fivell:
@timoschilling , rebased

docs/9-batch-actions.md
```diff
@@ -21,7 +21,7 @@ your desired batch action on all of them:
 ```ruby
 ActiveAdmin.register Post do
   batch_action :flag do |ids|
-    Post.find(ids).each do |post|
+    batch_action_collection.where(id: ids).each do |post|
```
timoschilling: we should use `find` like here https://github.com/activeadmin/activeadmin/pull/3775/files#diff-16da2146f502eb423c49bb455a83c2f7R70

timoschilling:
What are the open task?

Fivell:
@timoschilling, we started this PR, having conversation in #3764 , also I hope it will help implement #2375  in future

docs/9-batch-actions.md
```diff
@@ -21,7 +21,7 @@ your desired batch action on all of them:
 ```ruby
 ActiveAdmin.register Post do
   batch_action :flag do |ids|
-    Post.find(ids).each do |post|
+    batch_action_collection.find(id: ids).each do |post|
```
timoschilling: You need to remove id:

timoschilling:
#2375 could be a new PR. Is there something to do before merging this?

Fivell:
@timoschilling , #2375  must be new PR, I don't feel like I can do this because of luck UI skills , hope somebody can help with it. 
before merging this we should decide if this collection of appliers is what we need for default batch actions collection https://github.com/Fivell/activeadmin/commit/d3b25b06318e5b256bf3f71d6b2ed69573896c4a#diff-95a1987f84cc2650e828b2d27c9a2f7fR29

Please take a look

timoschilling:
Feedback:
1. Should we include `includes`? This can speed up the destroying of relations `has_many :orders, dependent: :destroy`
2. Should we add a `COLLECTION_APPLIES` constant to the `BatchController`? This will allow the customization.

Questions:
1. Do you want to write some more test?

Fivell:
@timoschilling, I can't understand about ```includes```. In real life has_many relations are not needed for rendering , that's why there can't be such includes on resource level. 
 
We can also add ```COLLECTION_APPLIES``` for ```BatchController```, but it is also included as a part of  ```DataAccess```  module. 

We already can call 
 ```find_collection(only: [])``` if we don't wan't to use ```batch_colleciton```

timoschilling:
`includes` prevent n+1 queries.

What I mean is:
```ruby
class ActiveAdmin::BatchActions::Controller
  COLLECTION_APPLIES = [:authorization_scope, :collection_decorator, :filtering, :scoping, :includes]

  def batch_action_collection
    find_collection(only: COLLECTION_APPLIES)
  end
end
```

You had many fights with the doc :wink:

Fivell:
@timoschilling , I know that but I can't find real example when we should add ```include``` for batch_collection , when common scenario is changing some attribute of each element of collection.  I mean why we need make additional sql query for include if we don't care about it in all cases?

timoschilling:
The most common Scenario:
Given:
```ruby
class Blog
 has_many :posts, dependent: :destroy
end

class Post
 has_many :comments, dependent: :destroy
end
```

DB:
2 Blogs with 2 Post with 2 Comments each

Exec batch delete of both with includes:
```sql
SELECT * FROM blogs WHERE id IN (1, 2);
SELECT * FROM posts WHERE blog_id IN (1, 2);
SELECT * FROM comments WHERE post_id IN (1, 2, 3, 4);
```

Exec batch delete of both without includes:
```sql
SELECT * FROM blogs WHERE id IN (1, 2);
SELECT * FROM posts WHERE blog_id = 1;
SELECT * FROM posts WHERE blog_id = 2;
SELECT * FROM comments WHERE post_id = 1;
SELECT * FROM comments WHERE post_id = 2;
SELECT * FROM comments WHERE post_id = 3;
SELECT * FROM comments WHERE post_id = 4;
```

Or a other example:
A Shop with Orders, each Order has n Items and a batch action "Send Invoice", here we can use a `includes(:items)`

Fivell:
@timoschilling, lets look at example with blog/posts/comments 
and think how resource for blog will look like.

```ruby
ActiveAdmin.register Blog do 
  
    index do
       column :name
       column :description
       column :created_at 
    end
    
end
```
or something like this. What are saying is that for destroy optimisation I should add ```includes``` applier (but it doesn't exist at all on resource level)

ok lets add it and see what we have after

```ruby
ActiveAdmin.register Blog do 
   includes :posts
   
    index do
       column :name
       column :description
       column :created_at 
    end
    
end
```
ok so now we have faster batch destroy, but what about index ???  Index action uses includes applier also. We are loading posts every time on index page because index page uses ```find_collection``` without arguments.
 (there can be hundreds or thousands of posts) but why we are doing it ?  

Your example is not generic, to fix it we just need to override default destroy batch action, how it is already described in documentation, in other cases we bring optimisation to one place  , but remove it in another one.

Fivell:
To clarify, ActiveAdmin DSL doesn't allow us to add ```includes``` applier for batch_action and not use it for simple index rendering

timoschilling:
Ok you are right in the Blog example `includes` don't make sense for the index action.

But let us do this:
```ruby
class ActiveAdmin::BatchActions::Controller
  COLLECTION_APPLIES = [:authorization_scope, :collection_decorator, :filtering, :scoping, :includes]

  def batch_action_collection
    find_collection(only: COLLECTION_APPLIES)
  end
end
```

Fivell:
@timoschilling , ok let's see what travis will tell

Fivell:
@timoschilling , you're also hinting about docs,  how do you think I can improve it?

timoschilling:
:+1: for
```ruby
      def batch_action_collection(only = COLLECTION_APPLIES)
        find_collection(only: only)
      end
```

Fivell:
@timoschilling , what next?

timoschilling:
Sorry I missed your last commit. I think we can merge or miss I something that you want to do?

Fivell:
@timoschilling , I think it can be merged, or I should rebase and squash commits?

timoschilling:
good work

Issue #3781, opened 2015-02-05
Travis is failing, because of a dependency change


timoschilling:
The Travis build is failing, because of a dependency change.

76f194ebe33303fb12dd0dbecb069e27fa8a8de9 02.02.2015 [runs](https://travis-ci.org/activeadmin/activeadmin/builds/49184873)
41043fa9aecc47f04eefc6322eea89a7bd4b0876 05.02.2015 [fails](https://travis-ci.org/activeadmin/activeadmin/jobs/49634917)

The only [change](https://github.com/activeadmin/activeadmin/compare/76f194ebe33303fb12dd0dbecb069e27fa8a8de9...41043fa9aecc47f04eefc6322eea89a7bd4b0876) between this both commits is a typo fix in a markdown file.

I think it happens because of a dependency change.

strivedi183:
@timoschilling I think I found the dependency change, rspec-rails was bumped to version 3.2.0 on 03.02.2015

Sent a PR to fall back to rspec-rails 3.1.x and the tests pass on travis

PR #3783, opened 2015-02-05
Rework development mode code reloading

chancancode:
This commit improved the code reloading mechanisms in a few ways:

1. Avoid reloading the routes unless the admin files has changed
2. Avoid reloading the routes twice in the same request if they were already
   reloaded by Rails
3. Respect the `reload_classes_only_on_change` config

In a big application with lots of routes, reloading them would be very costly, so it's something that should be avoided if possible.


chancancode:
before

after

seanlinsley:
Thanks so much for working to improve reloading performance!

In your big application, how long did initial load take before, and how much has this shaved off?

chancancode:
Before (baseline – no files changed, so nothing is reloaded):

before-baseline

Before (routes changed – `touch config/routes.rb`)

before-routes-changed

After (baseline – no files changed, so nothing is reloaded):

after-baseline

After (routes changed – `touch config/routes.rb`)

after-routes-changed

For our purpose, we are only interested in the page generation time (which is approximated by the TTFB/time-to-first-byte). So I think we are looking at a ~30% improvement here.

chancancode:
(The build failure seems unrelated, it appears to be failing for the same reason as the latest commit on master)

timoschilling:
Good work!!!

I can confirm that there are failing test, before your change, see #3781. I hope some can fix that before we merge this, to get the cucumber tests running.

I hope, I or @seanlinsley or some else find some time to test it locale.

timoschilling:
please rebase this commit, #3781 is closed

chancancode:
Rebased!

timoschilling:
@chancancode thanks for the good work :+1: 
Maybe you can take a look onto #697 and help us there too.

PR #3790, opened 2015-02-12
Change #map and #flatten methods call to #flat_map

davydovanton:


seanlinsley:
thanks :cat:

davydovanton:
@seanlinsley thanks you too :tada:

PR #3808, opened 2015-02-21
Improving html tables and columns

stefanvignir:
Those are SCSS modifications, improving the display of tables and columns.


stefanvignir:
Woah. Hound is not happy about this. I'm postponing this until this is resolved.

Issue #3804, opened 2015-02-20
Batch action form not filtered (Possible XSS injection)


Antonin-Deniau:
When I create a basic form like this:
```ruby
batch_action :invoice, form: {
      client_name: Client.pluck(:id, :name),
      another_field: :datepicker
    } do |ids, inputs|
    # Some code ...
end
```

The function pluck fetch all client, but if I put some script tag in the name of my client like this:
```ruby
Client.pluck(:name, :id) 
=> [['Client1', 1] , ['Client2', 2]] 
=> [['<script>alert("xss")</script>', 1] , ['Client2', 2]]
```

In my html code I obtain that:
```html
<option value="2"><script>alert("xss")</script></option>
```

PR #3811, opened 2015-02-23
Fix XSS in modal

Antonin-Deniau:
When I create a basic form like this:
```ruby
batch_action :invoice, form: {
      client_name: Client.pluck(:id, :name),
      another_field: :datepicker
    } do |ids, inputs|
    # Some code ...
end
```

The function pluck fetch all client, but if I put some script tag in the name of my client like this:
```ruby
Client.pluck(:name, :id) 
=> [['Client1', 1] , ['Client2', 2]] 
=> [['<script>alert("xss")</script>', 1] , ['Client2', 2]]
```

In my html code I obtain that:
```html
<option value="2"><script>alert("xss")</script></option>
```

This patch filter html tags and fix the XSS

### DIFF
--- a/app/assets/javascripts/active_admin/lib/modal_dialog.js.coffee
+++ b/app/assets/javascripts/active_admin/lib/modal_dialog.js.coffee
@@ -20,8 +20,8 @@ ActiveAdmin.modal_dialog = (message, inputs, callback)->
       "'": '&#39;'
       "/": '&#x2F;'
 
-    escapeHtml = (string)->
-      String(string).replace /[&<>"'\/]/g, (s)->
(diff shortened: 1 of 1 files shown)


Antonin-Deniau:
I added the space before the arrows ([2c48ee0](https://github.com/Antonin-Deniau/activeadmin/commit/2c48ee05992ac2bba24a1d73c95842bfe5920087))

timoschilling:
Writing your own escape code is a bad practice. Maybe jQuery has a solution for that or we need to do that in Ruby.

timoschilling:
This is related to #3804

Issue #3100, opened 2014-04-21
has_many and check_boxes : css issue


afriqs:
Given:
```ruby
 f.has_many :practicings, heading: nil, allow_destroy: true, new_record: true do |p|
   p.input :disciplines
end
```
![select](https://cloud.githubusercontent.com/assets/318464/2754243/b3110d94-c94f-11e3-8362-4a4ad4477f79.png)

The select box is well displayed for a habtm relationship. But given:
```ruby
 f.has_many :practicings, heading: nil, allow_destroy: true, new_record: true do |p|
   p.input :disciplines, as: :check_boxes
end
```
![check_boxes](https://cloud.githubusercontent.com/assets/318464/2754256/efaa3a3c-c94f-11e3-823b-f8ee0e61cfdb.png)

The check boxes are stuck to the left.

Note : Rails 4.0.4 and ActiveAdmin 1.0.0.pre (65aae6b68e78)

pencilcheck:
+1 Same issue

timoschilling:
closed by #3812

PR #3812, opened 2015-02-24
Fixes 3100

davidb583:
See https://github.com/activeadmin/activeadmin/issues/3100

The padding set to 0 prevents block from displaying on the right of the title.

Radio and checkboxes have the problem, simply corrected by removing the padding:0 propety.

I checked in simple & nested forms, and visually with browserStack, works well from IE9, works with firefox 3.6+, chrome, safari. 

All tests pass.

I created static pages so that you can visually see the result:

http://davidb583.github.io/activeadmin-static-3100-bugfix-1/
http://davidb583.github.io/activeadmin-static-3100-bugfix-2/

### DIFF
--- a/app/assets/stylesheets/active_admin/_forms.scss
+++ b/app/assets/stylesheets/active_admin/_forms.scss
@@ -114,7 +114,6 @@ form {
         float: left;
         width: 100%;
         margin: 0;
-        padding: 0;
       }
     }
   }


timoschilling:
Before:
edit_car___test_rails_app410_and_audi-media-tv_ _working_copy__dev_ _12_changed_files_
After:
edit_car___test_rails_app410

timoschilling:
:+1:

PR #3824, opened 2015-02-27
Update and validate CSS vendor prefixes

acconrad:
These vendor prefixes were out of date and no longer valid, creating CSS warnings. This pull request removes the unnecessary prefixes, as well as corrects some shortcuts that were not valid CSS. This primarily concerns `background`; you should be using `background-color` and `background-image` when dealing with gradients, as `background` is technically not a valid CSS property to be handling just the gradients.

For reference checking, please see [CSS3Please](http://css3please.com/) for further commentary as to the validity of the changes in this pull request.

### DIFF
--- a/app/assets/stylesheets/active_admin/mixins/_shadows.scss
+++ b/app/assets/stylesheets/active_admin/mixins/_shadows.scss
@@ -1,22 +1,18 @@
 @mixin shadow($x: 0, $y: 1px, $blur: 2px, $color: rgba(0,0,0,0.37)) {
   box-shadow: $x $y $blur $color;
-  -moz-box-shadow: $x $y $blur $color;
   -webkit-box-shadow: $x $y $blur $color;
 }
 
 @mixin no-shadow {
   box-shadow: none;
-  -moz-box-shadow: none;
   -webkit-box-shadow: none;
 }
 
 @mixin inset-shadow($x: 0, $y: 1px, $blur: 2px, $color: #aaa) {
   box-shadow: inset $x $y $blur $color;
-  -moz-box-shadow: inset $x $y $blur $color;
   -webkit-box-shadow: inset $x $y $blur $color;
 }
 
 @mixin text-shadow($color: #fff, $x: 0, $y: 1px, $blur: 0) {
   text-shadow: $color $x $y $blur;
 }
-
--- a/app/assets/stylesheets/active_admin/mixins/_shadows.scss
+++ b/app/assets/stylesheets/active_admin/mixins/_shadows.scss
@@ -1,16 +1,13 @@
 @mixin shadow($x: 0, $y: 1px, $blur: 2px, $color: rgba(0,0,0,0.37)) {
   box-shadow: $x $y $blur $color;
-  -webkit-box-shadow: $x $y $blur $color;
 }
 
 @mixin no-shadow {
   box-shadow: none;
-  -webkit-box-shadow: none;
 }
 
 @mixin inset-shadow($x: 0, $y: 1px, $blur: 2px, $color: #aaa) {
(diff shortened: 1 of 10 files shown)


app/assets/stylesheets/active_admin/mixins/_shadows.scss
```diff
@@ -1,22 +1,18 @@
...
 @mixin inset-shadow($x: 0, $y: 1px, $blur: 2px, $color: #aaa) {
   box-shadow: inset $x $y $blur $color;
-  -moz-box-shadow: inset $x $y $blur $color;
   -webkit-box-shadow: inset $x $y $blur $color;
 }
 
 @mixin text-shadow($color: #fff, $x: 0, $y: 1px, $blur: 0) {
   text-shadow: $color $x $y $blur;
 }
-
```
timoschilling: keep the last line

timoschilling:
Please remove IE 6, 7 and 8 hacks too

timoschilling:
Wich Browser Versions are affected from your change?

app/assets/stylesheets/active_admin/mixins/_shadows.scss
acconrad: hmmm it was only supposed to cut the extra line, there should be an extra line after this...line 19 should still be there

timoschilling: never mind if this is the case, viewing the last line online sucks sometimes

acconrad:
The hacks have been removed. Technically no browsers should be affected, as these were deprecated vendor prefixes.

timoschilling:
I mean which old versions we exclude now.

acconrad:
Well so via this change, IE6-8 will not have background gradients, but everything else should work as expected. IE9+, and latest Opera, Firefox, Safari and Chrome should operate fine (this includes the mobile versions of Safari and Chrome). Not sure if this affects any of the proprietary Android browser offshoots that utiliize WebKit, but they should be fine too.

timoschilling:
I have expect a answer like: FF 10+, S 7+. But forget the question, since the most one have autoupdate.

timoschilling:
Why you don't remove this:
-webkit-linear-gradient
-webkit-box-shadow

acconrad:
`-webkit-box-shadow` is for Android 2.3+, iOS 4.0.2-4.2, Safari 3-4

`-webkit-linear-gradient` is for Chrome 10-25, iOS 5+, Safari 5.1+

acconrad:
So in other words, these changes mean your officially supported browsers are:

IE 10+ (8+ if you don't include the linear gradients)
Safari 3+
Firefox (all)
Opera (all)
Chrome 10+
Mobile Safari 3+
Android 2.3+

timoschilling:
This page looks a bit more up to date:
http://caniuse.com/#search=box-shadow
http://caniuse.com/#search=linear-gradient

I think we can rip off them all

timoschilling:
We have some of this:
-webkit-font-smoothing: antialiased;
What do you think about removing that?

acconrad:
Okay just realize that by dropping the vendor prefixes for box shadow and linear gradient (esp the gradient one), you're throwing out Safari 6 and Android 4.3, which are pretty modern. 

[This post](http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/) tends to believe that font smoothing is a bad thing. For now I will remove all of the prefixes.

timoschilling:
I think you miss some `box-shadow` in `_utilities.scss`

timoschilling:
Safari 6 is 4 1/2 Year old, I think it's ok to throwing it out.
Android 4.3 is not the newest too, but sure it's still in use, but I think a missing shadow is no problem.

Issue #3807, opened 2015-02-21
Option to include boilerplate code in new resource generation


dpmccabe:
When you create a new resource, the default capabilities for that resource (index columns, filters, form fields, etc.) are all of the ones for that model. This makes perfect sense: You get everything unless you specify otherwise. This is accomplished by the following file:

    ActiveAdmin.register Partner do
    end

Now what if I just want to remove a single column from the index table, show page, filters, and form? Well, great, now I have to write all of those blocks from scratch because the generator gave me nothing to start with. This is a _lot_ of tedious work if there are many columns in the database table. I've used AA on a dozen projects and I've wasted so much time copying boilerplate code from other resource files into one I just created, manually entering index table columns, specifying filters, etc., just to make a slightly less full version of the default resource.

A new resource includes all possible information and features by default. A new resource _file_ includes no information and features. This makes no sense. Is AA's philosophy to provide an admin tool that lets developers start from everything or start from nothing? It can't be both simultaneously.

I understand that the point of the blank resource file is to ensure the actual pages work regardless of what changes in the underlying model. But who actually uses blank resource files? I've never done it and any production app is going to have some customization in every resource file.

The generator should at least provide a sensible default template that includes boilerplate code using the model's current columns for the index columns, filters, form fields, and show fields.

ccallebs:
@dpmccabe I like this idea, but the question remains what should be included and what should be omitted?

I suppose *everything* could be generated (and just commented out, with documentation/explanations). Deleting code is certainly easier than writing it.

dpmccabe:
I'm thinking that whatever you see when you load the actual pages should be what appears in the code. This might not be the default behavior, but rather something available in an alternate generator.

ccallebs:
@dpmccabe I think I get what you mean, but there's a *lot* that we could show. For example, the `controller do` block could be populated with each of the actions. That seems like overkill.

I'd propose each of the pages (index, show, form), the filters, and the action visibility.

timoschilling:
closed by #3825

PR #3825, opened 2015-02-28
Optional Boilerplate Generation

ccallebs:
Proof of concept branch for #3807 

@timoschilling Is this something we want to pursue? Will output the following (as-is, I imagine we would want to change the copy):

``` RUBY
ActiveAdmin.register Post do

# Available Filters
#
# filter :id
# filter :title
# filter :body
# filter :published_at
# filter :author
# filter :position
# filter :custom_category
# filter :starred
# filter :foo
# filter :created_at
# filter :updated_at

# Sample Index
#
# index do
#   selectable_column
#   id_column
#   column :id
#   column :title
#   column :body
#   column :published_at
#   column :author
#   column :position
#   column :custom_category
#   column :starred
#   column :foo
#   column :created_at
#   column :updated_at
#   actions
# end
end
```

Of course, this wouldn't be enabled by default. The syntax could be something like `rails g active_admin:resource --include-boilerplate`

### DIFF
--- a/lib/generators/active_admin/resource/templates/admin.rb
+++ b/lib/generators/active_admin/resource/templates/admin.rb
@@ -1,17 +1,26 @@
 ActiveAdmin.register <%= class_name %> do
 
-# Available Filters
-#<% class_name.constantize.new.attributes.keys.each do |attr| %>
-# filter :<%= attr.gsub(/_id$/, '') %><% end %>
+# Limit actions available to your users by adding them to the 'except' array
+# actions :all, except: []
 
-# Sample Index
-#
+# Add or remove filters (you can use any ActiveRecord scope) to toggle their
+# visibility in the sidebar
+<%= @formatter.filters %>
+
+# Add or remove columns to toggle their visiblity in the index action
 # index do
 #   selectable_column
-#   id_column<% class_name.constantize.new.attributes.keys.each do |attr| %>
-#   column :<%= attr.gsub(/_id$/, '') %><% end %>
+#   id_column
+<%= @formatter.columns %>
 #   actions
 # end
+
+# Add or remove rows to toggle their visiblity in the show action
+# show do |<%= class_name.downcase %>|
+<%= @formatter.rows %>
+# end
+
+# Add or remove fields to toggle their visibility in the form
 
 <% if Rails::VERSION::MAJOR == 4 || defined?(ActionController::StrongParameters) %>
   # See permitted parameters documentation:
@@ -27,6 +36,4 @@ ActiveAdmin.register <%= class_name %> do
   #   permitted
   # end
 <% end %>
-
-
 end
--- a/lib/generators/active_admin/resource/templates/admin.rb
+++ b/lib/generators/active_admin/resource/templates/admin.rb
@@ -1,5 +1,18 @@
 ActiveAdmin.register <%= class_name %> do
 
+# Available Filters
+#<% class_name.constantize.new.attributes.keys.each do |attr| %>
+# filter :<%= attr.gsub(/_id$/, '') %><% end %>
+
+# Sample Index
+#
+# index do
+#   selectable_column
+#   id_column<% class_name.constantize.new.attributes.keys.each do |attr| %>
+#   column :<%= attr.gsub(/_id$/, '') %><% end %>
+#   actions
+# end
+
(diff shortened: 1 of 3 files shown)


lib/generators/active_admin/resource/templates/admin.rb
```diff
@@ -1,5 +1,18 @@
 ActiveAdmin.register <%= class_name %> do
 
+# Available Filters
+#<% class_name.constantize.new.attributes.keys.each do |attr| %>
+# filter :<%= attr.gsub(/_id$/, '') %><% end %>
```
timoschilling: 1. There should be a way without initialize a instance to get the attributes
2. The code style can be nicer:
```ruby
<% class_name.constantize.new.attributes.keys.each do |attr| -%>
# filter :<%= attr.gsub(/_id$/, '') %>
<% end -%>
```

lib/generators/active_admin/resource/templates/admin.rb
```diff
@@ -1,5 +1,18 @@
...
+
+# Sample Index
+#
+# index do
+#   selectable_column
+#   id_column<% class_name.constantize.new.attributes.keys.each do |attr| %>
+#   column :<%= attr.gsub(/_id$/, '') %><% end %>
```
timoschilling: same like in line 5

lib/generators/active_admin/resource/templates/admin.rb
```diff
@@ -15,4 +28,5 @@
   # end
 <% end %>
```
timoschilling: This part should be on top of the file

timoschilling:
I am not averse to this.

Maybe we can provide the same for scopes, show and form.

And it will be cool to something like:
```
rails g active_admin:resource --add-boilerplate
# or
rails g active_admin:resource --update-boilerplate
```

lib/generators/active_admin/resource/templates/admin.rb
ccallebs: @timoschilling I'll probably extract the functionality out into a common class. I didn't know about `-%>`, thanks!

ccallebs:
@timoschilling The plan was to add boilerplate for those things, as well as the visibility actions. I just couldn't find a quick way to nab the model associations and wanted to show a proof of concept to get approved before I went further. I'll work on this some more this weekend and ping you again when it's ready to be reviewed.

timoschilling:
Go for it I like the idea! :+1:

jancel:
+15 ?! I like.

On Sun, Mar 1, 2015 at 3:58 PM, Timo Schilling <notifications@github.com>
wrote:

> Go for it I like the idea! [image: :+1:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/activeadmin/activeadmin/pull/3825#issuecomment-76633549>
> .
>



-- 
Jeff Ancel

(314) 703-8829 - Main

www.jeffancel.com

ccallebs:
@timoschilling @seanlinsley I have the filters/show/index/actions complete, as well as the flag to turn boilerplate generation on. Any suggestions on how to generate the form? With various model associations, I don't see that being a simple task.

I could keep it simple and only provide boilerplate for top-level attributes, but it'd be useful to include everything.

javierjulio:
This is excellent @ccallebs thanks! Personally, I say make this the default. Is there any reason we don't want too?

jancel:
Thumbs up for default!

On Monday, April 13, 2015, Javier Julio <notifications@github.com> wrote:

> This is excellent @ccallebs <https://github.com/ccallebs> thanks!
> Personally, I say make this the default. Is there any reason we don't want
> too?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/activeadmin/activeadmin/pull/3825#issuecomment-92518996>
> .
>


-- 
Jeff Ancel

(314) 703-8829 - Main

www.jeffancel.com

timoschilling:
:-1: for default
but :+1: for a new expect option for `filter` and `index` or `columns`

seanlinsley:
This should definitely be added to the changelog, and the getting started documentation.

javierjulio:
The boilerplate will no longer be behind a flag and is now the default. The only requirement is providing a resource name. The template will be uncommented so it directs users to configure their resource page and not rely on the defaults-rendering.

PR #3823, opened 2015-03-05
Thread-unsafe use of url_helpers

hasghari:
There are two files that are thread unsafe by virtue of using `Rails.application.routes.url_helpers` in threaded code: 

1. [lib/active_admin/devise.rb:39](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/devise.rb#L39)
2. [lib/active_admin/resource/routes:110](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/resource/routes.rb#L110)

This is actually due to a bug in MRI (https://github.com/puma/puma/issues/647)


timoschilling:
:+1: Great work, like to see that someone use the Github Issue to PR function!

Issue #3838, opened 2015-03-06
Rails 4.2 still complains about deprecated ActiveRecord #column_for_attribute


rdj:
Active Admin produces a deprecation warning with Rails 4.2.1.rc3:

```DEPRECATION WARNING: `#column_for_attribute` will return a null object for non-existent columns in Rails 5. Use `#has_attribute?` if you need to check for an attribute's existence. (called from is_boolean? at /var/jenkins/.rvm/gems/ruby-2.2.1@project/bundler/gems/active_admin-96c6db140e57/lib/active_admin/views/components/table_for.rb:120)```

This was previously reported in #3487 (fixed c8e54d2) and then amended in a way that reintroduced the deprecation warning (48ccc3db). 

That last commit provides justification for a why the method should be used. The problem remains, however, that anybody using Active Admin will deal with the noise of the deprecation warning. 

Rails provides a way to silence deprecation warnings on a particular section of code when you've decided you know better. Now is the time to use it.

rdj:
Ok, since the PR has been closed, can we review what to actually do here? The commit history I outlined here shows that the aa team seems to have considered and rejected actually taking action to remove the deprecated action. While I agree that silencing deprecation warnings is a bad idea in general, I think that as a library you expect other people to use in their own rails apps, ignoring the deprecation warnings is even worse. As a library, deciding this actively determining to waste uncountable people's time in the future, as each and every rails app that uses active admin investigates this deprecation warning, deal with their own internal policies about how their apps are not allowed to have deprecation warnings, and has to decide whether to drop activeadmin or create a private fork that silences it, etc. 

So if we could revisit the previous commits here, can we agree that it would be preferable to have a fix that works correctly and does not emit the deprecation warning? 

@gonzedge - you contributed 48ccc3d — I have read the commit message, but I don't completely follow it. Possibly multiple rails versions are involved here, but you didn't specify.

Also, I get tons of failures on my MacBook running 10.10.2 when I try to run the test suite. I see Travis says the build is good? I looked at the contribution guidelines and it doesn't make it sound like running the tests is that hard; should I be able to get them working on my machine or do they only run on Linux or some other caveat I didn't see?

rdj:
Oh, yeah, pull in @timoschilling since he closed the PR. Really would like to figure out something positive to do here rather than just ignoring the problem.

seanlinsley:
[...]

PR #3839, opened 2015-03-06
Silence deprecation warning for correctly-used Rails4.2 #column_for_attribute

rdj:
#3838


timoschilling:
Sorry, but suppressing a Deprecation warning is a bad practice.

Issue #3777, opened 2015-01-30
Labels suddenly stopped working on login screen


aaronsakowski:
I can't figure out why on a recent deploy my admin login screen suddenly started looking like this. Can someone point me in the right direction? Rails 4.0 and 4.1 (tried with both) and ActiveAdmin 1.0.0.pre.

![login neighborly](https://cloud.githubusercontent.com/assets/963659/5969841/18fe60da-a7de-11e4-929a-162f5ec3649c.png)

timoschilling:
please show the markup of one of the input label's and of the button.

aaronsakowski:
Sure. Here's the markup of the main elements of the form. The labels are being rendered like so: ```&lt;p&gt;active_admin.devise.email.title&lt;/p&gt;```

    <form accept-charset="UTF-8" action="/admin/login" class="formtastic admin_user" id="session_new" method="post" novalidate="novalidate"><div style="display:none"><input name="utf8" type="hidden" value="&#x2713;" /><input name="authenticity_token" type="hidden" value="cnfXPtoClHe/MriL2Z5CMZrYuIvxtXjc8wegdTodDLE=" /></div><fieldset class="inputs"><ol><li class="email input required autofocus stringish" id="admin_user_email_input"><label class="label" for="admin_user_email">&lt;p&gt;active_admin.devise.email.title&lt;/p&gt;<abbr title="required">*</abbr></label><input autofocus="autofocus" id="admin_user_email" maxlength="255" name="admin_user[email]" type="email" value="" />

    </li><li class="password input required stringish" id="admin_user_password_input"><label class="label" for="admin_user_password">&lt;p&gt;active_admin.devise.password.title&lt;/p&gt;<abbr title="required">*</abbr></label><input id="admin_user_password" maxlength="128" name="admin_user[password]" type="password" />

    </li><li class="boolean input optional" id="admin_user_remember_me_input"><input name="admin_user[remember_me]" type="hidden" value="0" /><label class="" for="admin_user_remember_me"><input id="admin_user_remember_me" name="admin_user[remember_me]" type="checkbox" value="1" />&lt;p&gt;active_admin.devise.login.remember_me&lt;/p&gt;</label>

    </li></ol></fieldset><fieldset class="actions"><ol><li class="action input_action " id="admin_user_submit_action"><input name="commit" type="submit" value="&lt;p&gt;active_admin.devise.login.submit&lt;/p&gt;" /></li></ol></fieldset></form>

  
[...]

Issue #3061, opened 2014-04-09
ViewsHelpers methods not available in ActiveAdmin with include_all_helpers = false


aprescott:
In a project with

``` ruby
config.action_controller.include_all_helpers = false
```

set in `config/application.rb` (different from the default value of `true`), view helpers are not global across all views. This can be desired behaviour to keep things scoped.

However, with this setting, ActiveAdmin can no longer see methods defined in `ActiveAdmin::ViewsHelper`.

``` ruby
# dummy example
module ActiveAdmin::ViewsHelper
  def final_cost(item)
    item.on_sale? ? 10 : 50
  end
end
```

``` ruby
ActiveAdmin.register Item do
  # ...

  index do
    # ...

    column :cost do |item|
      final_cost(item) # not available with include_all_helpers = false
    end

    # ...
  end
end
```

This fails with

```
ActionView::Template::Error (undefined method `final_cost' for #<ActiveAdmin::Views::IndexAsTable:0x00000008ab47c8>):
```

This seems like a bug, since helpers that are defined for ActiveAdmin should be available to ActiveAdmin even if the config setting is set to `false`.

The only work-around I found was to define the helper "inline" with the `register` content:

``` ruby
ActiveAdmin.register Item do
  # ...

  controller do
    def final_cost(item)
      item.on_sale? ? 10 : 50
    end
    helper_method :final_cost
  end

  index do
    # ...

    column :cost do |item|
      final_cost(item)
    end

    # ...
  end
end
```

Obviously this isn't great for organising helper methods, and doesn't allow ActiveAdmin-global helpers.

reicheltd:
@jhenkens your approach works for me in 1.0.0.pre2. thx

PR #3840, opened 2015-03-07
Revert "remove kaminari theme"

killthekitten:
Reverts activeadmin/activeadmin#3653

The reason is #1126

### DIFF
--- a/app/views/kaminari/active_admin/_first_page.html.erb
+++ b/app/views/kaminari/active_admin/_first_page.html.erb
@@ -1,0 +1,11 @@
+<%# Link to the "First" page
+  - available local variables
+    url:           url to the first page
+    current_page:  a page object for the currently displayed page
+    total_pages:   total number of pages
+    per_page:      number of items to fetch per page
+    remote:        data-remote
+-%>
+<span class="first">
+  <%= link_to_unless current_page.first?, raw(t 'views.pagination.first'), url, remote: remote %>
+</span>
--- a/app/views/kaminari/active_admin/_gap.html.erb
+++ b/app/views/kaminari/active_admin/_gap.html.erb
@@ -1,0 +1,8 @@
+<%# Non-link tag that stands for skipped pages...
+  - available local variables
+    current_page:  a page object for the currently displayed page
+    total_pages:   total number of pages
+    per_page:      number of items to fetch per page
+    remote:        data-remote
+-%>
+<span class="page gap"><%= raw(t 'views.pagination.truncate') %></span>
--- a/app/views/kaminari/active_admin/_last_page.html.erb
+++ b/app/views/kaminari/active_admin/_last_page.html.erb
@@ -1,0 +1,11 @@
+<%# Link to the "Last" page
+  - available local variables
+    url:           url to the last page
+    current_page:  a page object for the currently displayed page
+    total_pages:   total number of pages
+    per_page:      number of items to fetch per page
+    remote:        data-remote
+-%>
+<span class="last">
+  <%= link_to_unless current_page.last?, raw(t 'views.pagination.last'), url, remote: remote %>
+</span>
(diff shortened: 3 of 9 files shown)


dmitry:
:+1: good point.

killthekitten:
@timoschilling any thoughts?

timoschilling:
I understand why it was in, but I still think it should be keep out. The reason is why should we carry a copy of the template with us, if there is a template in kaminari. I know it can be happen that the app overwrites this template, but then the developer can simply use this template as a theme.

Some equal examples:
- If you turn of `include_all_helpers`, you need to include the devise helper by your self (#3061)
- If you overwrite a bourbon mixin, you break the activeadmin css (is happened to me)
- If you use redcloth it breaks the devise views (#3777)
All cases can be fixed by activeadmin, but if we care about all thinks we have a massive code base which is hard to maintain.

Thats why my opinion ist *keep it simple as it can*.

Maybe @seanlinsley has a other opinion, then we can talk about it.

killthekitten:
@timoschilling that's one of the reasons, why `ActiveAdmin` development is abandoned by community. There always comes a moment, when it's much simpler to bootstrap an admin interface from scratch than to deal with inexplicable behaviors of something so called `Ruby on Rails framework for creating elegant backends for website administration`.

Tool should be predictable and respect the real world use cases. Look around, we're using `ruby` and `rails` here. 

That's about pretty much everything:
https://github.com/activeadmin/activeadmin/issues/3846

timoschilling:
That you link to #3846 shows your problem, you want a autonomic full featured drop in framework. That's not possible without self building all components. Thats why we use third party gems in activeadmin. If you use components you need to deal with there behavior. You can't start fixing the bug of component A in component B, thats what you want in #3846. Nearly the same here, your app breaks the default kaminari theme, so you need to fix it in your scope, for example by naming your theme and not overwriting it.

jhanggi:
I've just run into this same issue. The front end of our site has a custom template built by our designers. Now the pagination styling in ActiveAdmin is busted.

Kaminari does have built-in templates, which I understand the desire to use them so you don't need to maintain your own. 

What would you think about exposing the Kaminari theme name as a configuration option? That way, you don't have to maintain your own versions, and those of us who do need to override it for the front-end have a way to keep a sane default. Or is there a way to force AA to use the gem's version of the templates?

Thanks.

timoschilling:
I have already tried to force the loading of the gem template version. But actually that is not supported by kaminari and/or ActionView.

I have tried this:
```ruby
Kaminari::Engine.paths["app/views"].first
# => "/Users/captainhagbard/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/kaminari-0.16.3/app/views"
```

javierjulio:
@timoschilling the Kaminari maintainer (amatsuda) seems very active on accepting pull requests. Should we file an issue? Or have something we could submit? I can try to help.

killthekitten:
@javierjulio you can try to add an option of using gem's template without adding it to project

timoschilling:
@javierjulio sure, it would be nice if kaminari has a option for that.

javierjulio:
@timoschilling @killthekitten this must be [the pull request](https://github.com/amatsuda/kaminari/pull/668) we need to resolve this issue! Adds a path prefix option for the Kaminari views generator. Since I'm doing the CSS overhaul I'd like to modify the ActiveAdmin pagination setup or would we prefer to just keep the default?

whatcould:
Not sure why we couldn't just leave the paginatinon templates in! Almost two years later and there's still no way to override the pagination in my app without breaking the ActiveAdmin pagination, short of specifying it every time.

I get arguments about maintainability, but if that's what you're worried about, first of all, these templates don't really change. But also, it's easier to maintain the styles if you freeze the HTML by copying the templates!

PR #3841, opened 2015-03-07
Allow max line length of 120 symbols

killthekitten:
It's a modern value, since most of us are using modern laptops with wide screens, and it seems to be a standard of the active admin's codebase.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -9,3 +9,6 @@ java_script:
 
 scss:
   enabled: true
+
+Metrics/LineLength:
+  Max: 120


timoschilling:
80 chars is not only about the screen size, it's about the readability too.
Not the chars per line are important, the informations per line are important.

killthekitten:
@timoschilling I agree, but it makes development process more complex now

timoschilling:
I think 80 chars are easier to read and you doing something wrong if you need more then 80 chars per line. It's the same as a method should not more then X lines or a class should not more then X lines.

Do you know main projects thats have a line limit over 80?

killthekitten:
@timoschilling I do not insist on it, but I think main projects are too inert for changes like that. For example RubyMine IDE has a default limit of 120 symbols. And here is  a screenshot of ordinary ActiveAdmin spec:

timoschilling:
Some other opinions?

timoschilling:
I'm closing this while no one else has requested that.

Issue #3847, opened 2015-03-09
Index option `pagination_total: false` does not actually prevent the `count(*)`


losvedir:
The [documentation](http://activeadmin.info/docs/3-index-pages.html#index-pagination) says:

> If you have a very large database, you might want to disable SELECT COUNT(*) queries caused by the pagination info at the bottom of the page:
> ``` ruby
> ActiveAdmin.register Post do
>   index pagination_total: false do
>     # ...
>   end
> end
> ```

However, while that option changes the _display_ of the pagination, it does **not** actually prevent the `COUNT (*) from "table"` from occurring. This one line was causing us sometimes up to a second of page load time for some of our large tables.

I instrumented it to be notified when a `count (*)` was issued, via:
``` ruby
ActiveSupport::Notifications.subscribe("sql.active_record") do |_, _, _, _, details|
  if details[:sql] =~ /SELECT COUNT\(\*\) FROM "messages"/
    puts caller.join("\n")
    puts "*" * 50
  end
end
```

Then I loaded up our `/admin/messages` table, and sure enough, the count happened:
```
   (150.0ms)  SELECT COUNT(*) FROM "messages"
.../.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/activesupport-3.2.21/lib/active_support/notifications/fanout.rb:47:in `call'
.../.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/activesupport-3.2.21/lib/active_support/notifications/fanout.rb:47:in `publish'
.../.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/activesupport-3.2.21/lib/active_support/notifications/fanout.rb:25:in `block in publish'
.../.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/activesupport-3.2.21/lib/active_support/notifications/fanout.rb:25:in `each'
.../.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/activesupport-3.2.21/lib/active_support/notifications/fanout.rb:25:in `publish'
.../.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/activesupport-3.2.21/lib/active_support/notifications/instrumenter.rb:25:in `ensure in instrument'
.../.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/activesupport-3.2.21/lib/active_support/notifications/instrumenter.rb:25:in `instrument'
[...]

Issue #2638, opened 2013-10-31
We still aren't able to prevent catch-all COUNT queries for index pages


seanlinsley:
As @Fivell [pointed out](https://github.com/gregbell/active_admin/pull/2333#issuecomment-27518116), even after #2333 we're still hitting the database to get a count of all records, which is expensive if you have millions of records. This time around, it's an issue with Kaminari itself.

Specifically, we call [`paginate`](https://github.com/amatsuda/kaminari/blob/d5757227d7fafacefeab3d585f41c11933bd46db/lib/kaminari/helpers/action_view_extension.rb#L17) which ends up calling [`total_pages`](https://github.com/amatsuda/kaminari/blob/d5757227d7fafacefeab3d585f41c11933bd46db/lib/kaminari/models/page_scope_methods.rb#L21).

This has apparently been a [long-standing issue](https://github.com/amatsuda/kaminari/issues/240) with Kaminari, and some people have [suggested fixes](https://github.com/vhochstein/kaminari/commit/9db1e616c84327562173a9b392695e0b39011723), but no one has yet submitted an actual pull request.

Another reference: http://engineering.nulogy.com/posts/avoiding-redundant-counting/

seku:
confirm `pagination_total: false` doesn't do the job. Still `count` queries are performed.
AA 1.0.0pre

Soleone:
just noticed the same as @seku. we just updated to rails 4.2 using AA 1.0.0pre and getting database errors on production when it's trying to do a COUNT query:

```
ActiveRecord::StatementInvalid: Mysql2::Error: Lost connection to MySQL server during query: SELECT COUNT(*) FROM `my_table`

[PROJECT_ROOT]/vendor/bundle/ruby/2.1.0/gems/activerecord-4.2.0/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:299 → query
[PROJECT_ROOT]/vendor/bundle/ruby/2.1.0/gems/activerecord-4.2.0/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:299 → block in execute
[PROJECT_ROOT]/vendor/bundle/ruby/2.1.0/gems/activerecord-4.2.0/lib/active_record/connection_adapters/abstract_adapter.rb:466 → block in log
[PROJECT_ROOT]/vendor/bundle/ruby/2.1.0/gems/activesupport-4.2.0/lib/active_support/notifications/instrumenter.rb:20 → instrument
[...]

PR #3848, opened 2015-03-09
Fix pagination_total index option to prevent SELECT count(*) queries

losvedir:
Addresses https://github.com/activeadmin/activeadmin/issues/3847.

The `pagination_total` option is supposed to prevent expensive `SELECT count(*)` queries from being performed when viewing a resource's index page.

However, while this option was hiding the display of the total number of pages, it did not actually prevent the query from being performed.

There were two reasons:

First, in `#build_pagination`, the line: `text_node paginate collection, options` would trigger Kaminari to call `#total_pages` on the collection, which would trigger the query. Excerpt of kaminari code:

``` ruby
def paginate(scope, options = {}, &block)
  options[:total_pages] ||= options[:num_pages] || scope.total_pages
```

By passing in the `:total_pages` option, we short circuit this assignment and avoid the query.

The downside is we do so by hardcoding in an essentially random total pages count of 100. I chose the number 100, because I wanted to ensure ellipses would appear in Kaminari's "Page 1, 2, 3, ..., Next, Last" links, no matter how many links are configured to be displayed. However, if the resource only had 2 pages worth of items, then clicking the "3" would result in an empty page. That said, without querying the resource, there's no way to know how many pages there should be.

Second, in `#page_entries_info`, the call `collection.num_pages` also triggered a `SELECT count(*)`. I re-arranged the conditionals to avoid that call if `@display_total` is false, while trying to keep the behavior the same.

### DIFF
--- a/lib/active_admin/views/components/paginated_collection.rb
+++ b/lib/active_admin/views/components/paginated_collection.rb
@@ -97,6 +97,17 @@ module ActiveAdmin
         options = {}
         options[:param_name] = @param_name if @param_name
 
+        if !@display_total
+          # The #paginate method in kaminari will query the resource with a
+          # count(*) to determine how many pages there should be unless
+          # you pass in the :total_pages option. We issue a query to determine
+          # if there is another page or not, but the limit/offset make this
+          # query fast.
+          offset = collection.offset(collection.current_page * @per_page.to_i).limit(1).count
+          options[:total_pages] = collection.current_page + offset
+          options[:right] = 0
+        end
+
         text_node paginate collection, options
       end
 
@@ -117,22 +128,30 @@ module ActiveAdmin
           entries_name = I18n.translate key, count: collection.size, default: entry_name.pluralize
         end
 
-        if collection.num_pages < 2
-          case collection_size
-          when 0; I18n.t('active_admin.pagination.empty',    model: entries_name)
-          when 1; I18n.t('active_admin.pagination.one',      model: entry_name)
-          else;   I18n.t('active_admin.pagination.one_page', model: entries_name, n: collection.total_count)
+        if @display_total
+          if collection.num_pages < 2
+            case collection_size
+            when 0; I18n.t("active_admin.pagination.empty",    model: entries_name)
+            when 1; I18n.t("active_admin.pagination.one",      model: entry_name)
+            else;   I18n.t("active_admin.pagination.one_page", model: entries_name, n: collection.total_count)
+            end
+          else
+            offset = (collection.current_page - 1) * collection.limit_value
+            total  = collection.total_count
+            I18n.t "active_admin.pagination.multiple",
+                   model: entries_name,
+                   total: total,
+                   from: offset + 1,
+                   to: offset + collection_size
           end
         else
+          # Do not display total count, in order to prevent a `SELECT count(*)`.
+          # To do so we must not call `collection.num_pages`
           offset = (collection.current_page - 1) * collection.limit_value
-          if @display_total
-            total  = collection.total_count
-            I18n.t 'active_admin.pagination.multiple', model: entries_name, total: total,
-              from: offset + 1, to: offset + collection_size
-          else
-            I18n.t 'active_admin.pagination.multiple_without_total', model: entries_name,
-              from: offset + 1, to: offset + collection_size
-          end
+          I18n.t "active_admin.pagination.multiple_without_total",
+                 model: entries_name,
+                 from: offset + 1,
+                 to: offset + collection_size
         end
       end
 
--- a/spec/unit/views/components/paginated_collection_spec.rb
+++ b/spec/unit/views/components/paginated_collection_spec.rb
@@ -198,9 +198,10 @@ describe ActiveAdmin::Views::PaginatedCollection do
       end
 
       describe "set to false" do
-        let(:pagination) { paginated_collection(collection, pagination_total: false) }
-
         it "should not show the total item counts" do
+          expect(collection).not_to receive(:num_pages)
+          expect(collection).not_to receive(:total_pages)
+          pagination = paginated_collection(collection, pagination_total: false)
           info = pagination.find_by_class('pagination_information').first.content.gsub('&nbsp;',' ')
           expect(info).to eq "Displaying posts <b>1 - 30</b>"
         end
--- a/features/index/pagination.feature
+++ b/features/index/pagination.feature
@@ -43,17 +43,21 @@ Feature: Index Pagination
     When I am on the index page for posts
     Then I should not see pagination
 
-    Scenario: Viewing index with pagination_total set to false
-      Given an index configuration of:
-      """
-        ActiveAdmin.register Post do
-          index :pagination_total => false do
-          end
+  Scenario: Viewing index with pagination_total set to false
+    Given an index configuration of:
+    """
+      ActiveAdmin.register Post do
+        config.per_page = 10
+        index :pagination_total => false do
         end
-      """
-      Given 100 posts exist
-      When I am on the index page for posts
-      Then I should see pagination with 4 pages
-      Then I should see "Displaying Posts 1 - 30"
-      And I should not see "Displaying Posts 1 - 30 of 100 in total"
+      end
+    """
+    Given 11 posts exist
+    When I am on the index page for posts
+    Then I should see "Displaying Posts 1 - 10"
+    And I should not see "11 in total"
+    And I should see the pagination "Next" link
 
+    When I follow "Next"
+    Then I should see "Displaying Posts 11 - 11"
+    And I should not see the pagination "Next" link
--- a/features/step_definitions/pagination_steps.rb
+++ b/features/step_definitions/pagination_steps.rb
@@ -5,3 +5,11 @@ end
 Then /^I should see pagination with (\d+) pages$/ do |count|
   expect(page).to have_css '.pagination span.page', count: count
 end
+
+Then /^I should see the pagination "Next" link/ do
+  expect(page).to have_css "a", text: "Next"
+end
+
+Then /^I should not see the pagination "Next" link/ do
+  expect(page).to_not have_css "a", text: "Next"
+end


losvedir:
Do you want me to modify the line lengths? They're mostly from code I just moved around, and I didn't want to be too invasive. My comments are longer than 80 characters since I didn't realize the length restriction. I can change the line breaks if you want.

losvedir:
Okay, I changed my comments to be within 80 characters, and changed single quotes to double.

timoschilling:
note: related to #2638

> The downside is we do so by hardcoding in an essentially random total pages count of 100. I chose the number 100, because I wanted to ensure ellipses would appear in Kaminari's "Page 1, 2, 3, ..., Next, Last" links, no matter how many links are configured to be displayed. However, if the resource only had 2 pages worth of items, then clicking the "3" would result in an empty page. That said, without querying the resource, there's no way to know how many pages there should be.

Sorry but I can't merge that, that's a hack and not a solution. In my eyes this is not acceptable.

There are some unexpected behaviors if you trie all variants of `index pagination_total: false do` and `config.paginate = false`.

BTW: In my test app I still see count queries:
```sql
SELECT COUNT(count_column) FROM (SELECT  1 AS count_column FROM `companies`  WHERE (user_id IS NOT NULL) LIMIT 10000 OFFSET 0) subquery_for_count
```

Have you read #2638? I'm not very familiar with #2638 but in my understanding it's a kaminari bug.

losvedir:
Apologies, I didn't see that other issue. I tried searching for `pagination_false` and it didn't come up.

> There are some unexpected behaviors if you trie all variants of index pagination_total: false do and config.paginate = false.

I'm sorry, I don't understand this. Can you explain a bit? And if you disable pagination then won't that blow up activeadmin if you have millions of records (the impetus for turning `pagination_total: false`?

> ``` 
> SELECT COUNT(count_column) FROM (SELECT  1 AS count_column FROM `companies`  WHERE (user_id IS NOT NULL) LIMIT 10000 OFFSET 0) subquery_for_count
> 111
> ```

Yeah, I get these, too, but they run instantaneously for me. The `LIMIT` there narrows it down to just 10,000 records (and in my configuration it's just 30). I think it's to determine how many records are being displayed on the current page, rather than the total number of records?

The query to look out for is `SELECT count(*) FROM "companies"`. Before my patch that one query can take upwards of seconds if you have millions of records.

> Sorry but I can't merge that, that's a hack and not a solution. In my eyes this is not acceptable.

Thank you for the blunt feedback. I'll see if I can find another approach.

losvedir:
Okay, I've looked into this some more, including #2638.

Contrary to what #2638 says, I'd argue that this is _not_ a bug in Kaminari, but rather a feature request. It _is_ a bug in activeadmin, because the documentation advertises this as a feature that removes the `count (*)` and has some code intending that effect, but does not actually do so. I poked around the Kaminari codebase, and unfortunately don't feel I could cleanly implement the feature in the time I'm willing to spend.

While Kaminari does not have an "infinite pages" feature, it _does_ let you specify the `total_pages` as an option, so it doesn't have to do the query. That's what my PR took advantage of, although I arbitrarily set it to `100`.

I looked at #2638 and @azach links to a clever approach called `activeadmin_fuzzy_paginate`, which I think is a more elegant solution than the way I did. If I changed my PR to use his approach would that suffice?

Let me explain the issue and the approach:

Here's a screenshot of the pagination component:

screen shot 2015-03-10 at 11 11 53 am

The activeadmin code that generates this is:
``` ruby
def build_pagination
  options = {}
  options[:param_name] = @param_name if @param_name
  text_node paginate collection, options
end
```
The pagination component can be configured via the `options` that are passed in.

There are two parts to the pagination component that are relevant here:
* The links `96, 97, 98, 99, 100`. `98` is the current page, and the quantity of links on each side is determined in Kaminari via `options[:inner_window] || Kaminari.config.window`.
* The links `9999, 1000`. These are the links starting from `total_pages` and counting backwards. The number of links is determined by `options[:right]`.

So the clever approach is still to send in `options[:total_pages]` so Kaminari doesn't calculate it (like I did in my PR). But instead of hardcoding `100` as the total number of pages, we say the total is the "current page + the window". Since by definition we can't know the total number of pages without a `count(*)` this basically implements "infinite pages" by always saying there are at least `window` pages more than whatever we're on. We also set `options[:right] = 0` so we don't display the links to the last few pages.

``` ruby
options[:total_pages] = collection.current_page + Kaminari.config.window + 1
options[:right] = 0
```

To be clear there is still one small downside: Kaminari seems to generate the "Last" link no matter what. The screenshot from my proposed fix looks like this:

screen shot 2015-03-10 at 12 36 13 pm

If you click "Last", then the pagination on that page will look the same except now `14` will be the "current page", and it will count up `15, 16, 17, 18, ...`.

So, at this point, I see two possibilities:
* I update my PR with the approach above.
* We wait for someone to implement a new "infinite pages" feature in Kaminari which does basically the above, but also removes the `Last` link. Then we pass in `options[:infinite_pages] = true` or something here. Activeadmin lives with a bug until then, and we bump the required kaminari version in the gemspec when it happens. But I don't have the time or understanding of Kaminari codebase to do this, unfortunately.

I hope this new approach is sufficiently "not hacky" so that you'll accept it, but I'll understand if not and close the PR. Please let me know.

losvedir:
Okay, since I had already worked out the code locally for the screenshots, I just pushed up my changes to the PR. Feel free to review and merge if you approve of the approach! :)

timoschilling:
First: :+1: for one of the best documented PR's

I still think it's to hacky.

But I have a idea:
What if we change this:
```ruby
options[:total_pages] = collection.current_page + Kaminari.config.window + 1
```
to this:
```ruby
next_page = collection.page(collection.current_page + 1).per(@per_page).limit(1).any?
options[:total_pages] = collection.current_page + (QUERY ? 1 : 0)
```

losvedir:
Oh, that's very interesting. I like it! Sorry I haven't had time to respond. I'll try out this approach over the weekend and update the PR.

lib/active_admin/views/components/paginated_collection.rb
```diff
@@ -117,22 +128,25 @@ def page_entries_info(options = {})
...
+            total  = collection.total_count
+            I18n.t "active_admin.pagination.multiple", model: entries_name, total: total,
               from: offset + 1, to: offset + collection_size
           end
+        else
+          # Do not display total count, in order to prevent a `SELECT count(*)`.
+          # To do so we must not call `collection.num_pages`
+          offset = (collection.current_page - 1) * collection.limit_value
+          I18n.t "active_admin.pagination.multiple_without_total", model: entries_name,
+            from: offset + 1, to: offset + collection_size
```
timoschilling: I know it's not your code style, but please change this and do the same on line 138

timoschilling:
There is a more easy version:
```ruby
offset = collection.page(collection.current_page + 1).per(@per_page).limit(1).count
options[:total_pages] = collection.current_page + offset
```

losvedir:
I looked into these approaches this weekend and unfortunately they didn't work. I put a `byebug` in the code right where those code portions go, to try them out and see what gets executed in the way of SQL:

```
(byebug) collection.page(collection.current_page + 1).per(@per_page).limit(1).any?
   (117.3ms)  SELECT COUNT(*) FROM "messages"
true
(byebug)
```
and
```
(byebug) offset = collection.page(collection.current_page + 1).per(@per_page).limit(1).count
   (108.0ms)  SELECT COUNT(*) FROM "messages"
989547
(byebug)
```

I believe it's the usage of Kaminari's `#page` and `#per`. Doing it with straight `#offset` and `#limit` works:
```
(byebug) collection.offset( collection.current_page * @per_page + 1).limit(1).count
   (2.1ms)  SELECT COUNT(count_column) FROM (SELECT 1 AS count_column FROM "messages" LIMIT 1 OFFSET 15001) subquery_for_count
1
(byebug)
```

However, I don't know enough about activeadmin: are we guaranteed that the `collection` in `PaginatedCollection` responds to `offset` and `limit`? Mine did, since I'm using postgres and activerecord, but I'm not sure what other data stores are supported.

If I can use `#offset` and `#limit`, I'll rewrite the PR to use them. Let me know.

If I can't use them, then I'm out of ideas. I'll put in one more plug for the PR's current approach which is to set the total pages as "current page + 1". The only downside is that when you're on the last page, it will still have a "Next" button. In practice, I don't think it's a problem, since we're talking about collections with tens of thousands of pages here, and if a user opts in to `pagination_false: true`, then they shouldn't expect us to actually know the last page.

timoschilling:
`collection` should be a ActiveRecord::Base or ActiveRecord::Relation object. It's possible but not allowed that `collection` returns something else. So we can assume that it has `offset` and `limit`.

losvedir:
Ah, great! I'll update the PR to use `offset` and `limit` when I have a chance.

timoschilling:
Need we more test for that?

timoschilling:
@losvedir any updates?

losvedir:
@timoschilling - Sorry, for the delay. I made the `offset` and `limit` changes as well as hash formatting (just now pushed up) but have been working on tests for the feature. I'm not familiar with cucumber myself so the tests are slow going, and I don't have a ton of time on the side to work on it. Will try to complete the tests this week.

timoschilling:
@losvedir :+1:

losvedir:
@timoschilling - Okay, ready for review! I've never used cucumber before; please let me know if I should be doing that a different way.

timoschilling:
There is a syntax error on Ruby 1.9 https://travis-ci.org/activeadmin/activeadmin/jobs/57748755#L303

losvedir:
Ah, bummer. I don't have Ruby 1.9 on this computer and it's slow to build for me. Will have to take a look tonight.

losvedir:
Ah, ruby 1.9.3 didn't like the unicode literal in my string! Fixed.

sillylogger:
I think this `collection.offset(collection.current_page * @per_page.to_i).limit(1).count` is great in that it's faster than actually counting the items in the collection, but what about leveraging the already queried `collection_size` from `ActiveAdmin::Helpers::Collection`?

As it stands there are 3 queries actually run against the table:

```
SELECT COUNT(count_column) FROM (SELECT  1 AS count_column FROM "users" LIMIT 50 OFFSET 0) subquery_for_count
SELECT COUNT(count_column) FROM (SELECT  1 AS count_column FROM "users" LIMIT 1 OFFSET 50) subquery_for_count
SELECT  "users".* FROM "users"  ORDER BY "users"."id" desc LIMIT 50 OFFSET 0
```

1) to count the number of items on the current page
2) to peek if there is 1 item on the next page
3) to load the items on the current page

I see two separate potential optimizations here:

  - What if we assume there is a next page if the current page as a complete collection?  There is a loss in information here no doubt... but if the goal is performance, this would eliminate a potentially expensive query.

  -  What about actually loading the items on the first `collection_size` call in `ActiveAdmin::Helpers::Collection`?  Subsequent calls to iterate over the items will hit the cache.

spec/unit/views/components/paginated_collection_spec.rb
```diff
@@ -198,9 +198,10 @@ def paginated_collection(*args)
       end
 
       describe "set to false" do
-        let(:pagination) { paginated_collection(collection, pagination_total: false) }
-
         it "should not show the total item counts" do
+          expect(collection).not_to receive(:num_pages)
+          expect(collection).not_to receive(:total_pages)
+          pagination = paginated_collection(collection, pagination_total: false)
```
rafaelsales: This spec doesn't assert on COUNT queries. If this was ever fixed, it seems like a regression was added: 
https://github.com/activeadmin/activeadmin/pull/5388

Issue #3543, opened 2014-10-30
I18n support in Javascript


timoschilling:
We have some issues related to Javascript with I18n:
#3445 Translation not possible with default batch actions modal dialog
#3526 Use translated Month names for AA datepicker
#1665 Localize date picker

  #3123 Datepicker_input value field is fixed on yyyy-mm-dd (#3124 not working PR for #3123)

One solution could be the usage of [`i18n-js`](https://github.com/fnando/i18n-js).

PR #3849, opened 2015-03-10
dynamic labels for modal dialog forms

AndrewSwerlick:
extending modal_dialog.js.coffee so that the input definition can include a seperate label option. This allows dynamic server side generated labels with consistent name fields

In particular, we're using this in a situation where some of our batch action forms have a popup with a form where the user can choose to apply the action to all records meeting the current filter criteria, not just the visible ones. The label for the checkbox input, needs to include the total count of records. This change allows us to do that. It's good for any situation where you need dynamic labels, particularly on batch actions.

### DIFF
--- a/app/assets/javascripts/active_admin/lib/modal_dialog.js.coffee
+++ b/app/assets/javascripts/active_admin/lib/modal_dialog.js.coffee
@@ -1,6 +1,14 @@
 ActiveAdmin.modal_dialog = (message, inputs, callback)->
   html = """<form id="dialog_confirm" title="#{message}"><ul>"""
-  for name, type of inputs
+  for name, definition of inputs
+    label = name
+    type = definition
+    if definition.type
+      type = definition.type
+
+    if definition.label
+      label = definition.label
+
     if /^(datepicker|checkbox|text)$/.test type
       wrapper = 'input'
     else if type is 'textarea'
@@ -12,7 +20,7 @@ ActiveAdmin.modal_dialog = (message, inputs, callback)->
 
     klass = if type is 'datepicker' then type else ''
     html += """<li>
-      <label>#{name.charAt(0).toUpperCase() + name.slice(1)}</label>
+      <label>#{label.charAt(0).toUpperCase() + label.slice(1)}</label>
       <#{wrapper} name="#{name}" class="#{klass}" type="#{type}">""" +
         (if opts then (
           for v in opts


AndrewSwerlick:
Note it's fully backwards compatible with existing behavior. For dynamic labels you create a form definition that looks like this 
    {
      my_input: { label: "My custom label", type: "checkbox" }
    }

timoschilling:
I understand your intent for this change, but I don't like the interface of form generation (it's not your foult it starts before your PR).

Can you post some code how you use it and a screen shot how your example looks like.

AndrewSwerlick:
Sure we have a batch action that looks sort of like this

    batch_action "Add to data role", confirm: "Select a data role", form: -> {
        record_count = collection.limit(nil).count
        base = { common_input: [..input options..] }

        # if we have less than 5000 records let them apply to all the hidden pages, otherwise don't show that option
        if record_count < 5000
          base.merge(apply_to_hidden_pages: {
            type: "checkbox",
            label: "Apply to hidden pages(#{collection.limit(nil).count})"
          })
        else
          base
        end
      } 
    do |selection, inputs|
    	 # batch action code here ...
    end

It looks like this

screen shot 2015-03-10 at 9 52 53 am

AndrewSwerlick:
Note I'm aware this specific scenario is very related to issue #2375, but I see this as being a more general solution to any time you need a label that is dynamically generated. This could be particularly useful for nested resources where you want the label to have some information about the parent resource you're working with, etc.

Fivell:
@AndrewSwerlick,  :+1:

timoschilling:
Has someone else a other real world example? @AndrewSwerlick your example will be obsolete if #2375 has as implementation.

timoschilling:
Some other reason could be the translation of the field names, bat that should be handled by JS I18n integration (#3543)

Fivell:
>>Has someone else a other real world example? @AndrewSwerlick your example will be obsolete if #2375 has as implementation.

@timoschilling , #2375  is not implemented yet, (I'm not good in UI to develop it) but actually example provided by @AndrewSwerlick  is not working.

To make it work filters values should be passed to ```batch_action```

AndrewSwerlick:
Inside the batch action, I look for the input ``apply_to_hidden_pages`` and if it's true I access the ``batch_action_collection`` method. As noted in #3775 this method ensures the properly filtered collection is available throughout the lifetime of the request, so there is no need to pass the filters into the batch action. I have this working in my specific application and that method is behaving as expected and only returning the filtered records.

I agree about my specific use case being obsoleted by #2375

timoschilling:
With this PR the `:form` hash has 3 kinds of value:
1. a Symbol / String
2. a Array
3. a Hash containing a type with 1. or 2.

And I plan to add a I18n way for translating the labels.

That's a bit to much in my eyes. Thats why I ask for other realworld examples.

AndrewSwerlick:
It's fine if this doesn't fit into the long term vision for form management, and I agree that area probably needs a larger revision anyways. I implemented this because I needed something like #2375 today in the app I'm working on, and this seemed to be a simple, low impactful way to get it, since I could just monkey patch ``ActiveAdmin.modal_dialog``. I figured I'd submit a pull request since I saw some theoretical broader applications.

Personally I don't have any other real world examples.

luisbraga:
What are the current options to translate the labels?

waclock:
can someone merge this?

javierjulio:
Since it's been so long we'd need a new update on this. I think something like this can be helpful. I'd accept an update. Feel free to open a new PR. Thank you.

PR #3855, opened 2015-03-15
Change block.call to yield and delete excess argument in block

davydovanton:


timoschilling:
What's the intent for this change?

Please update the yard docs for that change
http://www.rubydoc.info/gems/yard/file/docs/Tags.md#yield

davydovanton:
@timoschilling `yield` more faster than `block.call` (in 5x times). If you are interested about this you can read this PR (rspec/rspec-rails#1308) and see [this](https://youtu.be/fGFM_UrSp70?t=10m28s) talk.
And I think that `menus` attribute is excess in block :smiley:

timoschilling:
Ok, I've been thinking that.

I talk about this change
https://github.com/davydovanton/activeadmin/commit/8df5027635d623e2fb19b8d88896e261c03b8be3#diff-8ee4e80c65b2a22317fa641a35bc00f9R113
It's no longer a Proc Attribute, now it's a `@yield` see http://www.rubydoc.info/gems/yard/file/docs/Tags.md#yield for details and examples.

davydovanton:
@timoschilling fixed docs, thanks for catching :sweat_smile:

timoschilling:
You are sure that it is `@yield [Proc]`, I think it is `@yield [MenuClass]`

davydovanton:
@timoschilling yes, you are right. I fixed this :tada:

timoschilling:
Sorry but the currect name is [`ActiveAdmin::Menu`](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/menu.rb#L15), `MenuClass` was a placeholder, while I didn't had the time to prove the right name. If you fix it, please reduce the commits to the first two.

davydovanton:
@timoschilling I'm sorry it took so long :sweat_smile:
I fixed doc and squashed commits

davydovanton:
@timoschilling thanks! :tada:

PR #3870, opened 2015-03-30
WIP: Nice error handling

bsvin33t:
Added a middleware to intercept errors that are raised in the namespace of activeadmin

Added a template for the exception page
Added a controller to respond with an ActiveAdmin layout exception page
**No Tests Added**


lib/active_admin/engine.rb
```diff
@@ -9,6 +9,11 @@ class Engine < ::Rails::Engine
       end
     end
 
+    initializer "active_admin.add_error_handler" do |app|
```
timoschilling: use `"active_admin.middleware"`

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,29 @@
...
+    class ErrorHandler
+
+      def initialize(app)
+        @app = app
+      end
+
+      def call(env)
+        @app.call(env)
+      rescue Exception => e
+        if ActiveAdmin.application.default_namespace.to_s == env['PATH_INFO'].split('/').second
```
bsvineeth: @timoschilling I feel the way that I'm checking if the request-response is going thur ActiveAdmin (and handling the exceptions)  is kind of not so great. I would like your thoughts here. What can I change/improve ?

bsvineeth: According to Rails/Engine docs, a rails app is supposed to create a separate middleware stack for the app and the engine. But this is not happening here. This is the reason, why I had to do it this way.

Reference: http://api.rubyonrails.org/classes/Rails/Engine.html#class-Rails::Engine-label-Configuration

timoschilling: This code only works for the default namespace. ActiveAdmin can have many. And maybe it don't work with a root namespace.

lib/active_admin.rb
```diff
@@ -120,6 +121,7 @@ def after_load(&block)
 # Require things that don't support autoload
 require 'active_admin/engine'
 require 'active_admin/error'
+require 'active_admin/middleware/error_handler'
```
timoschilling: This should be autoloaded too

features/error/error_page.feature
```diff
@@ -0,0 +1,24 @@
...
+
+  Scenario: For a 500 Error Page
+    Given a configuration of:
+    """
+      ActiveAdmin.register_page "Dashboard" do
+        content do
+          five_hundred_error
+        end
+      end
+      """
```
timoschilling: Please bring all into one line. See for example: https://github.com/bsvineeth/activeadmin/blob/29c61de9fb4a5f1ddd743b677f549b725857d0ca/features/action_item.feature#L11

features/error/error_page.feature
```diff
@@ -0,0 +1,24 @@
...
+    Given I am logged in
+    When I go to the dashboard
+    Then I should see the Active Admin layout
+    And I should see the page title "Error 500"
+
+  Scenario: For a 404 Error Page
+    Given an index configuration of:
+    """
+        ActiveAdmin.register Post
+      """
```
timoschilling: Same as line 11

timoschilling:
The Middleware should only be added in

timoschilling:
We should add the Middle ware not in test and development env.

And we need a different position in the Middleware stack:
```ruby
# ...
use Rack::Head
use Rack::ConditionalGet
use Rack::ETag
use Warden::Manager
use NewRelic::Rack::BrowserMonitoring
use NewRelic::Rack::AgentHooks
use NewRelic::Rack::ErrorCollector
use ActiveAdmin::Middleware::ErrorHandler
run FooBar::Application.routes
```
We have a problem in the current setup, NewRelic can no longer log errors for ActiveAdmin.

lib/active_admin/exception_app.rb
```diff
@@ -0,0 +1,74 @@
+module ActiveAdmin
+  class ExceptionController < ActionController::Base
```
timoschilling: This file should be named `exception_controller.rb`

bsvineeth:
I think it is safe to put our middleware below  `ActionDispatch::ShowExceptions`, what do you suggest?

I will add the `group: :production` in the initializer and add a stub to the feature spec, so that the testing would be possible.

bsvineeth:
Accidentally closed it.

timoschilling:
There is a much more easier way to display the error:
```ruby
ActiveAdmin.register_page "Error" do
  menu false

  content title: "Error" do
    "Error: #{env["STATUS"]}"
  end
end
```
```ruby
      def render_exception(env, exception)
        wrapper = ActionDispatch::ExceptionWrapper.new(env, exception)
        env["STATUS"] = wrapper.status_code
        Admin::ErrorController.action(:index).call(env)
      end
```

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def call(env)
+      @app.call(env)
+    rescue Exception => e
+      if ActiveAdmin.application.default_namespace.to_s == env['PATH_INFO'].split('/').second
+        render_exception(env, e)
+      else
+        @app.call(env)
+      end
+    end
```
timoschilling: a cleaner solution:
```ruby
      def call(env)
        @app.call(env)
      rescue Exception => exception
        if is_active_admin_error?
          render_exception(env, exception)
        else
          raise exception
        end
      end

      private

      def is_active_admin_error?
        ActiveAdmin.application.default_namespace.to_s == env['PATH_INFO'].split('/').second
      end
```
In your version you double call the application in case of a error. Thats not good for two reasons, first it coats time, second after the first call you have a different situation that maybe raise a different error.

timoschilling:
> I think it is safe to put our middleware below ActionDispatch::ShowExceptions, what do you suggest?

I'm not sure what do you mean with "below", below in which order? It's easier to say before or after `ActionDispatch::ShowExceptions` in the context of a middleware position.

> I will add the group: :production in the initializer and add a stub to the feature spec, so that the testing would be possible.

Let us build a option for the error page, the values can be `true`, `false` or a proc. The default is:
```ruby
-> { !(Rails.env.test? || Rails.env.development?) }
```

Then you don't need the `group: :production`, it works with a `staging` env (which is not the case with a `group: :production`) and it's simple to stub in the tests.

lib/active_admin/middleware/error_handler.rb
bsvineeth: Registering an error page and using it to show error is not allowed in Activeadmin.

I tried it. I am seeing an exception raised by the base controller, that `AbstractController::ActionNotFound`

https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/base_controller.rb#L28

This is the reason why I went with using ActionController for the exception controller.

My first choice was to use the base controller, so that I would need not have to implement all the helpers provided by it. But because of the exception, I did not continue with it.

bsvineeth:
I am thinking of doing something like this:
``` ruby
use Rack::Sendfile
use ActionDispatch::Static
use Rack::Lock
use #<ActiveSupport::Cache::Strategy::LocalCache::Middleware:0x007fb31b271228>
use Rack::Runtime
use Rack::MethodOverride
use ActionDispatch::RequestId
use RequestStore::Middleware
use Rails::Rack::Logger
use ActionDispatch::ShowExceptions

use ActiveAdmin::ErrorHandler

use ActionDispatch::DebugExceptions
use ActionDispatch::RemoteIp
use ActionDispatch::Reloader
use ActionDispatch::Callbacks
use ActiveRecord::Migration::CheckPending
use ActiveRecord::ConnectionAdapters::ConnectionManagement
use ActiveRecord::QueryCache
use ActionDispatch::Cookies
use ActionDispatch::Session::CookieStore
use ActionDispatch::Flash
use ActionDispatch::ParamsParser
use Rack::Head
use Rack::ConditionalGet
use Rack::ETag
use Warden::Manager
run TestRailsApp421::Application.routes
```

timoschilling:
:+1:

lib/active_admin/middleware/error_handler.rb
timoschilling: I have tried it and it has worked. The controller name depends on the active admin namespace name. Maybe you need to check your test setup. Take a look on `rake routes`

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    private
+
+    def render_exception(env, exception)
+      wrapper = ActionDispatch::ExceptionWrapper.new(env, exception)
+      env["STATUS"] = wrapper.status_code
```
timoschilling: please use `env["active_admin.original_error"]` here and move the `status_code` call up to the page.

lib/active_admin/base_controller.rb
```diff
@@ -59,6 +59,10 @@ def active_admin_namespace
     end
     helper_method :active_admin_namespace
 
+    def status_code
+      ActionDispatch::ExceptionWrapper.new(env, env["active_admin.original_error"]).status_code
```
timoschilling: This method shouldn't placed in the BaseController or is there a specific reason for that?

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,54 @@
...
+  class ErrorHandler
+
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      @env = env
+      @app.call(env)
+    rescue Exception => exception
```
timoschilling: That's ok

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,54 @@
...
+    def stringified_namespaces
+      ActiveAdmin.application.namespaces.collect { |namespace| namespace.name.to_s }
+    end
+
+    def current_namespace
+      @env['PATH_INFO'].split('/').second
+    end
+
+  end
+end
```
timoschilling: New Line missing

lib/active_admin/base_controller.rb
bsvineeth: I will take it out. The reason why I placed it in the base controller is so that I don't have any logic in the view. I was unable to find any other place to put it.

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,54 @@
...
+    def is_active_admin_error?
+      stringified_namespaces.include?(current_namespace)
+    end
+
+    def use_active_admin_error_page?
+      if ActiveAdmin.application.use_active_admin_error_page.is_a?(Proc)
+        ActiveAdmin.application.use_active_admin_error_page.call
+      else
+        ActiveAdmin.application.use_active_admin_error_page
+      end
```
timoschilling: Maybe we can use [`render_in_context`](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/view_helpers/method_or_proc_helper.rb#L82) here

timoschilling: We can remove `use_active_admin_error_page?` and add the middleware only if ActiveAdmin.application.use_active_admin_error_page or ActiveAdmin.application.use_active_admin_error_page.call is true

lib/active_admin/view_factory.rb
```diff
@@ -23,6 +23,5 @@ class ViewFactory < AbstractViewFactory
               edit_page:  ActiveAdmin::Views::Pages::Form,
               layout:     ActiveAdmin::Views::Pages::Layout,
               page:       ActiveAdmin::Views::Pages::Page
-
```
timoschilling: readd this line to get a clean diff

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def render_exception(exception)
+      @env["active_admin.original_error"] = exception
```
timoschilling: `@env["active_admin.original_error"] = ActionDispatch::ExceptionWrapper.new(env, exception)`

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,54 @@
...
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      @env = env
+      @app.call(env)
+    rescue Exception => exception
+      if is_active_admin_error? && use_active_admin_error_page?
+        render_exception(exception)
```
timoschilling: change it to `render_exception(exception, env)`, then we don't need a `@env` var.

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,54 @@
...
+      else
+        ActiveAdmin.application.use_active_admin_error_page
+      end
+    end
+
+    def render_exception(exception)
+      @env["active_admin.original_error"] = exception
+
+      @env["action_dispatch.request.parameters"] =
+          {controller: "#{current_namespace}/error", action: "index"}
```
timoschilling: Way do we need this?

features/error/error_page.feature
```diff
@@ -0,0 +1,28 @@
...
+
+  Scenario: For a 500 Error Page
+    Given a configuration of:
+      """
+      ActiveAdmin.application.use_active_admin_error_page = true
+      ActiveAdmin.register_page "Dashboard" do
+        content do
+          five_hundred_error
+        end
+      end
```
timoschilling: This block need's to more spaces

lib/active_admin/base_controller.rb
bsvineeth: Another problem is that when I hit the error page, I am going to see a 500 instead of a 404. This is because the routes to `namespace/error` exists. I have to think of something to stop this.

lib/active_admin/middleware/error_handler.rb
bsvineeth: If I don't put in line 36, the `AbstractController::ActionNotFound` exception will be raised by the base controller, for every errored action that is not index to its corresponding controller.


Here I'm modifying the `params` hash to make the `BaseController`'s  `before_filter :only_render_implemented_actions` understand that the request is coming from the `index` action of `Admin::ErrorController`

The cleaner way would have been to create a new `Rack::Request` object and do a `update_param`. But this was not working. So, I did it this way

lib/active_admin.rb
```diff
@@ -56,6 +56,8 @@ module ActiveAdmin
   autoload :ViewFactory,              'active_admin/view_factory'
   autoload :ViewHelpers,              'active_admin/view_helpers'
   autoload :Views,                    'active_admin/views'
+  autoload :ErrorHandler,             'active_admin/middleware/error_handler'
+
```
timoschilling: Please remove this line

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,51 @@
...
+    def stringified_namespaces
+      ActiveAdmin.application.namespaces.collect { |namespace| namespace.name.to_s }
+    end
+
+    def current_namespace(env)
+      env['PATH_INFO'].split('/').second
+    end
+
+  end
+end
```
timoschilling: Add a empty line at the end

lib/active_admin/error_pages/error_pages.rb
```diff
@@ -0,0 +1,9 @@
+ActiveAdmin.after_load do |app|
+  app.namespaces.each do |namespace|
+    namespace.register_page "Error" do
```
timoschilling: Change the name to `Exception`.

lib/active_admin/middleware/error_handler.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    private
+
+    def is_active_admin_error?(env)
+      stringified_namespaces.include?(current_namespace(env))
+    end
+
+    def use_active_admin_error_page?
+      render_in_context(ActiveAdmin.application, ActiveAdmin.application.use_active_admin_error_page)
```
timoschilling: `render_in_context(self, ActiveAdmin.application.use_active_admin_error_page)` should work

timoschilling:
I want some tests for `is_active_admin_error?`.

lib/active_admin/application.rb
```diff
@@ -110,6 +110,13 @@ def initialize
                                       :email,
                                       :to_s ]
 
+    # Active Admin overrides the default error pages of rails with it's own
+    # To enable it, it takes either a boolean or a proc which can be evaluated to a boolean
+    # By default, it is assigned the proc -> { !(Rails.env.test? || Rails.env.development?) }
+
```
timoschilling: Please remove this empty line.

Fivell:
@bsvin33t , are you going to complete this ?

bsvin33t:
@Fivell :( 

I was unable to get the errors app working on activeadmin rails engine(back in the day).
I don't think I can wrap this up now.

Fivell:
@timoschilling , if you're sure this https://github.com/activeadmin/activeadmin/pull/3870#issuecomment-88063782 works we may add this to wiki and close this

javierjulio:
Lets close this out as I don't think this is necessary. We use the Rails template for showing an error and I don't think we'd want this in by default as it could easily make its way into a production app. If anything @Fivell lets stick with documenting the solution @timoschilling came up.

bsvin33t:
@javierjulio If you guys think it is going to be useful, I can try to create a short writeup about configuring an errors app into active admin.

PR #3873, opened 2015-04-01
Make prepend (and append) filters available in ActiveAdmin.setup

### LINKED ISSUE #3872: Want to use prepend_before_filter in ActiveAdmin.setup
I have wanted to add an authentication filter on the top of everything, even skip login redirection.

In a regular Rails app, you can achieve this by using [`prepend_before_filter`](http://apidock.com/rails/v3.2.13/AbstractController/Callbacks/ClassMethods/prepend_before_filter). However, I found that current ActiveAdmin just delegate some part of filters, thus I couldn't use `prepend_before_filter`.

I've just added the missing methods [here](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/application.rb#L219) and it works fine. I'll submit a pullreq later.

tnj:
This fixes #3872. :beers:

### DIFF
--- a/spec/unit/controller_filters_spec.rb
+++ b/spec/unit/controller_filters_spec.rb
@@ -12,33 +12,30 @@ describe ActiveAdmin::Application do
     ]
   end
 
-  it 'before_filter' do
-    controllers.each{ |c| expect(c).to receive(:before_filter).and_return(true) }
-    application.before_filter :my_filter, only: :show
-  end
+  expected_actions = (
+    prefixes = %w(skip append prepend) << nil
+    positions = %w(before around after)
+    suffixes = %w(filter)
+    base = %w(skip_filter)
+    if Rails::VERSION::MAJOR >= 4
+      suffixes += %w(action)
+      base += %w(skip_action_callback)
+    end
 
-  it 'skip_before_filter' do
-    controllers.each{ |c| expect(c).to receive(:skip_before_filter).and_return(true) }
-    application.skip_before_filter :my_filter, only: :show
-  end
+    prefixes.each_with_object(base) do |prefix, stack|
+      positions.each do |position|
+        suffixes.each do |suffix|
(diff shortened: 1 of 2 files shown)


timoschilling:
1. I missing the specs.
2. We should support Rails 4 `before_action` too.
3. We should build a more generic code like:
```ruby
prefixes = [nil, "skip", "append", "prepend"]
positions = ["before", "around", "after"]
subfixes = ["filter", "action"]
prefixes.each_with_object([]){|prefix, stack| positions.each{|position| subfixes.each{|subfix| stack << [prefix, position, subfix].compact.join("_") } } }

```

tnj:
Hey @timoschilling,

Rather than generating method names from hardcoded values, enumerating the public instance methods from [`AbstractController::Callbacks::ClassMethods`](http://apidock.com/rails/v4.1.8/AbstractController/Callbacks/ClassMethods) could be much more robust against the future changes on Rails.

I've revised the code in this manner, and added tests based on your example code to ensure the existence of methods what we are expecting regardless the Rails modification.

spec/unit/controller_filters_spec.rb
```diff
@@ -12,33 +12,29 @@
...
+        end
+      end
+    end
+  )
+
+  expected_actions.each do |action|
+    it action do
+      controllers.each { |c| expect(c).to receive(action).and_return(true) }
+      application.public_send action, [:my_filter, { only: :show }]
+    end
```
timoschilling:
```ruby
it action do
  args = [:my_filter, { only: :show }]
  controllers.each { |c| expect(c).to receive(action).with(args) } # I think `.and_return(true)` is not needed
  application.public_send action, args
end
```

tnj: :+1: Agreed.

timoschilling:
Good work!
Please apply my suggested chance and squash your commits.

tnj:
Thanks :+1:  I squashed the commits.

PR #3874, opened 2015-04-02
Minor feature: Except trailblazer param on form builder

fernandes:
Trailblazer injects current_user on params[:current_user] to use authentication and process user-related stuff inside Operations.

So I'm suggesting remove params[:current_user] using except: current_user once it does not affect ActiveAdmin stuff and would really help on Trailblazer integration, I based this decision checking that `page` (Kaminari) and `q` (Ransack) are removed the same way I'm suggesting with `current_user`

ps: I looked the tests, but just found about except behavior, nothing about what to take off using except


lib/active_admin/filters/forms.rb
```diff
@@ -65,7 +65,7 @@ def active_admin_filters_form_for(search, filters, options = {})
           buttons = content_tag :div, class: "buttons" do
             f.submit(I18n.t('active_admin.filters.buttons.filter')) +
               link_to(I18n.t('active_admin.filters.buttons.clear'), '#', class: 'clear_filters_btn') +
-              hidden_field_tags_for(params, except: [:q, :page])
+              hidden_field_tags_for(params, except: [:q, :page, :current_user])
```
fernandes: @timoschilling in my opinion will be very ugly align the operands like hound said... what do you think?

timoschilling: You can ignore @houndci in this case

fernandes: so... its ready to be merged :stuck_out_tongue_winking_eye: 

if you see any place to add tests for it (I haven't), please tell so I can do it, otherwise we can go for it ;)

timoschilling:
I'm not satisfied with this change. The reason is that `q` and `page` are common parameter names / use cases and they are used by gems on wich active admin directly depend. I'm against hardcoded changes to supporting third party code.

I think it's not a clean solution from trailblazer to inject `current_user` into the params hash.

You can monkey patch `hidden_field_tags_for` in your `activetrail` gem.

fernandes:
thats right @timoschilling thank you for your consideration and attention :)

fernandes:
btw, can we just isolate the except method on something like this

```ruby
buttons = content_tag :div, class: "buttons" do
  f.submit(I18n.t('active_admin.filters.buttons.filter')) +
   link_to(I18n.t('active_admin.filters.buttons.clear'), '#', class: 'clear_filters_btn') +
   hidden_field_tags_for(params, except: except_hidden_fields)
end

def except_hidden_fields
  [:q, :page]
end
```

just to make the monkey patch easier once [hidden_fields_tags_for](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/view_helpers/form_helper.rb#L11) is not so simple, wouldn't like to make a big copy-paste... if not I can handle it here anyway

thank you

fernandes:
amazing man! thank you very much @timoschilling !

PR #3878, opened 2015-04-06
making `columns` works for Page and Resource

bsvin33t:
Calculating the column styles and margins only if the child element is of type column.

### DIFF
--- a/lib/active_admin/views/components/columns.rb
+++ b/lib/active_admin/views/components/columns.rb
@@ -93,19 +93,24 @@ module ActiveAdmin
         all_margins_width = margin_size * (span_count - 1)
         column_width = (100.00 - all_margins_width) / span_count
 
-        children.each_with_index do |col, i|
+        columns.each_with_index do |column, i|
           is_last_column = i == (columns_count - 1)
-          col.set_column_styles(column_width, margin_size, is_last_column)
+          column.set_column_styles(column_width, margin_size, is_last_column)
         end
       end
 
       def columns_span_count
         count = 0
-        children.each {|column| count += column.span_size }
+        columns.each do |column|
+          count += column.span_size
+        end
 
         count
       end
 
+      def columns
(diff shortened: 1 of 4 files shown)


lib/active_admin/views/components/columns.rb
```diff
@@ -93,15 +93,21 @@ def calculate_columns!
...
       end
 
       def columns_span_count
         count = 0
-        children.each {|column| count += column.span_size }
+        children.each do|child|
+          if child.is_a?(Column)
+            count += child.span_size
+          end
+        end
```
timoschilling: You can simplify Line 97 and Line 107 by building a `columns` method, wich looks like:
```ruby
def columns
  children.select { |child| child.is_a?(Column) } 
end
```

timoschilling:
Good work. Please apply the Hound comments.

timoschilling:
There is no need to merge / pick the master into your PR. Please rebase your PR on master and squash all commits.

lib/active_admin/views/components/columns.rb
```diff
@@ -93,19 +93,24 @@ def calculate_columns!
         all_margins_width = margin_size * (span_count - 1)
         column_width = (100.00 - all_margins_width) / span_count
 
-        children.each_with_index do |col, i|
+        columns.each_with_index do |column, i|
           is_last_column = i == (columns_count - 1)
-          col.set_column_styles(column_width, margin_size, is_last_column)
+          column.set_column_styles(column_width, margin_size, is_last_column)
```
timoschilling: I'm not sure that this code does the same as the old one. What `children` can be there that are not a `Column`?

bsvin33t: maybe something like a `para` ?

timoschilling: I don't know, but if the tests are working keep it as it is.

Issue #3867, opened 2015-03-27
AA thinks "false" means "empty" in show view


dpmccabe:
If you have a boolean attribute on a resource and have a row for that attribute in your `show` view, AA will display "true" for `true` and "empty" for `false`. Try the following:

    show do
      attributes_table do
        row :id
        row :true do |a|
          true
        end
        row :false do |a|
          false
        end
      end
    end

This yields the following:

![screen shot 2015-03-27 at 5 50 07 pm](https://cloud.githubusercontent.com/assets/527578/6878314/034092f6-d4aa-11e4-8b2f-804906192ae3.png)

This is using PostgreSQL. I'm not sure if the bug exists in other DBs.

It's not a problem on the `index` view, though, where `true` and `false` are correctly interpreted as `status_tag('yes')` and `status_tag('no')`, respectively. I think `status_tag` should also be the default behavior in show, but at the very least it shouldn't be confusing `false` with `empty`/`nil`, which are certainly different things in Ruby, falsiness aside.

bsvin33t:
Hi @timoschilling I have created a PR for this bug in https://github.com/activeadmin/activeadmin/pull/3882

dpmccabe:
Thanks, @bsvin33t, I've confirmed the patch is working on my original sample code and in an actual resource.

PR #3882, opened 2015-04-08
AA thinks "false" means "empty" in show view

bsvin33t:
Calling render in display name for all resources that are not nil
Self is assigned to render_in_context only when the context is nil
Added Spec to check if false returns a `false` string when display_name is called on it
Added Cucumber specs to cover the same scenarios in the integration testing.

Ref #3867

### DIFF
--- a/lib/active_admin/view_helpers/method_or_proc_helper.rb
+++ b/lib/active_admin/view_helpers/method_or_proc_helper.rb
@@ -80,7 +80,7 @@ module MethodOrProcHelper
   # passing it the proc. This evaluates the proc in the context of the reciever, thus changing
   # what `self` means inside the proc.
   def render_in_context(context, obj, *args)
-    context ||= self # default to `self`
+    context = self if context.nil? # default to `self` only when nil
     case obj
     when Proc
       context.instance_exec *args, &obj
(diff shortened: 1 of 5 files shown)


lib/active_admin/view_helpers/method_or_proc_helper.rb
```diff
@@ -80,7 +80,7 @@ def render_or_call_method_or_proc_on(obj, string_symbol_or_proc, options = {})
   # passing it the proc. This evaluates the proc in the context of the reciever, thus changing
   # what `self` means inside the proc.
   def render_in_context(context, obj, *args)
-    context ||= self # default to `self`
+    context = self if context.nil? # default to `self` only when nil
```
timoschilling: have you tried `render_in_context(context = self, obj, *args)`

spec/unit/view_helpers/display_name_spec.rb
```diff
@@ -31,10 +31,12 @@
...
 
-  [nil, false].each do |type|
-    it "should return nil when the passed object is #{type.inspect}" do
-      expect(display_name type).to eq nil
-    end
+  it "should return nil when the passed object is nil" do
+    expect(display_name nil).to eq nil
+  end
+
+  it "should return `false` when the passed objct is false" do
```
timoschilling: Please use 'false', `false` is a FalseClass instance, 'false' is a String

lib/active_admin/view_helpers/method_or_proc_helper.rb
bsvin33t: I'm getting a syntax error when I use `render_in_context(context = self, obj, *args)`

PR #3883, opened 2015-04-08
Fixed unexpected page title behavior on the show page of resources

### LINKED ISSUE #3860: Incorrect page title for models with a to_s method
It seems that defining a `to_s` method on a model will forcefully set the page title to the fallback "(model name) #(id)" instead of using any of the methods defined in `display_name_methods`.

Consider the following model:
    
    class Chart < ActiveRecord::Base
      #...

      def name
        "test"
      end
      alias_method :to_s, :name
    end

The page title and heading for "/admin/charts/1" will be "Chart #1" instead of "test".

This only seems be happening with the page title/heading on the `show` page. `name` gets used everywhere else, breadcrumbs included. `show title: :name do [...]` would fix it, but seems unnecessary. 

I'm using ActiveAdmin 1.0.0.pre1 and Rails 4.2.0.

wolfemm:
Fixes #3860.

`title == resource.to_s` prevented a user-defined to_s method on resources from being used as the page title, including any of the other display_name_methods aliased to it.

Currently the only workaround is to add `show title: :to_s`, which goes against the conventional nature of AA that I like so much.

Even with my change, the kernel's to_s will not be used because of the more comprehensive check in [DisplayHelper#display_name_method_for](https://github.com/activeadmin/activeadmin/blob/6bde1f57ec5f9b724c21e5af146678f4ade839bc/lib/active_admin/view_helpers/display_helper.rb#L30).

I can confirm that removing to_s from display_name_methods will not fix the problem, because if a method aliased to to_s is used `title == resource.to_s` will be true and result in the fallback.

### DIFF
--- a/lib/active_admin/views/pages/show.rb
+++ b/lib/active_admin/views/pages/show.rb
@@ -35,7 +35,7 @@ module ActiveAdmin
         def default_title
           title = display_name(resource)
 
-          if title.blank? || title == resource.to_s
+          if title.blank?
             title = "#{active_admin_config.resource_label} ##{resource.id}"
           end
 


Fivell:
@timoschilling ,  after merging this why we need  ```default_title```  method at all ?
it  was used as convention to display nice page title and breadcrumb title when you have no ```display_name``` in model 
```
title = "#{active_admin_config.resource_label} ##{resource.id}"
```
After merging this, it uses ```to_s``` everywhere 

like 
```
#<Order:0x007fb70fe71008>
```

PR #3886, opened 2015-04-09
truncate display_name_methods values

timoschilling:
There are some methods in that list which are to specific for some use cases. We should setup only `display_name` as ActiveAdmin default and `to_s` as fallback.

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -101,14 +101,7 @@ module ActiveAdmin
 
     # Active Admin makes educated guesses when displaying objects, this is
     # the list of methods it tries calling in order
-    setting :display_name_methods, [ :display_name,
-                                      :full_name,
-                                      :name,
-                                      :username,
-                                      :login,
-                                      :title,
-                                      :email,
-                                      :to_s ]
+    setting :display_name_methods, [:display_name, :to_s]
 
     # == Deprecated Settings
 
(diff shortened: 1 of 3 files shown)


mattvague:
@timoschilling Was this causing an issue for you somewhere?

dmitry:
:+1: becuase it's now more predictable. In other cases, you are always can add more names to the settings array.

timoschilling:
The old one is sometimes confusing. I share @dmitry's opinion. The new List ist more explicit.

I think about to rename `display_name` to  `active_admin_display_name` or something like that. To avoid conflicts and to decelerate for what the method is on the model.

I'm waiting with the merge for more discussion.

mattvague:
Alright, fair enough. What if instead of changing `display_name` to `active_admin_display_name`, we added a DSL method to resources and fall back on `#display_name`? Something like:

```
ActiveAdmin.register User do
  display_name_method :name
```

I just think it'd be better to avoid mixing ActiveAdmin-related presentation logic into an app's models. Thoughts?

mattvague:
(This is probably beyond the scope of this PR though, and something I would be happy to tackle)

timoschilling:
@mattvague I like the idea of not mixing ActiveAdmin-related presentation logic into models! But what is with Models which are not registered in ActiveAdmin? How should they be configured?

mattvague:
@timoschilling That's a good question. Would it make sense to just rely on `#display_name` and any other methods registered in `display_name_methods` for these models? If people want AA-specific display names for models that they don't want in the menu they would have 2 options:

1. Register something like `active_admin_display_name` in `display_name_methods`
2. Register a resource, specify `display_name_method` and hide the resource from the menu

mattvague:
@timoschilling Thoughts?

seanlinsley:
I really don't think this is something that should change, when we haven't even gotten out 1.0.0. People have relied on this behavior for many years now, and there's no good way to warn them that it's changed. If we're going to change this, it should wait until 2.0.

javierjulio:
@timoschilling we are moving forward with breaking changes for master (for a v2). Are you able to update the conflicts so we can review for merging in?

PR #3862, opened 2015-03-24
Overhaul CSS for maintainability and customization

javierjulio:
I've seen mention about CSS overhauls but nothing seems active, if anything years old. I'm streamlining the CSS to make it easier to maintain, improve and customize. This should make it much easier to move to a framework like Bootstrap or Foundation as well.

Since this was a long lived branch, several changes that were made here have been pulled out and merged into master to make this smaller. We've cut the number of commits to almost half!

- [x] Remove excessive styles e.g. gradients, shadows and keep default theme basic
- [x] Remove unnecessary styles as many are either defaulted, have no effect or are repetitive
  - [x] Clean up typography
  - [x] Replace strip table even/odd classes with nth-child selectors 
  - [x] Fix calc in forms due to box-sizing:border-box global change
  - [x] Clean up modal dialog styles
- [x] Reduce selector specificity (3 ids and 1 class in one selector 😩)
- [x] Replace id based selectors with class based ones but keeping ids in HTML in place
- [x] Rename CSS class index_grid to index_as_grid and index_table to index_as_table
- [ ] Simplify table_tools CSS to use button group approach (index_list.rb, scopes.rb)
- [x] Remove excessive status_tag classes and focus only on yes/no
- [x] Streamline Formtastic CSS (I've done the best I can here in just CSS, would to need to tweak HTML)
- [x] Configure component styles using variables
- [x] Summary of removed assets and removed SASS variables for the changelog
- [x] Remove any inline styles
- [x] update print stylesheet

## How can you help?

- use my branch to test out how the changes are affecting your admin 

## Changelog

- Streamlined CSS: removing outdated techniques, styles that have no effect, etc.
- Modern theme with simpler selectors and defaults (less CSS) allowing easier customization.
- Any un-styled buttons need classes now: `button button-default` or `button button-primary`.
- Action Items no longer set default button styles, you can customize as necessary.
  - Apply class of `button button-default` for a default button, or `button button-primary` for the primary button style.
- `box-sizing: border-box` now applied globally to all elements.
- Removed orderable.png, nested_menu_arrow.gif and nested_menu_arrow_dark.gif images and replaced with CSS based arrows. #3612 
- Removed dropdown_menu JS positioning (CSS can do this just fine).
- Some internal Sass mixins have been removed: border-colors, sans-family, etc.
- Added "commented on" to EN locale (we need translations).

PR #3893, opened 2015-04-13
Cancel buttons are now unstyled

knu:
Styles for cancel buttons are mistakenly removed in 6ebb4a7d, which this PR fixes.

### DIFF
--- a/app/assets/stylesheets/active_admin/_forms.scss
+++ b/app/assets/stylesheets/active_admin/_forms.scss
@@ -244,8 +244,9 @@ form {
   fieldset.buttons li, fieldset.actions li {
     float:left;
     padding: 0;
-  }
 
+    &.cancel a { @include light-button; }
+  }
 }
 
 // -------------------------------------- Sidebar Forms


timoschilling:
Thanks :+1:

knu:
That was quick, thanks!

Looking at the buttons closely, there seem to be some minor style differences between submit buttons and cancel buttons (namely border-radius and height), so I'd appreciate if you could fix them as appropriate.

timoschilling:
In #3862 @javierjulio is doing a big overhaul of the css, maybe he can include that in his changes.

knu:
OK, I'm in no particular hurry and will be looking forward to it.

javierjulio:
@timoschilling yes its still in place in my repo. The problem was that "remove Iconic" branch removed this bit of CSS, my guess is you did a replace all on anything icon related and since the light-button and icon-size were on the same line it wiped both.

I plan to update it so we have a reusable button class rather than target each set of buttons individually. This way it'll be easier for when people customize ActiveAdmin they can just reuse a simple CSS class rather than having to write selectors and using our internal mixins. Nothing wrong with that bit on mixins but I think a button is such an essential part we can make it easier to access and use.

@knu thanks for finding and fixing this!

Issue #3901, opened 2015-04-21
If filter is Relational Model it would be better, if Search Status Bar will show us 'title' of filter then only 'ID'.


Shoroh:
If filter is relations then Search Status Bar will try to show (title|name|value|id) instead of just ID. Let me show on examples.

Before:
![screen shot 2015-04-21 at 17 11 17](https://cloud.githubusercontent.com/assets/1865984/7254144/63543416-e86b-11e4-866c-4dea382c8065.png)

After:
![screen shot 2015-04-21 at 21 15 03](https://cloud.githubusercontent.com/assets/1865984/7254161/8449cd52-e86b-11e4-81f7-748b3bb027c5.png)

Patch at activeadmin/lib/active_admin/filters/resource_extension.rb#search_status_section:

```ruby
active.filters.each do |filter|
  li do
    if filter.body =~ /ID/i
      begin
        klass = filter.body.split(' ').first.constantize
        method = klass.inspect.scan(/(title|name|value|id)/).last[0].to_sym
        value = klass.find(filter.value).send(method)
        span "#{klass}: "
        b value
      rescue
        span filter.body
        b filter.value
      end
    else
      span filter.body
      b filter.value
    end
  end
end
```

Do not judge me strictly for the code, I'm a Junior )

timoschilling:
I don't understand what you want or what your problem is.

timoschilling:
Now I think I know what you want. I close this we now have #3902

PR #3902, opened 2015-04-21
If filter is relations then Search Status Bar will try to show (title|name|value|id) instead of just ID.

Shoroh:
If filter is relations then Search Status Bar will try to show (title|name|value|id) instead of just ID. Let me show on examples.

Before:
screen shot 2015-04-21 at 17 11 17

After:
screen shot 2015-04-21 at 21 15 03

NOTE: text copied from #3901 by @timoschilling


lib/active_admin/filters/resource_extension.rb
```diff
@@ -174,8 +174,21 @@ def search_status_section
                 else
                   active.filters.each do |filter|
                     li do
-                      span filter.body
-                      b filter.value
+                      if filter.body =~ /ID/i
+                        begin
+                          klass = filter.body.split(' ').first.constantize
+                          method = klass.inspect.scan(/(title|name|value|id)/).last[0].to_sym
+                          value = klass.find(filter.value).send(method)
```
timoschilling: This 3 lines are looking to much hacky for me. Maybe @ccallebs knows a better solution.

Shoroh:
Let me try to explain )

Suppose User selects Area in filter: **Coconut Island**. Search Status Bar shows **Area ID equals 2** instead.

lib/active_admin/filters/resource_extension.rb
```diff
@@ -174,8 +174,21 @@ def search_status_section
                 else
                   active.filters.each do |filter|
                     li do
-                      span filter.body
-                      b filter.value
+                      if filter.body =~ /ID/i
+                        begin
+                          klass = filter.body.split(' ').first.constantize
```
ccallebs: @Shoroh What will happen with multi-word models?

lib/active_admin/filters/resource_extension.rb
ccallebs: @timoschilling I ran into the same issues as @Shoroh when trying to do this. The "correct" solution would be to have `active.filters` return a class in tandem with body/value. I can spend some more time looking at it this evening.

lib/active_admin/filters/resource_extension.rb
Shoroh: I see.
```ruby
klass = filter.body.split(' ID').first.gsub(' ','').constantize
```
?

lib/active_admin/filters/resource_extension.rb
Shoroh: @ccallebs 
You're right with this.
I didn't want to change other code a lot, thought this monkey-patch can help for the first time ))

timoschilling: @ccallebs that would be nice. I see the problem, but think that this is not the solution.

lib/active_admin/filters/resource_extension.rb
```diff
@@ -174,8 +174,21 @@ def search_status_section
                 else
                   active.filters.each do |filter|
                     li do
-                      span filter.body
-                      b filter.value
+                      if filter.body =~ /ID/i
+                        begin
+                          klass = filter.body.split(' ID').first.gsub(' ', '').constantize
+                          method = klass.inspect.scan(/(title|name|value|id)/).last[0].to_sym
```
ccallebs: @Shoroh Just curious whats with the `(title|name|value|id)` regex? Are you searching for a display name to use? If that's the case, we can just assume they're using the default `#display_name` method that ActiveAdmin requires.

Shoroh: Yes, that's the case )

Shoroh:
I'll be happy to see any comments about my last solution.

ccallebs:
@Shoroh @timoschilling I think that's a good refactor. I don't think we'll get much cleaner than that without some serious work. 

@Shoroh There are specs surrounding the Humanized class. You should add some for your edge cases here.

Shoroh:
Ok, I'll do it tomorrow.

lib/active_admin/filters/humanized.rb
```diff
@@ -9,8 +9,22 @@ def initialize(param)
...
+            rescue
+              nil
+            end
+      end
+
       def value
-        @value
+        @value =
+            begin
+              display_name(related_class.find(@value))
```
ccallebs: @Shoroh You should be able to do:

`related_class.find(@value).display_name`

If they don't have it defined, it'll just default to `@value` below.

Shoroh: I assume that I and people like me would not define this method often for any reason.
In fact it doesn't work at all if display_name won't define. I mean we won't get :title )
Are you sure?

ccallebs: @Shoroh That's a good point, I think your code is good as-is then. Would probably want to check with @timoschilling though.

dmitry:
I believe some sort of tests should be included in this PR.

Shoroh:
Better to test multi-word Model names also. It possible with Rspec Mocks Gem, or if create some additional model. What do you thing?

.travis.yml
```diff
@@ -31,6 +32,8 @@ matrix:
       env: "RAILS=4.1.8"
     - rvm: 2.1
       env: "RAILS=4.2.0"
+    - rvm: 2.1
+      env: "RAILS=4.2.1"
```
timoschilling: why do you want to test `4.2.0` and `4.2.1`?

Shoroh: @timoschilling 
There is no reason.
I thought it is a common test-flow for this gem )
Should I drop it?

timoschilling:
@Shoroh I think you can mock Models like [this](https://github.com/activeadmin/activeadmin/pull/3281/files#diff-8dfb4d0b8c040898f71715d2e5b071ebR34).

.travis.yml
timoschilling: Please drop it from this commit. I will change the `4.2.0` to `4.2.1`

spec/unit/filters/humanized_spec.rb
```diff
@@ -1,20 +1,46 @@
 require 'rails_helper'
 
 describe ActiveAdmin::Filters::Humanized do
-  let(:param) { ['category_id_eq', '1'] }
-
-  subject { ActiveAdmin::Filters::Humanized.new(param) }
+  let(:category) { Category.create!(name: 'Jazz') }
```
dmitry: Pretty strange rule for the @houndci

.travis.yml
Shoroh: Done.

Shoroh:
These tests are ok?

momolog:
@Shoroh anything I could do to help for this PR? @Fivell could you elaborate your latest request for changes a bit more?

lib/active_admin/filters/humanized.rb
```diff
@@ -9,8 +9,26 @@ def initialize(param)
         @value = param[1]
       end
 
+      def related_class
+        @related_class ||=
+            begin
```
Fivell: I think we can pass resource class to constructor instead of parsing

Fivell:
@momolog , if filter declared as 
```
filter :customer
```

active filters should display customers display_name


but when 
```
ActiveAdmin.register Logs do
  filter :customer_id, as: :numeric
end
```

there is not select dropdown. just text input for filtering numbers, in this case active filters should display customers id.

momolog:
@Fivell if I am not completely mistaken, is not possible to tell from the ransack query string, if `filter :customer_id, as: :numeric` or `filter :customer` was used.
But I would regard the first case (filtering by ID explicitly) the "unnormal" edge case and vote for always displaying the humanized version.

I would really love to see this merged but I currently do not see how I can help.

Fivell:
@momolog, so maybe implementation is wrong if it is handle only params, and ignore how filters are declared

momolog:
@Fivell the query is ambiguous and there is no way to change this short of adding more half-redundant query params to transport the *exact filter definitions* - or putting them into the session, both of which seem undesirable.

But in this situation one could at least chose the more sensible interpretation, which imho. would be the humanized version of both the body and the value wherever possible. My assumption and experience is that you almost never filter for bare IDs or integers.

The current state is pretty weird and my customers are **very** confused when they set a filter "Kunde: Alex" and the resulting current filter is "User ID gleicht 2". It just doesn't make any sense.

@timoschilling what do you think?

Fivell:
> @Fivell the query is ambiguous and there is no way to change this short of adding more half-redundant query params to transport the exact filter definitions - or putting them into the session, both of which seem undesirable.


Can you please explain what are you talking about? 

We already have filter definitions  so why we can't detect type of filter and render valid result?

momolog:
@Fivell you are of course right and i was talking confused. 
So would it be possible (and the right way) to read the models filter settings in the Humanized class ? 

Sorry for asking here but I am new to the active admin code base and there is no IRC.

momolog:
@Fivell While  the problem you point out is of course real, I would still argue that it is not introduced with this PR. The *current* `Filters::Humanized` implementation depends on params only (ignoring the filter definitions) and the PR just interprets the params more human-friendly.
It is an overall improvement without re-implementing `Filters::Humanized`.

Fivell:
@momolog, there is a bug of course, and yes it should be fixed, however I do  think this feature should be complete and handle all the cases or at least most of them. I will take a look next week on this and investigate what can be done  and what can be improved. In case I find solution new PR will be opened , in other case this PR will be rebased.

lib/active_admin/filters/humanized.rb
ccallebs: @Fivell Just to clarify, are you suggesting we move:

``` ruby
@body.split('_id')...
```
up to `ActiveAdmin::Filters::Active` and inject the corresponding classes into the filters array? Something like this:

``` ruby
      def build_filters
        filters = @params['q'] || []
        filters = filters.map { |f| f << deduce_resource_class(f) }
        filters.map { |param| Humanized.new(param) }
      end
```

Fivell: @ccallebs, I want to refactor this in future , to make it not dependent from from ```@body```, and inject active_admin_config and filters  to this part of  ActiveAdmin to make it work always and not to guess related_class.  However now I agree with @momolog, it is better to make it work not in all cases now and fix it in future, because I have no free time at the moment.

Fivell:
closed by #4951

Issue #3766, opened 2015-01-28
ActiveAdmin creates commets routes while commets are disabled


chumakoff:
I have set the option  'config.comments = false'  in my initializers/active_admin.rb file.   

But, including 'ActiveAdmin.routes(self)' to my config/routes.rb file creates the following routes, which should not be created:

  * admin/comments#index
  * admin/comments#create
  * admin/comments#show

Is there a way to avoid this ?

timoschilling:
No, not at the moment, while you can reenable comments pre resource.

Is the not used comments controller a problem for you?

chumakoff:
Yes. I have created my own Admin::CommentsController and it conflicts with ActiveAdmin's CommentsController. Changing the name under which comments are registered in initializers/active_admin.rb file solve the problem.  

config.comments_registration_name = 'AdminComment'

Thank you for quick answer !

PR #3903, opened 2015-04-21
Prevent not to define unnecessary path

kysnm:
I configured ActiveAdmin to `config.comments = false`.
And does not create active_admin_comments table.
But /admin/comments had defined.
When accessing to the admin_comments_path, raise a Mysql2::Error.

So I create a patch.
Please review this.
(Sorry it does not wrote a test. Please taught a suitable place)

related #3766

### DIFF
--- a/lib/active_admin/router.rb
+++ b/lib/active_admin/router.rb
@@ -56,6 +56,10 @@ module ActiveAdmin
           # Add on the namespace if required
           unless config.namespace.root?
             nested = routes
+
+            # Do not define unnecessary path
+            return if config.resource_name.route_key == 'comments' && ActiveAdmin.application.comments == false
+
             routes = Proc.new do
               namespace config.namespace.name do
                 instance_exec &nested


timoschilling:
Your code don't work, while `ActiveAdmin.application.comments` is only a global disable for the comment form in the resource. You can enable the comments per resource. Thats way we need the controller even if `ActiveAdmin.application.comments == false`.

If you find a way to disable the routes only if `ActiveAdmin.application.comments == false` and no resource enables the comments a new PR is welcome!

sa-wa-m:
Hi @timoschilling,

Thank you for maintaining awesome product.
I also want to remove unnecessary path.
It seems that all comments are able to be disabled by `config.comments = false` from [the line](https://github.com/activeadmin/activeadmin/blob/master/lib/generators/active_admin/install/templates/active_admin.rb.erb#L121) 
Which specification is correct in active admin???

timoschilling:
@sa-wa-m `config.comments = false` disable the comments global, but you can reenable them in each resource with `config.comments = true`

sa-wa-m:
@timoschilling 
I see. Thank you for your reply.

Issue #675, opened 2011-10-25
table_for inside `show` view doesn't use I18n


Zequez:
If you write the following:

```ruby
show do
  panel 'Comments' do
    table_for post.comments do
      column :user_name
      column :user_email
      column :content
    end
  end
end
```

It only humanize the keys, but it does not call I18n
For it to call I18n you have to do the following:

```ruby
show do
  panel 'Comments' do
    table_for post.comments, i18n: Comment do
      column :user_name
      column :user_email
      column :content
    end
  end
end
```

`table_for` inside the `show` view is just calling the default HTML builder function, so I think it should call specific DSL for the show view (I guess even a helper would do?) that behaves the same way as the `index` table.

seanlinsley:
This should be resolved by #3909

PR #3909, opened 2015-04-24
Detect resource class based on AR collection klass in table_for

kirillplatonov:
References to the issue #675 
Will work only with AR collections. I think this is OK for most cases. If you use custom array, you will still need to set i18n option.

### DIFF
--- a/lib/active_admin/views/components/table_for.rb
+++ b/lib/active_admin/views/components/table_for.rb
@@ -10,8 +10,9 @@ module ActiveAdmin
       def build(obj, *attrs)
         options         = attrs.extract_options!
         @sortable       = options.delete(:sortable)
-        @resource_class = options.delete(:i18n)
         @collection     = obj.respond_to?(:each) && !obj.is_a?(Hash) ? obj : [obj]
+        @resource_class = options.delete(:i18n)
+        @resource_class ||= @collection.klass if @collection.respond_to? :klass
         @columns        = []
         @row_class      = options.delete(:row_class)
 
--- a/spec/unit/views/components/table_for_spec.rb
+++ b/spec/unit/views/components/table_for_spec.rb
@@ -301,6 +301,51 @@ describe ActiveAdmin::Views::TableFor do
       end
     end
 
+    context "when i18n option is specified" do
+      before(:each) do
+        I18n.backend.store_translations :en,
+          activerecord: { attributes: { post: { title: "Name" } } }
+      end
+
+      let(:table) do
+        render_arbre_component assigns, helpers do
+          table_for(collection, i18n: Post) do
+            column :title
+          end
+        end
+      end
+
+      it "should use localized column key" do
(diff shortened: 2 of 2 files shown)


timoschilling:
:+1: good work
can you please apply the Hound comments, than it can be merged

timoschilling:
why you don't apply all comments?

lib/active_admin/views/components/table_for.rb
```diff
@@ -10,8 +10,8 @@ def tag_name
       def build(obj, *attrs)
         options         = attrs.extract_options!
         @sortable       = options.delete(:sortable)
-        @resource_class = options.delete(:i18n)
         @collection     = obj.respond_to?(:each) && !obj.is_a?(Hash) ? obj : [obj]
+        @resource_class = options.delete(:i18n) || @collection.klass rescue nil
```
kirillplatonov: @timoschilling should I replace it with?
```ruby
@resource_class = 
  begin
    options.delete(:i18n) || @collection.klass
  rescue NoMethodError
    nil
  end
```

kirillplatonov:
@timoschilling done

lib/active_admin/views/components/table_for.rb
```diff
@@ -10,8 +10,13 @@ def tag_name
...
         options         = attrs.extract_options!
         @sortable       = options.delete(:sortable)
-        @resource_class = options.delete(:i18n)
         @collection     = obj.respond_to?(:each) && !obj.is_a?(Hash) ? obj : [obj]
+        @resource_class =
+          begin
+            options.delete(:i18n) || @collection.klass
+          rescue NoMethodError
+            nil
+          end
```
timoschilling: we should handle this by:
```ruby
@collection.respond_to?(:klass)
```

kirillplatonov: Fixed. Can use also:
```ruby
@resource_class ||= @collection.klass if @collection.respond_to?(:klass)
```
But it will be less readable

kirillplatonov:
@timoschilling everything is ok?

lib/active_admin/views/components/table_for.rb
```diff
@@ -10,8 +10,11 @@ def tag_name
       def build(obj, *attrs)
         options         = attrs.extract_options!
         @sortable       = options.delete(:sortable)
-        @resource_class = options.delete(:i18n)
         @collection     = obj.respond_to?(:each) && !obj.is_a?(Hash) ? obj : [obj]
+        @resource_class = options.delete(:i18n)
+        if !@resource_class && @collection.respond_to?(:klass)
+          @resource_class = @collection.klass
+        end
```
seanlinsley: It might be better to write it like this:
```ruby
@resource_class ||= @colection.klass if @collection.respond_to? :klass

kirillplatonov: Ok. Changed it

seanlinsley:
Looks good, thanks!

PR #3918, opened 2015-04-28
Option to Prevent Devise Pages From Showing Up in Search Engine Results

tf:
Some of our customers get nervous when they see the sign-in page of their admin backend in google results. Including a `robots` meta tag with content `"noindex, nofollow"` in the head of the logged out pages prevents search engines from indexing these pages.

Add an application setting called `robots_meta_tag_for_logged_out_pages` to control whether this meta tag should be present.

I could see some alternatives to the route taken in this commit:

* Less control: Turn the setting into a boolean. Not sure though if someone might only want `"noindex"` etc.
* More control: Instead add a `meta_tags_for_logged_out_pages` setting to define arbitrary meta tags. Might be too open though.
* Find a better name for this option since it's a bit long.

Finally, one could argue that this could be handled via `robots.txt`. In our case though, we supply ActiveAdmin along with other Rails engines and this change would allow to settle this question once and for all instead of having to edit files in each individual project every time.

Looking forward to your feedback.

### DIFF
--- a/app/views/layouts/active_admin_logged_out.html.erb
+++ b/app/views/layouts/active_admin_logged_out.html.erb
@@ -14,6 +14,10 @@
 
   <%= favicon_link_tag ActiveAdmin.application.favicon if ActiveAdmin.application.favicon %>
 
+  <% if ActiveAdmin.application.robots_meta_tag_for_logged_out_pages.present? %>
+    <%= tag(:meta, name: 'robots', content: ActiveAdmin.application.robots_meta_tag_for_logged_out_pages) %>
+  <% end %>
+
   <%= csrf_meta_tag %>
 </head>
 <body class="active_admin logged_out <%= controller.action_name %>">
--- a/features/robots_meta_tag_for_logged_out_pages.feature
+++ b/features/robots_meta_tag_for_logged_out_pages.feature
@@ -1,0 +1,15 @@
+Feature: Robots Meta Tag for Logged Out Pages
+
+  Prevent Devise pages from showing up in search engine results.
+
+  Background:
+    Given a configuration of:
+    """
+      ActiveAdmin.register Post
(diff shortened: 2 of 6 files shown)


app/views/layouts/active_admin_logged_out.html.erb
```diff
@@ -14,6 +14,10 @@
 
   <%= favicon_link_tag ActiveAdmin.application.favicon if ActiveAdmin.application.favicon %>
 
+  <% if ActiveAdmin.application.robots_meta_tag_for_logged_out_pages.present? %>
```
timoschilling: you can short this to:
```ruby
ActiveAdmin.application.robots_meta_tag_for_logged_out_pages?
```

features/robots_meta_tag_for_logged_out_pages.feature
```diff
@@ -0,0 +1,15 @@
...
+    Given a configuration of:
+    """
+      ActiveAdmin.register Post
+      ActiveAdmin.application.robots_meta_tag_for_logged_out_pages = "noindex, nofollow"
+    """
+
+  Scenario: Logged out views show Favicon
+    Given I am logged out
+    When I am on the login page
+    Then the site should contain a meta tag with name "robots" and content "noindex, nofollow"
```
timoschilling: Please add a empty at the end of the file

timoschilling:
I like the idea of `meta_tags_for_logged_out_pages` and we should add `meta_tags` (for all other pages) too.

tf:
Alright. Shall `meta_tags` be included on logged out pages as well i.e. should there be `meta_tags_for_logged_in_pages`?

timoschilling:
`meta_tags` should only on logged out.

tf:
I think you meant "only on logged in"? See #3919.

timoschilling:
Yes, your are right.

PR #3919, opened 2015-04-28
Support for Custom Meta Tags

tf:
Add `meta_tags` and `meta_tags_for_logged_out_pages` settings to application.

### DIFF
--- a/lib/active_admin/views/pages/base.rb
+++ b/lib/active_admin/views/pages/base.rb
@@ -35,6 +35,10 @@ module ActiveAdmin
               text_node(favicon_link_tag(active_admin_namespace.favicon))
             end
 
+            active_admin_namespace.meta_tags.each do |name, content|
+              text_node(tag(:meta, name: name, content: content))
+            end
+
             text_node csrf_meta_tag
           end
         end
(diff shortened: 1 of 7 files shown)


lib/active_admin/views/pages/base.rb
```diff
@@ -35,6 +35,10 @@ def build_active_admin_head
               text_node(favicon_link_tag(active_admin_namespace.favicon))
             end
 
+            active_admin_application.meta_tags.each do |name, content|
```
timoschilling: Here we need to use `active_admin_namespace`, like in line 35

timoschilling:
What do you think about making `config.meta_tags_for_logged_out_pages = { robots: 'noindex, nofollow' }` default?

tf:
Well, for us it would be great.  Some projects might *want* the sign up/recover password pages to be indexed by google, though.

timoschilling:
I think the most ActiveAdmin use case is not public. If someone will public his pages he can reset the value.

lib/active_admin/views/pages/base.rb
tf: Makes sense.

tf:
Ok. I've amended the commit to make `noindex, nofollow` the default.

timoschilling:
Please apply the one last comment

tf:
I fixed two hound comments. But now there are two others in the cucumber step definitions. Not sure though if breaking cucumber regexes into multiple lines is an improvement...

timoschilling:
It's ok like it is. I merge it after travis pass

tf:
:+1:

PR #3925, opened 2015-05-01
#3044: Use `find_by` instead of `find_by_xxx`

### LINKED ISSUE #3044: Audit the code to use `public_send` wherever possible
Anywhere possible, we should be using `public_send`. Everywhere else, we need to open a ticket for the given dependency, or find another way to implement the same behavior.

avokhmin:
By #3044 , #3090 and according to http://guides.rubyonrails.org/4_2_release_notes.html#adequate-record

Also, should help to prevent:

```
NoMethodError - private method `find_by_id' called for #<Class:0x00000110288708>:
  activeadmin (1.0.0.pre1) lib/active_admin/resource.rb:147:in `find_resource'
```

### DIFF
--- a/spec/unit/resource_spec.rb
+++ b/spec/unit/resource_spec.rb
@@ -232,7 +232,11 @@ module ActiveAdmin
       let(:resource) { namespace.register(Post) }
       let(:post) { double }
       before do
-        allow(Post).to receive(:find_by_id).with('12345') { post }
+        if Rails::VERSION::MAJOR >= 4
+          allow(Post).to receive(:find_by).with("id" => "12345") { post }
+        else
+          allow(Post).to receive(:find_by_id).with("12345") { post }
+        end
       end
 
       it 'can find the resource' do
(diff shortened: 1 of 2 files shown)


avokhmin:
Ok, ready for review.

seanlinsley:
:+1: looks good, but the commits are a bit confusing. Would you mind squashing them together?

spec/unit/resource_spec.rb
```diff
@@ -232,7 +232,11 @@ def config(options = {})
       let(:resource) { namespace.register(Post) }
       let(:post) { double }
       before do
-        allow(Post).to receive(:find_by_id).with('12345') { post }
+        if Rails::VERSION::MAJOR > 3
```
timoschilling: I think it's more clear if you write `>= 4`

timoschilling:
Please squash the last commits too

avokhmin:
Updated, sorry for delay, I had a vacation...

PR #3927, opened 2015-05-03
Allow ApplicationController and helpers to reload in development environment

jvaill:
This fixes https://github.com/activeadmin/activeadmin/issues/697.

The problem is that `PageController` and its ancestors (up until `::InheritedResources::Base`) don't reload by default and thus don't end up inheriting from future changes made to `ApplicationController`. It turns out that autoloading a class with `ActiveSupport::Autoload` does nothing to hook into Rails' reloading mechanism.

This PR moves the controllers to `app/controllers/` which allows Rails to take over its autoloading and its reloading.

A good way to test this is with @timoschilling's minimal test case:
https://github.com/activeadmin/activeadmin_helper_reload_bug

:tada:


seanlinsley:
It seems like there should be a way to allow these files to be reloaded without moving them out of the `lib` directory.

By the way, last year there was a PR that did something similar, that didn't end up being merged: #2906.

jvaill:
Just quickly brainstorming, off the top of my head I can think of a few other feasible solutions:

1. Add the entire `lib/` directory to `ActiveSupport::Dependencies.autoload_paths` and let Rails handle autoloading and reloading. This has the disadvantage of unloading all files inside of `lib/` on each request.
2. Move the controllers to a `lib/controllers/` directory and add that to `ActiveSupport::Dependencies.autoload_paths`.
3. [See if we can use `ActiveSupport::Dependencies.explicitly_unloadable_constants` to get Rails to explicitly unload the controllers after each request.](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/dependencies.rb#L61)
4. ? :smile: 

If we'd rather not shuffle files around, then looking into the third option might be worth it.

timoschilling:
I agree with @seanlinsley, so I'm closing this, but I will copy your ideas over to #697.

javierjulio:
@jvaill thanks. Your changes have been reincorporated in #8180 for now. I've confirmed they fix the helper bug reproductions locally. Just encountering another issue I'd need help with. We'll see if we can finally resolve this. Thank you.

PR #3929, opened 2015-05-03
remove TableFor#is_boolean, detect boolean by attribute type

### LINKED ISSUE #3838: Rails 4.2 still complains about deprecated ActiveRecord #column_for_attribute

Fivell:
do not use database table column to detect boolean attribute

simplified #2753 
fixes #3838

### DIFF
--- a/lib/active_admin/views/components/table_for.rb
+++ b/lib/active_admin/views/components/table_for.rb
@@ -109,19 +109,11 @@ module ActiveAdmin
         elsif item.respond_to? :[]
           item[data]
         end
+        value = status_tag(value.to_s) if [true, false].include?(value)
         value = pretty_format(value) if data.is_a?(Symbol)
-        value = status_tag value     if is_boolean? data, item
         value
       end
-
-      def is_boolean?(data, item)
-        if item.respond_to? :has_attribute?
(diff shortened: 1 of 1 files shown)


Fivell:
this doesn't work for boolean but nil attributes

seanlinsley:
Removing functionality isn't the right way forward. #3930 should resolve the issue.

PR #3930, opened 2015-05-04
Fix the column_for_attribute deprecation warning for the last time

seanlinsley:
#3838

Turns out we couldn’t use `type_for_column` since:
- it’s not intended as a public method
- it would return a null object if the column had the default value (nil)

for example:
```
expected: "No"
     got: ""

(compared using ==)
 (RSpec::Expectations::ExpectationNotMetError)
./features/step_definitions/table_steps.rb:96:in `assert_cells_match'
./features/step_definitions/table_steps.rb:79:in `block (2 levels) in assert_tables_match'
./features/step_definitions/table_steps.rb:75:in `each_index'
./features/step_definitions/table_steps.rb:75:in `block in assert_tables_match'
./features/step_definitions/table_steps.rb:74:in `each_index'
./features/step_definitions/table_steps.rb:74:in `assert_tables_match'
./features/step_definitions/table_steps.rb:115:in `/^I should see the "([^"]*)" table:$/'
features/index/index_as_table.feature:252:in `Then I should see the "index_table_posts" table:'

Failing Scenarios:
cucumber features/index/index_as_table.feature:244 # Scenario: Sorting
```

### DIFF
--- a/lib/active_admin/views/components/table_for.rb
+++ b/lib/active_admin/views/components/table_for.rb
@@ -116,10 +116,8 @@ module ActiveAdmin
       end
 
       def is_boolean?(data, item)
-        if item.respond_to? :has_attribute?
-          item.has_attribute?(data) &&
-            item.column_for_attribute(data) &&
-            item.column_for_attribute(data).type == :boolean
+        if item.class.respond_to? :columns_hash
+          column = item.class.columns_hash[data.to_s] and column.type == :boolean
         end
       end
 
--- a/features/decorators.feature
+++ b/features/decorators.feature
@@ -17,12 +17,14 @@ Feature: Decorators
           column(:id)
           column(:title)
           column(:decorator_method)
+          column(:starred)
         end
       end
     """
     When I am on the index page for posts
     Then I should see "A method only available on the decorator"
     And I should see "A very unique post"
+    And I should see "No"
 
   Scenario: Show page with decorator
     Given a configuration of:
--- a/features/step_definitions/factory_steps.rb
+++ b/features/step_definitions/factory_steps.rb
@@ -5,9 +5,9 @@ end
 
 Given /^(a|\d+)( published)?( unstarred|starred)? posts?(?: with the title "([^"]*)")?(?: and body "([^"]*)")?(?: written by "([^"]*)")?(?: in category "([^"]*)")? exists?$/ do |count, published, starred, title, body, user, category_name|
   count     = count == 'a' ? 1 : count.to_i
-  published = Time.now          if published
+  published = Time.now              if published
   starred   = starred == " starred" if starred
-  author    = create_user(user) if user
+  author    = create_user(user)     if user
   category  = Category.where(name: category_name).first_or_create if category_name
   title   ||= "Hello World %i"
   count.times do |i|
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -106,7 +106,6 @@ After do
 end
 
 Before do
-
   begin
     # We are caching classes, but need to manually clear references to
     # the controllers. If they aren't clear, the router stores references
@@ -119,6 +118,14 @@ Before do
     p $!
     raise $!
   end
+end
+
+# Force deprecations to raise an exception.
+# This would set `behavior = :raise`, but that wasn't added until Rails 4.
+ActiveSupport::Deprecation.behavior = -> message, callstack do
+  e = StandardError.new message
+  e.set_backtrace callstack
+  raise e
 end
 
 # improve the performance of the specs suite by not logging anything
--- a/lib/active_admin/batch_actions/controller.rb
+++ b/lib/active_admin/batch_actions/controller.rb
@@ -30,7 +30,6 @@ module ActiveAdmin
         :filtering,
         :scoping,
         :includes,
-        :collection_decorator
       ].freeze
 
       def batch_action_collection(only = COLLECTION_APPLIES)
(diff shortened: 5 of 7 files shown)


lib/active_admin/views/components/table_for.rb
```diff
@@ -115,10 +115,8 @@ def render_data(data, item)
       end
 
       def is_boolean?(data, item)
-        if item.respond_to? :has_attribute?
-          item.has_attribute?(data) &&
-            item.column_for_attribute(data) &&
-            item.column_for_attribute(data).type == :boolean
+        if item.class.respond_to? :columns_hash
+          column = item.class.columns_hash[data.to_s] and column.type == :boolean
```
seanlinsley: > Line is too long. [81/80]

@timoschilling I'd prefer the maximum to be 90 or 100 characters. Do you have a strong opinion either way?

> Use `&&` instead of `and`.

Heh, static code analysis can only understand so much.

seanlinsley:
It's been a while since I've contributed. Did Travis always run the test suite twice?

screen shot 2015-05-03 at 8 33 53 pm

Fivell:
@seanlinsley , I think in case when  record is decorated  ```item.class.respond_to? :columns_hash``` will always give false

timoschilling:
@seanlinsley your commit run's as a ActiveAdmin repo commit (continuous-integration/travis-ci/push) and as a PR commit (continuous-integration/travis-ci/pr).

seanlinsley:
@Fivell that might be true, in which case we need more tests!

Fivell:
@seanlinsley,  I can add cucumber failing/or not  here  (https://github.com/activeadmin/activeadmin/blob/master/features/index/index_as_table.feature)  feature if you need it

lib/active_admin/views/components/table_for.rb
timoschilling: @seanlinsley There was already a discussion about that in #3841. I think the 120 chars from #3841 are to much, but 90-100 could be ok. In the past I have seen 80 chars as a "soft limit", in cases where a code rewrite is ugly then a long line, I have preferred the long line.

timoschilling: @seanlinsley does the test covers the difference between `&&` and `and`? Not that some will change it and the test don't show it.

deivid-rodriguez:
@Fivell Are you still up for adding that failing test?

timoschilling:
@Fivell maybe you can continue the work on this PR?

seanlinsley:
I'm working adding that test now.

> @seanlinsley your commit run's as a ActiveAdmin repo commit (continuous-integration/travis-ci/push) and as a PR commit (continuous-integration/travis-ci/pr). That's happens only if a ActiveAdmin contributor creates a PR from a ActiveAdmin branch.

It seems like we should make it so `continuous-integration/travis-ci/pr` only happens on master, and on version branches like `0-6-stable`.

seanlinsley:
Okay, this is ready for review.

seanlinsley:
@Fivell this PR stops decorating the batch action collection, because I couldn't figure out how to get around the Draper deprecation from calling `find` on a collection decorator. Unless I'm mis-remembering, before #3775 we didn't decorate batch actions. Why did you add that in the first place?

Fivell:
@seanlinsley , how do you think isn't it confusing when you're declaring decorator, you still have some places where it is decorated and some places where it isn't.

seanlinsley:
It seems like the only way we could continue to decorate the collection is by doing the filtering for the user, which would be a breaking change.
```ruby
batch_action :foo do |collection|
  collection.each{ |record| record.update archived: true }
end
```

We might want to do that for AA 2.0, but for now I don't think we can support decoration.

Fivell:
@seanlinsley , maybe we can introduce new specific  method filtered_collection on controller level to make it possible to use for AA users who want to have decorated items, other can use ```batch_collection``` collection which will be documented about it contains undecorated items ?

seanlinsley:
I don't think it makes sense to provide both `filtered_collection` and `batch_collection`. If you're applying decoration to Active Admin, there's no reason to provide an undecorated version.

A nice way to provide the new DSL without needing to deprecate anything would be to still pass the IDs as a block argument, but provide `collection` as a method that would automatically apply filtering and decoration.
```ruby
batch_action :archive do |ids|
  ids == [1,2,3]
  collection.is_a? Draper::CollectionDecorator # with filtering applied
```
Is there a reason we would use `batch_collection` instead of `collection`?

Fivell:
@seanlinsley , sorry not clear for me how it helps

fabn:
What's the status of this PR? Does it solve the deprecation warning? Thanks.

fabn:
For other seeking for a solution until this is merged. I imported this in my code using a monkey patch, ugly but effective

```ruby
module ActiveAdmin
  # Module used to define patches using Module#prepend
  module Patches
    # Original module
    module Views
      # Monkeypatch for https://github.com/activeadmin/activeadmin/pull/3930
      module TableFor
        protected

        def is_boolean?(data, item)
          if item.class.respond_to? :columns_hash
            (column = item.class.columns_hash[data.to_s]) and column.type == :boolean
          end
        end
      end

    end
  end
end

ActiveAdmin::Views::TableFor.prepend ActiveAdmin::Patches::Views::TableFor
```

seanlinsley:
@fabn this does indeed fix the deprecation. I'm going to go ahead and merge it, moving the decorator discussion to #4211

Issue #101, opened 2011-05-26
Navigation URLs aren't generated correctly when using Passenger's RackBaseURI


joevandyk:
I have the Rails 3 application containing active_admin mounted as a suburi (i.e. http://domain.com/sub-uri/).

The links in ActiveAdmin below the navigation header work properly (http://domain.com/sub-uri/admin/path). 

But the links in the header navigation section don't include the suburi.  They look like http://domain.com/admin/path.

armahillo:
Encountered this problem with AA 0.6.0 and Rails 3.2. 

This fix:
https://github.com/activeadmin/activeadmin/issues/101#issuecomment-22273869

correctly resolved the issue fo rme.

betesh:
@armahillo 's solution worked for me.  Actually, all I needed was the 1st part--didn't need to use `SetEnv RAILS_RELATIVE_URL_ROOT=/foo`

alfonsocora:
The only solution that worked for me with Rails 4.2 and AA from github is to manually set the menu urls like mentioned here https://github.com/activeadmin/activeadmin/issues/101#issuecomment-22273050

pacuna:
yeah, not the best solution but also fixed the issue for me

semen225:
my solution:

- config/environment.rb
```ruby
Rails.application.routes.default_url_options[:script_name] = ENV['RAILS_RELATIVE_URL_ROOT']
```

- app/controller/application_controller.rb
```ruby
protected
	def after_sign_in_path_for(resource)
		ENV['RAILS_RELATIVE_URL_ROOT']
	end

	def after_sign_out_path_for(resource)
		new_user_session_path
	end

	def authenticate_user!
		if user_signed_in?
			super
		else
			redirect_to new_user_session_path
		end
	end
```

app mounted in sub-uri "/record"

- nginx.conf
```
server {
        passenger_user web;
        passenger_group web;
        listen       80;
        server_name  localhost;
        client_max_body_size 20M;
        root   /home/web/ror_apps/share/static;
        passenger_enabled on;
        passenger_base_uri /record;
        rails_spawn_method smart;
        rails_env production;

        location ~ ^/(i|js|css)/ {
            expires 1d;
        }

        error_page   500 502 503 504  /50x.html;
        location = /50x.html {
            root   html;
        }

    }
```

betesh:
@semen225 2 suggestions:
In `authenticate_user!`, right before the redirect, add this:

[...]

PR #3937, opened 2015-05-09
Acknowledge relative url root

coderdan:
Allows resource URLs to acknowledge the Rails relative_url_root setting.

### DIFF
--- a/lib/active_admin/resource/menu.rb
+++ b/lib/active_admin/resource/menu.rb
@@ -26,7 +26,7 @@ module ActiveAdmin
         {
           id: resource_name.plural,
           label: proc{ resource.plural_resource_label },
-          url:   proc{ resource.route_collection_path(params) },
+          url:   proc{ (controller.config.relative_url_root || '') + resource.route_collection_path(params) },
           if:    proc{ authorized?(:read, menu_resource_class) }
         }
       end


timoschilling:
I can't merge this for two reasons.
1. Your change is to hacky
2. https://github.com/activeadmin/activeadmin/issues/101#issuecomment-63240288

coderdan:
OK, perhaps rather than just closing you could give a suggestion as to a "non-hacky" solution? I don't know your code and not sure what sort of a solution you are looking for.

I'm trying to help improve your product.

coderdan:
Any suggestions @timoschilling ? A PR is supposed to be a dialogue to a fix, no?

timoschilling:
I think you can't do it in a not hacky way, while the "problem" is in rails. ActiveAdmin do some mapping and call rails routes helpers

Please read this comments, to get a summery of the discussion:
- https://github.com/activeadmin/activeadmin/issues/101#issuecomment-63240288 `relative_url_root` is deprecated
- https://github.com/activeadmin/activeadmin/issues/101#issuecomment-22273869 fix
- https://github.com/activeadmin/activeadmin/issues/101#issuecomment-25058713 a other @activeadmin/owners close meaning

This is not a ActiveAdmin bug, Rails route helpers are returning a wrong value. But you can configure Rails to returning the right value see https://github.com/activeadmin/activeadmin/issues/101#issuecomment-22273869.

I hope you now understand why I close this PR.

PR #3959, opened 2015-05-23
update zh-TW locales

Neson:
Trying to make zh-TW locales catch up with the English version.


timoschilling:
Pleas squash this commits into one

Neson:
OK, done!

PR #3963, opened 2015-05-25
Add a new option `:download_url_options` to paginated_collection()

knu:
Suppose you are building a user show page with a panel that lists a paginated collection of the user's posts.  In this case, `paginated_collection()` will generate download links with URLs based on the user show page, which do not work.

Using this new option, you can make them work by specifying the correct controller, action and parameters to download the collection.

```
ActiveAdmin.register User do
  show do |user|
    # ...
    panel Post.model_name.human, id: :posts do
      paginated_collection(user.posts.page(params[:posts_page]).per(10),
                           param_name: :posts_page,
                           params: { anchor: 'posts' },
                           download_url_options: {
                             controller: 'admin/posts',
                             action: 'index',
                             params: { user_id: user.id }
                           }) do
      # ...
    end
  end
end
```

### DIFF
--- a/lib/active_admin/view_helpers/download_format_links_helper.rb
+++ b/lib/active_admin/view_helpers/download_format_links_helper.rb
@@ -7,9 +7,13 @@ module ActiveAdmin
         div class: "download_links" do
           span I18n.t('active_admin.download')
           formats.each do |format|
-            a format.upcase, href: url_for(params: params, format: format)
+            a format.upcase, href: download_url_for(params: params, format: format)
           end
         end
+      end
+
+      def download_url_for(options = {})
+        url_for(options)
       end
 
       def self.included base
--- a/lib/active_admin/views/components/paginated_collection.rb
+++ b/lib/active_admin/views/components/paginated_collection.rb
@@ -41,6 +41,7 @@ module ActiveAdmin
         @params         = options.delete(:params)
         @param_name     = options.delete(:param_name)
         @download_links = options.delete(:download_links)
+        @download_url_options = options.delete(:download_url_options) || {}
         @display_total  = options.delete(:pagination_total) { true }
         @per_page       = options.delete(:per_page)
 
@@ -60,6 +61,10 @@ module ActiveAdmin
         else
           super
         end
+      end
+
+      def download_url_for(options = {})
+        url_for(options.merge(@download_url_options))
       end
 
       protected
--- a/lib/active_admin/views/components/paginated_collection.rb
+++ b/lib/active_admin/views/components/paginated_collection.rb
@@ -41,7 +41,7 @@ module ActiveAdmin
         @params         = options.delete(:params)
         @param_name     = options.delete(:param_name)
         @download_links = options.delete(:download_links)
-        @download_url_options = options.delete(:download_url_options) || {}
+        @download_url_options = options.delete(:download_url_options) { {} }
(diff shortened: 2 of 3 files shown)


lib/active_admin/view_helpers/download_format_links_helper.rb
```diff
@@ -7,11 +7,15 @@ def build_download_format_links(formats = self.class.formats)
...
-            a format.upcase, href: url_for(params: params, format: format)
+            a format.upcase, href: download_url_for(params: params, format: format)
           end
         end
       end
 
+      def download_url_for(options = {})
+        url_for(options)
+      end
+
```
timoschilling: Why this method with is only an alias?

lib/active_admin/views/components/paginated_collection.rb
```diff
@@ -62,6 +63,10 @@ def add_child(*args, &block)
         end
       end
 
+      def download_url_for(options = {})
+        url_for(options.merge(@download_url_options))
+      end
+
```
timoschilling: It look's like this method is not used, or have I missed something?

knu: This definition overrides DownloadFormatLinksHelper#download_url_for.

lib/active_admin/views/components/paginated_collection.rb
```diff
@@ -41,6 +41,7 @@ def build(collection, options = {})
         @params         = options.delete(:params)
         @param_name     = options.delete(:param_name)
         @download_links = options.delete(:download_links)
+        @download_url_options = options.delete(:download_url_options) || {}
```
timoschilling: change this to:
@download_url_options = options.delete(:download_url_options, {})

lib/active_admin/view_helpers/download_format_links_helper.rb
knu: This method will be overridden by ActiveAdmin::Views::PaginatedCollection when included by the class, so you can omit this.  However, I thought it should be defined here just in case this helper module gets included by other view component classes.

lib/active_admin/views/components/paginated_collection.rb
knu: You mean `options.delete(:download_url_options) { {} }` ?

timoschilling: It's ok, too

knu: Done!

knu:
@timoschilling Is there anything necessary to improve or consider before this can be merged?

knu:
Anyone out there who wants this?  You can even put multiple `paginated_collection`'s on a page, so I think there should be a way to tweak download links.

knu:
@timoschilling Is there any chance of getting this merged?  If there's not enough interest, that's perfectly fine and I'll just make a monkey patch for my personal use.  Thanks for your time and consideration!

timoschilling:
I understand your problem, but it's a edge case and the solution is a to much.

knu:
Ouch.

knu:
For future viewers, there is no easy way to monkey-patch this.  Just specify `download_links: false` and place links yourself.

timoschilling:
We can leaf this PR open, maybe we find a better solution

PR #3970, opened 2015-05-28
require formtastic_i18n

debbbbie:
I found that locale of `formtastic_i18n` do not load auto!
I then spent many hours to found it out, even create a temp gem `formtastic-i18n`, and study how `ransack` auto works for i18n.

And finally I succeed, is just `require 'formtastic_i18n'`

:smile_cat:


lib/active_admin.rb
```diff
@@ -6,6 +6,7 @@
 require 'bourbon'
 require 'kaminari'
 require 'formtastic'
+require 'formtastic_i18n'
```
debbbbie: @houndci 
Do you mean `require "formtastic_i18n"`

timoschilling: @debbbbie @houndci is just a bot ;)

PR #3977, opened 2015-06-05
Customize Comment Menu

whitehat101:
Instead of Removing the Comments completely, I wanted to control their position in the default menu.

This PR creates a `comments_menu` config option that is passed to the `menu` in `ActiveAdmin::Comment`. This gives users full control over this item's position in the menu.

### DIFF
--- a/lib/active_admin/orm/active_record/comments.rb
+++ b/lib/active_admin/orm/active_record/comments.rb
@@ -7,6 +7,7 @@ require 'active_admin/orm/active_record/comments/resource_helper'
 ActiveAdmin::Application.inheritable_setting :comments,                   true
 ActiveAdmin::Application.inheritable_setting :show_comments_in_menu,      true
 ActiveAdmin::Application.inheritable_setting :comments_registration_name, 'Comment'
+ActiveAdmin::Application.inheritable_setting :comments_menu, {}
 
 # Insert helper modules
 ActiveAdmin::Namespace.send :include, ActiveAdmin::Comments::NamespaceHelper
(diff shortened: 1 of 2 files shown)


timoschilling:
I like your idea, but I will simplify it.

What do you think about:
- removing `inheritable_setting :show_comments_in_menu`
- changing `inheritable_setting :comments_menu, {}` or `inheritable_setting :comments_menu, false`
- changing
```erb
  # == Admin Comments
  #
  # This allows your users to comment on any resource registered with Active Admin.
  #
  # You can completely disable comments:
  # config.comments = false
  #
  # You can change the name under which comments are registered:
  # config.comments_registration_name = 'AdminComment'
  #
  # You can disable the menu item for the comments index page:
  # config.comments_menu = false
  #
  # You can customize the comment menu
  # config.comments_menu = { parent: 'Admins', priority: 1 }
```

whitehat101:
I like the your thoughts about combining the settings. The (only) advantage to my PR as-is is that it will be 100% backwards compatible with the existing configuration options.

If you're not concerned about compatibility, merging these options sounds excellent.

timoschilling:
I think, thats only a little brake, that we can do in a major release.

snowmoonsoftware:
What's the status of this? Can we get it merged?

timoschilling:
After some apply my comment https://github.com/activeadmin/activeadmin/pull/3977#issuecomment-121171839

drn:
@timoschilling - added another commit that addresses your comments https://github.com/activeadmin/activeadmin/pull/4187
would love to get this merged in as well

seanlinsley:
Closing this in favor of #4187

Issue #3994, opened 2015-06-16
Ransack definition issue upgrading activeadmin from 1.0.0.pre to 1.0.0.pre1


swaroopmurthy:
We have model classes defined as a separate gem from the activeadmin app. Upon upgrading from 1.0.0.pre to 1.0.0.pre1 we get the following error:

    Error Message: undefined method `ransack' for #<#<Class:0x4bf754f0>:0x2aec155>

It's trying to call "ransack" on a presenter object. The object doesn't have "ransack" nor "search" methods, please document how to upgrade.

Fivell:
@swaroopmurthy . did you try master branch ? 
I think this was fixed here https://github.com/activeadmin/activeadmin/commit/49eea30553f12ee5e516392470faa0bb15459009

swaroopmurthy:
@Fivell wow, that does work. please discard the pull request. thank you :+1:

PR #3995, opened 2015-06-16
Fix for undefined ransack method #3994

swaroopmurthy:
Temporary fix for undefined ransack, is there a more elegant way of doing this?


timoschilling:
closed like by @swaroopmurthy in https://github.com/activeadmin/activeadmin/issues/3994#issuecomment-112549880 requested

PR #4004, opened 2015-06-23
Fix typos in #render_in_context documentation

davydovanton:


davydovanton:
@timoschilling thanks :star2:

PR #4018, opened 2015-07-02
batch_action_path includes :q and :scope params

### LINKED ISSUE #4000: filtering does not get applied to batch_action_collection
When performing a batch action on a filtered set of records, batch_action_collection still returns all records. This appears to be because there's nothing in the BatchActionForm that ensures that `params[:q]` will be included in the post request, either in the action url or as a hidden input. 

Is there something I should be doing to ensure batch_action_collection is filtered, or is this a bug?

Fivell:
fixes #4000


timoschilling:
Please apply houndci

Fivell:
@timoschilling , ready

Issue #3914, opened 2015-04-27
How to handle :symbol values?


matthewrudy:
We use [simple_enum](https://github.com/lwe/simple_enum) for enums.

Example

``` ruby
class User < ActiveRecord::Base
  as_enum :gender, female: 1, male: 0
end
```

But using this with the latest active admin
``` ruby
show do
  row "gender", :gender
end
```

This raises an exception `can't define singleton` (or something like that).

The key causes is `pretty_format` that has no implementation for symbols.

``` ruby
def pretty_format(object)
  case object
  when String, Numeric, Arbre::Element
    object.to_s
  when Date, Time
    localize object, format: :long
  else
    if defined?(::ActiveRecord) && object.is_a?(ActiveRecord::Base) ||
       defined?(::Mongoid)      && object.class.include?(Mongoid::Document)
      auto_link object
    else
      display_name object
    end
  end
end
```

My solution was to hack it;

``` ruby
module ActiveAdmin::ViewHelpers::DisplayHelper
  alias :pretty_format_before_symbol_hack :pretty_format
  def pretty_format(object)
    if object.is_a?(Symbol)
      object.to_s
    else
      pretty_format_before_symbol_hack(object)
    end
  end
end
```
But a nicer solution would be welcome :)

timoschilling:
There is something on the road, take a look on the [pretty_format_api](https://github.com/activeadmin/activeadmin/tree/idea/pretty_format_api) branch.

matthewrudy:
Ah, great.
So we'd just register a formatter.

That's a good solution.

adambrod:
Same thing started failing for me on activeadmin 1.0.0.pre1 when we upgraded to ruby 2.2. I'll put in the stack trace to help others track this down. Using the hack from @matthewrudy gives us a workaround. Thanks!

```
 Failure/Error: click_on "Update Retailer"
     ActionView::Template::Error:
       can't define singleton
     # ./vendor/bundle/ruby/2.2.0/gems/activeadmin-1.0.0.pre1/lib/active_admin/view_helpers/method_or_proc_helper.rb:86:in `instance_exec'
     # ./vendor/bundle/ruby/2.2.0/gems/activeadmin-1.0.0.pre1/lib/active_admin/view_helpers/method_or_proc_helper.rb:86:in `render_in_context'
[...]

PR #4023, opened 2015-07-08
Enable pretty_format to handle Symbols

adambrod:
Very basic solution to https://github.com/activeadmin/activeadmin/issues/3914 - if the model returns a Symbol, just call to_s on it as we do for numbers and strings.


matthewrudy:
That's simple enough.
I don't know why I didn't PR it before.

Guess I thought there was a reason you handled symbols that way.

:+1:

PR #3978, opened 2015-07-14
Form Builder has_many doesn't work with has_one relationship

timoschilling:
The documentation for forms (5-forms.md) says a has_many can be used with a has_one relationship, and shows f.has_many :comment as an example. As far as I can tell, this doesn't work at all when creating a new record.

You'll get an exception `undefined method '<<' for nil:NilClass` when you make the `f.has_many` call. This exception is from lib/active_admin/form_builder.rb:67, because `contents` variable is nil. `contents` is the result of formtastic/helpers/inputs_helper.rb `inputs_for_nested_attributes`, whose last line is a call to action_view/helpers/form_helper.rb's `field_for` with only one parameter (the second parameter, `record_object`, is nil). `fields_for` calls  `fields_for_with_nested_attributes` which ends up failing because the `association` parameter is nil.

....

It's broken and I'm tired. It works if I use a has_many relationship. If it's not supposed to work then the documentation needs to be updated.

### DIFF
--- a/lib/active_admin/form_builder.rb
+++ b/lib/active_admin/form_builder.rb
@@ -61,7 +61,7 @@ module ActiveAdmin
         end
         
         template.assign(has_many_block: true)
-        contents = without_wrapper { inputs(options, &form_block) }
+        contents = without_wrapper { inputs(options, &form_block) } || "".html_safe
 
         if builder_options[:new_record]
           contents << js_for_has_many(assoc, form_block, template, builder_options[:new_record], options[:class])
--- a/spec/support/rails_template.rb
+++ b/spec/support/rails_template.rb
@@ -38,13 +38,20 @@ inject_into_file 'app/models/blog/post.rb', %q{
 }, after: 'class Blog::Post < ActiveRecord::Base'
 
 
+generate :model, "profile user_id:integer bio:text"
 generate :model, "user type:string first_name:string last_name:string username:string age:integer"
 inject_into_file 'app/models/user.rb', %q{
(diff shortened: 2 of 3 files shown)


sfoop:
I'm not sure if I should open a new ticket for this, but it's still not fixed properly. The problem is if you have code like this,
```ruby
f.has_many :stuff do |has_many_form|
  # code goes here
end
```
The `has_many_form.object` will be nil. As a workaround you can do this,

```ruby
f.has_many :stuff, for: [:stuff, f.object.stuff] do |has_many_form|
  # code goes here
end
```

Even with this I have an issue with it, because there will be a Remove button added to the relation even though I've supplied `allow_destroy: false`, due to the code on line 82 (the `has_many_actions` function's first line of code). There isn't a way of preventing that Remove button from showing.

buermann:
I have neither problem with the has_one relationships in my app using the syntax of the first example.

sfoop:
I've changed the way the form and some of the data is structured since I wrote that, and doing a quick check it now works if I use the first form (－‸ლ) I'm debugging a new project that's not mine; if I'm able to trip it up again I'll update.

lucaDgt:
`allow_destroy: false` is still not working to me,  for an `has_one` relationship. I cannot hide that 'Remove' button appearing.

buermann:
Yeah the "remove" button is for editing new, unpersisted records, which can always be removed. The allow_destroy: false option removes the "delete" checkbox for existing records.

PR #3792, opened 2015-02-13
WIP: Support for JRuby-9000

kennethkalmer:
Will bring in the combined efforts of previous attempts to get activeadmin running on jruby-9000, slated to be the next major JRuby release.

- [x] Get dependencies installed
- [x] Get the tests running
- [ ] Fix any test failures
- [x] Hook up to CI
- [ ] Prosper

PR #4032, opened 2015-07-17
Jruby compat

deivid-rodriguez:
Supersedes #3792


deivid-rodriguez:
@brennovich The travis build timed out. The `-X+O` flag is too slow. I've removed it for now. The build was not failing for me without and maybe they've fixed the original issue you were having upstream...

deivid-rodriguez:
It worked!! It only failed for Rails 3.2.22 but this Rails version is excluded from the build for MRI 2.2, so I guess you don't expect to support that combination. Since jruby-head is 2.2 compatible, I've excluded that Rails version too.

Now the build should be green, let's cross fingers.

deivid-rodriguez:
yay! :smile:

brennovich:
\o/

brennovich:
Weird, I didn't look closely, but it seems that the memory leak specs were removed, otherwise they shouldn't be passing as jRuby does not have ObjectSpace emulation enabled by default (only with -X+O)

deivid-rodriguez:
You're totally right. The specs are there but I guess they're not getting run, because running them separately fail without the flag.

I'm on it, the idea is to enable the flag for that single test only.

deivid-rodriguez:
Ok, I think I'm now happy with this. Sorry for the `hound` noise.

Let's see what `travis` says and let me know your feedback!

brennovich:
Wow! Great work, well documented commits :)

Thanks \o/

timoschilling:
Good job!

deivid-rodriguez:
Thanks!

seanlinsley:
:heart: :heart: :heart: I'm so glad to see this finally happened

PR #4051, opened 2015-08-03
Configure button radius via SCSS variables

jclay:
This makes it easy to adjust the rounded corners on the buttons if you're not a fan of the default style.

A user wanting to remove the rounded corners can simply add the following:
```scss
$basic-button-radius: 0px; //all buttons
$tabs-radius: 0px; //navigation tabs
```


timoschilling:
I need to close this, while it will be conflict with #3862 wich will be merged son and already has this feature.

PR #4060, opened 2015-08-08
Update es.yml

kennym:
Provide spanish translations for `comments.created_at` and `comments.updated_at`

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -72,6 +72,7 @@ es:
       labels:
         destroy: "Borrar"
     comments:
+      created_at: "Fecha de creación"
       resource_type: "Tipo de recurso"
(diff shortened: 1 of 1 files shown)


timoschilling:
`updated_at

timoschilling:
`updated_at` is not used by the `en.yml` so we can't add it for `es.yml`. Please update you PR to add only `created_at` or create a new PR.

kennym:
@timoschilling done.

I wonder why we don't add `updated_at` to the local `en.yml`?

timoschilling:
Can you please squash both commits and force push your branch.
`updated_at` is not needed anywhere, so it is not added.

kennym:
@timoschilling done! :)

PR #4082, opened 2015-08-19
CSV Builder: iterate until we find an empty page

matthewrudy:
Current functionality of the CSV builder:
- ask kaminari for the `total_pages`
- iterate over that number of pages

Issues:
- `total_pages` issues a `count`
- our table has many millions of records and a `count` is very slow

Solution:
- iterate over kaminari pages
- break when we hit an empty page

### DIFF
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -53,11 +53,9 @@ module ActiveAdmin
         csv << CSV.generate_line(columns.map{ |c| encode c.name, options }, csv_options)
       end
 
-      (1..paginated_collection.total_pages).each do |page|
-        paginated_collection(page).each do |resource|
-          resource = controller.send :apply_decorator, resource
-          csv << CSV.generate_line(build_row(resource, columns, options), csv_options)
-        end
+      each_resource do |resource|
+        resource = controller.send :apply_decorator, resource
+        csv << CSV.generate_line(build_row(resource, columns, options), csv_options)
       end
 
       csv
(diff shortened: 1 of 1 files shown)


matthewrudy:
An alternative would be to use `find_each` and skip kaminari entirely.
Not sure if the order of a CSV matters

matthewrudy:
For extra information, we have a kaminari hack in place:
``` ruby
module Kaminari::ActiveRecordRelationMethods
  def total_count(column_name = :all, options = {})
    @total_count ||= begin
      c = except(:offset, :limit, :order)
      c = c.except(:includes) unless references_eager_loaded_tables?

      self.connection.unprepared_statement do
        query_plan = self.connection.execute("EXPLAIN #{c.to_sql}").first["QUERY PLAN"]
        query_plan.match(/.*rows=(\d+).*/)[1].to_i
      end
    end
  end
end
```

This makes calculating totals very fast on our large datasets
but makes the total inaccurate.

That's fine for pagination,
but means the CSV generation is incorrect.

seanlinsley:
> An alternative would be to use find_each and skip kaminari entirely.
>Not sure if the order of a CSV matters

We actually used `find_each` before #3375.

seanlinsley:
The desire in #3375 was to be able to iterate over a collection that had a `group by` clause applied, hence ordering by ID as `find_each` does wouldn't work. Seems like we should make pagination configurable. For example:
```ruby
csv paginate_with: :kaminari
csv paginate_with: :find_each
csv paginate_with: ->collection{ collection.each{ |r| yield r }  }

matthewrudy:
@seanlinsley thanks for your time.

I like your solution,
if it can be made to work.

javierjulio:
Closing in favor of #4209 as its been agreed upon that its the better solution. Although both seem to have gone stale.

Issue #3693, opened 2014-12-12
Numbering items in Index as a Table report


bnussey:
Is it possible to get the sequential number of each item in ActiveAdmin?

So it would look like this:

    | Number | Name  |

    | 1      | Bill  |

    | 2      | Smith |  

I know this can be achieved with each_with_index but don't think there is a way to use this with ActiveAdmin.

Could something be added in as controller do ?

I want to add this functionality into an Index as a Table report so for example

    index do
        column "Number" do |a|
           # Some way of specifying index number         
       end
       column "Name" do |a|
           a.name
       end
    end

Thanks

seanlinsley:
Sounds like this has been implemented in #4093

PR #4093, opened 2015-08-23
Added index_column to display the position of the row in the total collection.

cadetstar:
This is a proposed solution to fulfill the Feature Request activeadmin/activeadmin#3693

It only extends IndexAsTable for now, but could be built upon to do the same in other index styles.

### DIFF
--- a/docs/3-index-pages/index-as-table.md
+++ b/docs/3-index-pages/index-as-table.md
@@ -55,6 +55,8 @@ index do
 end
 ```
 
+## Defining Actions
+
 To setup links to View, Edit and Delete a resource, use the `actions` method:
 
 ```ruby
--- a/docs/3-index-pages/index-as-table.md
+++ b/docs/3-index-pages/index-as-table.md
@@ -55,8 +55,6 @@ index do
 end
 ```
 
-## Defining Actions
-
 To setup links to View, Edit and Delete a resource, use the `actions` method:
 
 ```ruby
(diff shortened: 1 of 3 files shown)


lib/active_admin/views/index_as_table.rb
```diff
@@ -244,6 +254,12 @@ def selectable_column
           end
         end
 
+        def index_column
+          column '#', class: 'col-index', sortable: false do |resource|
+            @collection.offset_value + @collection.index(resource) + 1
```
timoschilling: I wold like to see the `+1` configurable:
```ruby
def index_column(offset = 1)
```

timoschilling: Then the developer can make a `0` or `1` based index

lib/active_admin/views/index_as_table.rb
```diff
@@ -108,6 +108,16 @@ module Views
...
+    # In addition, you can insert the position of the row in the greater collection by using the index_column special command:
+    #
+    # ```ruby
+    # index do
+    #   selectable_column
+    #   index_column
+    #   column :title
+    # end
+    # ```
+    #
```
timoschilling: Can you add the same to the docs please.

docs/3-index-pages/index-as-table.md
```diff
@@ -55,8 +55,6 @@ index do
 end
 ```
 
-## Defining Actions
```
timoschilling: Why did you remove this headline?

cadetstar: That was removed when I ran rake yard, but I didn't change the source file in that location.  I can add it back, but it had already been removed from the source documentation.

timoschilling:
please make a rebase on master

PR #4113, opened 2015-09-05
Ruby 2.2.2 fixes

ka8725:
These are fixes for https://github.com/activeadmin/activeadmin/pull/4112

Sorry those was an old branch.

### DIFF
--- a/lib/active_admin/helpers/collection.rb
+++ b/lib/active_admin/helpers/collection.rb
@@ -3,13 +3,13 @@ module ActiveAdmin
     module Collection
       # 1. removes `select` and `order` to prevent invalid SQL
       # 2. correctly handles the Hash returned when `group by` is used
-      def collection_size(c = collection)
+      def collection_size(c = collection())
         c = c.except :select, :order
 
         c.group_values.present? ? c.count.count : c.count
       end
 
-      def collection_is_empty?(c = collection)
+      def collection_is_empty?(c = collection())
         collection_size(c) == 0
       end
     end


ka8725:
These are fixes for https://github.com/activeadmin/activeadmin/pull/4112

Sorry those was an old branch.

lib/active_admin/helpers/collection.rb
```diff
@@ -3,14 +3,14 @@ module Helpers
     module Collection
       # 1. removes `select` and `order` to prevent invalid SQL
       # 2. correctly handles the Hash returned when `group by` is used
-      def collection_size(c = collection)
-        c = c.except :select, :order
+      def collection_size(collection)
+        collection = collection.except :select, :order
 
-        c.group_values.present? ? c.count.count : c.count
+        collection.group_values.present? ? collection.count.count : collection.count
```
ka8725: How do you count i heret? :smile:

timoschilling:
What do you want to fix with this change?

lib/active_admin/helpers/collection.rb
```diff
@@ -3,14 +3,18 @@ module Helpers
     module Collection
       # 1. removes `select` and `order` to prevent invalid SQL
       # 2. correctly handles the Hash returned when `group by` is used
-      def collection_size(c = collection)
-        c = c.except :select, :order
+      def collection_size(collection)
+        collection = collection.except :select, :order
```
seanlinsley: Instead of changing the API, it seems better to do this:
```ruby
def collection_size(collection = collection())

File: lib/active_admin/helpers/collection.rb [hunk 3]
```diff
@@ -3,13 +3,13 @@ module Helpers
     module Collection
       # 1. removes `select` and `order` to prevent invalid SQL
       # 2. correctly handles the Hash returned when `group by` is used
-      def collection_size(c = collection)
+      def collection_size(c = collection())
```
ka8725: Right, it works either. What I actually fixing here are warnings like [these](https://github.com/rails/rails/pull/18160/files). Actually activeadmin doesn't work on Ruby >= 2.2.0 without this fix at all.

lib/active_admin/helpers/collection.rb
timoschilling: :+1:

timoschilling:
Please make a rebase on master

seanlinsley:
I'm on Ruby 2.2.3 and I'm not getting any warnings, so it seems this change isn't needed.

PR #4116, opened 2015-09-08
Don't pass encoding_options to CSV

keichan34:
CSV will raise an error when you pass it an unrecognized key, but it's
used in #encode.

This patch filters the encoding_options CSV option out of the hash that is passed to CSV.

### DIFF
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -49,13 +49,17 @@ module ActiveAdmin
       end
 
       if options.delete(:column_names) { true }
-        receiver << CSV.generate_line(columns.map{ |c| encode c.name, options }, options)
+        receiver << CSV.generate_line(
+          columns.map { |c| encode c.name, options },
+          options.except(:encoding_options))
       end
 
       (1..paginated_collection.total_pages).each do |page_no|
         paginated_collection(page_no).each do |resource|
-           resource = controller.send :apply_decorator, resource
-           receiver << CSV.generate_line(build_row(resource, columns, options), options)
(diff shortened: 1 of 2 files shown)


keichan34:
I've fixed the Hound errors, but I don't think the test failures have been caused by this patch?

lib/active_admin/csv_builder.rb
```diff
@@ -48,14 +48,20 @@ def build(controller, receiver)
         receiver << byte_order_mark
       end
 
+      options_for_csv = options.reject { |k, _v| k == :encoding_options }
+
       if options.delete(:column_names) { true }
-        receiver << CSV.generate_line(columns.map{ |c| encode c.name, options }, options)
+        receiver << CSV.generate_line(
+          columns.map { |c| encode c.name, options },
+          options_for_csv)
```
seanlinsley: It's cleaner to use [`except`](http://api.rubyonrails.org/classes/Hash.html#method-i-except):
```ruby
options_for_csv = options.except :encoding_options
```

But I'd prefer this code style:
```ruby
columns = columns.map{ |c| encode c.name, options }
receiver << CSV.generate_line(columns, options.except(:encoding_options))
```

seanlinsley:
Please add a test for this.

timoschilling:
and please make a rebase on master

keichan34:
Added a couple tests, rebased on master, and changed to use `options.except`.

timoschilling:
It look's like your change don't run on ruby 1.9

keichan34:
Ah yes. Sorry about that, tests are running on 1.9 now.

Issue #2692, opened 2013-11-11
Handle huge associations


seanlinsley:
AA users are commonly frustrated by our naive building of dropdowns everywhere for a given association, because that option isn't scalable when you have millions of records.

The most obvious option is to allow the developer to blacklist certain models or tables:

``` ruby
ActiveAdmin.setup do |config|
  config.huge_db_tables = %w[tags categories user_actions]
end
```

The question then becomes: what do you replace it with?

As far as I can tell, there are three options:
- just remove it
- use [AJAX](https://github.com/meltingice/ajax-chosen)
- use Rails' `pluck` method to just grab a name and an ID

So which would be best, as far as integrating into the official repo? (both for filters and forms)

drush:
@igorbernstein Your approach looks pretty interesting.  Have you gotten anyone else to collaborate with you on completing the spec prototype?  Is this code in a branch/PR somewhere?

ryanwaggoner:
+1

franklsf95:
Would be a really nice addition. Now I need to manually exclude all the huge associations from generating huge dropdown lists.

osmancrush:
+1

benoittgt:
+1

Epiju:
+1

wkoffel:
+2

thinkalike-dev:
+1

kiote:
+1

dmitry:
Plus one, +1 or :+1: cannot help. Instead please tell how would you like to use it and how the DSL should look like.

kiote:
DSL [here](https://github.com/activeadmin/activeadmin/issues/2692#issuecomment-29355831) seems to be pretty clear, what do you think?

aaronchi:
+1 for this DSL

mwlang:
Is anything being done with this or is it sitting idle?

dmitry:
@mwlang nothing done since: https://github.com/activeadmin/activeadmin/issues/2692#issuecomment-29355831

raldred:
Proposed DSL seems fairly clean, I'd be happy to proceed. It seems consensus is all good. I'm interested in picking this up to implement, does anyone already have this WIP?

mwlang:
I thought about doing it myself, but I'm tight on time at the moment due to project deadlines.  However, if you get started on it, you've got a keen collaborator/tester in me.

dmitry:
@raldred :+1: could be tested by me aswell.

[...]

PR #4125, opened 2015-09-17
Add configuration for including association filters by default

danielevans:
In many situations loading all records of a model for default association filters is unreasonable.

This change adds a `include_default_association_filters` setting which can be used to disable the inclusion of association filters by default without affecting other default filters.

### DIFF
--- a/lib/active_admin/filters/resource_extension.rb
+++ b/lib/active_admin/filters/resource_extension.rb
@@ -107,7 +107,11 @@ module ActiveAdmin
 
       # @return [Array] The array of default filters for this resource
       def default_filters
-        default_association_filters + default_content_filters + custom_ransack_filters
+        [].tap do |result|
+          result.concat default_association_filters if namespace.include_default_association_filters
+          result.concat default_content_filters
+          result.concat custom_ransack_filters
+        end
       end
 
       def custom_ransack_filters
--- a/lib/active_admin/filters/resource_extension.rb
+++ b/lib/active_admin/filters/resource_extension.rb
@@ -107,11 +107,11 @@ module ActiveAdmin
 
       # @return [Array] The array of default filters for this resource
       def default_filters
-        [].tap do |result|
-          result.concat default_association_filters if namespace.include_default_association_filters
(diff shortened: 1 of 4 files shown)


JustinAiken:
:+1:

lib/active_admin/filters/resource_extension.rb
```diff
@@ -107,7 +107,11 @@ def filter_lookup
 
       # @return [Array] The array of default filters for this resource
       def default_filters
-        default_association_filters + default_content_filters + custom_ransack_filters
+        [].tap do |result|
+          result.concat default_association_filters if namespace.include_default_association_filters
+          result.concat default_content_filters
+          result.concat custom_ransack_filters
+        end
```
seanlinsley: `tap` isn't beginner-friendly, because it isn't clear what it does. this would be cleaner:
```ruby
filters = []
filters << default_association_filters if namespace.include_default_association_filters
filters << default_content_filters
filters << custom_ransack_filters
filters

File: spec/unit/filters/resource_spec.rb [hunk 2]
```diff
@@ -32,6 +32,13 @@
     expect(resource.filters).to be_empty
   end
 
+  it "should return the defaults without associations if default association filters are disabled on the namespace" do
```
seanlinsley: this could be rewritten as:
> should exclude association filters when disabled

lib/active_admin/filters/resource_extension.rb
danielevans: I was aiming more for cleanliness than beginner friendliness, I apologize. I've altered my code.

I will still stick with `concat` since `<<` produces a different and undesirable result and I think it is preferable to avoid variable names which match method names so I will leave out changing the variable name to `filter`.

seanlinsley:
Whenever #2692 gets built this setting won't be needed anymore, but for now it makes sense. Thanks for the PR!

Adamantish:
@danielevans Thanks for this. Got the line in our config file commented out and ready to go when we upgrade rails. Just had to do our own override in the meantime.

PR #4129, opened 2015-09-20
4128 add indonesian translations

saveav:
Adding support for Indonesian translation/locale in ActiveAdmin

### DIFF
--- a/config/locales/id.yml
+++ b/config/locales/id.yml
@@ -1,0 +1,135 @@
+id:
+  active_admin:
+    dashboard: Dashboard
+    dashboard_welcome:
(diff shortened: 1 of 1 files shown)


timoschilling:
I can't accept this PR, provide it only with the locale change

saveav:
@timoschilling OK, I can force push it

saveav:
Hi @timoschilling, I already made the changes. But part of the test failed, due to RAILS 3.

timoschilling:
The Rails 3 test don't fail because of your change, every thing is fine with you PR now

PR #4130, opened 2015-09-21
Support jquery-ui-rails version 4.* and 5.*

pcreux:
I came across several apps where the main pain point adding ActiveAdmin was to upgrade to the latest `jquery-ui-rails` as it was affecting the front-end. I believe that removing the version of `jquery-ui-rails` ActiveAdmin depends on makes it easier to add ActiveAdmin to an existing app.

Looking at https://github.com/activeadmin/activeadmin/commit/7b9ad75a93f5edaacdfef09cbf5e1d999826a952 I figured that we lock the dependency to `jquery-ui-rails` version **5+** because the load path is different from **4+**. Looking at the discusion on https://github.com/sstephenson/sprockets/issues/90 I figured that we can support `jquery-ui-rails` **4.*** and **5.*** by taking advantage of the `require_asset` method.

Here are two screenshots where I run ActiveAdmin on `jquery-ui-rails` **4.*** and **5.***:

**jquery-ui-rails 4.***
screen shot 2015-09-21 at 12 06 32 pm

**jquery-ui-rails 5.***
screen shot 2015-09-21 at 12 05 40 pm

I also changed the test suite to run against `jquery-ui-rails 4.0+` on Rails 3.2 and `jquery-ui-rails 5.0+` on Rails 4.* to ensure that we support both versions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,9 @@ require File.expand_path 'spec/support/detect_rails_version', File.dirname(__FIL
 rails_version = detect_rails_version
 gem 'rails', rails_version
 
+jquery_ui_rails_version = rails_version > "4" ? "~> 5.0" : "~> 4.0"
+gem 'jquery-ui-rails', jquery_ui_rails_version
+
 gem 'execjs', '~> 2.4.0' # ~> 2.5.0 works only for Ruby > 2.0
 
 # Optional dependencies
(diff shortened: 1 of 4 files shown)


pcreux:
Tests are currently failing on Travis CI on all branches for some reason. Reverting back to an old passing commit does not fix it.

seanlinsley:
I wasn't aware this was possible. Thanks @pcreux :smile_cat:

PR #4136, opened 2015-09-23
disable email notifications from travis

timoschilling:
@seanlinsley what do you think?


seanlinsley:
Why do you want to disable them? I like being notified if my build fails, so I know to go back to the PR and figure out what went wrong.

PR #4139, opened 2015-09-25
adding tests for public methods of index_as_blog.rb

jethroo:
adding tests for index_as_blog.rb leverage coverage from ~ 20 % to ~ 60%


timoschilling:
:+1:

PR #4140, opened 2015-09-28
Add german translation for the Comments model

kalkin:
This translation is used also in the top menu. You could use this also for `de_CH` it should be the same.

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -132,3 +132,8 @@ de:
       block: "Liste"
       grid: "Gitter"
       blog: "Blog"
+  activerecord:
+    models:
+      comment:
+        one: Kommentar
+        other: Kommentare


timoschilling:
We don't add model translations to our locale files

seanlinsley:
Shouldn't we in this case, since comments are a feature provided by Active Admin?

timoschilling:
I don't think so, while you can change the comment model.

kalkin:
@seanlinsley Well IMHO i'm on the same page with you. That's the reason why i did this PR.
@timoschilling Why? You never even tried to explain this.

timoschilling:
@kalkin in the past we had translation for thousand of thinks that are not came from ActiveAdmin, until I cleaned them up.
For example here is a PR form me that is merged by @seanlinsley which removes the `AdminUser` translations. Which are only exist in `de/de-ch` until that PR. https://github.com/activeadmin/activeadmin/pull/3416/files

We have done the same with keys from formtastic and devise. And now we have a much more cleaner locale file.

kalkin:
@timoschilling Thank you for your response. You are right you should not provide locales for devise and co. IMHO comments is a feature of ActiveAdmin so it's different, but I think the one liner isn't worth discussing :).

PR #4145, opened 2015-10-02
Rails 5: In fields_for, use to_unsafe_h on params before flat_map if necessary

tjgrathwell:
See http://eileencodes.com/posts/actioncontroller-parameters-now-returns-an-object-instead-of-a-hash/

PR #4146, opened 2015-10-02
Lock shoulda matchers to 2.8.0

tjgrathwell:
shoulda-matchers 3 drops suport for Rails 3.x and Ruby 1.9,
making it incompatible with the ActiveAdmin test matrix


tjgrathwell:
Should fix the failing build in #4145 (and any new PRs after that)

PR #4144, opened 2015-09-30
Update 0-installation.md

hobbes37:
Fixed installation instructions so that users can log in as the default admin after installing.

PR #4151, opened 2015-10-07
A few notes tho make things easier on installation

khebbie:


docs/0-installation.md
```diff
@@ -41,9 +41,11 @@ Now, migrate your database and start the server:
 
 ```sh
 rake db:migrate
+rake db:seed
 rails server
 ```
 
+*Please note if the server was running, you need to restart it for active admin to work.*
```
timoschilling: That's are Rails basics, we don't add thinks like that.

seanlinsley:
This issue has been resolved by a similar PR: #4144

PR #3640, opened 2014-11-21
WIP: new travis setup with appraisal

timoschilling:
This PR uses the new Travis build environment like provided by @joshk in #3595.
From now one we use [`appraisal`](https://github.com/thoughtbot/appraisal) to manage our gemfiles for different versions. The `RAILS` env variable no longer works.

TODOs:
- [ ] write some line in the docs about appraisal
- [ ] changelog entry
- [ ] make `script/locale` working with it

PR #4161, opened 2015-10-16
Wholesale Test Infrastructure Upgrade

pboling:
I promise I will make a better commit message.  Want to see what Travis does with all these changes while I sleep.  I am getting the same five failures on every Ruby against every version of Rails.

I upgraded to latest Rubies and latest Rails patches, along with updating to latest patch level of many gems.  I haven't looked into the 5 failures at all, but some upgrade is causing them I think.

This is the first step toward an Appraisal-based testing structure.

CC @seanlinsley as promised.


pboling:
So Travis failed very quickly!  I still have some more work to do removing the dependency on ENV["RAILS"].

tasks/test.rake
```diff
@@ -13,31 +15,27 @@ end
...
+      cmd "BUNDLE_GEMFILE='gemfiles/Gemfile.rails-#{fuzzy_rails_version}' bundle exec cucumber -p class-reloading features"
     end
-
-    cmd "./script/use_rails #{current_version}" if current_version
   end
 
   desc "Run the full suite against the important versions of rails"
   task :major_supported_rails do
-    run_tests_against *TRAVIS_RAILS_VERSIONS
+    run_tests_against *%(3.2.x 4.1.x 4.2.x)
```
timoschilling: Can we grep that from the gemfile names?

.travis.yml
```diff
@@ -11,23 +11,18 @@ install:
 script:
   - bundle exec rake test_with_coveralls
 rvm:
-  - 1.9
```
timoschilling: Way you drop 1.9?

tasks/test.rake
pboling: Definitely.  I question is we keep this whole `namespace :test do` section though.  So much of this manual code will be obviated by Appraisal (not yet in the PR, but will be soon).

.travis.yml
pboling: There were 90+ failures on 1.9, and I didn't want to focus on them.  I'll bring it back once I have it working elsewhere, and see what the problems are.  Most likely need to scope some gem versions to `platforms: [:ruby_19]` (or similar) in the Gemfile.

tasks/test.rake
timoschilling: Yes, Appraisal is better then this manual code!

seanlinsley:
Hi @pboling, first off thanks for putting so much effort into this. However, I'm not completely sold on Appraisal. Sure, a lot of other projects use it, but how exactly is it beneficial? Looking at the diff, all it seems to have done is add extra duplication and complexity.

pboling:
@seanlinsley I haven't actually added appraisal yet.  This was just me trying to de-couple some things that would need to be decoupled.

(Some of) The benefits are:
- You will be able to test against multiple Ruby / Rails / DB Adapters combinations.
- Technically you won't need to delete and recreate the test rails apps as much.
- You will know, as will your users, with greater clarity, what the gem version dependencies are for each Ruby / Rails combination.

I am not finished with this PR (obviously, failing build).  I'll work on it more tonight and see how far I get.

pboling:
Also of note: I think there is yet significant complexity that I will be able to, and need to, remove when I add appraisal for real.  This may be a several week effort, but I'm ok with that.  I see you are trying to clean up PRs and issues right now, but at least let me get this closer to done before you decide on it.  I'll go through the change set at that point and make notes on the changes and why I think they are beneficial.

timoschilling:
@seanlinsley some reasons are that can switch to the travis container based infrastructure, which brings faster test runs, and we can use the travis cache system with that. The cache system works then with PR's too. See #3640 and #3595

timoschilling:
we should take a look on https://github.com/grosser/wwtd and this [RubyConf talk](https://youtu.be/tqLbgS0fw5c?t=17m39s)

seanlinsley:
@pboling are you still planning on working on this?

pboling:
I will close it, and if I get time to pick it up again I will.

timoschilling:
:disappointed:

pboling:
It is sorely needed, but I let too much time pass and got lost in the complexity of the changes when I tried to work on it a bit.  Will need a solid block of time to work on it.  Still hoping to get to it though.

Issue #3571, opened 2014-11-04
extract components


timoschilling:
Change proposal:
I would like to extract some parts of active admin into its one gems to get a cleaner code base.
This a the parts that I want to extract:
- Authorization:
  - CanCan Adapter
  - Pundit Adapter
- Authentication (Devise)
- Comments

The code of the extracted components, can be cleaner through this change. For example comments don't need to life inside the active record orm.

Whats your meaning on this?

seanlinsley:
I don't think this is valuable. Active Admin is monolithic, but that in itself isn't bad. If there are some parts that need to be designed better, that can be done without separating them into their own gems.

I think having a ton of add-on gems is a bad idea for the same reason that I think that microservices are a bad idea: code out of sight, is out of mind. Our test suite won't run on it, so we won't know when it breaks. Additionally, having to switch between multiple git repos to read through and understand the code adds extra burden on the reader.

pboling:
Well if you provide the integration layer that the external gems plug into then you test that your API on this side does what it needs, and the other can and should be out of sight and out of mind.  It reduces, ideally, the cognitive load required to grok this codebase, because you don't have to deal with as many responsibilities.

Many gems follow this pattern, Rails, Rspec, Devise.  Also nice is with defined plugin APIs other people can build additional plugins for the tool, sort of like I did with React-rails here: https://github.com/pboling/react-rails-benchmark_renderer

timoschilling:
@seanlinsley would we really support any decorator gem, any authentication gem, any orm gem, any what ever gem?

timoschilling:
Things like https://github.com/activeadmin/activeadmin/pull/4175#issuecomment-151694771 are one reason more, to move adapters out of active admin.

seanlinsley:
@pboling I'm not against having an API that lets people easily plug in their own code; that's just good design. But for any gem that meets the qualifications I mention :arrow_double_down:, I'd prefer to keep the integration in this repo.

> @seanlinsley would we really support any decorator gem, any authentication gem, any orm gem, any what ever gem?

If the integration isn't too complex, someone else is willing to write it and open a PR, and it's a common enough gem that more people than the PR author would actually use the integration, I see no reason why it shouldn't be part of Active Admin proper.

> Things like #4175 (comment) are one reason more, to move adapters out of active admin.

[...]

PR #4165, opened 2015-10-19
Integrate ActiveAdmin with reform

bolshakov:
In order to provide ability to perform ActiveAdmin specific validations, I integrated ActiveAdmin with reform gem, which implements form object.

Usage:

```ruby
# app/models/post.rb
  # has title, content, and image_url
end

# app/forms/post_form.rb
class PostForm < Reform::Form
  property :title, validates: { presence: true, length: { minimum: 10 } }
  property :content, validates: { presence: true }
  property :image_url, validates: { presence: true }
end

# app/admin/post.rb
ActiveAdmin.register Post do
  form_class PostForm

  index do
    column :title
    column :image
    actions
  end
end
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'cancan'
 gem 'devise'
 gem 'draper'
 gem 'pundit'
+gem 'reform'
 
 # Utility gems used in both development & test environments
 gem 'rake', require: false
--- a/docs/15-forms.md
+++ b/docs/15-forms.md
@@ -1,0 +1,31 @@
+# Form Objects
+
+Active Admin allows you to use the form objects to provide ActiveAdmin-specific
+validations of a resource. [Reform](https://github.com/apotonick/reform)
+
+## Example usage
+
+```ruby
+# app/models/post.rb
+class Post < ActiveRecord::Base
+  # has title, content, and image_url
+end
+
+# app/forms/post_form.rb
+class PostForm < Reform::Form
+  property :title, validates: { presence: true, length: { minimum: 10 } }
+  property :content, validates: { presence: true }
+  property :image_url, validates: { presence: true }
+end
+
+# app/admin/post.rb
+ActiveAdmin.register Post do
+  form_class PostForm
+
+  index do
+    column :title
+    column :image
+    actions
+  end
+end
+```
--- a/features/edit_page.feature
+++ b/features/edit_page.feature
@@ -123,7 +123,7 @@ Feature: Edit Page
 
       ActiveAdmin.register Post do
         form_class PostForm
-        permit_params :title, :body
+        permit_params :title, :body if Rails::VERSION::MAJOR == 4
 
         form do |f|
           f.inputs do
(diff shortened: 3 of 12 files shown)


timoschilling:
I like the idea and I'm a user of Reform too. But I think this should not be part of the core. Something like that should be a autonomic gem.

If you are interested in creating `activeadmin-reform`, the @activeadmin org will give you a home for that.

bolshakov:
I'm understand the motivation not adding this to core, but there is a couple of pros for that:

1. ActiveAdmin does not have point of extractions for that reason now. We can add them.
2. I don't want maintain test infrastructure for different versions of Rails and ActiveAdmin (AA does it perfectly!). I don't know how to solve this problem.

Do you know good example for inspiration?

timoschilling:
I understand your reasons. You can take a look on the repos of @activeadmin-plugins and @unmantained-activeadmin-plugins. https://github.com/activeadmin/activeadmin/pull/4161 tries to simplify the test setup.

Maybe you can do something like this:
In your Gem:
1. build a generator which modifies the AA test app for your need
In a setup.rb of your gem:
1. clone ActiveAdmin into `activeadmin-reform/test/activeadmin`
2. run `cd test/activeadmin && rake:setup`
3. run `rails g modify-aa-tests`
On each test run:
1. run `rake`
2. run `cd test/activeadmin && rake`
You can do the same on travis.

timoschilling:
I think since #4166 this can be closed now?

bolshakov:
Sure, thank you for feedback!

bolshakov:
Hi @timoschilling I factored [activeadmin_reform gem](https://github.com/bolshakov/activeadmin_reform) Could you review? Would you mind to move it under @activeadmin org?

timoschilling:
@bolshakov I have leaved you some comments on the repo.
@seanlinsley what do you think about moving that into the @activeadmin org?

seanlinsley:
I would actually prefer that this be part of Active Admin, instead of a separate gem. Sure it imposes more of a maintenance burden in the future, but it:
- keeps the integration becoming out of date with AA or Reform (because our test suite runs often)
- makes it easier for people to get started with AA

If the maintenance burden is too much, we should add more maintainers :smile: 

@timoschilling what are your reasons for thinking it should be a separate gem? It's a very small number of changes we're talking about here:

screen shot 2015-10-27 at 9 45 56 am

bolshakov:
@seanlinsley I thinks it's a good idea to have an ActiveAdmin with basic features and plugins which add advanced functionality. So the more we have such extensions the more ActiveAdmin extendable :)

seanlinsley:
@bolshakov the problem is, realistically, no one is going to maintain those add-on gems. So we'll have a graveyard of extensions, most of which require monkeypatches to get working with the latest version of Active Admin and the other gem.

timoschilling:
@seanlinsley that gem don't make monkeypatches

timoschilling:
@seanlinsley I'm with @bolshakov, ActiveAdmin should be focused and not be overloaded with features.

timoschilling:
@seanlinsley that a look on #3571 and the feedback

seanlinsley:
> @seanlinsley that gem don't make monkeypatches

I mean monkeypatches on the add-on gem, because it no longer works with the latest Active Admin, and the other gem it's built to integrate.

timoschilling:
@seanlinsley that way we can join that gem into the @activeadmin org, so someone can create a PR to make it work.

PR #4166, opened 2015-10-20
Allow to apply custom decorations

bolshakov:
As discussed earlier here #4165, I made some changes to make AA more extendable. 
This PR does not add any new functionality, but allows to easily embed AA with custom decorations.


timoschilling:
:+1:

PR #4176, opened 2015-10-27
Revert "truncate changelog"

seanlinsley:
This reverts commit e3a7354c0196743e2462bdc27c591e8285d110b0.

@timoschilling I'd prefer to keep the entire history in a single file. It's easier to link to, and search through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,3 @@
-# Changelog
 ## 1.0.0 [☰](https://github.com/activeadmin/activeadmin/compare/v0.6.3...master) (unreleased)
 
 ### Breaking Changes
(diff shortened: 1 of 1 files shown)


timoschilling:
I don't think some.

seanlinsley:
What do you mean?

For linkability: it means that even years-old blog posts still bring you to the data you were looking for (as well as from Stack Overflow).

For searchability: it means that you can search for feature changes across multiple releases, which is useful if you're upgrading by multiple versions at once.

seanlinsley:
The only downside is that the changelog is long. Which doesn't seem like an issue, since visiting the page starts you at the top.

timoschilling:
I have split the file, while the 1.0.0 changelog and 0.6 changelog are out of sync

timoschilling:
ok, let us merge this, but then you need to care about the change file sync.

seanlinsley:
Okay, just a minute. I'll update this PR with the 0.6.x changelog.

PR #4183, opened 2015-11-03
allow proc as parent menu label

ishumilova:
allow any Proc in menu parent: DSL option, for example you can use in in case of broken I18n
```
menu parent: -> { I18n.t('active_admin.menu_parent') }
```


timoschilling:
We need tests for that.

seanlinsley:
Hi @ishumilova, please take a look at the [menu item documentation](https://github.com/activeadmin/activeadmin/blob/c6510db25864c869256f5ba34120c5f7c8ada825/docs/2-resource-customization.md#dynamic-parent-menu-items). That shows how to use translations as labels for parent menu items.

PR #4185, opened 2015-11-09
Update comment.rb

Ropeney:
Belongs to author but unable to set author; added to attr_accessible.


timoschilling:
Note: This is for v0.6

seanlinsley:
The author isn't meant to be set via parameters. Instead we use the [current user](https://github.com/activeadmin/activeadmin/blob/c6510db25864c869256f5ba34120c5f7c8ada825/lib/active_admin/orm/active_record/comments.rb#L41-L44) that's logged in. It's meant to be a security feature.

PR #4196, opened 2015-11-13
fix for issue #4185. Notification and alert styling within logged out layout missing

JoeWoodward:
Minor tweak to fix a rendering issue within the `app/views/layouts/active_admin_logged_out.html.erb`

screen shot 2015-11-13 at 8 17 21 pm

### DIFF
--- a/app/assets/stylesheets/active_admin/components/_flash_messages.scss
+++ b/app/assets/stylesheets/active_admin/components/_flash_messages.scss
@@ -25,13 +25,25 @@ body.logged_in {
 
 body.logged_out {
   .flash {
-    @include no-shadow;
-    @include text-shadow(#fff);
-    background: none;
-    color: #666;
+    @include gradient(#f7f1d3, #f5edc5);
+    @include text-shadow(#fafafa);
+    border-bottom: 1px solid #eee098;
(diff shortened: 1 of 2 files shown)


timoschilling:
Need to wait for #3862

Fivell:
@javierjulio , what do you think about this?

javierjulio:
@Fivell I've done the same but I didn't move the flash message div in the html, although that's because I changed the CSS so its rendered at the top stretching full screen which is the same result as here essentially, I just got there a different way.

Would you rather hold off on this and put the time into #3862 instead? I believe what you, @deivid-rodriguez, and @varyonic want me to do is slice #3862 in half and squash some commits (which I just learned to do this week and am really enjoying 😏) and do a new PR with just those changes? If that's the direction you want let me know and I'll set aside some time to test that out.

javierjulio:
Going to close out as what we have is fine for the login form. In #3862 we've done essentially the same as here, made it component based as it should be, avoided any duplication in the process and also removed unnecessary styles. Not your fault since the CSS wasn't done this way but I want to close this out as this just introduces more changes that I don't think are important. They will change anyway in v2.

Issue #4182, opened 2015-11-03
Ability to set a custom count method


matthewrudy:
For our app we have certain tables which are so big calling `MyModel.count` takes over a second.
And conceivably using some combination of search params could cause a table scan, and possibly crash the database.

So we need a way to override the `total_count` method from Kaminari.
Our current approach is to override the count globally

``` ruby
module Kaminari::ActiveRecordRelationMethods
  def total_count(column_name = :all, options = {})
    @total_count ||= begin
       proprietary_count_method
    end
  end
end
```

This works, but it's a hack, and being global in nature, we have no way to selectively opt-in/out.

So I'm wondering, what's the best way to avoid calling `total_count`?

Options:

A. Enable custom pagination methods

``` ruby
ActiveAdmin.config.pagination_method_name ||= Kaminari.config.page_method_name
```

B. Introduce custom counting methods

``` ruby
ActiveAdmin.config.count_method_name ||= :total_count
```

C. Something else?

matthewrudy:
Looking more into this, I saw this Issue #4133.

So I think we just need a good way to override all counts.

timoschilling:
closed as double of #4133

Issue #4133, opened 2015-09-22
Equivalent of `scoped_collection` for index table, but not scope counts?


ahmacleod:
I frequently need to customize the scope for the index table by including associations, joins for sorting, custom count queries, etc. I've been using `scoped_collection` for this, but then my scope counts are computed using this more expensive query. Is there a way to specialize the scope for just the index table without overriding the controller?

timoschilling:
I think there is no way.

@seanlinsley ???

timoschilling:
We have something for [`includes`](https://github.com/timoschilling/activeadmin/blob/master/docs/2-resource-customization.md#eager-loading)

ahmacleod:
Ok thanks; I'll try monkey patching for now.

Fivell:
@ahmacleod , don't know if it can help only an idea. There is an option to create methods for scopes.
For instance
```ruby
def scoped_collection
  MyModel.special_scope 
end
```
and than on model level

```ruby
class MyModel
   scope :special_scope, -> { joins(:sometable)) }   do 
      def count
         1000000
      end 
   end
end
```

seanlinsley:
You can override `ActiveAdmin::Views::Scopes#get_scope_count`, or `ActiveAdmin::Helpers::Collection#collection_size`. You might be able to use `@collection_before_scope` as a hack to grab the collection before too many queries are added to the chain.

As far as speeding up scope counts in general, I created #4132 recently.

Fivell:
@seanlinsley , 
ActiveAdmin uses more expensive count queries than for instance kaminari.
Maybe ActiveAdmin can just use implementation from kaminari https://github.com/amatsuda/kaminari/blob/3974ae5f39ce31bfca28d17eccd8daf3ca263aa3/lib/kaminari/models/active_record_relation_methods.rb#L12
(it should except also includes from relation )

in  ```collection_size(c = collection) ``` method ?

Fivell:
@ahmacleod , is this still an issue?

varyonic:
I think overriding `ActiveAdmin::Views::Scopes#get_scope_count` mentioned above is a good suggestion.

PR #4200, opened 2015-11-17
Improvement of collection_size helper

Bishop:
AR count method generates bad sql for relation with limit, like `SELECT COUNT(count_column) FROM (SELECT  1 AS count_column FROM "table" LIMIT 30 OFFSET 0) subquery_for_count` - it is really slow for big tables.
And we can get size of loaded collection without additional query. Loading collection and count entries is a best solution for relations with limit.

On my real DB, "table" has ~ 50_000_000 rows: 

```
explain analyze SELECT count(*) FROM "table" WHERE <some_condition>;
Planning time: 0.393 ms
Execution time: 0.315 ms

explain analyze SELECT COUNT(count_column) FROM (SELECT  1 AS count_column FROM "table" WHERE <some_condition> LIMIT 50 OFFSET 0) subquery_for_count
Planning time: 0.217 ms
Execution time: 18246.767 ms
```

### DIFF
--- a/lib/active_admin/helpers/collection.rb
+++ b/lib/active_admin/helpers/collection.rb
@@ -4,9 +4,12 @@ module ActiveAdmin
       # 1. removes `select` and `order` to prevent invalid SQL
       # 2. correctly handles the Hash returned when `group by` is used
       def collection_size(c = collection)
-        c = c.except :select, :order
-
-        c.group_values.present? ? c.count.count : c.count
+        if c.loaded?
+          c.size
+        else
+          c = c.except :select, :order
+          c.group_values.present? ? c.count.count : c.count
+        end
       end
 
       def collection_is_empty?(c = collection)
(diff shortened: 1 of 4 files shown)


nicodeceulaer:
I'm hitting the same problem. 
Is this problem still actively being investigated?

I see 3 different issue tickets debating the same problem: #4133, #4182 and this one. But neither of the three seems to be getting a lot of attention.

Fivell:
@varyonic , I don't think this is a bug, we need some tag for optimisations

Fivell:
@Bishop ,

> And we can get size of loaded collection without additional query. Loading collection and count entries is a best solution for relations with limit.



Can you show what part of active admin makes additional sql query?

Fivell:
@timoschilling , I think this logic should be in pagination layer

Bishop:
`SELECT COUNT(count_column) FROM (SELECT 1 AS count_column FROM "table" LIMIT 30 OFFSET 0) subquery_for_count`

I don't see this evil query with last master code. So problem was fixed. I would like to find which change solved it, and i think it doesn't need to merge this MR.

Fivell:
@Bishop ,  would be great to see which commit fixed it.

PR #4209, opened 2015-11-22
allow CSV pagination to be customizable

seanlinsley:
This fixes the performance issues that #3375 introduced by switching from `find_each` to Kaminari, since Kaminari has to do a SQL count. For those that want that behavior they can enable it by setting `paginate_with: :kaminari`, but the fastest option should be the default for Active Admin.

I chose this approach instead of #4082 because databases don’t guarantee any particular order when requesting records, so it’s possible that the resulting CSV won’t include all records.

### DIFF
--- a/docs/4-csv-format.md
+++ b/docs/4-csv-format.md
@@ -56,3 +56,25 @@ environments where CSV streaming is disabled, you can change this setting:
 
 config.disable_streaming_in = ['development', 'staging']
 ```
+
+# Pagination
+
+By default we use `find_each` to paginate through your records because it's the
+best option made available by Rails. Note that it relies on the table's primary
+key, and orders the collection by ID. If you would like to configure either of
+those, you can use the `paginate_with` option.
+
+```ruby
+ActiveAdmin.register Post do
+  controller do
+    def collection
+      Post.order created_at: :desc
+    end
+  end
+
+  csv paginate_with: :kaminari
+end
+```
+
+You can also provide any custom pagination by passing a proc/lambda. The only
+requirement is that it needs to respond to `each`.
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -26,13 +26,17 @@ module ActiveAdmin
       end
     end
 
-    attr_reader :columns, :options, :view_context
+    attr_reader :columns, :options, :byte_order_mark, :column_names, :csv_options
 
     COLUMN_TRANSITIVE_OPTIONS = [:humanize_name].freeze
 
     def initialize(options={}, &block)
       @resource = options.delete(:resource)
-      @columns, @options, @block = [], options, block
+      @block    = block
+      @options  = ActiveAdmin.application.csv_options.merge options
+      @byte_order_mark = @options.delete :byte_order_mark
+      @column_names    = @options.delete(:column_names) { true }
+      @csv_options     = @options.except :encoding_options
     end
 
     def column(name, options={}, &block)
@@ -40,14 +44,10 @@ module ActiveAdmin
     end
 
     def build(controller, csv)
-      @collection  = controller.send :find_collection, except: :pagination
-      columns      = exec_columns controller.view_context
-      options      = ActiveAdmin.application.csv_options.merge self.options
-      bom          = options.delete :byte_order_mark
-      column_names = options.delete(:column_names) { true }
-      csv_options  = options.except :encoding_options
+      @collection = controller.send :find_collection, except: :pagination
+      columns     = exec_columns controller.view_context
 
-      csv << bom if bom
+      csv << byte_order_mark if byte_order_mark
 
       if column_names
         csv << CSV.generate_line(columns.map{ |c| encode c.name, options }, csv_options)
@@ -67,7 +67,7 @@ module ActiveAdmin
       @view_context = view_context
       @columns = [] # we want to re-render these every instance
       instance_exec &@block if @block.present?
-      columns
+      @columns
     end
 
     def build_row(resource, columns, options)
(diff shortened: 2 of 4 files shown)


seanlinsley:
CC @kushkella, @matthewrudy

seanlinsley:
The only commit that's important to the feature is 70ed39040f47a5089501f9d5b10cde7d065f3628

seanlinsley:
Looks like something in this PR is broken for Rails 3.2.

kushkella:
We have the same problem with our production rails project, and we decided to switch to a custom pagination technique that worked using `offset` and `limit` and did not do `count` the entire collection. The idea is to keep finding the next batch until the size the batch returned is less than the batch_size provided. Something like this should work:

```ruby
def get_each(relation, opts = {})
  batch_size   = opts[:batch_size] || 1000
  next_batch   = true
  batch_offest = 0
  while next_batch
    records = relation.offset(batch_offest).limit(batch_size)
    records.each { |record| yield record }
    batch_offest += batch_size
    next_batch = records.size == batch_size
  end
end
```

kushkella:
And I want to add that AA counts the collection on all index pages to show the pagination controls, so a normal user would see performance issues in viewing the index pages as well unless he is using CSV export separately from Index pages.  Thanks!

kushkella:
And the other reason why we I switched to Kaminari was to support custom sort order in CSV export. `find_each` always uses `ORDER BY id` which is why I switched to Kaminari since it preserves the custom sort order. And I believe currently if the users doesn't sort by certain column, the default sort order is applied to guarantee that DB returns records in predictable order.

seanlinsley:
> And I believe currently if the users doesn't sort by certain column, the default sort order is applied to guarantee that DB returns records in predictable order.

There isn't a default sort order, so no ordering is applied in the most common scenario:
```ruby
User.public_send(Kaminari.config.page_method_name, 1).per(1000).to_a; nil
  User Load (1.8ms)  SELECT  "users".* FROM "users" LIMIT 1000 OFFSET 0

**Comment by Fivell** [comment]
@kushkella  , sorting by any other column except id can causes problems, because between batch selects ordering can be changed (for instance updated_at can be changed for some records and so on).
As result you will have duplicated rows in your csv and/or missed  rows.

**Comment by Fivell** [comment]
@seanlinsley , how do you think maybe forcing sorting by primary key when downloading csv should be default behaviour in AA ?

**Comment by kushkella** [comment]
@seanlinsley <https://github.com/seanlinsley> I believe that
`config.default_sort_order` is applied to the collection passed to CSV,
otherwise the index page would have similar issue where it would show
different records when page reloads. But again if the CSV generator is used
independently, this might not be true. I think your approach is probably
the right one.

@igor Ah, that is the behavior I removed earlier :) But I agree with Igor
that things can change and sorting on the primary key is the only way to
ensure all records are returned which is why I've been doing an additional
sort by primary key in our production application. But one needs to have
proper indexes in place to have it working without any performances issues.

On Mon, Nov 23, 2015 at 9:50 PM, Igor Fedoronchuk <notifications@github.com>
wrote:

> @seanlinsley <https://github.com/seanlinsley> , how do you think maybe
> forcing sorting by primary key when downloading csv should be default
> behaviour in AA ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/activeadmin/activeadmin/pull/4209#issuecomment-158993375>
> .
>



-- 
Regards,
Kush Kella

**Comment by matthewrudy** [comment]
I like this as a solution to our problem.

(actually in general we're trying to find ways to sensibly avoid ever doing counts on our big tables)

File: docs/4-csv-format.md [hunk 1]
```diff
@@ -56,3 +56,25 @@ environments where CSV streaming is disabled, you can change this setting:
...
+      Post.order created_at: :desc
+    end
+  end
+
+  csv paginate_with: :kaminari
+end
+```
+
+You can also provide any custom pagination by passing a proc/lambda. The only
+requirement is that it needs to respond to `each`.
```
seanlinsley: > The only requirement is the object you return needs to respond to `each`.

thoughtpunch:
@seanlinsley status?

seanlinsley:
@thoughtpunch the general answer to that question is "pay people to work on open source".

Alternatively, volunteer your own time to become a maintainer.

My current job does not afford me time to work on open source, so it is very difficult to find time to maintain Active Admin.

thoughtpunch:
@seanlinsley I was just curious what happened with this, as you went as far as to open a PR, write tests, etc but it never got merged. I thought it was an awesome approach. I'd be happy to try to take over the finish line if you are open to it :) 

Can you share what was left to complete to get this mergable?

seanlinsley:
I honestly don't remember; it's been too long. At the very least, there are test failures on this branch, and the master branch itself is currently failing.

I'm actually currently working on getting Draper working with Rails 5 (also getting its master branch passing). If you would like to open a PR to get the AA master branch passing, that would be really appreciated.

Fivell:
related https://github.com/kaminari/kaminari/pull/681
kaminari fixed this problem itself, we can somehow try to use this new feature ...

javierjulio:
@Fivell I didn't know about that Kaminari feature! Would be great to make that an option for AA that way we don't do counts on index pages. Maybe we leave the default to use pages but then have a config option where it would just show previous/next links instead?

Fivell:
@javierjulio , sounds good need to be investigated

matthewrudy:
I still maintain my fork that just iterates until it finds no more records.

```
gem 'activeadmin', github: 'gogovan/activeadmin',
                                 branch: 'ggv-1.0.0'
```

works fine for us.

sunny:
Perhaps related to #5862, since Pagy also has support for "countless" pagination.

PR #4213, opened 2015-11-23
Update 0-installation.md

barisizm:
Documentation have a little missing part.
If your rails version equal or bigger than 4.0.0, bundler throw an error like this: 
"Your Gemfile requires gems that depend on each other, creating an infinite loop. Please remove gem 'meta_search' and try again. "

And then i fetched activeadmin from master and solved the problem.


timoschilling:
The documentation is based on the released gem.

PR #4223, opened 2015-11-30
Not removing action_items container when empty

stefanvignir:
I am using Javascript to add buttons to the `action_items` bar `inside titlebar_right`.

Removing `action_items` completely when there are no action items generated by the backend is not a good idea where action_item buttons are added on the client side, as the script would need to check for the existence of the container every time.

Example JS/jQuery to add a print button:

```js
function addPrintButton() {
    $('.action_items').prepend('<span class="action_item"><a href="#" class="print_button">Print</a></span>');
    $('.print_button').click(function(e) {
        e.preventDefault();
        window.print();
    });
}

$(document).on('ready page:load', function(event) {
    addPrintButton();
});
```

### DIFF
--- a/lib/active_admin/views/title_bar.rb
+++ b/lib/active_admin/views/title_bar.rb
@@ -47,7 +47,7 @@ module ActiveAdmin
       end
 
       def build_action_items
-        insert_tag(view_factory.action_items, @action_items) if @action_items.any?
+        insert_tag(view_factory.action_items, @action_items)
       end
 
     end


timoschilling:
1. Why you don't use active admin's api to add the button?
2. How does this change affect the UI? Screenshots

stefanvignir:
@timoschilling In this case, the only sensible way to do this is JS, as the buttons I'm adding are only useful with JS enabled and I'm aiming for graceful fallback for the users who are unable to use JS or keep it disabled. (Yes. This is a actually a thing.)

The ActiveAdmin API is great for linking to member actions etc but it becomes a bit too hacky in this case as I would have to hide the JS-only buttons with CSS and then unhide them using JS. `javascript:` urls are also not best practice nowadays.

stefanvignir:
Basically, I felt like this was the most logical way to solve this.

stefanvignir:
Oh, and this does not affect the UI in any way.

stefanvignir:
screen shot 2015-11-30 at 16 30 38

So basically in my case, the little JS adds a print button next to the header and the same method could be used for any other JS-only functionality. Not the fanciest thing in the universe, but I thought it belonged in the master branch.

Fivell:
>>users who are unable to use JS
I though ActiveAdmin can't be used without js at all.

Fivell:
@stefanvignir , you can add member_action on server side and  custom css-class to member_action link to make it display:none; 
And then just show this existing hidden button with js instead of building new button.

stefanvignir:
@Fivell: How boring would my job be if there was only one way to do things?

seanlinsley:
Most of Active Admin does work without JS. Thanks for the PR

Issue #4230, opened 2015-12-09
Add search status panel name in I18n


marcomd:
Hi guys,
I would add the filter panel name:

```yaml
    search_status:
	  search_status_name: "Search status:"
```
/lib/active_admin/filters/resource_extension.rb

```ruby
  # row 132
  ActiveAdmin::SidebarSection.new I18n.t("active_admin.search_status.search_status_name", :default => 'Search status'), only: :index, if: -> { params[:q] || params[:scope] } do
```
I also added the new tag in the config/locale yml file for :en and :it

timoschilling:
name it `"active_admin.search_status.headline"` and remove the default.

marcomd:
There is already headline, could i name it "headname" ?

```yaml
    search_status:
      headname: "Search status:"
      headline: "Scope:"
      current_filters: "Current filters:"
      no_current_filters: "None"
```

timoschilling:
We should rename the current `headline` key to `current_scope` and use `headline` for your case.

timoschilling:
Would you like to provide a PR for that?

marcomd:
Sure for the PR
In my version there is "headline" should i rename it in current_scope or i have to wait ?

timoschilling:
rename it

marcomd:
There is another "headline" under "unsupported_browser" :frowning:
I would use a unique name to avoid confusion and errors, could i use "search_status_headline" ?

timoschilling:
No, in I18n you use a uniq name per scope not global.

marcomd:
When is a common tag name i prefer to use uniqe global name to facilitate find or rename operations like this but i follow you guide lines.
I also update all translation yaml if you agree

PR #4231, opened 2015-12-09
4230 search status name on filter sidebar

marcomd:
Add search status panel name in I18n #4230 

- [x] refactor `search_status.headline` to `search_status.current_scope` (all yml )
- [x] added new tag `search_status.headline` to :gb: and :it: yml

### DIFF
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -32,7 +32,7 @@ ar:
         greater_than: "أكبر من"
         less_than: "أقل من"
     search_status:
-      headline: "المجال:"
+      current_scope: "المجال:"
       current_filters: "المُرشحات الحاليّة:"
       no_current_filters: "بدون"
     status_tag:
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -32,7 +32,7 @@ de:
         greater_than: "Größer als"
         less_than: "Kleiner als"
     search_status:
-      headline: "Anwendungsbereich:"
+      current_scope: "Anwendungsbereich:"
       current_filters: "Aktuelle Filter:"
       no_current_filters: "Keine"
     status_tag:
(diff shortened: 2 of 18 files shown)


timoschilling:
you have a formating problem in the [it.yml](https://github.com/marcomd/activeadmin/blob/4230_search_status_name_on_filter_sidebar/config/locales/it.yml#L35) and [en.yml](https://github.com/marcomd/activeadmin/blob/4230_search_status_name_on_filter_sidebar/config/locales/en.yml#L35).

marcomd:
The usual skirmishes between windows and unix, sorry

marcomd:
Ok all tests passed, let me know if the updated spec is ok

marcomd:
Close #4230

seanlinsley:
The code now relies on both the `headline` and `current_scope` translations, but for the vast majority of translation files are now missing `headline`. @timoschilling I personally wouldn't have accepted this PR without the new translations, because it means that anyone who upgrades whose app isn't set up to use English or Italian ends up with a substandard experience.

I feel like it should be standard that any PR that modifies translations should ensure that existing translation files aren't left broken.

It's not even something that has to take a long time to do, since Google spreadsheets have a function to automatically translate phrases: http://www.labnol.org/internet/google-translate-for-spreadsheets/10086/

timoschilling:
This PR hasn't broke any translation! This PR just has renamed a translation and has add a new one.
The only think which is confusing was, that we have renamed `headline` to `current_scope` and have add a new `headline` key. That look like we have broke the old translations, but we don't.

BTW: #4304 has add the missing key for all locals

seanlinsley:
The problem is as you said: this PR renamed `headline` to `current_scope`, and then reused `headline` for another purpose. So any language that didn't have a `headline` translation added by this PR was effectively broken.

#4304 helps, but it doesn't cover all languages. For example: [`de.yml`](https://github.com/activeadmin/activeadmin/blob/master/config/locales/de.yml#L34) is missing the new translation.

timoschilling:
We have many key with are missing in different languages

timoschilling:
`de.yml` has now the key

PR #4233, opened 2015-12-10
Updated patch and gem version

ferdinandrosario:


timoschilling:
why should we only allow `1.7.x`?

ferdinandrosario:
not necessary, If you wish I will revert that. just wanted to ensure user are using the latest version ..
Shall I remove it 1.7.x?

timoschilling:
It is not our task to ensure that the user use the latest versions. We only say we need `> 1.3` while `1.2` has a bug or while we need a feature that is introduced in `1.3`. So, yes please revert that.

ferdinandrosario:
Done :+1:

PR #4235, opened 2015-12-10
add edit action to id_column if show action not available

illogikal:
Have added an additional conditional to id_column as my requirements for activeadmin doesn't include needing show views and thought it could be helpful?


timoschilling:
:+1: please remove the empty line and force push that branch

illogikal:
done!

Issue #1590, opened 2012-08-14
Order with Nulls Last


corybuecker:
I posted [a note in the AA Google Group](https://groups.google.com/forum/#!topic/activeadmin/rBbTtpFqky8/discussion%5B1-25%5D) to determine if this problem is actually a lack of understanding on my part. There was no answer there, and before I submit a pull request I want to make sure that I'm not missing something obvious.

When using PostgreSQL and AA, there is an issue with sorting a column that contains null values. [PostgreSQL places null values before non-null values](http://www.postgresql.org/docs/9.1/static/queries-order.html) when descending order is specified. This is opposite of MySQL's behavior.

I've made the follow change to the [sorting module](https://github.com/gregbell/active_admin/blob/master/lib/active_admin/resource_controller/collection.rb#L43), but I want to get some feedback before I submit it. Please let me know if there is a better way to approach this issue.

``` ruby
def sort_order(chain)
  params[:order] ||= active_admin_config.sort_order
  if params[:order] && params[:order] =~ /^([\w\_\.]+)_(desc|asc)$/
    column = $1
    order  = $2

    # if the database backend is PostgreSQL and the sort order is descending,
    # put null values after non-null values to emulate MySQL behavior
    if order == "desc" && ActiveRecord::Base.connection.instance_of?(ActiveRecord::ConnectionAdapters::PostgreSQLAdapter)
      order = "desc nulls last"
    end

    table  = active_admin_config.resource_table_name
    table_column = (column =~ /\./) ? column :
      "#{table}.#{active_admin_config.resource_quoted_column_name(column)}"

    chain.reorder("#{table_column} #{order}")
  else
    chain # just return the chain
  end
end
```

epitron:
I just ran across this issue today. @corybuecker's patch seems like a good solution. I'd be interested if this was merged into the codebase!

seanlinsley:
Is there a reason why anyone would desire the default behavior that Postgres provides? Should we make the behavior configurable in Active Admin?

`ActiveAdmin::OrderClause` really needs to be abstracted out into the [orm folder](https://github.com/activeadmin/activeadmin/tree/master/lib/active_admin/orm) so that we can have different logic for different databases. Though that might not happen as part of this ticket.

Fivell:
ransack issue https://github.com/activerecord-hackery/ransack/issues/443

epitron:
I've been using a forked version with that 3 line patch for over a year and it's been working fine and dandy.

[...]

PR #4242, opened 2015-12-17
Fix postgres NULL sorting

epitron:
Hello!

This is a tiny patch to fix issue #1590.

I used the ``@@postgres`` class variable to store the result of the check for the PostgreSQL ActiveRecord adapter. This might be a pointless optimization. I can switch it back to the `instance_of?` check. The only reason it might cause a problem is if people switch database adapters while the process is running. I don't know if that's ever actually done, however.

### DIFF
--- a/lib/active_admin/order_clause.rb
+++ b/lib/active_admin/order_clause.rb
@@ -3,6 +3,8 @@ module ActiveAdmin
     attr_reader :field, :order
 
     def initialize(clause)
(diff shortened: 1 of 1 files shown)


timoschilling:
Database specific behavior is not a task of active admin,

epitron:
Unfortunately, since you've replaced active record's order method with your own custom method, you've kind of taken responsibility for its sorting functionality, since the user has no way to fix this without monkey patching your code.

Postgresql is also a very popular database choice. 

Perhaps we could discuss an improved solution, such as extra options for the active admin sort option.

seanlinsley:
@epitron I just reopened #1590 so we can discuss a better solution.

PR #4247, opened 2015-12-21
add protocol to permitted_params for ssl users

randito78:


timoschilling:
Your `:protocol` key is application specific and should not be part of active admin

seanlinsley:
@timoschilling not necessarily. If `:protocol` is a parameter that Rails itself defines, it makes sense to always permit it.

timoschilling:
I don't think that rails adds something like `:protocol` to the `params`.

seanlinsley:
I can't find any documentation for this. And if this were to be merged, ideally it should have a test to go along with it. Closing for now.

Issue #4251, opened 2015-12-23
cucumber specs failing on master


bsy:
cucumber specs are failing with: 

`couldn't find file 'turbolinks' with type 'application/javascript' (ActionView::Template::Error)...`

timoschilling:
I think an the setup of your app is something wrong. What does your active_admin.js looks like?

PR #4252, opened 2015-12-23
add turbolinks

bsy:
addresses https://github.com/activeadmin/activeadmin/issues/4251


timoschilling:
ActiveAdmin don't use turbolinks, if some wont to use it, he can add it in his App.
We will solve your problem in #4251

PR #4268, opened 2016-01-14
Update activeadmin.gemspec

maokomioko:


timoschilling:
Why do you want this change?

maokomioko:
To keep up with up to date version of ransack and make '1' to true bug fixing happen faster.
The one from here https://github.com/activerecord-hackery/ransack/issues/593

timoschilling:
Thats not a relevant reason. A relevant reason is a something that active admin need with is only in `1.7` or a bug which is fixed in `1.7`. If non of that is the case we will keep the dependency as low as possible. For example we require Rails `3.2.X` and not `5.0.0-betaX` ;)

PR #4270, opened 2016-01-15
Use `status_tag` in AttributesTable#content_for to display booleans

jtomaszewski:
This makes it work nearly the same as in `ActiveAdmin::Views::TableFor`.

It's a bit duplicated code though. Maybe we should move it to `pretty_format` function?

P.S. If someone wants this change in his current rails app, u can implement it without changing the activeadmin gem using this gist: https://gist.github.com/fred/2574969 .

### DIFF
--- a/lib/active_admin/views/components/attributes_table.rb
+++ b/lib/active_admin/views/components/attributes_table.rb
@@ -79,8 +79,21 @@ module ActiveAdmin
       end
 
       def content_for(record, attr)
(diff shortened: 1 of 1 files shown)


dmoss18:
I believe this is resolved in this commit https://github.com/activeadmin/activeadmin/commit/3efb359533df46e76bb4355a02ad31e727fb0da3

seanlinsley:
Yeah, #4346 adds this.

PR #4275, opened 2016-01-19
Update 3-index-pages.md

edatrix:
add a line about how to access filter results. :page_facing_up:


timoschilling:
That has nothing to do with active admin

PR #4276, opened 2016-01-20
Allow usage of wrapper_html option in filters

fabien-michel:


seanlinsley:
Are we sure that `super` doesn't return something unexpected here?

fabien-michel:
From what I've tested, there are no visual differences.
Here the rendered HTML diff : 
Before : `<div class="filter_form_field filter_string select_and_search">`
After : `<div class="string input optional stringish filter_form_field filter_string select_and_search" id="q_insurance_company_claim_reference_input">`

PR #4287, opened 2016-01-25
Method #decorate? should use form.options[:decorate]

mrzasa:
I noticed that decorators are added to forms in `update` and `create` actions. Consider the following scenario: 
1. User tries to create a record that has a decorator.
2. User visits the `new` page. Record is not decorated.
3. User enters invalid data and submits.
4. Action `create` is fired, a form is rerendered, but this time the record is decorated.

I assume form should be not decorated in `create`/`update` by default. I have prepared a fix for that.

I had problems with running `rake test:major_supported_rails`: specs from  `features/strong_parameters.feature` fails on my machine on my branch and on master. Tried to debug it and I noticed that  `Before('@rails4')` is not executed, therefore specs intended to be run on Rails 4 are run on Rails 3. Do you have any clue what could be wrong on my env?

### DIFF
--- a/lib/active_admin/resource_controller/decorators.rb
+++ b/lib/active_admin/resource_controller/decorators.rb
@@ -28,7 +28,7 @@ module ActiveAdmin
 
       def decorate?
         case action_name
-        when 'new', 'edit'
+        when 'new', 'edit', 'create', 'update'
           form = active_admin_config.get_page_presenter :form
           form && form.options[:decorate] && decorator_class.present?
         else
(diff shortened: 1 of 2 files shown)


seanlinsley:
Thanks for the bugfix PR! I've confirmed that the tests are passing locally.


> Tried to debug it and I noticed that `Before('@rails4')` is not executed, therefore specs intended to be run on Rails 4 are run on Rails 3. Do you have any clue what could be wrong on my env?

That's a problem with Cucumber 3: #4308. The Gemfile has been updated to require Cucumber 2 for now.

PR #4299, opened 2016-02-04
using AUTH constants for menu generation

### LINKED ISSUE #4297: Change menu building to use Auth::READ constant variable
Hello, I'd like to suggest the following small change, but wanted to submit it as an issue before providing a PR. 

We are using the cancan adapter and need to namespace our permissioning with `:admin_read`, `:admin_write`, etc. to differentiate between admin abilities and in-app abilities. So we've monkey-patched the below

```ruby
module ActiveAdmin
  module Authorization
    READ    = :admin_read
    CREATE  = :admin_create
    UPDATE  = :admin_update
    DESTROY = :admin_destroy
  end

  Auth = Authorization
end
```

However, when AA builds the menu in the admin portal it runs this bit of code: 

```ruby
     # activeadmin/lib/active_admin/resource/menu.rb
      def default_menu_options
        # These local variables are accessible to the procs.
        menu_resource_class = respond_to?(:resource_class) ? resource_class : self
        resource = self
        {
          id: resource_name.plural,
          label: proc{ resource.plural_resource_label },
          url:   proc{ resource.route_collection_path(params) },
          if:    proc{ authorized?(:read, menu_resource_class) }
        }
      end
```

I'd like to change this to:

```ruby
     # activeadmin/lib/active_admin/resource/menu.rb
      def default_menu_options
        # These local variables are accessible to the procs.
        menu_resource_class = respond_to?(:resource_class) ? resource_class : self
        resource = self
        {
          id: resource_name.plural,
          label: proc{ resource.plural_resource_label },
          url:   proc{ resource.route_collection_path(params) },
          if:    proc{ authorized?(Auth::READ, menu_resource_class) }
        }
      end
```

Any potential issues? Seems pretty safe to me, but would like your blessing.

daino3:
closes #4297 

Admittedly, I can't 100% say this doesn't break something.  I pulled the latest down from master, and it seems the spec suite is in a weird state. I'm getting errors simply from the randomization of the suite.  Fixing the specs is outside the scope of this PR and this reallllly shouldn't change anything. It's also used in our application as a monkey-patch and hasn't failed us.

### DIFF
--- a/lib/active_admin/resource/menu.rb
+++ b/lib/active_admin/resource/menu.rb
@@ -27,7 +27,7 @@ module ActiveAdmin
           id: resource_name.plural,
           label: proc{ resource.plural_resource_label },
           url:   proc{ resource.route_collection_path(params) },
-          if:    proc{ authorized?(:read, menu_resource_class) }
+          if:    proc{ authorized?(Auth::READ, menu_resource_class) }
         }
       end
 


timoschilling:
:+1: will be merged after the travis run

seanlinsley:
For some reason it doesn't look like Travis ever ran for this PR. After fixing the test suite on master (c46167c683d416267f7b8ac010a977c72ca5bfc1), I ran the tests on this PR locally and it all passed.

Thanks for the PR!

PR #4310, opened 2016-02-09
Add gem `i18n-tasks` to maintain locale files

yhirano55:
I found many missing keys in `config/locale/*.yml`, so I added gem `i18n-tasks` and set up this to easy maintainable all locale files.
## What's i18n-tasks ?

[glebm/i18n-tasks](https://github.com/glebm/i18n-tasks)

> Manage translation and localization with static analysis, for Ruby i18n

I think really useful gem.
## Feature
- find missing keys: `bundle exec i18n-tasks missing`
- find missing keys and filling value in base locale (en): `bundle exec i18n-tasks add-missing`
- normalize keys (sorting a-z) all locale files: `bundle exec i18n-tasks normalize`
- If it found missing keys, spec will be failure: `bundle exec rspec spec/i18n_spec.rb`
## Next Step
- We should fill (or translate) all missing keys: `bundle exec i18n-tasks missing`
- Enable pending example `spec/i18n_spec.rb` after filling all missing keys.

### DIFF
--- a/spec/unit/i18n_spec.rb
+++ b/spec/unit/i18n_spec.rb
@@ -10,3 +10,26 @@ Dir.glob('config/locales/*.yml') do |locale_file|
     it { is_expected.to be_a_subset_of 'config/locales/en.yml' }
   end
 end
+
+if RUBY_VERSION >= '2.1' && !RUBY_PLATFORM.include?('java')
+
+  require 'i18n/tasks'
+
+  describe 'I18n' do
+    let(:i18n) { I18n::Tasks::BaseTask.new }
+    let(:missing_keys) { i18n.missing_keys }
+
+    pending 'FIXME: Run these examples after filling all missing keys' do
+      it 'does not have missing keys' do
+        expect(missing_keys).to be_empty,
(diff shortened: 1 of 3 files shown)


yhirano55:
I have thrown away commit that it executed `bundle exec i18n-tasks add-missing`.

spec/unit/i18n_spec.rb
```diff
@@ -10,3 +10,26 @@
     it { is_expected.to be_a_subset_of 'config/locales/en.yml' }
   end
 end
+
+if RUBY_VERSION >= '2.1' && !RUBY_PLATFORM.include?('java')
```
glebm: i18n-tasks is tested with JRuby, did it not work for you?

timoschilling:
This setup is to complicated.

seanlinsley:
i18n-tasks can use [Google Translate](https://github.com/glebm/i18n-tasks#google-translate-missing-keys) to fill in missing translations. I think we should consider it some more.

deivid-rodriguez:
Closing in favor of #4932.

PR #4317, opened 2016-02-14
Enhance PunditAdapter to support pundit_user hooks

syndbg:
Adds support for https://github.com/elabs/pundit#customize-pundit-user

What yet needs to be done:

* [ ] add tests regarding new functionality.

I'm not confident in my knowledge regarding the tests setup and test ideology here.
Any help on how to test this will be very helpful.


timoschilling:
Why this change? Whats the difference between `current_active_admin_user` and `pundit_user`?

lib/active_admin/base_controller/authorization.rb
```diff
@@ -68,11 +68,24 @@ def authorize_resource!(resource)
...
+                                               else
+                                                 current_active_admin_user
+                                               end
+
+                                        adapter.new(active_admin_config,
+                                                    user)
+                                      else
+                                        adapter.new(active_admin_config,
+                                                    current_active_admin_user)
+                                      end
```
timoschilling: This don't follow the adapter pattern. Every adapter related code should be placed in the adapter.

lib/active_admin/base_controller/authorization.rb
```diff
@@ -68,11 +68,24 @@ def authorize_resource!(resource)
       end
 
       # Retrieve or instantiate the authorization instance for this resource
-      #
       # @return [ActiveAdmin::AuthorizationAdapter]
       def active_admin_authorization
-        @active_admin_authorization ||=
-         active_admin_authorization_adapter.new active_admin_config, current_active_admin_user
+        return @active_admin_authorization if @active_admin_authorization
```
timoschilling: This is a bad pattern, write it:
```ruby
@active_admin_authorization ||= begin
  ...
end
```

syndbg:
@timoschilling A common way to support a custom `user` method throughout the whole app and not specifically ActiveAdmin.

timoschilling:
You can set the user auth method via `config.current_user_method`, `current_active_admin_user` is only a internal method which calls the `config.current_user_method`.

I think there is no reason for your change anymore or do you think different?

seanlinsley:
I agree with @timoschilling, it doesn't make sense to use different users between Active Admin, and Pundit.

AvaelKross:
I need this feature too. My case: I want to pass an extra variable to the pundit policy initializer. Normally I can do it by overriding `pundit_user` method in `ApplicationController`. But ActiveAdmin is ignoring this method..

gdufloux:
I agree with @timoschilling and @seanlinsley, and there's a non intrusive alternative with a custom adapter.

@AvaelKross you should be able to manage it with a custom class for your pundit user like

```rb
# app/lib/custom_pundit_user.rb (or whatever name)
class CustomPunditUser
  attr_reader :user, :extra_stuff

  def initialize(user: nil, extra_stuff: nil)
    @user = user
    @extra_stuff = extra_stuff
  end
end
```

that your can instanciate with your extra stuff (e.g. some context arround or whatever thing you need)

```rb
# app/controllers/application_controller.rb
class ApplicationController
  private

  def pundit_user
    CustomPunditUser.new(user: current_user, extra_stuff: "here")
  end
end
```

and then adjust with your own adapter for active admin (inheriting from PunditAdapter) to fit your pundit expectatioin

```rb
# config/initializers/active_admin.rb (or whatever place before ActiveAdmin.setup)
module ActiveAdmin
  class CustomPunditAdapter < PunditAdapter
    def initialize(resource, user)
      @resource = resource
      @user = CustomPunditUser.new(user: user, extra_stuff: "here")
    end
  end
end

ActiveAdmin.setup do |config|
  config.authorization_adapter = ActiveAdmin::CustomPunditAdapter
end
```

Hope it helps !

timoschilling:
@gdufloux thx for the good suggestion!

Issue #4150, opened 2015-10-05
1.0.0.pre2 Could not translate for key [ransack.predicates.contains]


istana:
Hi, I get this error. Does anybody know what it means?

````
RuntimeError in Admin::Users#index Showing /home/vagrant/.rbenv/versions/
2.1.6/lib/ruby/gems/2.1.0/gems/activeadmin-1.0.0.pre2/app/views/active_admin/
resource/index.html.arb where line #2 raised: ------------------ Could not
translate for key [ransack.predicates.contains] Extracted source
(around line #6): 4 5 6 7 Rails.logger.error(message) 
warn message raise message }
````

timoschilling:
double of #3958

istana:
I looked into this for a bit and found out that long forms of predicates are not translated

Issue #3958, opened 2015-05-22
Predicates are i18n'ed, but ransacked scopes aren't


k3rni:
Even when i18n translations for attributes are provided, they remain untranslated in the "Search status" box. This is understandable for custom ransackers, but surprising for simple ones that match attribute names directly.

In [lib/filters/humanized.rb](/activeadmin/activeadmin/blob/master/lib/active_admin/filters/humanized.rb), `parse_parameter_body`, the predicate is correctly dismantled into its field name and operator. But then (in `body`) the field name is presented as is, with no i18n applied.

What it should do, is look it up under:
1. `ransack.attributes.#{resource_class_name}.#{scope_name}`. This is actually supported by ransack itself, through `Ransack::Translate.attribute`, albeit used only in Ransack's own form helper.

and then, as a fallback:
1. `activemodel.attributes.#{resource_class_name}.attributes.#{scope_name}`
2. same as above, looking up the class hierarchy
3. same as the two above, but under activerecord
4. something like `active_admin.ransack_scopes.#{resource_class_name}.#{scope_name}`

Fivell:
@k3rni , please try master branch now,

k3rni:
Testing with a new trivial app on Rails 5.0.2. Can confirm it correctly displays translated attribute name both in filter box and search status box. Also traverses model class hierarchy correctly. However, only translations provided under `activerecord` scope work, `activemodel` and `ransack.attributes` don't; this is not a big issue.

Closing, because the issue is solved. Thanks!

PR #4330, opened 2016-02-29
translate custom Ransack predicates

istana:
The issue is described here https://github.com/activeadmin/activeadmin/issues/4150 and here https://github.com/activerecord-hackery/ransack/pull/627.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -134,3 +134,10 @@ en:
       block: "List"
       grid: "Grid"
       blog: "Blog"
+    ransack:
+      predicates:
+        equals: "equals"
+        contains: "contains"
+        starts_with: "starts with"
+        ends_with: "ends with"
+


timoschilling:
we don't accept 3 party keys in our locales

istana:
Ah, ok. I thought it originates from activeadmin.

seanlinsley:
Those should be `active_admin.filters.predicates` key. For example:
```yaml
ar:
  active_admin:
    filters:
      buttons:
        filter: "تصفية"
        clear: "تفريغ التصفية"
      predicates:
        contains: "يحتوي"
        equals: "متساوي"
        starts_with: "يبدأ بـ"
        ends_with: "ينتهي بـ"
        greater_than: "أكبر من"
        less_than: "أقل من"

seanlinsley:
Actually, they already are for English. Looks like the issue is #3958. I'm not sure I understand the problem after my quick reading of the ticket, but we're going to need to think through a solution.

istana:
now I understand why it is a duplicate of #3958

Issue #3529, opened 2014-10-28
Date range filter doesn't cover the last day


timoschilling:
refs #2374 and activerecord-hackery/ransack#458

![date filter](https://cloud.githubusercontent.com/assets/165599/4809100/54243088-5ea8-11e4-86c6-9be135707327.jpg)

With this UI we create a request like `?q[date_lteq]=2014-10-29`, put this queries the db with ````appointments`.`date` <= '2014-10-29 00:00:00'````, but it should be ````appointments`.`date` <= '2014-10-29 23:59:59'````, otherwise it not cover the last date of the range.

We could solve this via a hidden field for `date_lteq` witch adds the `23:59:59` via JS, like we do it with the predict select.

chrp:
@timoschilling - here's our temporary solution for your problem: 

Change date_range filter to use a custom ransack predicate:
```ruby
module ActiveAdmin
  module Inputs
    module Filters
      class DateRangeInput < ::Formtastic::Inputs::StringInput
        def lt_input_name
          "#{method}_lteqdate"
        end
      end
    end
  end
end
```
And define that predicate in an intializer or next to that monkey patch: 
```ruby
Ransack.configure do |config|
  config.add_predicate 'lteqdate',
    arel_predicate: 'lteq',
    formatter: -> (v) { v.strftime '%Y-%m-%d 23:59:59.9999' }
end
```
So ```lteqdate``` provides a more intuitive ```lteq``` for dates.

keithlayne:
@chrp - I like your approach - my previous workaround for this stopped working when I started using master.  However, your formatter caused problems for me considering timezones.  I ended up using this predicate instead, and it seems to do what I want:

```ruby
config.add_predicate 'lteqdate',
  arel_predicate: 'lt',
  formatter: -> (v) { v + 1.day }
```

chrp:
@keithlayne Thanks, your approach is indeed timezone proof when there are differences between application time and UTC.

kaelumania:
+1

nitsujri:
+1

seanlinsley:
It seems like the only reasonable solution is @chrp's. If someone would like to open a PR, we'd be happy to review it.

groony:
#4340

hanumakanthvvn:
I also stuck on the same issue, any update on the PR.

lucasfais:
[...]

PR #4341, opened 2016-03-11
3529 fix date range cover last day

groony:


groony:
#3529

timoschilling:
please rename `lteqdate` to `lteq_date` and provide the same for `gteq`

lib/ransack_ext.rb
```diff
@@ -9,4 +9,8 @@
   {'equals'=>'eq', 'greater_than'=>'gt', 'less_than'=>'lt'}.each do |old,current|
     config.add_predicate old, arel_predicate: current
   end
+
+  config.add_predicate 'lteqdate',
+  	arel_predicate: 'lt',
+  	formatter: -> (v) { v + 1.day }
```
timoschilling: Way not `v + 1.day - 1.second`, like in the previews commit?

groony:
@timoschilling done

lib/ransack_ext.rb
groony: For  ``v + 1.day - 1.second`` predicate **lteq**, but for ``v + 1.day`` predicate **lt**
I think second variant better,

lib/ransack_ext.rb
```diff
@@ -9,4 +9,11 @@
   {'equals'=>'eq', 'greater_than'=>'gt', 'less_than'=>'lt'}.each do |old,current|
     config.add_predicate old, arel_predicate: current
   end
+
+  config.add_predicate 'gteq_date',
+    arel_predicate: 'gteq'
+
+  config.add_predicate 'lteq_date',
+  	arel_predicate: 'lt',
+  	formatter: -> (v) { v + 1.day }
```
timoschilling: Way not v + 1.day - 1.second, like in the previews commit?

groony: For  ``v + 1.day - 1.second`` predicate **lteq**, but for ``v + 1.day`` predicate **lt**
I think second variant better,

timoschilling: ok

timoschilling:
I will think about this later this day or at the weekend

groony:
ok, tnx

seanlinsley:
The test suite is currently failing

groony:
@seanlinsley now only for rails 5 and there is suite that has no relation to this pull request

seanlinsley:
Thanks for that @groony. Would you mind squashing this into a single commit, and also adding a test ensuring that it actually fixes the problem? That test would make sense in [`filters.feature`](https://github.com/activeadmin/activeadmin/blob/master/features/index/filters.feature). This will probably require updating the step definitions to let you set a custom `created_at` time. Let me know if you need help.

lib/ransack_ext.rb
```diff
@@ -9,4 +9,11 @@
   {'equals'=>'eq', 'greater_than'=>'gt', 'less_than'=>'lt'}.each do |old,current|
     config.add_predicate old, arel_predicate: current
   end
+
+  config.add_predicate 'gteq_date',
+    arel_predicate: 'gteq'
```
timoschilling: the `gteq` value should be converted to `<date> 00:00:00`

bestori:
PR seems valid, why does travis fail?

constXife:
So?

timoschilling:
merged by #4504

PR #4345, opened 2016-03-13
WIP: Routing specs refactor

seanlinsley:


timoschilling:
looks good!

seanlinsley:
The tests are failing though. I've had this code locally for a while. I haven't been able to figure out the failure, so I'm creating a PR so I don't forget about it. Also perhaps someone else has an idea on how to fix it.

seanlinsley:
The problem is that the `memory_spec.rb` starts failing on namespace objects. It seems to be related to #697 and https://github.com/rails/rails/issues/12195. I'm not sure how to fix it, and it's probably not worth it at this point.

For now, I've merged the important parts of this PR into #4346.

PR #4347, opened 2016-03-14
Updated to work with activeadmin-axlsx

emmanueldd:


timoschilling:
Why do you think that this is necessary?

PR #4346, opened 2016-03-13
Standardize value formatting for attributes_table and table_for

seanlinsley:
This is a subset of #2562

PR #4356, opened 2016-03-18
feat(Column): Add support for formatting date column

emilioeduardob:
This PR allows for passing a format or using the global date/time
format. I'm not sure whether to call the options `format` or `date_format`

### DIFF
--- a/docs/3-index-pages/index-as-table.md
+++ b/docs/3-index-pages/index-as-table.md
@@ -40,6 +40,16 @@ index do
 end
 ```
 
+Column also accepts a `format` option for overriding the global date or time format
+
+```ruby
+index do
+  selectable_column
+  column :title
+  column :created_at, format: :short
+end
+```
+
 Sometimes that just isn't enough and you need to write some view-specific code.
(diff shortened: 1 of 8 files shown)


timoschilling:
There is already the `block` for format the value.

I don't like this PR, @seanlinsley what do you think?

seanlinsley:
I think this would make sense. It would be similar to allowing `status_tag: true` to be passed, which I've considered in the past, but never added.

This does need to be rebased because of #4346.

Please also update the documentation with this new feature.

seanlinsley:
@emilioeduardob: ping

emilioeduardob:
Hey! great, I'll rebase and will add documentation for this

emilioeduardob:
How about now @seanlinsley :) ?

javierjulio:
Closing out as its also gotten stale. With @timoschilling on this one. I don't think this is a good change because at the end of the day the block feature solves this and any other customization, for example, timezones. If we were to provide format here and use a block, users could expect both to work as well. I think we ultimately solve the issue at hand.

PR #4362, opened 2016-03-24
Remove outdated monkeypatch

seanlinsley:
The comments say this was needed for Rails 3.0, which we no longer support. I haven't actually run the full test suite to see if it actually needs this code, so definitely don't merge this until Travis runs.


seanlinsley:
#4365 replaces this PR

PR #4365, opened 2016-03-26
Don't create cucumber Rails environments

seanlinsley:
This simplifies our test setup, making it easier to understand. The previous version created a cucumber and cucumber_with_reloading Rails environment, but I removed that in favor of setting a custom environment variable.


seanlinsley:
Travis is getting this error with Rails 4.2 on Ruby 1.9:
> Gem::InstallError: mime-types-data requires Ruby version >= 2.0.

I wish Bundler could automatically find an older version instead of us having to specify it...

timoschilling:
There is no older version of `mime-types-data` which supports Ruby < 2.
We need to look `mime-types` to `< 3` for Ruby < 1.9

PR #4390, opened 2016-04-12
Rails 5: Migration is missing "[5.0]" when auto created

bestori:
Add Support for Rails 5 style migrations (in comments) when running generator.


timoschilling:
I would be a better way to switch from a migration template to a `rails migration CreateComments ...` command

lib/generators/active_admin/install/templates/migrations/create_active_admin_comments_rails5.rb
```diff
@@ -0,0 +1,19 @@
+class CreateActiveAdminComments < ActiveRecord::Migration[5.0]
```
seanlinsley: This would be better:

```ruby
parent_class = ActiveRecord::Migration
parent_class = parent_class[5.0] if Rails::VERSION::MAJOR >= 5

class CreateActiveAdminComments < parent_class

**Comment by bestori** [comment]
I agree. Fixed accordingly.

**Comment by bestori** [comment]
I agree with @seanlinsley . Fixed accordingly.

**Comment by n-studio** [comment]
Build fails with the following error during the migration: (https://travis-ci.org/activeadmin/activeadmin/jobs/132374288)
```
== 20160523200040 CreateActiveAdminComments: migrating ========================
-- create_table(:active_admin_comments)
   -> 0.0016s
-- add_index(:active_admin_comments, [:namespace])
   -> 0.0010s
-- add_index(:active_admin_comments, [:author_type, :author_id])
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Index name 'index_active_admin_comments_on_author_type_and_author_id' on table 'active_admin_comments' already exists
/home/travis/build/activeadmin/activeadmin/bundle/ruby/2.3.0/gems/activerecord-5.0.0.rc1/lib/active_record/connection_adapters/abstract/schema_statements.rb:1144:in `add_index_options'
/home/travis/build/activeadmin/activeadmin/bundle/ruby/2.3.0/gems/activerecord-5.0.0.rc1/lib/active_record/connection_adapters/abstract/schema_statements.rb:717:in `add_index'
```

I suspect a bug in Rails 5.0.0.rc1.

Workaround: assign a non standard name to the indexes
```
class CreateActiveAdminComments < ActiveRecord::Migration[5.0]
  def change
    create_table :active_admin_comments do |t|
      t.string :namespace
      t.text   :body
      t.string :resource_id,   null: false
      t.string :resource_type, null: false
      t.references :author, polymorphic: true
      t.timestamps
    end
    add_index :active_admin_comments, [:namespace]
    add_index :active_admin_comments, [:author_type, :author_id], name: 'index_active_admin_comments_on_author'
    add_index :active_admin_comments, [:resource_type, :resource_id], name: 'index_active_admin_comments_on_resource'
  end
end
```

PR #4408, opened 2016-04-26
Sort load path files in ruby

NickLaMuro:
Description
-----------
Different OS/OS Versions return a different ordering for the same directory contents.  Because of this, we can't rely on the given operating system when autoloading files in the same order we expect in different environments, which can lead to startup issues between development and production.  

A full example of these inconsistencies can be found in the following gist:  https://gist.github.com/NickLaMuro/dd35ea9241bfb7e651037ec3bae7f8b3

Changes
-------
By first sorting the result of `Dir["#{path}/**/*.rb"]`, we make sure that the files are returned in the same order for each environment, regardless of OS inconsistancies.

We are also `uniq`ing the files returned after the sort, which retains the order defined in the `load_paths`, but removes any duplicates from entire list.  It is possible that doing that will lead to potentially unintendened concequences for the user, so that might be something that is not wanted.

I ran the tests locally, but a few tests were failing `rake spec` and `rake cucumber` before and after my changes were applied.  Because it seemed unrelated to what I was doing, I have left them failing for now since it seemed to be unrelated to my changes, and my additions don't seem to cause new failures.

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -217,7 +217,7 @@ module ActiveAdmin
 
     # Returns ALL the files to be loaded
     def files
-      load_paths.flatten.compact.uniq.flat_map{ |path| Dir["#{path}/**/*.rb"] }
+      load_paths.flatten.compact.uniq.flat_map{ |path| Dir["#{path}/**/*.rb"].sort }.uniq
     end
 
     def router
--- a/spec/unit/application_spec.rb
+++ b/spec/unit/application_spec.rb
@@ -121,6 +121,42 @@ describe ActiveAdmin::Application do
       FileUtils.touch(test_file)
       expect(application.files).to include(test_file)
     end
+
+    context "given a complex admin directory" do
+      before(:each) do
+        @tmpdir = Dir.mktmpdir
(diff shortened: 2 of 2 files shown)


timoschilling:
On which system it don't work?

NickLaMuro:
@timoschilling Not a question of "which system" in this case, but the difference between systems and it not being obvious what the error was between the two.

In the case I dealt with (using the example from the gist I linked to in the description), some modules were defined in the `concerns` directory to share functionality between parts of the admin views in the base `app/admin` directory.

In development, using the default `load_paths` value of `['app/admin']`, the globing of the files returned them in an order where the files in the `concerns` directory were loaded before the files that required them.  But in production after a ruby and a AMI update (we are on AWS in this particular scenario), the files that required the `concerns` were loaded before the concerns themselves.


Basically what I am fixing here is making dev/prod consistent in the load order so this would be caught on development instead of on production.  For myself, I was chasing my tail with this because I assumed it was a Ruby upgrade issue, and not a OS default sorting issue.  The Gist is about the best way that I can explain it, but let me know if I need to provide more clarification.

PR #4427, opened 2016-05-11
num_pages is deprecated

stadia:
will be removed in Kaminari 1.0. Please use total_pages instead.
so i changed name of method

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,6 +62,7 @@ group :test do
   gem 'database_cleaner' if rails_version != '> 5.x'
   gem 'guard-rspec', require: false
   gem 'jasmine'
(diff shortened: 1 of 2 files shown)


baberthal:
Nice work!

[The check on line 47](https://github.com/stadia/activeadmin/blob/757b5cfc762c3b2b79a2ace9bfe3983a54b9af09/lib/active_admin/views/components/paginated_collection.rb#L47) still checks to see if collection responds to `#num_pages`, so you may want to change it to `#total_pages` so an exception isn't raised once the collection no longer responds to `#num_pages`.

stadia:
thank u
and now, test is broken because capybara is wrong.
so changed version to downgrade

stadia:
double of #4458

PR #4441, opened 2016-05-20
capybara's matcher is wrong result

stadia:
latest version of capybara has wrong result. i don't know why.
anyway i changed version to downgrade


stadia:
#4492

PR #4445, opened 2016-05-24
Empty check boxes control is not saved

### LINKED ISSUE #4444: Empty check boxes control is not saved (as empty array)
On check boxes control; uncheck all boxes, save; empty array is not saved

sandfoxme:
Attempt to fix #4444 

I removed a hidden field suppressor but it still only works with explicit ```:hidden_fields => true```


sandfoxme:
default behavior is broken in formtastic
https://github.com/justinfrench/formtastic/issues/1200

however active admin should not restrict the hidden_fields option, so I hope it still will be merged

timoschilling:
can you explain what you want to fix?

sandfoxme:
It seems it works now

PR #4448, opened 2016-05-26
support turbolinks in has_many.js

kissrobber:
has_many function is broken.

### DIFF
--- a/app/assets/javascripts/active_admin/lib/has_many.js.coffee
+++ b/app/assets/javascripts/active_admin/lib/has_many.js.coffee
@@ -1,4 +1,4 @@
-$ ->
+$(document).on 'load DOMContentLoaded', ->
   # Provides a before-removal hook:
   # $ ->
   #   # This is a good place to tear down JS plugins to prevent memory leaks.
@@ -52,9 +52,7 @@ $ ->
   $(document).on 'change','.has_many_container[data-sortable] :input[name$="[_destroy]"]', ->
     recompute_positions $(@).closest '.has_many'
 
-  init_sortable()
   $(document).on 'has_many_add:after', '.has_many_container', init_sortable
-
 
 init_sortable = ->
   elems = $('.has_many_container[data-sortable]:not(.ui-sortable)')
@@ -63,6 +61,8 @@ init_sortable = ->
     handle: '> ol > .handle',
     stop:    recompute_positions
   elems.each recompute_positions
+
+$(document).on 'load DOMContentLoaded page:load turbolinks:load', init_sortable
 
 recompute_positions = (parent)->
   parent     = if parent instanceof jQuery then parent else $(@)


kissrobber:
Until this PR is merged, I'm overriding has_many.js.coffee by http://stackoverflow.com/questions/16435428/how-to-override-single-files-from-gem-assets-for-assetsprecompile

varyonic:
Anyone else using this workaround, agree with this solution?

app/assets/javascripts/active_admin/lib/has_many.js.coffee
```diff
@@ -52,9 +52,10 @@ $ ->
   $(document).on 'change','.has_many_container[data-sortable] :input[name$="[_destroy]"]', ->
     recompute_positions $(@).closest '.has_many'
 
-  init_sortable()
   $(document).on 'has_many_add:after', '.has_many_container', init_sortable
 
+$(document).on 'ready page:load', ->
```
ccallebs: On Rails 5, `page:load` becomes `turbolinks:load`. I believe it would be better to wrap the document ready closure in a function and call it like so:

``` coffeescript
ready_hook = ->
  # current content of $ ->

$(document).on 'ready', ready_hook
$(document).on 'page:load', ready_hook
$(document).on 'turbolinks:load', ready_hook
```

This *should* cover all cases, but I haven't tested it.

ccallebs: Additionally, you could include it using the initial syntax, as in:

``` coffeescript
$(document).on 'ready page:load turbolinks:load', ->
  # ...
```

I'm impartial between the two.

kissrobber: Thank you. I'm going to fix it.

kissrobber:
it's working fine now 👍

Fivell:
@ccallebs , can you approve this PR or not?

javierjulio:
I'm not that familiar with Turbolinks but I do understand how it works at a high level so I would expect these changes to be necessary. I have not used it though so its best to get review from someone else. The thing that I'm confused on is why its just the sorting that this changes and why not other parts since those other handlers affect the DOM too?

javierjulio:
We can't merge this in since its using a deprecated jQuery ready handler. It should use the `ready` method instead.

@kissrobber we can accept but we need to change this to use `$(document).ready()` method instead. Very similar changes to what you see in #4682.

kissrobber:
@javierjulio Thank you. but `ready()` does not solve this.
What I want to do is the function L1 is called only after first page load, and the function `init_sortable` is called on every turbolink page load.
`ready()` is seems to called on every turbolink page load.

kissrobber:
I'm going to try to use `on('load DOMContentLoaded')` mentioned in #4682 instead of the deprecated handler.

kissrobber:
Fixed.
Tested with jQuery 1 & 2 & 3 👍

javierjulio:
@kissrobber thanks so much for following through! We appreciate your help as we'd very much like to get this through for v1.1.

I don't think this is right though. We shouldn't be using this as its not advised too. You mentioned `ready()` was called on every turbolink page load which sounds wrong. It would only ever fire once so I think something could be wrong with the test app you might have used. Keep in mind it needs to be tested with and without turbolinks as it has to work for both cases. 

We should stick to using `ready()` here over the `on('load DOMContentLoaded')` as advised by jQuery. I think the sortable call needs to be changed to:

```js
$(document)
  ready(init_sortable)
  .on 'page:load turbolinks:load', init_sortable
```

With that one change does at least the sorting part work? Also assuming you have kept the removal of `init_sortable` from the first function since its now being called in the sample above for both cases (normal and turbolinks) and should work correctly.

After that I think we need to change the first line to just be:

```js
$(document).ready ->
```

And that should queue up everything normally when the page loads.

kissrobber:
@javierjulio Hi
I inspected my production code after reading your last comment.
By updating the turbolinks gem, this problem is solved.

I'm going to close this PR.
Thank you and sorry for taking your time.

javierjulio:
@kissrobber no problem at all, I'm glad we took our time and investigated. Thanks for continuing to follow up! 👍🏻❤️ Just to confirm, the only change you needed was to upgrade the Turbolinks gem? Does that mean you did **not** include the changes here or the changes in my comment?

kissrobber:
No, I did not include any changes 👍 
Old version Turbolinks gem with jQuery ver.1 seemed to break `ready()` and `$ ->`.

PR #4458, opened 2016-06-02
Deprecation warning fix for Kaminari

benbristow:


timoschilling:
What effect does this change have to the required kaminari version / which is the min version which supports `total_pages`?

sublimecoder:
@timoschilling looks like the changes was introduced in Kaminari 0.14.0  So the earliest supported version should be that one.

https://github.com/amatsuda/kaminari/blob/master/CHANGELOG.rdoc#0140

timoschilling:
Ok, we are already limited to Kaminari 0.15

spec/unit/views/components/paginated_collection_spec.rb
```diff
@@ -212,7 +212,7 @@ def paginated_collection(*args)
 
       describe "set to false" do
         it "should not show the total item counts" do
-          expect(collection).not_to receive(:num_pages)
+          expect(collection).not_to receive(:total_pages)
           expect(collection).not_to receive(:total_pages)
```
timoschilling: this row is doubled

stadia:
#4427

sublimecoder:
@benbristow Can you fix the line duplication so we can get this merged in to fix the deprecation warnings?

PR #4463, opened 2016-06-06
Fix Kaminari deprecation message

fredngo:
Fixes the following:

DEPRECATION WARNING: num_pages is deprecated and will be removed in Kaminari 1.0. Please use total_pages instead. (called from page_entries_info at (…)/lib/active_admin/views/components/paginated_collection.rb:135)


fredngo:
Never mind, others have already submitted PRs for this

PR #4468, opened 2016-06-09
delegate parameters to controller

kvokka:
Earlier I did not found the way to pass variable to controller, like with this example
```
    ActiveAdmin.register Foo do
      #some code  
      bar = 42
    
      controller do
        def scoped_collection
          # ??? bar
        end
      end
    end
```

and with this patch you can use
```
    ActiveAdmin.register Foo do
      #some code  
      bar = 42
    
      controller(bar) do |bar|
        @bar= bar
        def scoped_collection
          self.class.send :class_eval, '@bar' # 42
        end
      end
    end
```

Maybe there is easier way, I'll be very grateful for it.

### DIFF
--- a/lib/active_admin/dsl.rb
+++ b/lib/active_admin/dsl.rb
@@ -70,8 +70,8 @@ module ActiveAdmin
     #
     #   end
     #
-    def controller(&block)
-      @config.controller.class_exec(&block) if block_given?
+    def controller(*args,&block)
+      @config.controller.class_exec(*args,&block) if block_given?
       @config.controller
     end
 
--- a/lib/active_admin/dsl.rb
+++ b/lib/active_admin/dsl.rb
@@ -71,9 +71,16 @@ module ActiveAdmin
     #   end
     #
     def controller(*args,&block)
-      @config.controller.class_exec(*args,&block) if block_given?
+      if block_given?
+        @config.controller.class_exec(*args) do
+          args_names = block.parameters.map(&:last)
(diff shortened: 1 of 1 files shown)


timoschilling:
Defining variables on a DSL level looks like a bad idea to me.

Do you have a real world example why you want to do that?

kvokka:
I understand, that it will be much better to avoid this, but I do not understand how to avoid it in this case

```
[Foo, Bar].each do |base_class|
ActiveAdmin.register base_class do
  controller(base_class) do |base_class|
    def scoped_collection
      @ext_base_class = base_class
      self.class.send(:class_eval,'@ext_base_class').includes(:event, :customer, :seat_sector, ticket_journals: :ticket)
    end
  end
```

This 2 models are similar view an AA and different logic in model, I can not use STI (STI provide bugs with inhered classes, if STI type is used for polymorphic association).

I understand, that this hack will pollute DSL scope, but, it is easiest way to implement. As second Idea is the variant to upgrade it, and use for it custom method, defined in controller, so futore code will be cleaner, and all external variables will be placed at 1 object. At this point i just want to know, may be I can use something for this, which is already done?

Thank you

kvokka:
I've update the PR, so now you can run this way:
```
  controller(42,'fff') do |foo,bar|
    def scoped_collection
      foo # 42      
    end
  end
```

In this implementation you do not pollute DSL scope.

timoschilling:
I this is too much hacky code for a too much edge case

PR #4469, opened 2016-06-09
Remove deprecated Kaminari method num_pages

arr-dev:
Replace it with total_pages

Deprecation notice https://github.com/amatsuda/kaminari/blob/9d550c737a72fc1ddee71e979d358704593a5a6f/CHANGELOG.rdoc#0170

This also reduced number of spec failures to only 4, from 23.


timoschilling:
double of #4458

Issue #4376, opened 2016-04-01
Breadcrumbs should not link to show action with to_param


BrahimDa:
I have this code

```ruby
ActiveAdmin.register MyModel do
  actions :all, :except => :show
  ...
end
```

in case i use "to_param" (in model)  in order to use permalink instead of the "id" in the url, the breadcrumb genarte the same error like [#736](https://github.com/activeadmin/activeadmin/issues/736) issue

timoschilling:
wich active admin version do you use?

BrahimDa:
i'm in 1.0.0.pre2

bestori:
@BrahimDa , thanks for reporting! 

Can you provide a short gist link or code so we could review it?

Thanks!

timoschilling:
https://github.com/activeadmin/activeadmin/pull/4485#issuecomment-228703698

Issue #279, opened 2011-07-11
Routing Errors When to_param is Overwritten


stungeye:
I overwrote `to_param` in a model to [implement friendly urls](http://www.miguelsanmiguel.com/2011/03/17/slug-that-slash).  Active Admin can no longer link to the admin urls that require ids without throwing a routing error.

party.rb:

```ruby
# Should give me urls like this: /parties/23/name-of-party
def to_param
  "#{id}/#{slug}"
end

def slug
  self.name.parameterize
end
```

routes.rb:
```ruby
resources :parties, :constraints => { :id => /[0-9]+\/.+/ }
```

I believe this a different problem than the other `to_param` issues listed below as I'm getting a routing error, not a active record loading error. Specifically, I get a routing error when any active admin page attempts to build a link that requires an id.

Is there any way to force active admin to ignore the overwritten to_param method?

Other issue that involve to_param:

#35: https://github.com/gregbell/active_admin/issues/35
#62: https://github.com/gregbell/active_admin/issues/62
#191: https://github.com/gregbell/active_admin/issues/191

lukerollans:
This works perfectly for me :+1: 

Thanks @mdoyle13 

```
controller do
    defaults :finder => :find_by_slug
end
```

timoschilling:
I'm closing this, while it's a inherited_resources bug and not a ActiveAdmin one.

This should be the correct way to handle it:
```ruby
controller do
  defaults :finder => :find_by_slug
end
```

aprescott:
I think there's more than one problem being discussed on this issue, and `defaults finder: :find_by_slug` only addresses one of them.

As I see it:

1. Some people have a custom `to_param` which they need to tie back into their ActiveAdmin pages, so that things can be looked up properly.
2. Some people have a custom `to_param` but don't want it to have any effect on ActiveAdmin pages, so that everything is based on ID.

(2) causes a bunch of problems. It means that you have to specify `.id` on any `link_to`s in the view templates, but it also means that for stuff like `index` blocks, you have to reimplement the link to "View" a particular record.

```ruby
column "" do |foo|
  link_to "View", admin_foo_path(foo.id)
end
```

That's very hacky and isn't very sustainable in the face of future changes to how ActiveAdmin renders the index table, or handling permissions for which links are visible by a particular admin user. (It's also fairly cringeworthy that other solutions posted here include dynamically changing method definitions on the fly.)

[...]

PR #4485, opened 2016-06-21
fix links with model using to_params

ThomasSevestre:
Some links do not work when using to_params.
For instance when you click to show an instance from the list.

This might be related to #4376


timoschilling:
If you overwrite `to_params` you brake Rails behavior.

ThomasSevestre:
Overwriting `to_params` is very common and used to be supported by active admin.

Please check this issue :
https://github.com/activeadmin/activeadmin/issues/279

PR #4498, opened 2016-06-27
Fixes #4342 Corrects missing filters in #index when using a filter. (…

### LINKED ISSUE #4342: No current filters when use filters in sidebar
When i use filters i see **no current filters** in sidebar.

armahillo:
@timoschilling -- Re my other PR -- this is just the one feature. 

Previously: Filters applied would say 'none' even if filters were applied, due to mismatch in params hash expectation :q vs. 'q'.

### DIFF
--- a/lib/active_admin/filters/active.rb
+++ b/lib/active_admin/filters/active.rb
@@ -16,7 +16,7 @@ module ActiveAdmin
       private
 
       def build_filters
-        filters = @params[:q] || []
+        filters = @params[:q] || @params["q"] || []
         filters.map{ |param| Humanized.new(param) }
       end
 


timoschilling:
Rails 4 or 5?

armahillo:
Rails 4

timoschilling:
I think your try is wrong way. We should fix this commit https://github.com/armahillo/activeadmin/commit/4effc7be0f509752eadeb09118f977b05d53399a, `@params` need to be a `ActionController::Parameters` or `ActiveSupport::HashWithIndifferentAccess`. Maybe we don't need to call `to_unsafe_h` in https://github.com/armahillo/activeadmin/commit/4effc7be0f509752eadeb09118f977b05d53399a#diff-282009a178f2f0176aca3b953fdc4406R11.

Maybe you like to create a PR going that way.

PR #4499, opened 2016-06-27
Fixes #2374 where end date does not include terminal day

### LINKED ISSUE #2374: Datetime field search result doesn't include its end date.
When I choose from 2013-07-23 to 2013-07-26 for a datetime field with a jQuery date_picker on a resource index page, the request parameters it creates is `"q"=>{"created_at_gte"=>"2013-07-23", "created_at_lte"=>"2013-07-26"}` which is what I intended to search.

However the generated sql is 

```
SELECT COUNT(*) 
  FROM `users` 
WHERE `users`.`guest` = 0 
    AND (`users`.`created_at` >= '2013-07-22 15:00:00') 
    AND (`users`.`created_at` <= '2013-07-25 15:00:00')
```

which doesn't include 2013-07-26 but searches from 2013-07-23 to 2013-07-25. (My time zone is +9) So I have to tell my colleagues who use an ActiveAdmin-based admin that they should select +1 day in order to get their desired result.

Is it a correct behaviour or am I missing something?

armahillo:
There were a number of fixes discussed for this, this was the one that fit best.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -31,6 +31,7 @@ en:
         ends_with: "Ends with"
         greater_than: "Greater than"
         less_than: "Less than"
+        lteqdate: "Before"
     search_status:
       headline: "Search status:"
       current_scope: "Scope:"
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -31,6 +31,7 @@ es:
         ends_with: "Termina con"
         greater_than: "Mayor que"
         less_than: "Menor que"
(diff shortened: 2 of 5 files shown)


dlackty:
A similar fix has been merged as #4499.

timoschilling:
@dlackty you mean #4504

timoschilling:
@armahillo I have preferred @dlackty's PR over yours, while the code was cleaner and has based on the older PR #4341

PR #4504, opened 2016-06-30
Fix ActiveAdmin::Inputs::Filters::DateRangeInput cover last day

### LINKED ISSUE #3529: Date range filter doesn't cover the last day

dlackty:
This is a rebased version of #4341 and should close #3529.
Thanks @groony for the work!

### DIFF
--- a/lib/active_admin/inputs/filters/date_range_input.rb
+++ b/lib/active_admin/inputs/filters/date_range_input.rb
@@ -15,12 +15,12 @@ module ActiveAdmin
         end
 
         def gt_input_name
-          "#{method}_gteq"
+          "#{method}_gteq_date"
         end
         alias :input_name :gt_input_name
 
         def lt_input_name
-          "#{method}_lteq"
+          "#{method}_lteq_date"
         end
 
         def input_html_options(input_name = gt_input_name)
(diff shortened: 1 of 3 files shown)


timoschilling:
Thanks to @groony and you

bartoszkopinski:
@timoschilling @dlackty I believe this breaks filtering for `date` fields. This is the resulting query for range `2016-05-02` - `2016-05-02`:

```
SELECT  1 AS count_column FROM "dates" WHERE ("dates"."starts_on" >= '2016-05-01 23:00:00.000000' AND "dates"."starts_on" < '2016-05-03') LIMIT 30 OFFSET 0
```
It tries to use date time ranges for date fields resulting in timezone conversion. Then `2016-05-01 23:00:00.000000` gets downcasted to `2016-05-01` and the results are invalid.

My temporary workaround was to monkeypatch two methods:

```ruby
ActiveAdmin::Inputs::Filters::DateRangeInput.class_eval do
  def gt_input_name
    case column.type
    when :date
      "#{method}_gteq"
    else
      "#{method}_gteq_date"
    end
  end

  def lt_input_name
    case column.type
    when :date
      "#{method}_lteq"
    else
      "#{method}_lteq_date"
    end
  end
end
```

Should I open a separate issue?

timoschilling:
@bartoszkopinski yes please

PR #4510, opened 2016-07-03
WIP: Upgrade test infrastructure

dlackty:
In order to simplify contribution & development process, I plan to:
- [ ] Use appraisal to test against multiple Rails versions
- [ ] Remove custom S3 cache and use Travis CI cache instead
- [ ] Create a shared Rails test app instead of generating during test process

See also: #3640 and #4161


.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: ruby
 sudo: false
-install:
+cache: bundler
+bundler_args: --without development
+before_install:
   - gem install bundler # use the very latest Bundler
-  - bundle install --without development --path=./bundle
-  - bundle clean # delete now-outdated gems
+after_isntall:
```
timoschilling: `after_install`

timoschilling:
Please see also https://github.com/activeadmin/activeadmin/pull/3595

timoschilling:
>  Create a shared Rails test app instead of generating during test process

Is this really a good idea? Every Rails version has its own scaffold. Maybe we can simple cache the generated app in travis.

timoschilling:
any updates?

PR #4514, opened 2016-07-05
Updated to Kaminari 0.17.0

marclennox:
Without this update, any apps using ActiveAdmin are prevented from using the latest release of `has_scope` (0.7.0).


timoschilling:
`s.add_dependency 'kaminari', '~> 0.15'` allow it to use any version between `0.15` and `1.0`

timoschilling:
There will be come a new `inherited_resources` for us. https://github.com/josevalim/inherited_resources/pull/411

marclennox:
Got it, thanks.

PR #4515, opened 2016-07-06
Specifying kaminari version for Rails 5 is no longer needed

5t111111:
Specifying the branch of kaminari for using Active Admin with Rails 5 seems to be no longer needed since kaminari 0.17 is released.


seanlinsley:
Note that when updating the readme you also need to update the Gemfile. I'm currently doing that 😄

PR #4527, opened 2016-07-14
Fix Cucumber specs for Devise

merqlove:
Subj. 
Seems `devise` and `arbre` is updated.


timoschilling:
Can you please provide a separate PR for the arbre fix, I want to merge that.

The devise fix can break with the next devise version. We should go a other way there.
We should inject our own text into the devise translation or we should read the devise translations and match against that. Then we are safe for future changes in devise.

merqlove:
Ok

merqlove:
Hi, it is. Prepared for locales.

timoschilling:
👍 just waiting for travis

PR #4528, opened 2016-07-14
Add Sprockets 4.x support. Update Rails 5 gem list in README.

merqlove:
Better support for Rails5 & new Sprockets 4.
Also added ability to test Sprockets with Rails 4.2+.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'test-unit', '~> 3.0' if rails_major == '3'
 
 if rails_major == '5'
   # Note: when updating this list, be sure to also update the README
+  gem 'sass-rails',          github: 'rails/sass-rails'
   gem 'inherited_resources', github: 'activeadmin/inherited_resources'
   gem 'ransack',             github: 'activerecord-hackery/ransack'
 end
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -43,6 +44,10 @@ group :development do
   gem 'rack-mini-profiler'                  # Inline app profiler. See ?pp=help for options.
   gem 'flamegraph', platforms: :mri         # Flamegraph visualiztion: ?pp=flamegraph
 
+  # Flamegraph dependency
+  gem 'stackprof', platforms: [:mri_21, :mri_22, :mri_23], require: false
+  gem 'fast_track', platforms: [:mri_19, :mri_20], require: false
+
```
timoschilling: why we need this?

merqlove: Due: https://github.com/SamSaffron/flamegraph/issues/19
It fails in dummy(it is when i tried with sprockets 4).

merqlove:
Timo, thank you a lot!
Just tried our Angular SPA & its monolith with AA backend, seems Sprockets 4 did the trick.
Speed similar with `Webpack` :)

PR #4529, opened 2016-07-14
Fix permitted params in dummy apps, for Rails 3, 4, 5

merqlove:


timoschilling:
Why do you that we need this? The tests are working

merqlove:
Have you tried to run dummy apps?
It fails when we try to save something in post, cat..'s, ...

merqlove:
Also in Rails 3, exclusively, seed fails within dummy.

PR #4534, opened 2016-07-15
Depend only on specific Rails components

lest:
Rails 5 introduces Action Cable that comes with additional dependencies.
Not all applications use Action Cable, so developers can choose to
include only required Rails components to their Gemfiles to avoid having
unused dependencies.

Removing the dependency on `rails` meta gem will allow using ActiveAdmin
in the applications where the developers decided to opt-out of some of
the Rails components.


timoschilling:
👍 

@seanlinsley what do you think?

PR #4538, opened 2016-07-19
Add back pretty formatting of non-symbol attributes.

kaspth:
3efb3595, switched around a bunch of code, but also scoped `pretty_format`-ing to only
symbol attributes.

This broke formatting in our app. In Active Admin v1.0.0.pre2 code like:

```ruby
show do
  attributes_table do
    row :mobile_subscription do
      subscription # ActiveRecord::Base object.
    end
  end
end
```

would auto link the returned model, but pre4 merely returns an `inspect` like output.

I think scoping `pretty_format` to only symbol attributes was an oversight as the
method clearly knows how to handle lots of different objects.

This refines `format_attribute` so we can still get nice status tags for boolean
fields, but get pretty formatted output for the rest of the values like pre2 and
earlier did.

### DIFF
--- a/lib/active_admin/view_helpers/display_helper.rb
+++ b/lib/active_admin/view_helpers/display_helper.rb
@@ -42,9 +42,14 @@ module ActiveAdmin
 
       def format_attribute(resource, attr)
         value = find_value resource, attr
-        value = pretty_format value                    if attr.is_a? Symbol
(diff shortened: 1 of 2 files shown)


kaspth:
Okay, I was a little too quick to call it an oversight for my own good. Sorry about that! ❤️

I do think it would be nice to do someting about this, since the behavior in pre4 doesn't match the one in pre2 as shown above.

PR #4551, opened 2016-07-29
Fix bug that was occluding active filter display.

martinstreicher:
Fixes https://github.com/activeadmin/activeadmin/issues/4342

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -74,5 +74,5 @@ group :test do
   gem 'sqlite3', platforms: :mri
   gem 'activerecord-jdbcsqlite3-adapter', platforms: :jruby
   gem 'poltergeist'
-  gem 'tins', '1.6.0'
+  gem 'tins', '< 1.3.4', platforms: :ruby_19
 end
--- a/Gemfile
+++ b/Gemfile
@@ -74,4 +74,5 @@ group :test do
   gem 'sqlite3', platforms: :mri
   gem 'activerecord-jdbcsqlite3-adapter', platforms: :jruby
   gem 'poltergeist'
+  gem 'tins', '1.6.0'
 end
--- a/lib/active_admin/filters/active.rb
+++ b/lib/active_admin/filters/active.rb
@@ -16,7 +16,7 @@ module ActiveAdmin
       private
 
       def build_filters
-        filters = @params[:q] || []
+        filters = @params['q'] || []
         filters.map{ |param| Humanized.new(param) }
       end
 


timoschilling:
interesting approach

martinstreicher:
I suppose I could make a more minimal change and change the lookup to use a string rather than a symbol, but in general, I hate having to know which is which. So, I just made it indifferent.

timoschilling:
whats happen if you use a `ActionController::Parameters` on Rails 5?

martinstreicher:
Hey, Tim, not sure. I am currently using this code on Rails 4, too. Would such a change make it tied to Rails 5?

timoschilling:
`(rails.version == 5 ? ActionController::Parameters : ActiveSupport::HashWithIndifferentAccess).new`

timoschilling:
Never mind, that would not work.

timoschilling:
But this could work:
```ruby
params = params.respond_to?(:to_unsafe_h) ? params.to_unsafe_h : params
@params = ActiveSupport::HashWithIndifferentAccess.new params
```

martinstreicher:
The previous line with to_unsafe_hash is superfluous.

martinstreicher:
The failures here are now related to 1.9. I got the other specs to pass. I'll look at some of the recent master builds and see if I can lock the gem that's causing the issue to a specific version. Thoughts?

martinstreicher:
Hmm... the `Gemfile.lock` is the same as in `activeadmin/master`. I get an error saying that tins 1.11.0 cannot be built because it requires Ruby 2.0. Any ideas?

Gemfile
```diff
@@ -74,4 +74,5 @@ group :test do
   gem 'sqlite3', platforms: :mri
   gem 'activerecord-jdbcsqlite3-adapter', platforms: :jruby
   gem 'poltergeist'
+  gem 'tins', '1.6.0'
```
Fivell: @martinstreicher , try 

>   gem 'tins', '< 1.3.4', platforms: :ruby_19

martinstreicher: Will do. 

What does that statement do on Ruby 2.*? 

Is it effectively ignored in Rubies later than 1.9 so that later versions of the _tins_ gem (`1.11.0`) are installed?

Committing now.

martinstreicher: That made 1.9 pass, but now there are different issues with JRuby. 

https://travis-ci.org/activeadmin/activeadmin/builds/149576226

timoschilling:
We don't accept PR's with gems like `tins`!

martinstreicher:
I did not add tins, it was already a dependency. I simply had to make it stick to a version. 

I don’t think comments like this are helpful.

martinstreicher:
I see a more recent commit fixed the dependencies on `tin` and listen. #4627

martinstreicher:
Ignore.

PR #4553, opened 2016-08-01
Update filter example to use "_action" in name

javierjulio:
Since this gem supports Rails 4 and 5 now going forward I thought it would be better despite this being commented out that it should use the current accepted filter name `before_action` since anything suffixed with `_filter` is deprecated in Rails 5.


timoschilling:
We still support Rails 3.x, so this will brake other thinks. But we keep this for ActiveAdmin 2.0, which drops Rails 3.x.

PR #4558, opened 2016-08-04
Don't reload routes twice on admin code changes.

kaspth:
If someone makes a change in an `app/admin` file, Rails will trigger a
routes reload by eval'ing `routes.rb` which in turn will call `routes`
on the Active Admin application.

Later, because we've inserted ourselves into the reloading through the
`to_prepare` and `to_complete` callbacks, `to_prepare` will unload our
namespaces while `to_complete` will retrigger a routes and application
reload.

Thus we can reliably skip the first routes reload whenver the reloader
is `updated?`. Because `execute_if_updated` is called if it is updated
the routes will be reloaded at that time. When `execute_if_updated` is
reloading the routes the reloader is no longer `updated?`. Finally, if
the code change wasn't in `app/admin` our reloader won't be `updated?`
there either.

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -239,7 +239,13 @@ module ActiveAdmin
     # One-liner called by user's config/routes.rb file
     def routes(rails_router)
       load!
-      router.apply(rails_router)
+
+      # If our reloader has been updated, we know it will trigger
+      # a routes reload later. By then the reloader will no longer
+      # consider itself updated and the routes will safely be reloaded.
+      unless defined?(@routes_reloader) && @routes_reloader.updated?
+        router.apply(rails_router)
+      end
     end
 
     # Adds before, around and after filters to all controllers.
@@ -260,6 +266,8 @@ module ActiveAdmin
       controllers.push *Devise.controllers_for_filters if Dependency.devise?
       controllers
     end
+
+    attr_writer :routes_reloader # :nodoc:
 
   private
 
@@ -310,6 +318,7 @@ module ActiveAdmin
         routes_reloader = app.config.file_watcher.new([], admin_dirs) do
           app.reload_routes!
         end
+        ActiveAdmin.application.routes_reloader = routes_reloader
 
         app.reloaders << routes_reloader
 


lib/active_admin/application.rb
```diff
@@ -250,6 +256,8 @@ def controllers_for_filters
       controllers
     end
 
+    attr_writer :routes_reloader # :nodoc:
```
kaspth: Don't intend this to a public API, but I'm unsure how you keep methods in the public namespace from showing up in documentation.

kaspth: Alternatively we could return the capture the reloader with `@routes_reloader = attach_reloader` in `prepare!`. That will just force a local variable capture into `attach_reloader`. Let me know which you prefer 😁

timoschilling:
can some rebase this to get the tests green

kaspth:
@timoschilling rebased 👍 — I'll keep an eye on the build ❤️

lib/active_admin/application.rb
```diff
@@ -233,7 +233,13 @@ def router
     # One-liner called by user's config/routes.rb file
     def routes(rails_router)
       load!
-      router.apply(rails_router)
+
+      # If our reloader has been updated, we now that it will trigger
```
varyonic: know

kaspth: Knowing what I now know, I'd have spelled this differently 🤓😄

kaspth:
Not sure why exactly this change causes the build to fail, if you have pointers I'll gladly take them 👍 

I'll try to investigate soon.

deivid-rodriguez:
@kaspth I'm working on #4805 and I just run into this same failure. Once #4805 lands, rebasing this PR should get it green.

deivid-rodriguez:
For the record, the fix for this particular failure should be https://github.com/activeadmin/activeadmin/pull/4805/commits/69ebfe66757c2c18e7b68e51c0333039b4ecb778.

kaspth:
@deivid-rodriguez I love you 😄

deivid-rodriguez:
@kaspth https://github.com/rails/rails/pull/28038 I love you too :heart_eyes:

deivid-rodriguez:
@kaspth Could you rebase this now?

kaspth:
@deivid-rodriguez rebased! Thanks for fixing the order dependence in the tests 😊

kaspth:
Well, some of the builds turned green 😄

Fivell:
@deivid-rodriguez , there so much failed specs, that looks not like random issue

deivid-rodriguez:
I wanted to have a look but didn't get the chance. But yes... although some times one failures triggers a lot of the failures...

Fivell:
@deivid-rodriguez , let's rerun build and see if something changes

deivid-rodriguez:
@kaspth Fixed one more issue affecting this in #4852. Could you try rebasing one more time?

kaspth:
@deivid-rodriguez you rock, thanks so much! I've rebased again 🤘

kaspth:
All green! 💸💸💸

Fivell:
@kaspth , any idea how to add spec for this to not break it in future ?

kaspth:
@Fivell not entirely sure, but here's a high-level quite-handwavy version:

Generate a template app to keep in the test folder, copy that to a tmp folder, cd into it, overwrite `@routes_reloader` with one that supers but also tracks the number of times we've triggered reloading, then boot a rails server, issue a request to the server, overwrite an admin file, issue the request again, then assert that the mocked `@routes_reloader`'s `reload` was only called once.

Fivell:
@kaspth, that will be too much I think :)

Fivell:
@varyonic , let's give this a try ?

deivid-rodriguez:
We have a `class-reloading` cucumber profile. Maybe it has the right setup to test this? Have someone tested this manually anyways? I mean other than @kaspth of course :)

javierjulio:
Sounds like a lot but it could be doable. If what @deivid-rodriguez suggests is right then worth a try. Otherwise I've rummaged through a lot of Rails to trust @kaspth's work. His input here is very valuable so I'm fine with taking it as is but I understand the concern about wanting a test.

kaspth:
I haven't used cucumber before, so I'm afraid trying my hand at a full stack test is a bit much for me at the moment. Though I've just enabled PR edits from maintainers so if you want to give it a go you can! 😊

We've also just upgraded our app to Active Admin 1.0 — works wonderfully, thanks! 💪

deivid-rodriguez:
I'm fine with taking this as is as well! But our cucumbers seem to have the right setup to test this, so I wanted to at least give it a try.

deivid-rodriguez:
@kaspth I had a look at this, but can't get it to work as I expect. This is what I'm doing:

* Generate a sample application and edit `config/routes.rb` dropping `puts "Reloading routes"` on top of it.
* Boot it and log into the admin.
* Edit a file in `app/admin`.
* Reload the page.

I seem to always get the message printed twice both with and without your changes. Any ideas?

kaspth:
I’m no longer with the project that required this change so I’ll just close this. Thanks for all the help and reviews and sorry for not getting back to you! ❤️

PR #4586, opened 2016-08-19
Allow line breaks in sortable column headers

gkop:
Currently, we allow line breaks in non-sortable column headers but not in sortable column headers. It's important to allow line breaks in all column headers because then we can squeeze more columns in without compromising on descriptive column headers. This change allows line breaks in sortable column headers as well as as non-sortable column headers.

In order to accomplish this with relatively clean CSS, I broke the existing sprite into three sprites. I think this is totally acceptable as ActiveAdmin is quite light on its use of sprites overall, and with the direction of HTTP/2 the extra requests are not a big deal.


varyonic:
Do you have before and after screenshots?

gkop:
before

after

(see the 3rd column from the right, "Days since placement")

javierjulio:
Going to close this out as we are very close to v2 and this will be addressed there. I don't want to introduce this because of the extra images. I know why its being done but I don't want to add more code on top of that. This will be fixed by using css arrows and to allow wrapping.

PR #4590, opened 2016-08-20
Custom Controller Actions: Allow create an action with the same name, that handles multiple HTTP verbs  a more natural way

chumakoff:
When somebody wants to create an action, that handles multiple HTTP verbs, they might write code like this:

          ActiveAdmin.register Post do
            member_action :action_name, method: :get do
              # do something on GET request
            end

            member_action :action_name, method: :post do
              # do something on POST request
            end
          end

This looks fine but doesn't work as expected. There is no magic  )
The second declaration will override the first one.

Furthermore, there will be 2 routes generated (GET, POST), that is strange. 
If there are two routes, why one behaviour?
At least, If action is overridden, then previous route should be removed.

This PR adds a lot of magic and automatically turns the above code into something like this:

          ActiveAdmin.register Post do
            member_action :action_name, method: [:get, :post] do
              if request.post?
                # do something on POST request
              else
                # do something on GET request
              end
            end
          end

In case when action name and http verb both the same as previous declared,  the latest declaration overrides the previous ones.

This works for both member and collection actions and with any combination of actions and http verbs.

This PR also fixes a bug. It is described in this commit https://github.com/activeadmin/activeadmin/commit/a7154b2faab4b3d0dbdff95baaef1e840dafb49f

### DIFF
--- a/lib/active_admin/controller_action.rb
+++ b/lib/active_admin/controller_action.rb
@@ -7,8 +7,8 @@ module ActiveAdmin
       @http_verb = resolve_http_verb(options)
     end
 
-    def remove_http_verb(verb)
-      self.http_verb = Array(http_verb) - Array(verb)
+    def remove_http_verbs(verbs)
+      self.http_verb = http_verb - verbs
     end
 
     private
@@ -17,13 +17,9 @@ module ActiveAdmin
 
     def resolve_http_verb(options)
       method = options[:method]
-      return :get unless method
+      return [:get] unless method
 
-      if method.is_a?(Array)
-        method.map(&:to_sym).uniq
-      else
-        method.to_sym
-      end
+      Array(method).map { |m| m.downcase.to_sym }.uniq
     end
   end
 end
--- a/lib/active_admin/controller_action.rb
+++ b/lib/active_admin/controller_action.rb
@@ -1,12 +1,29 @@
 module ActiveAdmin
   class ControllerAction
-    attr_reader :name
+    attr_reader :name, :http_verb
+
     def initialize(name, options = {})
-      @name, @options = name, options
+      @name = name.to_sym
+      @http_verb = resolve_http_verb(options)
     end
 
-    def http_verb
-      @options[:method] ||= :get
+    def remove_http_verb(verb)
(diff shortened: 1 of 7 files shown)


timoschilling:
Your PR is a oversized solution for an not existing Problem.

chumakoff:
I think there are at least two problems:

1) Creating several actions with the same name causes canceling previous action definitions, but doesn't cause cancelling previous routes. As a result, there are working routes, which are useless and may cause a problem. Action duplication might happen accidentally as well as a developer might specify different behaviours for different verbs of the same action in a separate block on purpose, expecting it to work this way. What if a developer creates a DELETE action, but accidentally (or on purpose) uses an action name, that was already specified with GET method? The old GET url will work and lead to deletion.

2) Creating several actions with the same name and verb causes routes duplication.

As for the first problem, I see 3 possible ways how to handle it:
  1. Raise an exception or produce a warning to let a developer know, that an action was defined already.
  2. Cancel both previous action definitions and routes.
  3. Add not only routes, but also specified definitions to the actions. 

As for the second problem, there is no doubt it should be resolved.

What do you think?

timoschilling:
There should be only one action per name. In a plain old rails controller you even can't do this:
```ruby
class FooController
  def delete
  end
  def delete
  end
```
The second `delete` overwrites the first one.

chumakoff:
That's true. But the problem is not that one action rewrites another. The problem is in routes. If an action is rewritten, the routes should be also rewritten. Otherwise, the new action will be available through the old route, which may cause a problem like the one I told about in my previous comment.

timoschilling:
The 'problem' with the doubled routes only happens when you call `member_action` twice with the same name. Why you should do that?

chumakoff:
As I already said, there are two possible scenarios:
  1. action duplication might happen accidentally
  2. a developer might specify different behaviours for different verbs of the same action in a separate block on purpose, expecting it to work this way.

timoschilling:
Handling 2 HTTP verbs in one action is an anti pattern. If someone want to do that he can make a `if` in the method. But the clean way is to handle it via two different methods.

PR #4591, opened 2016-08-21
Add the ability to specify Symbol and Proc as `allow_destroy` parameter of `f.has_many`

chumakoff:
When I generate nested has_many forms, I might whant to set the `allow_destroy` parameter 
depending on the child record.

For example: I want to not allow delete comments, that was created by an admin, 
or I want to not allow delete articles that are published.

I added the ability to specify Symbol and Proc as `allow_destroy` parameter.

If a Symbol is specified, the child object's method will be called. 

    # this will create 'delete' checkbox only for articles which #draft? method returns true
    f.has_many :articles, allow_destroy: :draft? do |a|

If a Proc is specified, the given blok will be calculated  for the child object.
The object is passed as the parameter of the block.

    f.has_many :comments, allow_destroy:  -> (comment) { !comment.author.admin? } do |c|

### DIFF
--- a/lib/active_admin/form_builder.rb
+++ b/lib/active_admin/form_builder.rb
@@ -59,7 +59,7 @@ module ActiveAdmin
           block.call has_many_form, index
           template.concat has_many_actions(has_many_form, builder_options, "".html_safe)
         end
-        
+
         template.assign(has_many_block: true)
         contents = without_wrapper { inputs(options, &form_block) } || "".html_safe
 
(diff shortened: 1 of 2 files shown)


lib/active_admin/form_builder.rb
```diff
@@ -137,5 +139,15 @@ def js_for_has_many(assoc, form_block, template, new_record, class_string)
...
+    def allow_destroy?(object, allow_destroy_option)
+      case allow_destroy_option
+      when Symbol
+        object.public_send allow_destroy_option
+      when Proc
+        allow_destroy_option.call(object)
+      else
+        allow_destroy_option
+      end
+    end
```
timoschilling: You should use this https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/view_helpers/method_or_proc_helper.rb here

chumakoff:
I edited the PR in order to use existing helper method from here https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/view_helpers/method_or_proc_helper.rb

This also allows use String as parameter.

chumakoff:
@timoschilling  If the parameter is not [String, Symbol, false, true, nil], should it be considered as true ? What do you think?

timoschilling:
@chumakoff yes, and since #4593 you can even more simplify your code

timoschilling:
It should be possible to remove `has_many_allow_destroy?` and direct use `call_method_or_proc_on`.

chumakoff:
Done.

timoschilling:
good work. can you please rebase your work on master.

chumakoff:
Thank you.  I've rebased the branch

timoschilling:
one last thing, can you write some lines in the docs?

chumakoff:
Ok, I will do it. Thank you for merging the PR!

PR #4426, opened 2016-05-10
Fix #4424, issue with new_with_cast from 6488d63

shaicoleman:

PR #4597, opened 2016-08-24
Revert "Fix #4424, issue with new_with_cast from 6488d63"

### LINKED ISSUE #4424: new_with_cast no longer works (regression)

timoschilling:
Reverts activeadmin/activeadmin#4426


alexgerstein:
@timoschilling In Rails 4, passing an STI subclass type to a build_association is supported. However, with this revert, ActiveAdmin builds the resource without passing the type attribute causing the child validations to be skipped.

This behavior was working as expected on pre2, but the revert is preventing an upgrade to pre4. Are there any plans to re-enable this?

PR #4619, opened 2016-09-08
Decorators now receive the action name as context

codyrobbins:
I’ve found when using decorators that I often want to decorate a value differently depending on whether it's being displayed in a form or not. For example, dollar amounts might be stored in the database as decimal numbers and displayed on the index and show pages with a dollar sign and cents but should be rendered in the form without the dollar sign and no cents.

Draper supports a context attribute for decorators and this change passes the current action name as the context when decorating resources. This will enable the decorator to decorate attributes differently depending on the page they are being shown on.

### DIFF
--- a/lib/active_admin/resource_controller/decorators.rb
+++ b/lib/active_admin/resource_controller/decorators.rb
@@ -5,7 +5,7 @@ module ActiveAdmin
     protected
 
       def apply_decorator(resource)
-        decorate? ? decorator_class.new(resource) : resource
+        decorate? ? decorator_class.new(resource, context: action_name) : resource
       end
 
       def apply_collection_decorator(collection)


Fivell:
I think this is already can be done by 
```ruby
controller do 
  def apply_decorator(resource)
      #custom logic for decoration
  end   
end
```
Don't think we should include this patch to master. 
@timoschilling what do you think?

timoschilling:
I agree with @Fivell

codyrobbins:
@Fivell I disagree that pasting boilerplate into every file is a preferable solution than ActiveAdmin taking advantage of a Draper feature that exists for this very reason. Why wouldn’t you want the context passed automatically given it’s such a trivial change?

Fivell:
@codyrobbins,  thanks for your reply. 
if you want to see context support to be merged your PR should cover more cases.
  - context was designed to be hash (not only action can be passed but also parent object if it has belongs_to resource for instance, need to thinks about other cases to setup default context.)
  - context should be passed also to collection decorator
  -  there should be DSL to allow users modify context

Fivell:
@codyrobbins , btw you have full access to helpers inside decorator.

```ruby
h.action_name

h.params
```
etc

timoschilling:
@codyrobbins you don't need to paste that code into every file, you can inject it into `ActiveAdmin:: ResourceController`

codyrobbins:
Thanks for the tip regarding access to the action name from the helper object inside the decorator—I wasn’t aware of that and that obviates the immediate need I had for this change!

Fivell:
@codyrobbins , you're welcome

PR #4621, opened 2016-09-08
Pluralized DSL methods

codyrobbins:
One of my long-standing annoyances with ActiveAdmin is the cruft of multiple invocations of the same DSL method over and over again when defining things like columns, filters, scopes, and form inputs:

``` ruby
index do
  column(:name)
  column(:price)
  column(:enabled)
end

filter(:name)
filter(:description)
filter(:price)
filter(:enabled)

scope(:all)
scope(:enabled)
scope(:new)
```

This drives me insane because I should just be able to do

``` ruby
index do
  columns(:name, :price, :enabled)
end

filters(:name, :description, :price, :enabled)
scopes(:all, :enabled, :new)
```

So this change adds the following pluralized DSL methods:
- `ActiveAdmin::Views::IndexAsTable#columns`
- `ActiveAdmin::CSVBuilder#columns`
- `ActiveAdmin::Filters::DSL#filters`
- `ActiveAdmin::ResourceDSL#scopes`
- `ActiveAdmin::Views::ActiveAdminForm#multiple_inputs`

(I would much prefer to call `multiple_inputs` just `inputs` but since that method is already proxied to Formtastic and it's part of the Formtastic DSL for creating a group of inputs, I left it as is.)

Each method can take a hash of options as the final argument which will be applied to each invocation of the equivalent singular method:

``` ruby
columns(:name,
        :price,
        :enabled,
        sortable: false)
```

In addition, you can pass hashes instead of symbols to specify options for individual attributes:

``` ruby
columns(:name,
        {price:   {sortable: true},
         enabled: {class: :enabled},
        :description,
        sortable: false)
```

Obviously there might be some discussion around a change like this, but if it looks good I will take the time to get the tests passing and write some additional tests for the new functionality.

### DIFF
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -26,17 +26,29 @@ module ActiveAdmin
       end
     end
 
-    attr_reader :columns, :options, :view_context
+    attr_reader :all_columns, :options, :view_context
 
     COLUMN_TRANSITIVE_OPTIONS = [:humanize_name].freeze
 
     def initialize(options={}, &block)
       @resource = options.delete(:resource)
-      @columns, @options, @block = [], options, block
+      @all_columns, @options, @block = [], options, block
+    end
+
+    def columns(*attributes)
+      options = attributes.extract_options!
+
+      attributes.each do |attributes|
+        attributes = {attributes => {}} unless attributes.is_a?(Hash)
+
+        attributes.each do |attribute, attribute_options|
+          column(attribute, options.merge(attribute_options))
+        end
+      end
     end
 
     def column(name, options={}, &block)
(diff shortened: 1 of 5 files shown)


Fivell:
hi @codyrobbins,  is this work in progress?

codyrobbins:
@Fivell No, it’s not. If you’re expecting something else not included in the commit then please let me know!

Fivell:
@codyrobbins , docs and specs are missing for this feature

javierjulio:
Closing as this has gone stale. I'm not even sure this is worth doing, adding the overhead since very quickly you'd outgrow this usage and would need to define each individually. We can discuss further but ultimately I'm not sure this is worth supporting.

PR #4505, opened 2016-06-30
WIP: Use Travis CI's cache instead

dlackty:
**WIP**

Try to simplify & probably speed up build process. 
Now it takes more than 30 minutes to finish a CI build.

PR #4622, opened 2016-09-08
Add i18n content for date range filter

mcon85:
I think this was broken in 1.0.0.pre3 because of the change in
#4505. The Ransack configuration was changed to add gteq_date and

lteq_date as predicates but ActiveAdmin's i18n keys were not updated
with this content. Ransack does not provide default values. Without the
i18n content you get translation errors in the sidebar search status on
a results page.

With translation error:

screen shot 2016-09-08 at 6 18 25 pm

With addition of i18n values:

screen shot 2016-09-08 at 5 30 24 pm

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -30,7 +30,9 @@ en:
         starts_with: "Starts with"
         ends_with: "Ends with"
         greater_than: "Greater than"
+        gteq_date: "Greater than or equal to"
         less_than: "Less than"
+        lteq_date: "Less than or equal to"
     search_status:
       headline: "Search status:"
       current_scope: "Scope:"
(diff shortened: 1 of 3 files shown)


varyonic:
@mcon85 did you mean to reference #4504, not #4505?  I think you may need to rebase and adjust for #4705.

rainchen:
I think the translation using `gteq_date: "Greater than or equal to"` is too "technology".

How about using following translation:

```
        lteq_date: "Before"
        gteq_date: "After"
```

then tips for filters scope will be:

```
Created At after 2016-09-07
Created At before 2016-09-08
```

Fivell:
@rainchen , agree

javierjulio:
@mcon85 I'd like to get this out for v1.1! Can you please rebase master on your branch to resolve the conflicts and change the translations to use "before" and "after"?

javierjulio:
This has been addressed in #5096 as I'd like to get this in for v1.1. Thanks @mcon85! 👏🏻❤️

PR #4589, opened 2016-08-19
Fix Travis CI

dlackty:
Finally have some time to work on #4510 but found that CI failed on master.

- [x] Lock `tins` version for Ruby 1.9
- [ ] Disable `ruby-dep` warning

Issue #4556, opened 2016-08-04
activeadmin doesn't work on rails 4.0.+ and 4.1.+ with ruby 1.9.3


Fivell:
Hi guys, I'm testing active_admin_import agains multiple rails versions.
since commit  https://github.com/activeadmin/activeadmin/commit/4292aafa33910b705c3076f36e0cb8b2322dbca9
I have broken support for rails 4.0 and 4.1 

```
ActionView::Template::Error:
       undefined method `to_h' for {}:ActionController::Parameters
     # /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/activeadmin-1.0.0.pre4/lib/active_admin/batch_actions/resource_extension.rb:58:in `batch_action_path'
     # /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/activeadmin-1.0.0.pre4/lib/active_admin/batch_actions/views/batch_action_form.rb:18:in `build'
```

looks like method to_h was added in rails 4.2

http://apidock.com/rails/ActionController/Parameters/to_h

Can you please comment this? Is active_admin able to support rails < 4.2 ? 

link to travis job with Matrix Ruby 1.9.3 + Rails 4.0.13
https://travis-ci.org/activeadmin-plugins/active_admin_import/jobs/149591932

link to travis job with Matrix Ruby 1.9.3 + Rails  4.1.15
https://travis-ci.org/activeadmin-plugins/active_admin_import/jobs/149591937

timoschilling:
In general we support Rails >= 3.2. But while @seanlinsley has removed Rails 4.0 and 4.1 from the tests, we miss that problem. I will try to get back running tests for 4.0 and 4.1.

agentargyle:
I am running into this problem as well. 

As an aside, I thought I was using Ruby 2.3.0 - running ruby -v tells me I am, as does rvm list. But the default rails index page's "About your application’s environment" link reveals my rails app is somehow still using Ruby 1.9.3. So I am hoping getting that straightened out resolves this for me.

Issue #4369, opened 2016-03-30
Rails 5: Migration is missing "[5.0]" when auto created


bestori:
The automated migration created for creating admin comments is currently: 

**class CreateActiveAdminComments < ActiveRecord::Migration**

When it **should be**:

**class CreateActiveAdminComments < ActiveRecord::Migration[5.0]**

Instead.

seanlinsley:
How do we make this change so that it's backwards-compatible?

bestori:
@seanlinsley maybe create the migration based on rails version...?

bestori:
Opened a matching PR.

PR #4624, opened 2016-09-09
Add Appraisal

deivid-rodriguez:
Closes #3640.
Closes #4510.
Closes #4589.

What do you think about this? The build is passing but I really don't want to bother installing and fighting against Ruby 1.9.3 & Rails 3, that were EOLed maybe about 2 years ago. <strike>So I dropped support for those.</strike> 1.9.3 support was restored, actually.


deivid-rodriguez:
Supersedes #4589 and #4510.

What do you think about this? The build is passing but I really don't want to bother installing and fighting against Ruby 1.9.3 & Rails 3, that were EOLed maybe about 2 years ago. So I dropped support for those.

deivid-rodriguez:
To compensate, now all the other supported Ruby + Rails combinations are tested in Travis :)

deivid-rodriguez:
Just saw #4618, if you really want to keep 1.9.3 & Rails 3.2 support I can reuse the fix in there (or rebase on top of it if it's merged first).

dlackty:
Fantastic work!
I wished I could achieve this in #4510 but didn't have enough time to finish it.

deivid-rodriguez:
@dlackty I reused your work so thanks for getting it started! :)

deivid-rodriguez:
Ok, I think I'm now happy with this so I removed the [WIP].

Let me know how you want to proceed.

activeadmin.gemspec
```diff
@@ -10,10 +10,13 @@ Gem::Specification.new do |s|
...
 
-  s.files         = `git ls-files`.split("\n").sort
+  s.files         = `git ls-files -z`.split("\x0").reject do |f|
+    f.match(%r{^(spec|features)/})
+  end
+
   s.test_files    = `git ls-files -- {spec,features}/*`.split("\n")
 
-  s.required_ruby_version = '>= 1.9.3'
+  s.required_ruby_version = '>= 2.0'
```
timoschilling: Why this? We are supporting 1.9 currently

timoschilling:
Nice work, but we can't merge this with dropping Ruby 1.9 and Rails 3.2

activeadmin.gemspec
deivid-rodriguez: Because I'm proposing to drop support.

deivid-rodriguez:
Ok, my intention was to not work at all in the Ruby 1.9 + Rails 3.2 specs since I don't want to install or mess with those.

But since there's another PR where I fix is proposed, I'll copy the fix from there and see if it passes Travis.

timoschilling:
Please do that! At the moment this PR is grate, but useless

deivid-rodriguez:
Sure, I'll get to it! I'll split the changes to drop 1.9 + 3.2 support and leave them in a separate PR, in case you want that sometime in the future.

deivid-rodriguez:
I've now restored support for old versions.

I'll now let Travis run and fix issues as they appear. It'll take a while since the matrix now is pretty big (that's a good remainder of the amount of things you are trying to support at the same time).

deivid-rodriguez:
Oh well, looks like I'll have to install 1.9.3 after all... :disappointed: 

So what's your plan, release 1.0.0 and then drop support? If yes, why?

timoschilling:
Yes we will make a 1.0.0 Release with Ruby 1.9 and Rails 3.2.

We have many users with use one of that dependences and our code base supports both.

zloyrusskiy:
Is there way to speed up new release? I'm using Rails 5, and would like to install activeadmin without additional options in Gemfile.

deivid-rodriguez:
> Yes we will make a 1.0.0 Release with Ruby 1.9 and Rails 3.2.

The more I investigate the more obscure your version policy looks to me. It seems that you've been _intentionally_ supporting only 1.9 with Rails 3.2 and 4.2 for quite a while. So you basically kept support for the old 3.2 version while breaking it for 4.0 and 4.1 (see #4556 and #4254).

> We have many users with use one of that dependences and our code base supports both. 

Any numbers to backup this? As I said, support for 1.9.3 + (4.0 & 4.1) has been broken for 6 months and only a confused user not aware of being using 1.9.3 and a plugin mantainer have complained. 

Anyways, fixes (including #4556) coming.

deivid-rodriguez:
This is ready.

I ended up not removing the `listen` dependency. That dependency comes from `guard` but since the `Guardfile` was last changed back in 2011, I assumed none of the current developers are using it. Given the complexity of the current set of dependencies, I think it's better to remove any unnecessary dependencies to try to prevent similar failures in the future.

Summarizing here the list of related issues for autoclosing on merge:

* Closes #3640.
* Closes #4390.
* Closes #4510.
* Closes #4589.

* Fixes #4369.
* Fixes #4556.

deivid-rodriguez:
@timoschilling Rebased on top of current master and removed 2 commits. One was completely unrelated to the PR, the other one had already landed in master through another PR.

You asked me to restore 1.9.3 & Rails 3.2 support and I've done it. Anything else you need here?

deivid-rodriguez:
Rebased again, and fixed two new failures because test dependencies dropped ruby 1.9 support.

deivid-rodriguez:
Rebased again. Any news @timoschilling?

deivid-rodriguez:
Tried rebasing this again and found that the build is again broken because dependencies keep dropping 1.9.3 support. Once maintainers state they want this and they are up for reviewing and merging, I'll fix this.

deivid-rodriguez:
Since we now have a new mantainer, @varyonic, and someone else, @Fivell, got master green again. I rebased this PR one more time! :smiley:

deivid-rodriguez:
Picked up this PR again. I extracted some (mostly unrelated) things to separate PRs, rebased on top of latest changes in master and improved some stuff.

@varyonic, TravisCI, @timoschilling Please review.

deivid-rodriguez:
I'm splitting stuff out of this PR into smaller ones to make it more digestible. So it's not ready yet

deivid-rodriguez:
Added 2.4.0 to Travis matrix. Run it locally and got a scary segmentation fault though :S

deivid-rodriguez:
This PR was huge so I've been extracting changes to smaller PR's. The idea is to leave this one only for migrating to `Appraisals`. I will ping you when it's ready.

deivid-rodriguez:
Ok, this is now ready. This is just one extra change I would like to do which is committing all bundler lock files to source control. This way we ensure that all developers AND CI run against the exact same dependencies so we are sure all environments are consistent and our build doesn't break because of changes in dependencies. What do you think?

deivid-rodriguez:
Squashed and lock files added!

Gemfile.lock
```diff
@@ -0,0 +1,210 @@
+GEM
```
timoschilling: I'm not sure that this is a good idea, maybe move this comment to a separate PR.

deivid-rodriguez: @timoschilling What's your concern? This obviously needs more maintenance but I thinks it's definitely less trouble than continuously broken builds because of 3rd party dependency changes. I strongly believe it is a good idea to enforce environment consistency between all developers and CI.

timoschilling: @deivid-rodriguez I agree with you that it brings trouble with 3rd party dependency changes. But this is a feature and not a bug 😉 . If commit the lock would find this issues and often have PRs/commits with dependency updates. Nearly every gem don't check in there Gemfile.lock. Event `bundle gem`'s gem scaffold tells you not do that. I disagree with this change, until some more/better reasons or all other maintainers will agree with it. Please remove it and if you still want a discussion create a new separate PR.

deivid-rodriguez: I can understand that reason although I disagree with it. It think it's a similar debate to using random vs deterministic data for testing (hope you can see the analogy). Support for third party dependencies should be done proactively, it should never get in the middle of unrelated work.

In any case, I understand this is controversial so I'll remove the change!

And by the way, I was really positive you would agree with this because you committed all .lock files to source control when you first tried to add `Appraisal` in #3640. I see now that it was an accident :)

deivid-rodriguez: Example: https://github.com/activerecord-hackery/ransack/pull/755... Some third party gem issue is getting in the middle and now you are forced to either merge a red PR (like I've seen you doing here in the past), hold back the change or start trying to figure out the problem with bundler/third party dependency... no good.

timoschilling: Please make a new PR for a dedicated discussion.

deivid-rodriguez: Commit adding lock files removed! Will open a separate PR once this lands.

.travis.yml
```diff
@@ -12,7 +12,7 @@ before_install:
   - source script/install_phantomjs.sh
   - gem update --system # use the very latest Rubygems
   - rvm @global do gem uninstall bundler -a -x
-  - rvm @global do gem install bundler -v 1.14.2 # latest version known to work
+  - rvm @global do gem install bundler -v 1.14.3 # latest version known to work
```
deivid-rodriguez: Well, I introduced this because of the same reason I proposed to commit lock files. https://github.com/activeadmin/activeadmin/pull/4744 was broken because of a buggy bundler release. @varyonic [pointed out](https://github.com/activeadmin/activeadmin/pull/4744#issuecomment-274338610) that it was too bad that our build breaks because of that, so I agreed and locked the version in https://github.com/activeadmin/activeadmin/pull/4744/commits/482734534adb69ca4eb08c889bfb810708f3ac8f.

timoschilling:
@deivid-rodriguez please check travis

deivid-rodriguez:
Yes, I'm planning to do that after I get merged and rebase on top of my other PR's.

deivid-rodriguez:
Ok, I reviewed this. One line had been lost at some point (probably during one of the multiple rebase/commit extractions). Fixed in https://github.com/activeadmin/activeadmin/pull/4624/commits/cee824c340c099552d341eda07b43976121d3f19. I guess you want me to squash that one with https://github.com/activeadmin/activeadmin/pull/4624/commits/4f29ce5481b6f80a909d82b1451311689e186538, right?

timoschilling:
please rebase

deivid-rodriguez:
Done!

deivid-rodriguez:
Nice. There's one last improvement I wanted to add here but I forgot. Make it so that the default `rake` task runs the tests against the latest Rails version. I'll add this as a separate PR if you find it useful.

PR #4618, opened 2016-09-08
3280 added nested belongs to rebased

rdallasgray:
This is a rebase of #3280 against the current master. We need this functionality and we're keen to see the PR merged. I understand the original author may not have had time to complete the rebase.

This commit also:
- Removes the extraneous `PostComment` model from the original PR and uses the existing `Tagging` model for specs
- Adds a feature spec for the breadcrumb links with nested `belongs_to` resources

PR #4627, opened 2016-09-12
Pin tins and listen gems to make tests pass

rdallasgray:
See https://github.com/activeadmin/activeadmin/pull/4618

This pins the `tins` gem, to make tests pass under Ruby 1.9, and the `listen` gem, to make tests pass under JRuby.


timoschilling:
THX!

PR #4632, opened 2016-09-19
Added note about how to remove select

thesunwave:
on text input field filter with a one parameter.


varyonic:
Edited and rebased.

thesunwave:
Thank you :)

PR #4641, opened 2016-09-25
Add the ability to customize the footer.

blimmer:
By default, it will still default to the "Powered by Active Admin `<version>`", but this allows folks to customize that message.

We wanted to append additional environmental information to better understand bug reports from our internal customers. To accomplish this today, I created `app/admin/footer.rb`, but it would be nice to set this as a configuration value.

In determining how to do this with my app, I found several resources ([stack overflow](http://stackoverflow.com/questions/20464690/how-do-i-edit-or-override-the-footer-of-activeadmin), [gists](https://gist.github.com/pcreux/3058773), [blog posts](http://reverbhq.com/blog/2012/08/building-web-apps-with-activeadmin/)) which indicated to me that this would be a useful feature for people.

### DIFF
--- a/docs/1-general-configuration.md
+++ b/docs/1-general-configuration.md
@@ -169,3 +169,11 @@ ActiveAdmin.setup do |config|
   end
 end
 ```
+
+## Footer Customization
+
+By default, Active Admin displays a "Powered by ActiveAdmin" message on every
+page. You can override this message and show domain-specific messaging:
+```ruby
+config.footer = "MyApp Revision v1.3"
+```
--- a/features/footer.feature
+++ b/features/footer.feature
@@ -1,0 +1,28 @@
+Feature: Site title
+
+  As a developer
+  In order to customize the site footer
+  I want to set it in the configuration
+
+  Background:
+    Given I am logged in
+
+  Scenario: No footer is set in the configuration (default)
+    When I am on the dashboard
+    And I should see the default footer
+
+  Scenario: Set the footer in the configuration
+    Given a configuration of:
+    """
+      ActiveAdmin.application.footer = "MyApp Revision 123"
+    """
+    When I am on the dashboard
+    And I should see the footer "MyApp Revision 123"
+
+  Scenario: Set the footer to a proc
+    Given a configuration of:
+    """
+      ActiveAdmin.application.footer = proc { "Enjoy MyApp Revision 123, #{controller.current_admin_user.try(:email)}!" }
+    """
+    When I am on the dashboard
+    And I should see the footer "Enjoy MyApp Revision 123, admin@example.com!"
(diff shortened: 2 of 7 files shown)


Fivell:
what if this already works?
```ruby
ActiveAdmin.application.footer = proc {  render_to_string partial: 'footer' }
```
but I haven't check

Fivell:
@blimmer, ping

varyonic:
BTW I couldn't get anything similar to `= proc { ... }` to work.  It needs something like `ActiveAdmin.application.view_factory.footer = ...` a footer component.

blimmer:
I haven't had time to look into a partial and am not sure my timeline on availability to look. If this solution isn't acceptable it might make sense to close this PR and open an issue to track that implementation.

varyonic:
I mentioned preferring a partial, but I do think this is acceptable.

blimmer:
Agreed - it seemed like the investigation / implementation of the partial was blocking from the requested changes indication on the PR

timoschilling:
I close this PR, it should be implemented as parial or proc

varyonic:
@timoschilling I'm disappointed.  Next time I won't suggest a preference when I approve.

blimmer:
👍 , this is inline with how one customizes the header and other elements and provides a nice, simple implementation that's better than what we have now (which is no ability to customize the footer at all).

docs/1-general-configuration.md
```diff
@@ -169,3 +169,11 @@ ActiveAdmin.setup do |config|
   end
 end
 ```
+
+## Footer Customization
+
+By default, Active Admin displayed a "Powered by ActiveAdmin" message on every
```
varyonic: ...ActiveAdmin displays...

lib/generators/active_admin/install/templates/active_admin.rb.erb
```diff
@@ -269,4 +269,11 @@ ActiveAdmin.setup do |config|
   # of those filters by default here.
   #
   # config.include_default_association_filters = true
+
+  # == Footer
+  #
+  # By default, the footer shows the current Active Admin version. You can
+  # override the content of the fotter here.
```
varyonic: s/fotter/footer/

varyonic:
Fixed a couple of typos, squashed and force-rebased.

Fivell:
closes #4758

timoschilling:
Maybe some can make a PR to add the proc support in the docs and the initializer template

features/footer.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Site title
```
varyonic: s/title/footer/

varyonic:
Agreed, though much of this was copied from Greg's original site_title where this is not documented either.

gkop:
This is a nice improvement, thanks Ben and friends!

varyonic:
Ironically this was one of the topics of an old [post by Philippe](http://reverbhq.com/blog/2012/08/building-web-apps-with-activeadmin/).

Issue #2818, opened 2013-12-23
Dynamic scopes and filters


seanlinsley:
It's a common misconception (#2340, #2781) that scopes will behave dynamically, but in reality they're loaded once, when the server boots.

I'm open to DSL suggestions.

itkin:
+1

heaven:
Any luck with this?

gingerlime:
also wondering about this... in our particular case, we have a "heavy" scope, which I only want to run when it's explicitly needed. Otherwise, the index page would run a `count` on the scope to show how many entries there are, but I'd rather avoid it and only apply it with an action item or something... Would be great to be able to apply the scope only with the scope param, but this doesn't seem to work if the scope isn't specified.

PR #4661, opened 2016-11-08
Enable caller to pass a scope count to `scope` via `show_count`

Empact:
This is useful in situations where there are multiple scopes and the count
can be generated for all of them in a single query.

### DIFF
--- a/features/index/index_scopes.feature
+++ b/features/index/index_scopes.feature
@@ -77,6 +77,19 @@ Feature: Index Scoping
     And I should see the scope "All" with no count
     And I should see 3 posts in the table
 
+  @scope
+  Scenario: Viewing resources with a scope and scope count provided for a single scope
+    Given 3 posts exist
+    And an index configuration of:
+      """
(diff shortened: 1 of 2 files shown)


Empact:
Incidentally I ran into this bug while trying to use this approach: https://stackoverflow.com/questions/26530016/get-around-the-activeadmindatabasehitduringload

Does activeadmin have an answer for this?

timoschilling:
`:show_count` is to enable/disable the display of the scope count, not to set a value.

What you want to do works only with this stackoverflow hacks, we don't support or build API for hacks.

Maybe this will someday supported by #2818.

Empact:
> :show_count is to enable/disable the display of the scope count, not to set a value.

Would you prefer an alternative which uses a different key, e.g. `:count` which serves both roles? IMO `:show_count` is overly narrow in its construction.

> What you want to do works only with this stackoverflow hack

Yeah, I referenced it in search of an official alternative, which is #2818. I like the `scopes` block solution there, incidentally.

Issue #3409, opened 2014-09-08
Bug in 'seems_searchable?' / ransack-related


mrbrdo:
When I create a custom ransacker in my model:
```
  ransacker :last_signup_step do
    Arel.sql("company_meta->>'lastSignupStep'")
  end
```

And then a filter in ActiveAdmin:
```
filter :last_signup_step, as: :numeric
```

It works, but it always does the equals comparison, even if I select something else from the dropdown. It seems because the `select_and_search` css class is not added to the field.
This comes from this line from ActiveAdmin source: `(opts[:class] ||= '') << ' select_and_search' unless seems_searchable?`

The method `seems_searchable?` returns true if there is a ransacker defined with that name, I think that is causing the issue. Maybe it was bugged after ransack was introduced? I think the metasearch methods were named like `field_equals` but in ransacker it's just `field`.

seanlinsley:
`seems_searchable?` checks `ransacker?` specifically so you can use a ransacker with `as: :select`
```ruby
      class SelectInput < ::Formtastic::Inputs::SelectInput
        include Base

        def input_name
          return method if seems_searchable?

          searchable_method_name.concat multiple? ? '_in' : '_eq'
        end
```

The fact that Ransack doesn't support the different predicates seems like a Ransack bug, not an Active Admin bug.

seanlinsley:
Or maybe Ransack does support it and we're just not allowing it? I'm not sure. But it seems like #3510 more succinctly explains the desired feature, so I'm going to merge this ticket into it.

llawlor:
My temporary jQuery solution until this is fixed:

// when the document is ready
$(document).ready(function() {
  // add class 'select_and_search' to all elements with class 'filter_numeric'
  $('.filter_numeric').addClass('select_and_search');
});

whatcould:
Just a note to clear up the confusion above: I think this is an ActiveAdmin bug, not a Ransack bug. Ransackers **do** support the predicates, but when filters are created for a ransacker (eg `filter :name_reversed`), AA wouldn't create the correct type of filter, since it would assume the query name should be `name_reversed` instead of the predicated name, eg`name_reversed_cont`.

My fix works great for the simple ransacker use cases I found, including the one I added to the sample Rails app in ActiveAdmin; I guess it's possible that some ransackers (ones that take parameters?) do not use predicates?

PR #4663, opened 2016-11-12
Update filters to work with custom Ransackers

### LINKED ISSUE #3510: default filter should list custom ransackers
If there is a custom runsacker defined, like this:

``` ruby
  ransacker :fullname do
    Arel.sql("CONCAT(`firstname`, ' ', `lastname`)")
  end
end
```

There should be a automatic generated filter for `fullname`, like the filters for each other field.

whatcould:
It seems that ransackers (eg `ransacker :name_reversed`) are _not_ expected to be called without predicates (that is, they **are** called with predicates, eg: `name_reversed_cont`). So a ransacker filter without a predicate should not `seems_searchable?`.

Seems to me that custom ransackers would never have worked as intended with this code — an artifact of converting from metasearch without testing it in a sample application.

[fixes #3510 / #3409]

### DIFF
--- a/spec/unit/filters/filter_form_builder_spec.rb
+++ b/spec/unit/filters/filter_form_builder_spec.rb
@@ -4,6 +4,9 @@ class Post
   ransacker :custom_searcher do
     # nothing to see here
   end
+  ransacker :custom_searcher_numeric, type: :numeric do
+    # nothing to see here
+  end
 end
 
 describe ActiveAdmin::Filters::ViewHelper do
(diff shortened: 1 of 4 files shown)


Fivell:
@whatcould , can you rebase upstream ?

whatcould:
@Fivell rebased!

Fivell:
@whatcould , please rebase again and it will be merged.

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -424,7 +424,7 @@ def view.a_helper_method
     it "should work as string" do
       body = Capybara.string(filter :custom_title_searcher, as: :string)
       expect(body).to have_selector("input[name='q[custom_title_searcher_contains]']")
-      expect(body).to have_selector("input[name='q[custom_searcher_starts_with]']")
+      expect(body).to have_selector("input[name='q[custom_title_searcher_starts_with]']")
```
Fivell: bad merge... that should be selector for ```option``` , not ```input``` :disappointed:

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -423,8 +423,8 @@ def view.a_helper_method
 
     it "should work as string" do
       body = Capybara.string(filter :custom_title_searcher, as: :string)
-      expect(body).to have_selector("input[name='q[custom_title_searcher_contains]']")
-      expect(body).to have_selector("input[name='q[custom_title_searcher_starts_with]']")
+      expect(body).to have_selector("option[name='q[custom_title_searcher_contains]']")
+      expect(body).to have_selector("option[name='q[custom_title_searcher_starts_with]']")
```
Fivell: finally this should work

```
  it "should work as string" do
      body = Capybara.string(filter :custom_title_searcher, as: :string)
      expect(body).to have_selector("option[value=custom_title_searcher_contains]")
      expect(body).to have_selector("option[value=custom_title_searcher_starts_with]")
    end
```

PR #4671, opened 2016-11-29
Lock jquery-ui-rails dependency to '~> 5.0'

wvteijlingen:
The new jquery-ui-rails 6.0 version breaks Active Admin, so I've locked it to the ~> 5.0 version for now.


timoschilling:
fixed by #4674

PR #4675, opened 2016-11-29
Revert "Lock jquery-ui-rails dependency to '~> 5.0'"

timoschilling:
Reverts activeadmin/activeadmin#4671


rhryniow:
Any reason this was reverted ? Using 6.x version breaks application

PR #4676, opened 2016-11-30
Remove Ruby 1.9 support

rhymes:
Ruby 1.9 support has ended almost two years ago, February 2015 https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

I don't think AA should keep the support (also the build keeps failing because of 1.9)


timoschilling:
You are right but the 1.x will support 1.9. With 2.0 we will drop 1.9.

Issue #4548, opened 2016-07-24
has_many creation fails on current master. Works on pre4


djpate:
I tried using the current master to my project.

I have a has_many form such as.

```
f.inputs do
    f.has_many :product_sub_categories do |psc|
      psc.input :sub_category, :include_blank => false
      psc.input :_destroy, as: :boolean
    end
end
```

This form works fine on 0.6 and 1.0.0.pre4 but  on current master it fails to create because of duplicate key constraint on my model.

I put a breakpoint in data_acces.rb in the save_model method and my object for some reason was created with two identical sub categories even though the params sent to the controler were fine (only one).

```
Started POST "/admin/products" for 127.0.0.1 at 2016-07-23 19:12:44 -0700
  ActiveRecord::SchemaMigration Load (2.6ms)  SELECT `schema_migrations`.* FROM `schema_migrations`
Processing by Admin::ProductsController#create as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"7C/cAwEUuR0uC2nKP9DD5M0I/tcdntNF7lxnw08OgmTyRSyZI+Xe95EVdf8pTaKh77X2oTAguA6QWCOpO5HlOw==", "product"=>{"title"=>"ddoo", "description"=>"fooo", "image"=>#<ActionDispatch::Http::UploadedFile:0x00000007cf19b0 @tempfile=#<File:/tmp/RackMultipart20160723-10866-7qvkqr.jpg>, @original_filename="evan dessin-8.jpg", @content_type="image/jpeg", @headers="Content-Disposition: form-data; name=\"product[image]\"; filename=\"evan dessin-8.jpg\"\r\nContent-Type: image/jpeg\r\n">, "sku"=>"80", "price"=>"50", "reduced_price"=>"", "display_order"=>"100", "active"=>"1", "gender"=>"0", "color_ids"=>["", "2", "3"], "size_ids"=>["", "8", "9"], "product_sub_categories_attributes"=>{"0"=>{"sub_category_id"=>"1", "_destroy"=>"0"}}}, "commit"=>"Create Product"}
```

```
[3] pry(#<Admin::ProductsController>)> object.product_sub_categories
=> [#<ProductSubCategory:0x00000007d03480
  id: nil,
  product_id: nil,
  sub_category_id: 1>,
 #<ProductSubCategory:0x00000007bb6b18
  id: nil,
  product_id: nil,
  sub_category_id: 1>]
```

```
[...]

PR #4680, opened 2016-12-02
Re-enable STI build_association with type on parent resource

### LINKED ISSUE #4424: new_with_cast no longer works (regression)

alexgerstein:
Fixes #4424, but without the regression previously seen in #4548.


alexgerstein:
@timoschilling Can you please take a look at this when you have a chance? It would be useful to get STI working again without having to patch.

deivid-rodriguez:
Since the previous patch fixing this was reverted because it caused a regression, I'd welcome two tests here. One that ensures that the regression doesn't come back, and another one to ensure that this fix is not reverted in the future

Fivell:
rebase here #4883

PR #4682, opened 2016-12-02
DOM ready events not working for non-Turbolinks apps

mateusg:
Since `$(document).on('ready', handler)` has been deprecated as of jQuery 1.8 and removed in jQuery 3.0, these events have stopped working for apps without Turbolinks:

* selected a filter
screenshot 2016-12-02 18 35 14

* it didn't change the parameter
screenshot 2016-12-02 18 35 45

First I thought about using `on('load DOMContentLoaded')`, but the jQuery API says it's safer to use the `.ready()` method because it assures the handler is executed even if the event binding happens after the DOM is ready: https://api.jquery.com/ready.

I know this is not a very elegant solution, but the only other solution that I came up with was extending jQuery's `.ready()` method to bind Turbolinks' events, and I don't know if it's safe (I haven't tried).

Any thoughts?

My app is using:
- Ruby 2.3.1
- Rails 4.2.6
- The latest version of ActiveAdmin (master branch, commit a932d56b156eaf64beb86cb45f8f801b7933c42e)
- jquery-rails 4.0.5, but I have tried others and the problem persisted

### DIFF
--- a/app/assets/javascripts/active_admin/initializers/batch_actions.js.coffee
+++ b/app/assets/javascripts/active_admin/initializers/batch_actions.js.coffee
@@ -1,7 +1,11 @@
-$(document).on 'ready page:load turbolinks:load', ->
+onDOMReady = ->
   # In order for index scopes to overflow properly onto the next line, we have
   # to manually set its width based on the width of the batch action button.
   if (batch_actions_selector = $('.table_tools .batch_actions_selector')).length
     batch_actions_selector.next().css
       width: "calc(100% - 10px - #{batch_actions_selector.outerWidth()}px)"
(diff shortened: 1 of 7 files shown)


timoschilling:
Or maybe some users can test and approve this?

swrobel:
Just tested and this fixes the issue for me w/ jQuery 2 & 3

swrobel:
FYI, this works in active_admin.js as a horrible workaround for the time being:

```javascript
$(function() {
  $(document).trigger('ready');
});
```

javierjulio:
@mateusg thanks! 🙏🏻❤️

PR #4688, opened 2016-12-06
Wrap the action link text with spans that can be styled using C…

vraravam:
…SS to replace ther content with images/icons for better horizontal screenspace.

### DIFF
--- a/lib/active_admin/views/index_as_table.rb
+++ b/lib/active_admin/views/index_as_table.rb
@@ -348,13 +348,13 @@ module ActiveAdmin
 
         def defaults(resource, options = {})
           if controller.action_methods.include?('show') && authorized?(ActiveAdmin::Auth::READ, resource)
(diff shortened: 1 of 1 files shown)


vraravam:
@timoschilling - could you please review and merge this before 1.0.0? It would be an immense favor to get this done. I'm trying to get the following kind of UI:
icons-as-action-links

timoschilling:
1. We will not accept HTML Strings in our code, rewrite it with arbre
2. Why do you need the span? You can replace the link text/content via css.

vraravam:
I just realized that this is possible with the `view_link` css-style itself. The only thing missing would then be the on hover tooltip (since in my case, I am replacing the text with the image). Would you accept this PR if I removed the `span` and added the `title` attribute to the `a` tag?

timoschilling:
yes

vraravam:
Done - please review and merge?

timoschilling:
please squash your commits into one, then it can be merged

vraravam:
@timoschilling - done squashing :)

PR #4705, opened 2016-12-28
Fix #4533, use custom Ransack predicate for datetime, not date columns.

### LINKED ISSUE #4533: Range filters for "date" field broken after #4504
I believe #4504 breaks filtering for `date` fields. 
This is a resulting query for a range `2016-05-02` - `2016-05-02`:

```
SELECT  1 AS count_column FROM "dates" WHERE ("dates"."starts_on" >= '2016-05-01 23:00:00.000000' AND "dates"."starts_on" < '2016-05-03') LIMIT 30 OFFSET 0
```

It tries to use date time ranges for date fields resulting in timezone conversion. 
Then `2016-05-01 23:00:00.000000` gets downcasted to `2016-05-01` and the results are invalid.

My temporary workaround was to monkeypatch these two methods:

``` ruby
ActiveAdmin::Inputs::Filters::DateRangeInput.class_eval do
  def gt_input_name
    case column&.type
    when :date
      "#{method}_gteq"
    else
      "#{method}_gteq_date"
    end
  end

  def lt_input_name
    case column&.type
    when :date
      "#{method}_lteq"
    else
      "#{method}_lteq_date"
    end
  end
end
```

Let me know if you need more details

varyonic:
Fixes #4533 I was surprised there was no date column anywhere in the test schema so I altered published_at:datetime to published_date:date.  An alternative is to add a new column but that will require fixing various tests also.

### DIFF
--- a/features/development_reloading.feature
+++ b/features/development_reloading.feature
@@ -14,7 +14,7 @@ Feature: Development Reloading
       ActiveAdmin.register Post do
         if Rails::VERSION::MAJOR >= 4
           permit_params :custom_category_id, :author_id, :title,
-            :body, :position, :published_at, :starred
+            :body, :position, :published_date, :starred
         end
       end
     """
--- a/features/edit_page.feature
+++ b/features/edit_page.feature
@@ -12,7 +12,7 @@ Feature: Edit Page
       ActiveAdmin.register Post do
         if Rails::VERSION::MAJOR >= 4
           permit_params :custom_category_id, :author_id, :title,
-            :body, :position, :published_at, :starred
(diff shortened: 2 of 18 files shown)


varyonic:
@timoschilling I'm a little confused by this new review process.  You approved and master is merged into my branch instead of my branch merged to master?

timoschilling:
I have approved your Code, but forgot to merge it into master.
I have merged master into your branch while your branch was out of date with master.
fix__4533__use_custom_ransack_predicate_for_datetime__not_date_columns__by_varyonic_ _pull_request__4705_ _activeadmin_activeadmin

PR #4706, opened 2016-12-31
Make Activeadmin possible inside Rails Engine

victorlcampos:
Hi,
I'm trying to use Activeadmin inside my Rails Engine plugin, but somethings are not working well =(

When I tried to run the install guide I get this error:

lib/generators/active_admin/devise/devise_generator.rb:55:in `add_default_user_to_seed': undefined method `paths' for nil:NilClass (NoMethodError)


timoschilling:
That doesn't solve the root problem

victorlcampos:
Could you show me the right way?

Issue #3604, opened 2014-11-10
Remove inherited_resources


timoschilling:
`inherited_resources` will support Rails 4.2 for us, see https://github.com/activeadmin/activeadmin/pull/3193#issuecomment-62421649 for details.

To support Rails 5.0 in the future we need to remove `inherited_resources` and need to implement some features of `inherited_resources` in our base controller.

NOTE: This ticket is only for the replacement of `inherited_resources`, please don't post `inherited_resources` bugs and not no other off topics here!
And please no `+1` posts, you don't need to need to do that, it is a done thing, there is no way around on the long term.

Fire-Dragon-DoL:
Are there any chance you guys can take over the development (and the upgrade) of inherited_resources (and keep it as a separate gem, which provides a double bonus). I don't use inherited_resources myself but I definitely agree that responders and respond_with is not a full replacement for what inherited_resources does, and considering how important is this for activeadmin, it can be option (also code will adapt to activeadmin requirements, which is quite good).

bamorim:
@timoschilling So, today I was dealing with some problems with the belongs_to :optional and breadcumbs and so and It got me thinking about it.

Well, I have an idea that I like to discuss here. Its somewhat complicated but it's also robust, I think

My solution to this problem (as well as for routing, nested, breadcumb generation and so on) is to create an Directed Graph.

Well, the way it should work is to have a Resource Graph, that keeps all resources and related stuff (actions, namespaces, and so on)

For default, the Resource Graph is initiated with the Root namespace.
Some actions on the DSL acts directly on the Resource Graph.

For example: belongs_to add a relation from the parent to the actual resource and removes the root connection unless optional: true is passed.

After we have our resource graph defined, we can then generate the routes and the controllers.

The routes are really straight forward: We just run a Graph Search, checking for and removing/warning about cycles on the Graph "instance_exec-ing" the routes. For each time we find something (A Resource, a Page or a Singleton Resource, we run some different router)

For generating the controllers, we can a similar approach, but modifying the controller evertime a route is added. After that the controller will have a tree of forms of getting to the resource.

When the request comes, the controller finds the path from the root down to the main resource.
After that he will run the request chain and the last item of the chain will be the required resource.

For breadcrumb generator, it just follows the resource chain.

Well, this is just some toughts (and a little bit crazy) and it's not refined. It'd also be complicated to implement. However this should gives us a better control over our problem and maybe will allow us to move faster.

That's it.

[...]

PR #4713, opened 2017-01-05
Recommend `inherited_resources` "official" repo

deivid-rodriguez:
It seems like it has an extra fix that's not in the fork...

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ gem 'activeadmin', '~> 1.0.0.pre4'
 Active Admin master has preliminary support for Rails 5. To give it a try, these Gemfile changes may be needed:
 
 ```ruby
-gem 'inherited_resources', github: 'activeadmin/inherited_resources'
+gem 'inherited_resources', github: 'josevalim/inherited_resources'
 ```
 
 ### 0.6.x


timoschilling:
Thanks for the hint on the new commits, but I will keep the reference at our own fork, so we can apply new fixes until a new version of inherited_resources is released. I have pulled from the josevalim/inherited_resources repo and pushed to activeadmin/inherited_resources

deivid-rodriguez:
You're welcome. I disagree with tracking a fork by default, specially when it's less active than the official repo.

However, I think the best solution would be to transfer the `inherited_resources` repository to the `activeadmin` organization since it seems to be the most popular project still stuck with it. I think the current owners would agree with this since they consider it a legacy project, and you would get privileges and be able to release the gem yourselves. What do you think? Should we ping Rafael França?

varyonic:
FYI see also #3604 and #3193 and [IR#415](https://github.com/josevalim/inherited_resources/issues/415) I think ActiveAdmin organization maintaining a fork is a good compromise.

deivid-rodriguez:
Another option is to just vendor whatever code from inherited_resources activeadmin is depending on.

deivid-rodriguez:
The problem with the fork is that you can't release to rubygems unless you rename the gem. And you most likely want a release before 1.0.0.

varyonic:
Agreed.  I propose emailing the current owners of inherited resources (Jose, Rafael and Georges) and requesting (a) that one or more of us be [added as an owner](http://stackoverflow.com/questions/8487218/how-to-add-more-owners-to-a-gem-in-rubygem) of the inherited_resources gem on rubygems, and (b) that the original inherited_resources repository be transferred to activeadmin.  If we cannot have (b) then we continue to maintain our own fork, and if we cannot have (a) then we publish our fork with the name inherited_resoures_lts (long term support)
or some such.

deivid-rodriguez:
Sounds good to me. Except maybe for the case where you're forced to release a differently named gem. In that case, I'd consider vendoring. But let's worry about that only when/if that happens.

varyonic:
@timoschilling Are you willing to accept the original inherited_resources repo into activeadmin organization if @josevalim is willing to transfer it?

deivid-rodriguez:
So the repo was actually transferred, right? :tada: Did you also get rubygems' permissions, @timoschilling?

varyonic:
@deivid-rodriguez Yes. You can see the new owner on [rubygems](https://rubygems.org/gems/inherited_resources)

deivid-rodriguez:
Cooool.

Issue #4173, opened 2015-10-27
Pagination looks broken


chrisjeon:
Pagination looks broken. Is this type of pagination styling intentional?

<img width="1280" alt="screen shot 2015-10-26 at 9 43 03 pm" src="https://cloud.githubusercontent.com/assets/2433941/10748102/ae0868f8-7c2a-11e5-945b-bada48cbc78b.png">

timoschilling:
Pleas post the `<head>` tag and his content of that page

seanlinsley:
Closing due to a lack of details.

mishatatinets:
Same issue

chrisjeon:
Got a screenshot @mishatatinets ?

mishatatinets:
![screen shot 2016-01-02 at 3 59 16 pm](https://cloud.githubusercontent.com/assets/931951/12076493/e730ead2-b169-11e5-9a9b-575c4237c0b3.png)

mishatatinets:
Here is my head tag.
![screen shot 2016-01-02 at 4 04 43 pm](https://cloud.githubusercontent.com/assets/931951/12076510/a025b068-b16a-11e5-91cb-3c0f07a3ad4e.png)

seanlinsley:
What browser is that? Are you able to reproduce in other browsers?

Can you upload a test Rails app that we can run, that reproduces this issue?

chrisjeon:
Yeah, that's how mine looked like too. @seanlinsley I was using Chrome.

mishatatinets:
Both Safari and Chrome displays same things.
In pagination div I have a class  "pagination pagination-centered" 


![screen shot 2016-01-02 at 4 09 53 pm](https://cloud.githubusercontent.com/assets/931951/12076532/dc74a7e4-b16b-11e5-83b1-bb37620c78aa.png)

mishatatinets:
I also have kaminari folder in my view that renders this 

![screen shot 2016-01-02 at 4 33 17 pm](https://cloud.githubusercontent.com/assets/931951/12076584/e29a87ee-b16e-11e5-8909-5da0f27671d7.png)
![screen shot 2016-01-02 at 4 33 27 pm](https://cloud.githubusercontent.com/assets/931951/12076583/e299569e-b16e-11e5-823c-6887c923ca48.png)
 

I have to use it for other content  pagination.
Can this cost an issue?

seanlinsley:
[...]

PR #4722, opened 2017-01-11
Relax kaminari dependency for kaminari 1.0.0

sue445:
kaminari v1.0.0 is released!

* https://rubygems.org/gems/kaminari/versions/1.0.0
* https://github.com/kaminari/kaminari/blob/master/CHANGELOG.md#100

It seems no problem at my environment. (Ruby 2.4.0, Rails 5.0.1)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ if rails_major == '5'
 end
 
 platform :ruby_19 do # Remove this block when we drop support for Ruby 1.9
+  gem 'kaminari', '~> 0.15'
   gem 'mime-types', '< 3'
   gem 'nokogiri', '< 1.7'
   gem 'public_suffix', '< 1.5'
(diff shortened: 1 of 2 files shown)


timoschilling:
The tests are not working with ruby 1.9, while Kaminari 1.0 no longer supports RUby <2.0

We can merge this if https://github.com/kaminari/kaminari/pull/847 is merged and released

sue445:
🙇

timoschilling:
one other option is that we can set different kaminari versions in our `Gemfile` for Ruby 1.9 and >2

sue445:
hmm...

I fixed that use kaminari 0.x when travis uses ruby 1.9.
5dca341

varyonic:
Does this resolve #4173 ?

sue445:
No. This problem is not solved.

image

PR #4725, opened 2017-01-12
Better generated migrations

### LINKED ISSUE #4369: Rails 5: Migration is missing "[5.0]" when auto created

deivid-rodriguez:
Closes #4390.
Closes #4369.

### DIFF
--- a/lib/generators/active_admin/install/templates/migrations/create_active_admin_comments.rb
+++ b/lib/generators/active_admin/install/templates/migrations/create_active_admin_comments.rb
@@ -1,4 +1,7 @@
-class CreateActiveAdminComments < ActiveRecord::Migration
+parent_class = ActiveRecord::Migration
+parent_class = parent_class[5.0] if Rails::VERSION::MAJOR >= 5
+
+class CreateActiveAdminComments < parent_class
   def self.up
     create_table :active_admin_comments do |t|
       t.string :namespace
--- a/lib/generators/active_admin/install/templates/migrations/create_active_admin_comments.rb
+++ b/lib/generators/active_admin/install/templates/migrations/create_active_admin_comments.rb
@@ -9,7 +9,11 @@ class CreateActiveAdminComments < parent_class
       t.string :resource_id,   null: false
       t.string :resource_type, null: false
       t.references :author, polymorphic: true
-      t.timestamps
+      if Rails::VERSION::MAJOR >= 5
+        t.timestamps
+      else
+        t.timestamps null: false
+      end
     end
     add_index :active_admin_comments, [:namespace]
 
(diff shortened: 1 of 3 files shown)


lib/generators/active_admin/install/templates/migrations/create_active_admin_comments.rb
```diff
@@ -9,7 +9,11 @@ def self.up
       t.string :resource_id,   null: false
       t.string :resource_type, null: false
       t.references :author, polymorphic: true
-      t.timestamps
+      if Rails::VERSION::MAJOR >= 5
+        t.timestamps
+      else
+        t.timestamps null: false
```
Fivell: @deivid-rodriguez , maybe it will be reasonable to declare just 
 ``` t.timestamps null: false```  without if/else statement ?

deivid-rodriguez: Absolutely. The only advantage I see with the current version is that the version check can act as a reminder to leave the simpler `t.timestamps` once Rails 4.2 support is dropped.

deivid-rodriguez:
It's a good point @timoschilling, your solution is much cleaner.

timoschilling:
Maybe we can remove the template and use the rails migration generator at the create/install time

deivid-rodriguez:
I'm on it, almost ready!

deivid-rodriguez:
@timoschilling Let's wait for CI, but how about https://github.com/activeadmin/activeadmin/pull/4725/commits/8d6250751b13ae1c1d6323f5b1f7943488d000ff?

deivid-rodriguez:
Sweet, green build! Ready?

deivid-rodriguez:
@timoschilling Rebased this PR. Since you asked for the changes, could you review the updated version, @timoschilling?

timoschilling:
DISCUSSION: What's about removing the migration template and switching to something like
```ruby
def create_migrations
  generate "migration active_admin/comments namespace:text:index ..."
end
```

timoschilling:
DISCUSSION: What's about removing the migration template and switching to something like
```ruby
def create_migrations
  generate "migration active_admin/comments namespace:text:index ..."
end
```
@Fivell @varyonic @deivid-rodriguez

deivid-rodriguez:
@timoschilling I think it's a good idea, but can be implemented in a separate PR. In particular, we need a completely standard migration to be able to do that and right now it is not. We would need at least https://github.com/activeadmin/activeadmin/pull/4734 to be merged first.

PR #4729, opened 2017-01-14
Add doc about Formtastic collection_label_methods

### LINKED ISSUE #4143: Filters don't respect configured `display_name_methods`
When objects are referenced in filters, e.g. auto-generated association filters with dropdowns, their display names are apparently handled by Formtastic rather than ActiveAdmin itself, bypassing ActiveAdmin's `display_name_methods`.

The default `display_name_methods`: `[ :display_name, :full_name, :name, :username, :login, :title, :email, :to_s ]`

Formtastic's default `collection_label_methods`: `["to_label", "display_name", "full_name", "name", "title", "username", "login", "value", "to_s"]`

So if an object responds to `:to_label`, that will be displayed in filters rather than the output of any of ActiveAdmin's specified methods. Or if you've configured ActiveAdmin to prioritize a custom method, say `:admin_name`, that will never be used in a filter.

You can change the methods Formtastic uses in its configuration, but if you use Formtastic elsewhere in your app and want this behavior to be specific to ActiveAdmin, that's not an ideal solution.

In any case, this is confusing and worth noting somewhere.

ypresto:
Fixes #4143

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -117,6 +117,8 @@ module ActiveAdmin
 
     # Active Admin makes educated guesses when displaying objects, this is
     # the list of methods it tries calling in order
+    # Note that Formtastic also has 'collection_label_methods' similar to this
+    # used by auto generated dropdowns in filter or belongs_to field of Active Admin
     setting :display_name_methods, [ :display_name,
                                       :full_name,
                                       :name,


lib/active_admin/application.rb
```diff
@@ -117,6 +117,8 @@ def initialize
 
     # Active Admin makes educated guesses when displaying objects, this is
     # the list of methods it tries calling in order
+    # Note that Formtastic has also have `collection_label_methods` similar to this
```
varyonic: Note that Formtastic also has 'collection_label_methods' similar to this
used by auto generated dropdowns in filter or belongs_to field of Active Admin

ypresto:
Thanks! Fixed and force-pushed..!

Issue #691, opened 2011-10-27
Is it possible to mount active_admin in subdomain?


samitheberber:
I want to know if it is possible to have active_admin responding from http://admin.example.com instead of http://example.com/admin

If it is can you tell me how :)

I use rails 3.0 in this project.

Thanks in any case :)

PR #4731, opened 2017-01-17
make namespace route options configurable

### LINKED ISSUE #4467: Subdomain - Decouple path from namespace
Thanks guys for this great gem. This is purely a non-critical request/support.

We currently use activeadmin on: `example.com/admin`

We want to move it to: `admin.example.com/`

[This](https://github.com/activeadmin/activeadmin/issues/3911) and [this](https://github.com/activeadmin/activeadmin/issues/691) describe how to do that.

The problem is we already use the default `admin_` namespace for most things and when attempting to set the namespace to `false` the admin area clashes with the main app since `admin_organization_path` now looks just like our other `organization_path`.

Is it possible decouple the `namespace` from the `path`? i.e.

```
#routes.rb
ActiveAdmin.routes(self, path: '')
```

stereoscott:
This PR allows you to set options in your namespace config that are passed to the router. For example:

```ruby
    config.route_options = { path: '' }
  end
```

And in your `routes.rb`:
```ruby
constraints :subdomain => "admin" do
  ActiveAdmin.routes(self)
end
```

This would generate routes like:
```
admin_root  GET        /       admin/dashboard#index, { :subdomain => "admin" }
admin_users GET        /users  admin/users/#index, { :subdomain => "admin" }
```

This is nice because you can keep your admin configuration and controllers all namespaced, but you have greater controller over the route generation. This is very useful when you'd like to use a subdomain without the namespace in the path.

Note: We call `.dup` on the options when calling `namespace` because inside of the rails router `options.delete(:path)` is called and subsequent calls to `namespace.route_options` would return the modified hash. 

Closes #4467.

Related: 
 - https://github.com/activeadmin/activeadmin/issues/691
 - http://stackoverflow.com/questions/38954533/how-to-use-subdomain-with-activeadmin-on-rails

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 index download_links: ->{ can?(:view_all_download_links) || [:pdf] }
 ```
 * Comments menu can be customized via configuration passed to `config.comments_menu` [#4187][] by [@drn][]
+* Added `config.route_options` to namespace to customize routes [#4467][] by [@stereoscott[]]
 
 ### Security Fixes
 
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -80,6 +80,9 @@ module ActiveAdmin
     # Options that a passed to root_to.
     inheritable_setting :root_to_options, {}
 
+    # Options passed to the routes, i.e. { path: '/custom' }
+    inheritable_setting :route_options, {}
+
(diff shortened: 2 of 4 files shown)


Fivell:
@stereoscott , can you review failing specs ?

stereoscott:
@Fivell all good now. thanks!

Issue #34, opened 2011-05-13
Comments should work with models with String Ids


MikeAbner:
_Edit: Original title: "Form for tables with string id's throw an exception"_

PGError: ERROR:  invalid input syntax for integer: "xksS5NjE"
LINE 1: ...admin' AND ("active_admin_comments".resource_id = 'xksS5NjE'..."

Looks like the comments tables expect integer ids on the parent table but this isn't universally true.

PR #4734, opened 2017-01-17
FIX comments migration. Now using resource as references

krim:
In base case migration used a separated resource_id and resource_type. And resource_id was a STRING type column. It's terrible.
Now we will use normal references for resource such as author.


Fivell:
@varyonic, did you see why it was introduced? Take a look pls https://github.com/activeadmin/activeadmin/issues/34

Fivell:
even there is line in changelog about this https://github.com/activeadmin/activeadmin/blob/0-6-stable/CHANGELOG.md#bug-fixes-3

varyonic:
We should have had a test failure?

Fivell:
@varyonic , looks like yes but they doesn't work as expected

https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/features/comments/commenting.feature#L148

https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/spec/unit/comments_spec.rb#L86

https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/lib/active_admin/orm/active_record/comments/comment.rb#L24

However this changes was made long time ago, so I'm not even sure that issue is still actual.

deivid-rodriguez:
I think @krim is right, it's not good to have an `_id` column with a string type. I think the fact that those tests still pass mean this is no longer an issue in the versions of Rails currently supported. I would merge this.

Fivell:
@deivid-rodriguez , the fact that those tests still pass also can mean they never worked as expected.

Fivell:
@varyonic , @timoschilling , @deivid-rodriguez 
As I understood, root of that line was that active admin was going to support comments with foreign associations to objects with stringed primary keys, however there were not enough tests for this was added. 
If we're going to merge this, than we also should remove some pieces of code that are not needed anymore, for instance this method https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/lib/active_admin/orm/active_record/comments/comment.rb#L23
and specs that are not needed anymore

deivid-rodriguez:
Yep yep, basically revert https://github.com/activeadmin/activeadmin/pull/1063. The feature spec you link to could be wrong of course. It looks fine to me on at first glance though.

Fivell:
@timoschilling , @varyonic, @seanlinsley  so support of this case is good idea what do you think https://github.com/activeadmin/activeadmin/issues/34 ?

deivid-rodriguez:
@krim ping

varyonic:
I think this is neither a bug-fix nor a feature.

deivid-rodriguez:
You mean it won't make it to 1.0.0? Sounds good to me, just pinging the OP in case (s)he wants to get this done.

krim:
I was make separated user access for manage active_admin comments. And when i was use cancancan and check objects ids for comparison with ids for access i really surprised when comparison work wrong for me. So, when i debugged this issue i found, that integer ids compared with string ids. Then i made this patch.

deivid-rodriguez:
Hi @krim, thanks for getting back. I agree with this change. It would be nice to know when and where (AR, specific pg adapter?) the original problem was fixed, but it seems like too much archeology. The OP's logs in #34 show 1.9.2 and Rails 3.0, not sure anyone would like to go back there just to clarify this.

So I'd suggest that you rebase this PR and add a full revert of #1063, and you'll have my approval :wink:.

krim:
I'm all done, but travis doesn't start tests with commit 967e397 and still use 28f867b for test. @deivid-rodriguez  please restart test

deivid-rodriguez:
Isn't the check in the last commit the same as not checking anything? I mean, Rails 6 is not there yet, not even planned I think...

krim:
I have typo in commit message, but not in commit :)

deivid-rodriguez:
I don't understand, isn't `Rails::VERSION::MAJOR <= 5` true for every Rails version?

krim:
My fault, fixed.
https://travis-ci.org/activeadmin/activeadmin/jobs/211880876 need to restart. Travis memory problem

deivid-rodriguez:
@krim Yes, there was a problem with Travis, my fault. It should be fixed now. Could you try rebasing again? Now it should pass.

Fivell:
@krim , @deivid-rodriguez , I would like to see this merged , but something wrong is here with rebasing

deivid-rodriguez:
Well, the final diff seems good so squashing down to a single commit should work.

krim:
Ready to merge?

deivid-rodriguez:
Could you fix the commit history so the PR includes a single commit with the proposed change? If you can't fix this branch, feel free to create a new PR from a fresh branch off master and apply the change there.

krim:
New PR here: #4868

PR #2823, opened 2013-12-24
Add action_item localizer

roshats:
Now one can set custom button name and page title for new, edit, and destroy actions/pages rather than "New %{model}" and "Edit %{model}".

PR #4739, opened 2017-01-19
Document renaming action items

varyonic:
As requested in #2823

### DIFF
--- a/docs/2-resource-customization.md
+++ b/docs/2-resource-customization.md
@@ -94,6 +94,13 @@ ActiveAdmin.register Post do
 end
 ```
 
+## Renaming Action Items
+
+One can set custom button name and page title for new, edit, and destroy actions/pages
+at the resource level by providing a resource specific translation.
+For example to change 'New Offer' to 'Make an Offer'
+set `active_admin.resources.offer.new_model`.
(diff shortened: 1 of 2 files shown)


Fivell:
@varyonic , can you please drop a line about scopes too.
Here is the spec https://github.com/activeadmin/activeadmin/pull/2823/files#diff-9906165631b78175492690873f3d746fR142

deivid-rodriguez:
@varyonic This is great! :100:

By the way, do you have write access to the repo? If you do, you might be able to add new commits to #2823 so everything can be merged together.

docs/2-resource-customization.md
```diff
@@ -94,6 +94,13 @@ ActiveAdmin.register Post do
 end
 ```
 
+## Renaming Action Items
+
+One can set custom button name and page title for new, edit, and destroy actions/pages
+at the resource level by providing a resource specific translation.
+For example to change 'New Offer' to 'Make an Offer' 
+set `active_admin.resources.offer.new_model`
```
roshats: I think the period is missing at the end. 😄

varyonic:
@Fivell Added.

PR #4744, opened 2017-01-21
Use travis gem cache

deivid-rodriguez:
This should fix current problems with CI, first triggered in #2823.

Since the error was happening when clearing the cache, I initially thought that reusing the first commits in #4624 (which remove our custom gems cache replacing it with Travis built-in mechanism) might fix it. That didn't work so I ended up freezing the bundler to a working one, but I still think it's a good time to get the cache removal stuff reviewed and merged.

Let me know!


timoschilling:
If the build is ready, please push a empty commit to trigger a new build on Travis to verify that the cache works

deivid-rodriguez:
Done!

deivid-rodriguez:
Seems like it's working: https://travis-ci.org/activeadmin/activeadmin/jobs/194055501#L668-L794 (all "using", no "installing").

deivid-rodriguez:
There's already a [proposed fix](https://github.com/bundler/bundler/pull/5343) in `bundler` fixing the issue with the build, so it's probably better to just wait for the next bundler release.

varyonic:
So should we use the very latest bundler in Travis, or should we lock down the version?  Having our build broken by an upstream dependency like this is disruptive.

deivid-rodriguez:
You're right, let's lock it and upgrade it at our own pace. Currently locking it is a bit cumbersome since jruby seems to automatically include the latest bundler as a global gem. I'd like to improve https://github.com/activeadmin/activeadmin/pull/4744/commits/2daae98054ee5ad34f408c97eaafea78e7a016e5 to at least not hardcode the jruby version.

varyonic:
@deivid-rodriguez bundler 1.14.2 with fix has been released.

deivid-rodriguez:
Yes, it seems to be working. I improved a bit the way we lock the bundler version. @timoschilling Are you happy with this? I still want to remove the empty commit but want to ask if there's anything else.

deivid-rodriguez:
@timoschilling This is rebased and ready. Please review.

deivid-rodriguez:
Just pushed a simplified version of pinning the bundler version, copied from https://github.com/bbatsov/rubocop/commit/2e70d6fc45c29a71c7ae922c9599b69cc988dd9a.

PR #4747, opened 2017-01-24
Remove DataAccess#clean_search_params

Fivell:
ransack cleaning params by itslef, so no need to do it twice

### DIFF
--- a/lib/active_admin/resource_controller/data_access.rb
+++ b/lib/active_admin/resource_controller/data_access.rb
@@ -222,14 +222,8 @@ module ActiveAdmin
       # Applies any Ransack search methods to the currently scoped collection.
       # Both `search` and `ransack` are provided, but we use `ransack` to prevent conflicts.
       def apply_filtering(chain)
-        @search = chain.ransack clean_search_params
+        @search = chain.ransack(params[:q] || {})
         @search.result
-      end
-
-      def clean_search_params
-        q = params[:q] || {}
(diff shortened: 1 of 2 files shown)


Fivell:
refs https://github.com/activerecord-hackery/ransack/issues/158

Fivell:
@deivid-rodriguez, this is something that we haven't tested yet I think, let's add one of course

Fivell:
I will squash and rebase if ok

deivid-rodriguez:
@timoschilling If I understand this correctly, this should not mean any change over the current behaviour.

@Fivell Assuming that "add a failing test" means "add a passing test", I'm ok with this :stuck_out_tongue_closed_eyes:

Fivell:
@timoschilling , you're talking about active filters sidebar ? 
https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/filters/resource_extension.rb#L166

Fivell:
@deivid-rodriguez , sure, I mean  passing test 👍

Fivell:
@timoschilling   flter box is not related to this change, because clean_search_params was called only for apply_filters .

Active Filters Sidebar is handling this itself (actually is not handling this and can display empty search however it is still not related to this method)


But we have cleaner of search params on client side that make filter box not to display empty filters .

take a look https://github.com/activeadmin/activeadmin/blob/85e8f66b232a3b570e1bd6277f5741654af5aa6c/app/assets/javascripts/active_admin/initializers/filters.js.coffee#L13

PR #4748, opened 2017-01-24
fixed active filters for rails 4.2

### LINKED ISSUE #4342: No current filters when use filters in sidebar

Fivell:
[fixes #4342]

### DIFF
--- a/lib/active_admin/filters/active.rb
+++ b/lib/active_admin/filters/active.rb
@@ -8,12 +8,20 @@ module ActiveAdmin
 
       def initialize(resource_class, params)
         @resource_class = resource_class
-        @params = params.respond_to?(:to_unsafe_h) ? params.to_unsafe_h : params
+        @params = normalize_params(params)
         @scope = humanize_scope
         @filters = build_filters
       end
 
       private
+
+      def normalize_params(params)
+        if params.is_a?(HashWithIndifferentAccess)
+          params
+        else
+          params.to_unsafe_h
+        end
+      end
 
       def build_filters
         filters = @params[:q] || []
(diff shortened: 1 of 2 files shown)


lib/active_admin/filters/active.rb
```diff
@@ -8,13 +8,21 @@ class Active
...
-        @params = params.respond_to?(:to_unsafe_h) ? params.to_unsafe_h : params
+        @params = normalize_params(params)
         @scope = humanize_scope
         @filters = build_filters
       end
 
       private
 
+      def normalize_params(params)
+        if params.kind_of?(HashWithIndifferentAccess)
```
deivid-rodriguez: There's much more usages of `is_a?` than `kind_of?` in the codebase (35 vs 3 if I counted corrected), so maybe use `is_a?` to keep it more consistent.

deivid-rodriguez:
This makes sense to me, just added a very minor comment.

lib/active_admin/filters/active.rb
varyonic: `HashWithIndifferentAccess === params ? params : params.to_unsafe_h` should work also.

Fivell: @varyonic , it doesn't for rails 4.2

Loading development environment (Rails 4.2.7.1)
2.2.0 :001 > params = ActionController::Parameters.new(name: 'Test')
 => {"name"=>"Test"} 
2.2.0 :002 > params === HashWithIndifferentAccess
 => false

Fivell: in rails 4.2 ActionController::Parameters inherited from HashWithIndifferentAccess, it even have to_unsafe_h method, but it works not that way like in rails5

PR #4754, opened 2017-01-25
Review dependencies

deivid-rodriguez:
Some changes extracted from #4624, so it's easier for you to review.

### DIFF
--- a/spec/unit/page_spec.rb
+++ b/spec/unit/page_spec.rb
@@ -34,7 +34,11 @@ module ActiveAdmin
       end
 
       it "returns the singular, lowercase name" do
-        expect(config.resource_name.singular).to eq "chocolate i lØve you!"
+        if RUBY_VERSION >= '2.4.0'
+          expect(config.resource_name.singular).to eq "chocolate i løve you!"
+        else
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
Ruby 2.4.0 failed as expected. I added what I think it's should be our build matrix for 2.4.0 in https://github.com/activeadmin/activeadmin/pull/4754/commits/47d9a511749e1e4b7c9ceac8f5172ed82bd89307. Thoughts?

deivid-rodriguez:
Ready! Anyone?

spec/unit/page_spec.rb
```diff
@@ -34,7 +34,11 @@ def config(options = {})
       end
 
       it "returns the singular, lowercase name" do
-        expect(config.resource_name.singular).to eq "chocolate i lØve you!"
+        if RUBY_VERSION >= '2.4.0'
```
Paxa: Both lines are same, isn't it?

deivid-rodriguez: No, case is different!

timoschilling: The `Ø` is lowercase in 2.4 and uppercase in others.

PR #4762, opened 2017-01-26
configure codecov.io

timoschilling:
This is the default config but with PR comments disabled.


deivid-rodriguez:
I'm ok with codecov comments, but if you prefer to remove those, no big deal.

PR #4763, opened 2017-01-26
Improve local script

deivid-rodriguez:
Improves setting up the sample application.

### DIFF
--- a/script/local
+++ b/script/local
@@ -31,7 +31,7 @@ system "mkdir #{dir}" unless File.exist?(dir)
 # Create the sample rails app if it doesn't already exist
 unless File.exist? path
   args = %w[
-    -m\ spec/support/rails_template_with_data.rb
+    -m spec/support/rails_template_with_data.rb
     --skip-bundle
     --skip-git
     --skip-listen
--- a/script/local
+++ b/script/local
@@ -12,9 +12,9 @@ app stored in test-rails-app.
 
 Examples:
 
-./script/local server
-./script/local c
-./script/local rake db:migrate
+#{__FILE__} server
+#{__FILE__} c
+#{__FILE__} rake db:migrate
   EOF
   exit(1)
 end
--- a/script/local
+++ b/script/local
@@ -32,7 +32,6 @@ system "mkdir #{dir}" unless File.exist?(dir)
 unless File.exist? path
   args = %w[
     -m\ spec/support/rails_template_with_data.rb
-    --skip-gemfile
     --skip-bundle
     --skip-git
     --skip-listen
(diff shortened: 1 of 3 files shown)


script/local
```diff
@@ -51,4 +50,4 @@ RAILS_COMMANDS = %w{generate console server dbconsole g c s runner}
 args = RAILS_COMMANDS.include?(ARGV[0]) ? ["rails", ARGV].flatten : ARGV
 
 # Run the command
-exec "cd test-rails-app && GEMFILE=../Gemfile #{args.join(" ")}"
+exec "cd test-rails-app && #{args.join(" ")}"
```
timoschilling: Can you please explain why this change (Line 34, 53)?

tasks/test.rake
```diff
@@ -7,7 +7,6 @@ task :setup, :parallel do |t, opts|
     system 'mkdir -p spec/rails'
     args = %w[
       -m\ spec/support/rails_template.rb
```
timoschilling: This line should be `-m spec/support/rails_template_with_data.rb` like in `script/local`

timoschilling:
I like to see min 2 reviewers who has tested it local before merging this.

tasks/test.rake
deivid-rodriguez: Why? This is used for tests only, I think it's better like it is now... without data. I don't think the tests would even pass if we seed the test DB.

timoschilling: Sorry I miss read the code, I only looked on the code formating, with removing the `\`

script/local
deivid-rodriguez: Yes, I should probably squash https://github.com/activeadmin/activeadmin/pull/4763/commits/f495ef6d505fbaa876fbf1be8a293b6a39a8510f and https://github.com/activeadmin/activeadmin/pull/4763/commits/ad6707de6e93e52b1dabde2d6aedad19c8e4d7f1 since they are part of the same thing. There's plenty of reasons to do this:Previously I couldn't create the sample application because of the error in https://github.com/activeadmin/activeadmin/pull/4763/commits/ad6707de6e93e52b1dabde2d6aedad19c8e4d7f1 which is pretty self descriptive. I don't know what that `GEMFILE` environment variable does, I guess it makes the `rails` command pick up that `Gemfile`but it's definitely not the best way to go. Why create the sample application without a `Gemfile` and then point to an external `Gemfile` to make it work?

timoschilling: 👍

deivid-rodriguez:
@timoschilling Thanks for the review. I'm fine with more people trying this out. Just note that this is unlikely to break anything since... it's already broken! :stuck_out_tongue:

tasks/test.rake
deivid-rodriguez: Ah, ok. Well, it does make sense to make escaping consistent. I'll fix it.

deivid-rodriguez: Addressed in c22d062.

PR #4766, opened 2017-01-27
Use default_url_options in auto_link

roshats:
`auto_link` ignores `default_url_options` now.

Before patch:
screen shot 2017-01-27 at 10 06 02 am

After patch:
screen shot 2017-01-27 at 10 07 59 am

Also I'm not happy with how `batch_action_path` was implemented and how it's implemented now. Why can't we just define route?

### DIFF
--- a/lib/active_admin/batch_actions/resource_extension.rb
+++ b/lib/active_admin/batch_actions/resource_extension.rb
@@ -51,18 +51,6 @@ module ActiveAdmin
         @batch_actions = {}
       end
 
-      # Path to the batch action itself
-      def batch_action_path(params = {}, url_options = {})
-        uri = URI(route_collection_path(params, url_options))
-        uri.path += "/batch_action"
-
-        query = params.slice(:q, :scope)
-        query = query.permit! if query.respond_to? :permit!
-        query = query.to_h if Rails::VERSION::MAJOR >= 5
-        uri.query = [uri.query, query.to_param].compact.join('&') if query.present?
-        uri.to_s
-      end
-
       private
 
       # @return [ActiveAdmin::BatchAction] The default "delete" action
--- a/lib/active_admin/batch_actions/resource_extension.rb
+++ b/lib/active_admin/batch_actions/resource_extension.rb
@@ -52,12 +52,15 @@ module ActiveAdmin
       end
 
       # Path to the batch action itself
-      def batch_action_path(params = {})
-        path = [route_collection_path(params), "batch_action"].join("/")
+      def batch_action_path(params = {}, url_options = {})
+        uri = URI(route_collection_path(params, url_options))
+        uri.path += "/batch_action"
+
         query = params.slice(:q, :scope)
         query = query.permit! if query.respond_to? :permit!
         query = query.to_h if Rails::VERSION::MAJOR >= 5
-        [path, query.to_param].reject(&:blank?).join("?")
+        uri.query = [uri.query, query.to_param].compact.join('&') if query.present?
+        uri.to_s
       end
 
       private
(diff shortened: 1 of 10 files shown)


lib/active_admin/resource/routes.rb
```diff
@@ -37,32 +37,32 @@ def initialize(resource)
           @resource = resource
         end
 
-        def collection_path(params)
+        def collection_path(params, additional_params)
```
timoschilling: `additional_params` should be optional

lib/active_admin/batch_actions/resource_extension.rb
```diff
@@ -52,12 +52,15 @@ def clear_batch_actions!
       end
 
       # Path to the batch action itself
-      def batch_action_path(params = {})
-        path = [route_collection_path(params), "batch_action"].join("/")
+      def batch_action_path(params = {}, url_options = {})
+        uri = URI(route_collection_path(params, url_options))
+        uri.path += "/batch_action"
```
timoschilling: why not building a `batch_action_path` eq to `route_collection_path`?

lib/active_admin/resource/routes.rb
```diff
@@ -37,32 +37,32 @@ def initialize(resource)
...
 
-          routes.public_send route_name, *route_collection_params(params)
+          routes.public_send route_name, *route_collection_params(params), additional_params
         end
 
         # @return [String] the path to this resource collection page
         # @param instance [ActiveRecord::Base] the instance we want the path of
         # @example "/admin/posts/1"
-        def instance_path(instance)
+        def instance_path(instance, additional_params)
```
timoschilling: `additional_params` should be optional

lib/active_admin/resource/routes.rb
```diff
@@ -37,32 +37,32 @@ def initialize(resource)
...
 
-          routes.public_send route_name, *route_instance_params(instance)
+          routes.public_send route_name, *route_instance_params(instance), additional_params
         end
 
         # @return [String] the path to the edit page of this resource
         # @param instance [ActiveRecord::Base] the instance we want the path of
         # @example "/admin/posts/1/edit"
-        def edit_instance_path(instance)
+        def edit_instance_path(instance, additional_params)
```
timoschilling: `additional_params` should be optional

timoschilling:
> Why can't we just define route?

Yes we can

timoschilling:
Can you please explain a bit more which problem will be solved by this PR.

roshats:
> Can you please explain a bit more which problem will be solved by this PR.

Rails provides a possibility to set `default_url_options`. We use `default_url_options` to pass locale in url. Most of the urls generated by activeadmin use this settings but urls in `auto_link` ignore `default_url_options`.

Scenario with a problem:
1. Modify application code to set `I18n.locale` in `ApplicationController`. Add the following method to `ApplicationController`: 
```
def default_url_options
  { locale: I18n.locale }
end
```
2. Open an admin page.
3. Change language to Russian, for instance.
4. Click on any `auto_link`.
5. Language of a page will be changed to English (default).

Please see my screenshots in attachment. Mouse is hovered on 'admin@example.com' in Resource column.

timoschilling:
please rebase your branch, then I will merge

roshats:
@timoschilling done.

deivid-rodriguez:
> Also I'm not happy with how batch_action_path was implemented and how it's implemented now. Why can't we just define route?

You didn't include that here, right? Can you add a extra commit or a diff to see how it would look?

roshats:
@deivid-rodriguez , added in https://github.com/activeadmin/activeadmin/pull/4766/commits/931a9d0f9d1376e1594d070355e5c788d32c432b .

deivid-rodriguez:
Let's see what CI says, but it definitely looks better to me!

roshats:
Tests are green again.

timoschilling:
@deivid-rodriguez are you ok with this state?

deivid-rodriguez:
I like this fix a lot. It always bothered me that AA wouldn't respect my locale selection.

It just bothers me a bit that we have to change every method signature, we can't access the `url_options` from inside the route helpers, right?

timoschilling:
Not without doing dirty stuff

PR #4768, opened 2017-01-28
order clause refactoring, allow to use custom sql ordering strategies

### LINKED ISSUE #1590: Order with Nulls Last

### LINKED ISSUE #3085: custom sort on custom column in index ?
Hi,

I'm looking for a way to make a custom sort on a custom column in an index page.

``` ruby
ActiveAdmin.register User do
  index do
    id_column
    ...
    column 'Drivers' do |user|
      user.drivers_in_neighborhood.count
    end
  end
end
```

My User#drivers_in_neighborhood method is in User model and is doing SQL queries.
I just want to make the column "Drivers" sortable.

How to do that?

Thanks!

Fivell:
fixes #1590
fixes #3085
- [x] added ability to use custom class for handling ordering per resource
  ```
    ActiveAdmin.register Post do
      config.order_clause = MyOrderClause
    end
  ```
- [x] added dsl to customize ordering strategy per column
```
    ActiveAdmin.register User do
      order_by(:full_name) do |order_clause|
        ['COALESCE(NULLIF(last_name, ''), first_name), first_name', order_clause.order].join(' ')
      end
    end
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 #### Minor
 
+* Support for custom sorting strategies [#4768][] by [@Fivell][]
 * Stream CSV downloads as they're generated [#3038][] by [@craigmcnamara][]
   * Disable streaming in development for easier debugging [#3535][] by [@seanlinsley][]
 * Improved code reloading [#3783][] by [@chancancode][]
--- a/docs/3-index-pages/index-as-table.md
+++ b/docs/3-index-pages/index-as-table.md
@@ -164,6 +164,24 @@ index do
 end
 ```
 
+## Custom sorting 
+
+It is also possible to use database specific expressions and options for sorting by column
+
+```ruby
+order_by(:title) do |order_clause|
+   if order_clause.order == 'desc'
+     [order_clause.to_sql, 'NULLS LAST'].join(' ')
+   else
+     [order_clause.to_sql, 'NULLS FIRST'].join(' ')
+   end
+end
+
+index do
+  column :title
+end
+```
+
 ## Associated Sorting
 
 You're normally able to sort columns alphabetically, but by default you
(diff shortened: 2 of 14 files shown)


Fivell:
@timoschilling , take a look pls

Issue #4423, opened 2016-05-10
Allow arrays as collection for custom pages?


paulgoetze:
Would it make sense to handle a collection that is an Array in the [`ActiveAdmin::Helpers::Collection`](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/helpers/collection.rb)

This would enable you to use an Array instead of a `ActiveRecord::Relation` when creating a table in the content block, e.g.:

``` ruby
ActiveAdmin.register_page 'Page name' do
  content do
    collection = [] # <= this can be an Array of ActiveRecord models, e.g a paginated array

    table_for(collection, class: 'index_table') do
      column :foo
      column :bar
    end

    paginated_collection(collection, entry_name: 'Page name')
  end
end
```

At the moment this throws an `undefined method 'except'` error, because `#except` is called on the collection here: https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/helpers/collection.rb#L7

I guess something like 

``` ruby
def collection_size(c = collection)
  return c.count if c.is_a?(Array) # <= !

  c = c.except :select, :order

  c.group_values.present? ? c.count.count : c.count
end
```

would be sufficient.

Fivell:
@paulgoetze ,  you're welcome to create PR with this change and spec for it

paulgoetze:
@Fivell Wow, it’s been some time. I’m not using ActiveAdmin in any project anymore atm, but I figure it might be useful for some other folks anyway :)

So here, we go with the PR: #4770

I also added the use case from above as a cucumber Scenario.

Fivell:
@paulgoetze , come back and use AA again)

Fivell:
closed via #4770

PR #4770, opened 2017-01-28
Issue #4423: Allow array as collection

paulgoetze:
This allows you to use not only an ActiveRecord::Relation but also an Array as collection for code that calls ActiveAdmin::Helpers::Collection.collection_size.

Passing an Array might be needed in case of creating a table on a custom page with a paginated array as collection.

See https://github.com/activeadmin/activeadmin/issues/4423

### DIFF
--- a/features/registering_pages.feature
+++ b/features/registering_pages.feature
@@ -146,3 +146,26 @@ Feature: Registering Pages
     And I follow "Check"
     Then I should see the content "Chocolate lØves You Too!"
     And I should see the Active Admin layout
+
+  Scenario: Registering a page with paginated index table for a collection Array
+    Given a user named "John Doe" exists
+    Given a configuration of:
+    """
+    ActiveAdmin.register_page "Special users" do
+      content do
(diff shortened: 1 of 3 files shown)


features/registering_pages.feature
```diff
@@ -146,3 +146,26 @@ Feature: Registering Pages
...
     Then I should see the content "Chocolate lØves You Too!"
     And I should see the Active Admin layout
+
+  Scenario: Registering a page with paginated index table for a collection Array
+    Given a user named "John Doe" exists
+    Given a configuration of:
+    """
+    ActiveAdmin.register_page "Special users" do
+      content do
+        collection = Kaminari.paginate_array(User.all).page(1)
```
Fivell: @paulgoetze ,  ``` .page(params.fetch(:page, 1)) ``` more looks like real world usecase

paulgoetze: I just wanted to produce an Array. But I agree, might be better if it represents a real use case. Chnaged it.

paulgoetze:
:+1: :heart:  thanks, @Fivell & @varyonic for the fast review and merging!

PR #4771, opened 2017-01-29
form page use active admin routes instead of resource_path helpers

### LINKED ISSUE #4544: "No route matches PATCH" or "DELETE" when using a model called "Collection"
I'm dealing with a model called `Collection` in ActiveAdmin. `index, new, show, edit` methods work absolutely fine, but `update` and `destroy` do not, because the `/:id` seems to be being omitted from the end of paths.

For instance, here is the opening code from the form tag when editing one of these models:

```
<form novalidate="novalidate" class="formtastic collection" id="edit_collection" action="/collections" accept-charset="UTF-8" method="post">
  <input name="utf8" type="hidden" value="✓">
  <input type="hidden" name="_method" value="patch">
```

If I use the inspector to alter the `action` attribute to `"/collections/3"`, the model updates successfully when the submit button is clicked.

(nb. I am not using a global namespace for ActiveAdmin here.)
### Workaround

Register the model under a different name with ActiveAdmin, eg.:

```
ActiveAdmin.register Collection as: 'collection_thing' do
  ...
```

Fivell:
form page use active admin routes instead of resource_path helpers

fixes #4544

### DIFF
--- a/lib/active_admin/views/pages/form.rb
+++ b/lib/active_admin/views/pages/form.rb
@@ -17,13 +17,15 @@ module ActiveAdmin
         end
 
         def main_content
-          options = default_form_options.merge form_presenter.options
-
           if options[:partial]
             render options[:partial]
           else
             active_admin_form_for resource, options, &form_presenter.block
           end
+        end
+
+        def options
+          @options ||= default_form_options.merge(form_presenter.options)
         end
 
         private
@@ -36,7 +38,7 @@ module ActiveAdmin
         end
 
         def default_form_path
-          resource.persisted? ? resource_path(resource) : collection_path
+          resource.persisted? ? active_admin_config.route_instance_path(resource, url_options) : active_admin_config.route_collection_path(url_options)
         end
 
         def default_form_config
--- a/spec/unit/views/pages/form_spec.rb
+++ b/spec/unit/views/pages/form_spec.rb
@@ -1,12 +1,15 @@
 require 'rails_helper'
 
 RSpec.describe ActiveAdmin::Views::Pages::Form do
-  let!(:application) { ActiveAdmin::Application.new }
-  let(:namespace) { ActiveAdmin::Namespace.new(application, "Admin") }
   let!(:params) { { controller: "PostsController", action: "edit" } }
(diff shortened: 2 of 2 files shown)


Fivell:
travis
```
undefined method `user' for #<Post:0x18575c1e> (ActionView::Template::Error)
```

connected with #4618 

Post ```has one author```  instead of ```user```, but it doesn't work this way

Fivell:
@timoschilling , is that another bug?

timoschilling:
I don't know

Fivell:
@dallasgray, maybe you have idea why this happens ?

Fivell:
@rdallasgray, maybe you have idea why this happens?

rdallasgray:
@Fivell Nothing jumps out at me I'm afraid. Bear in mind @bamorim originated the code for nested `belongs_to` -- I just rebased and got tests passing.

lib/active_admin/resource/routes.rb
```diff
@@ -98,13 +98,16 @@ def route_name(resource_path_name, options = {})
 
 
         # @return params to pass to instance path
-        def route_instance_params(instance)
-          belongs_to_names.reverse.reduce([instance]) do |arr,name| 
-            arr + [arr.last.public_send(name)]
+        def route_instance_params(instance, params = {})
+          belongs_to_names.reverse.reduce([instance.to_param]) do |arr, name|
+            #TODO read from belongs_to config
+            param_key = "#{name}_id"
```
rdallasgray: Should this be a symbol as in line 111?

Fivell: @rdallasgray , this is just another try I will rollback last changes

Fivell:
@timoschilling , I think about continuing on this after we drop support rails 3.2, can't figure out how to make it work on all rails versions

timoschilling:
ok, lets wait

Fivell:
I will separate PRs after more deep investigation, I think it is better to create different PRs for different issues

javierjulio:
@Fivell based on your last comment then, should we close this out?

Fivell:
@javierjulio , this fix doesn't work for nested pages/resources and  requires belongs_to issues to be fixed first

PR #4772, opened 2017-01-30
Revert "Merge pull request #4618 

Fivell:
This reverts commit a8b3bef3a678919a69f808e7516219294a03f553, reversing
changes made to 9fb572a4f10353508e7f6876594a9989cf550367.

Hello @varyonic , @timoschilling , @deivid-rodriguez.
Hello @rdallasgray 
I want to discuss reverting of nested belongs to feature that was added before.

First of all I'm considering this to be a major feature and unfortunately  
   - we need features to cover crud operations for single belongs to 
   - we have broken default permitted_params  that doesn't support belongs_to at all
   - we have 2 different APIs for pages belongs_to that is disabled for no reason, and resource belongs_to that is already 2 level. 
   - we have some routing bugs that need to be fixed before
   - we already receiving issues related to multiple belongs to 
  
All of this became hard to reach with merging not well tested "nested belongs_to" feature that gives us new level of complexity. 

After fixing this problems we can return to nested belongs to feature
However this requires some cucumber  features to implement 
   - one of belongs_to can be optional 
   - create/edit/destroy nested resources for cases
      - one of belongs_to is optional
      - both are required
      - route in uri is different from actual activerecord's belongs_to declaration 
         

In general, let's make belongs_to work before thinking of nested belongs_to

### DIFF
--- a/features/belongs_to.feature
+++ b/features/belongs_to.feature
@@ -110,24 +110,6 @@ Feature: Belongs To
     When I follow "Posts"
     Then the "Posts" tab should be selected
 
-  Scenario: When the belongs to is nested
-    Given a configuration of:
-    """
-      ActiveAdmin.register User
-      ActiveAdmin.register Post do
-        belongs_to :user
-      end
-      ActiveAdmin.register Tagging do
-        belongs_to :user
-        belongs_to :post
-      end
-    """
-    When I go to the last author's last post's taggings
-    Then I should see a link to "Users" in the breadcrumb
-    And I should see a link to "Jane Doe" in the breadcrumb
-    And I should see a link to "Posts" in the breadcrumb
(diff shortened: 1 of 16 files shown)


deivid-rodriguez:
@Fivell Makes sense. Let's make current features solid, the worry about adding new stuff.

rdallasgray:
@Fivell That's a bit depressing but sure. I'll try to get some time to implement the feature specs you mentioned.

Fivell:
@rdallasgray, if you have time you can open work in progress PR and I will assist you, but we need fix bugs unlit 1.0.0 release

Fivell:
@rdallasgray did you have update with feature specs?

Fivell:
we will return to this feature after release

PR #4773, opened 2017-01-31
Tweak appraisal

deivid-rodriguez:
We've seen some segfaults in Travis with ruby 2.4.0 after introducing Appraisal.

They only happen when running things like Travis does it: setting the `BUNDLE_GEMFILE` environment variable and then running commands as usual. I had only tried the other way: running things directly through appraisal (`script/appraisal rails_50 rake test`, for example).

I could reproduce the segfaults locally and these changes seem to prevent them. I have no idea why tests passed in the previous PR, I guess they just happen randomly.

Let's see if Travis passes...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,20 +46,20 @@ matrix:
       gemfile: gemfiles/rails_50.gemfile
 
     - rvm: 2.4.0
-      env: RAILS=3.2.22
+      gemfile: gemfiles/rails_32.gemfile
 
     - rvm: 2.4.0
-      env: RAILS=4.0.13
+      gemfile: gemfiles/rails_40.gemfile
 
     - rvm: 2.4.0
-      env: RAILS=4.1.16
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
I investigated the decrease in test coverage and it's expected. The last coverage report (after merging #4624) incorrectly slightly increased coverage (probably because the segfaults caused some reports in a bad state to be sent). See for example: https://codecov.io/gh/activeadmin/activeadmin/src/a6c3e3782709b191931a5608b3a37a0ba5495999/lib/active_admin.rb#L75. Line 75 is incorrectly counted as covered. This PR just resets coverage back to normal.

So I think we should ignore the required check in this case. I think @timoschilling has superpowers to do that.

PS: I added one extra commit while investigating this (https://github.com/activeadmin/activeadmin/pull/4773/commits/a5553c9762f686c054f81ead226ea70cfe580adb). It removes some dead test code.

Issue #3556, opened 2014-11-03
permit_params does not include belongs_to


kmayer:
If I have a resource with a `belongs_to` relationship, the new action fails because the "parent" id is not whitelisted:

```ruby
class AdminUser < ActiveRecord::Base
  belongs_to :employer
end

class Employer < ActiveRecord::Base
  has_many :administrators
end
...
ActiveAdmin.register Administrator do
  belongs_to :employer

  permit_params :employer_id
end
ActiveAdmin.register Employer do
end
```

visiting `/admin/employer/1/users/new` raises an error:

```
ActionController::UnpermittedParameters in Admin::AdministratorsController#new
found unpermitted parameters: employer_id
```

And with some digging, it turns out that `employer_id` isn't wrapped in the `param_key` (https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/resource_dsl.rb#L50), so it isn't permitted.

once the generated `permitted_params` method is called by InheritedResource, the resulting list of keys are:

```
[:utf8, :_method, :authenticity_token, :commit, :id, {:administrator=>[:employer_id]}]
```

but the submitted parameters (to the `new` action) are:

```
{"action"=>"new", "controller"=>"admin/administrators", "employer_id"=>"1"}
```

As a workaround, I just overrode the `permitted_params` method in the `Administrator` resource:

```ruby
  controller do
    define_method :permitted_params do
      params.permit *active_admin_namespace.permitted_params, :employer_id,
                    :administrator => [:email, :role, :employer_id, :password, :password_confirmation]
    end
  end
```

macowie:
Just hit this too on 1.0.0pre2, so it's still around. The workaround above did, in fact, work around it.

timoschilling:
You need to list all needed attrs

codyrobbins:
I’m running into the same problem on 1.0.0pre4. As described in the original comment, creating new nested resources is not possible. Unless I’m misunderstanding something, I think this is indeed a bug and shouldn’t have been closed.

@timoschilling If this isn’t a bug would you mind explaining in a little more detail how this issue can be worked around without redefining `permitted_params` in the controller? Thanks!

Ruff9:
The problem is still here on 1.0.0pre4.

dmitry:
@Ruff9 can you post a simple code that reproduces a problem?

Ruff9:
Considering that ActionItem belongs_to Action:

`ActiveAdmin.register Action do
[...]

PR #4776, opened 2017-02-01
Permit params for belongs to

### LINKED ISSUE #4765: permit_params - Unpermitted parameter of related model
I have a Orders model and OrdersProducts model.   Multiple orders can be held inside OrdersProducts.  In my OrdersProducts I have the following:
```
belongs_to :order
permit_params :order,
              :order_id,
              :product_id,
              :item_id,
              :user_id,
              :order_type,
              :product_label
              #order_ids:[]
```

When I attempt to update or create I get the following issue:  
**Unpermitted parameter: order_id**

Looking at the logs everything looks find with the request, but the transaction is being rolled back because the param is not allowed.

**Log** 
```
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"ZS46Src+SHV9ufiZ5nEqiBnyIEq2DMS3CQL/f3rbqjNDJFAukuJ1bDOgW//0WwOJap2UR6bzw2nsvFCVCTCkqA==", "orders_products"=>{"product_label"=>"sdafasd", "order_type"=>"purchase", "product_id"=>"18", "item_id"=>"", "order_id"=>"42"}, "commit"=>"Create Orders products", "order_id"=>"42"}
```

Fivell:
fixes #4765 , #3556

### DIFF
--- a/features/belongs_to.feature
+++ b/features/belongs_to.feature
@@ -21,8 +21,33 @@ Feature: Belongs To
     And I should see "Displaying 1 Post"
     And I should see a link to "Users" in the breadcrumb
     And I should see a link to "Jane Doe" in the breadcrumb
-    When I follow "Edit"
-    Then I should see a link to "Hello World" in the breadcrumb
+
+  Scenario: Updating a child resource page
+    Given a configuration of:
+    """
+      ActiveAdmin.register User
+      ActiveAdmin.register Post do
+        belongs_to :user
+        permit_params :title, :body, :published_date if Rails::VERSION::MAJOR >= 4
+
+        form do |f|
+          f.inputs "Your Post" do
+            f.input :title
+            f.input :body
+          end
+          f.inputs "Publishing" do
+            f.input :published_date
+          end
+          f.actions
+        end
+      end
+    """
+    When I go to the last author's last post page
+    And I follow "Edit Post"
+    And I should see the element "form[action='/admin/users/2/posts/2']"
+    And I should see a link to "Hello World" in the breadcrumb
+    When I press "Update Post"
+    Then I should see "Post was successfully updated."
 
   Scenario: Viewing a child resource page
     Given a configuration of:
--- a/features/belongs_to.feature
+++ b/features/belongs_to.feature
@@ -43,11 +43,44 @@ Feature: Belongs To
       end
     """
     When I go to the last author's last post page
-    And I follow "Edit Post"
-    And I should see the element "form[action='/admin/users/2/posts/2']"
-    And I should see a link to "Hello World" in the breadcrumb
+    Then I follow "Edit Post"
+    Then I should see the element "form[action='/admin/users/2/posts/2']"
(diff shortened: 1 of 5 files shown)


lib/active_admin/resource.rb
```diff
@@ -135,6 +135,10 @@ def belongs_to_config
       @belongs_to ||= []
     end
 
+    def belongs_to_params
+      belongs_to_config.select{ |c| c.required? }.map{|c| c.to_param }
```
timoschilling: `map{ |c|` missing whitespace

Fivell: @timoschilling , whitespace  added

deivid-rodriguez: @timoschilling Isn't @houndci-bot supposed to make this kind of comments? Or is it disabled?

features/belongs_to.feature
```diff
@@ -21,8 +21,65 @@ Feature: Belongs To
...
+          f.inputs "Publishing" do
+            f.input :published_date
+          end
+          f.actions
+        end
+      end
+    """
+    When I go to the last author's last post page
+    And I follow "Edit Post"
+    And I should see the element "form[action='/admin/users/2/posts/2']"
```
deivid-rodriguez: I don't know much about cucumber but aren't expectations supposed to be preceded by `Then`? It doesn't read very well like this.

Fivell: @deivid-rodriguez , I don't know much too, I will change it.

features/belongs_to.feature
```diff
@@ -21,9 +21,67 @@ Feature: Belongs To
...
+          end
+          f.inputs "Publishing" do
+            f.input :published_date
+          end
+          f.actions
+        end
+      end
+    """
+    When I go to the last author's last post page
+    Then I follow "Edit Post"
```
deivid-rodriguez: hehe, well I meant to change the following line, not this one. This seems part of the "cucumber setup", so `And` looked fine to me here.

features/belongs_to.feature
```diff
@@ -21,9 +21,67 @@ Feature: Belongs To
...
+          end
+          f.inputs "Publishing" do
+            f.input :published_date
+          end
+          f.actions
+        end
+      end
+    """
+    When I go to the last author's posts
+    Then I follow "New Post"
```
deivid-rodriguez: Same here, I think `And` should be used.

features/belongs_to.feature
```diff
@@ -21,9 +21,67 @@ Feature: Belongs To
...
+    Then I should see the element "form[action='/admin/users/2/posts']"
+    Then I fill in "Title" with "Hello World"
```
deivid-rodriguez: This is setup again, so I think `When` should be used. I don't really know if all of this makes a difference, it just reads better in English IMO.

Fivell: @deivid-rodriguez , maybe it is better to review all features in separate PR?

PR #4781, opened 2017-02-07
Apply the suffix adjustment in collection_path also to batch_action_path

knu:
When the resource model name were an uncountable noun like "News", the URL helper method for the batch action would be suffixed with an extra `_index` suffix.

We have confirmed that this fixes the following error you get with the current implementation when ActiveAdmin renders a batch action button:

```
Showing .../vendor/bundle/ruby/2.3.0/bundler/gems/activeadmin-6463e2b717b2/app/views/active_admin/resource/index.html.arb
where line #2 raised:

undefined method `batch_action_admin_news_path' for ActiveAdmin::Helpers::Routes:Module
Did you mean?  batch_action_admin_news_index_path
               ...
```

### DIFF
--- a/lib/active_admin/resource/routes.rb
+++ b/lib/active_admin/resource/routes.rb
@@ -90,7 +90,7 @@ module ActiveAdmin
           suffix = options[:suffix] || "path"
           route = []
 
-          route << options[:action]           # "edit" or "new"
+          route << options[:action]           # "batch_action", "edit" or "new"
           route << resource.route_prefix      # "admin"
           route += belongs_to_names
           route << resource_path_name         # "posts" or "post"
--- a/lib/active_admin/resource/routes.rb
+++ b/lib/active_admin/resource/routes.rb
@@ -51,8 +51,11 @@ module ActiveAdmin
         end
 
         def batch_action_path(params, additional_params = {})
-          path = resource.resources_configuration[:self][:route_collection_name]
-          route_name = route_name(path, action: :batch_action)
+          route_name = route_name(
+            resource.resources_configuration[:self][:route_collection_name],
+            action: :batch_action,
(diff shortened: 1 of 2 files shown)


knu:
I'll follow up with some specs.

knu:
@timoschilling Added specs contributed by my colleague.

Fivell:
similar logic in IR https://github.com/activeadmin/inherited_resources/blob/95d5f93e92e757aefeba6515ec67148ea4e9dfdf/lib/inherited_resources/url_helpers.rb#L132

Fivell:
@knu, any update ?

knu:
@Fivell The URL helpers that IR generates are callees and AA is the caller here, and the naming rules in IR at this point are not accessible from AA.  Do you think we should modify IR before fixing this bug for a cleaner implementation?

Fivell:
@knu let's merge it but can you answer on my comments above before

knu:
@Fivell What comments are you referring to?

Fivell:
@knu, codereview comments

spec/support/rails_template_with_data.rb
```diff
@@ -1,9 +1,16 @@
 apply File.expand_path("../rails_template.rb", __FILE__)
 
-%w{Post User Category}.each do |type|
+%w{Post User Category News}.each do |type|
```
knu: You're right, apparently the news model in the template isn't used and it is unnecessary.  I'll remove the last commit shortly.

knu:
And the CI still passes!

Fivell:
@knu , please rebase PR

knu:
@Fivell Sure.  I've rebased this onto the current master, and there was no conflict to resolve.

knu:
Thanks!

Issue #4298, opened 2016-02-04
Missing field from model


tylercollier:
I'm using Active Admin (directly from github in Gemfile, so commit a6f6794c0fba). It's been working fine for a long time. Today I ran a migration to add a field to my Agent model. That field is not showing up in any of the active admin parts for that model (index, form, filter). I'm confident that Active Admin is connecting to the correct database, because if I use the following code in `app/admin/agent.rb`, it DOES show up, with correct values from the database:

```
index do
  selectable_column
  Agent.column_names.each do |c|
    column c.to_sym
  end
  default_actions
end
```

But if remove that code (meaning the columns should be the default), I do not see my field/column. I must do a similar trick on the form inputs.

Any ideas?

tylercollier:
I just added another field to the table in the database. The new field showed up as usual, without me having to do anything special.

In case it helps, the 'special' field, that didn't show up, was named `idx_id`. The one that showed up fine was `email_cc`.

seanlinsley:
Currently we use `content_columns` to decide which attributes to display:
```ruby
lib/active_admin/csv_builder.rb:
   23        new resource: resource do
   24          column :id
   25:         resource.content_columns.each { |c| column c.name.to_sym }
   26        end
   27      end

lib/active_admin/filters/resource_extension.rb:
  140        # Returns a default set of filters for the content columns
  141        def default_content_filters
  142:         if resource_class.respond_to? :content_columns
  143:           resource_class.content_columns.map{ |c| c.name.to_sym }
  144          else
  145            []

lib/active_admin/views/index_as_table.rb:
  234            selectable_column
  235            id_column if resource_class.primary_key # View based Models have no primary_key
  236:           resource_class.content_columns.each do |col|
  237              column col.name.to_sym
  238            end
```

`content_columns` is defined by Rails, and excludes any attributes with `_id` in the name:
```ruby
def content_columns
  @content_columns ||= columns.reject { |c| c.name == primary_key || c.name =~ /(_id|_count)$/ || c.name == inheritance_column }
end
```

That's not the best though. I think a more robust option would map through the associations the model has, and only exclude the attributes ending in `_id` if they're a foreign key on an association.

tylercollier:
Thanks for the great diagnosis. In my case, `idx_id` is not a foreign key. Well, it technically is, but not one known to Rails. So that would explain it.

You must have something special for associations. Because `content_columns` would exclude them right, since it excludes all `_id` columns.

[...]

PR #4784, opened 2017-02-08
 resource attributes layer (refactoring)

Fivell:
introducing correct handling foreign keys attributes on resource level.

Problems:

https://github.com/activeadmin/activeadmin/issues/4298#issuecomment-185467400 content_columns hardcoded in many places 

https://github.com/activeadmin/activeadmin/commit/f21cff30e7212706a39710e2c70ad0af8087ccea see Notes to commit message, example when renaming foreign key affects default show content

https://github.com/activeadmin/activeadmin/issues/4298#issuecomment-185849747  ActiveAdmin can't figure out if column is foreign key or not

https://github.com/activeadmin/activeadmin/commit/3efb359533df46e76bb4355a02ad31e727fb0da3#diff-b2d41db115eac24dba6f5e64bda8a690R53 introduced reflection parsing for each find_value method call because we still not know if this foreign key or not.

Idea is to initialise attributes for each resource and skip heavy magic in runtime 

@timoschilling , @varyonic ,  @deivid-rodriguez  before continue working on this, I would like to have feedback

### DIFF
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -22,7 +22,7 @@ module ActiveAdmin
     def self.default_for_resource(resource)
       new resource: resource do
         column :id
-        resource.content_columns.each { |c| column c.name.to_sym }
+        resource.content_columns.each { |c| column c }
       end
     end
 
@@ -105,7 +105,7 @@ module ActiveAdmin
 
       def humanize_name(name, resource, humanize_name_option)
         if humanize_name_option
-          name.is_a?(Symbol) && resource.present? ? resource.human_attribute_name(name) : name.to_s.humanize
+          name.is_a?(Symbol) && resource ? resource.resource_class.human_attribute_name(name) : name.to_s.humanize
         else
           name.to_s
         end
--- a/lib/active_admin/filters/resource_extension.rb
+++ b/lib/active_admin/filters/resource_extension.rb
@@ -109,7 +109,7 @@ module ActiveAdmin
       def default_filters
         result = []
         result.concat default_association_filters if namespace.include_default_association_filters
-        result.concat default_content_filters
+        result.concat content_columns
         result.concat custom_ransack_filters
         result
       end
@@ -121,7 +121,7 @@ module ActiveAdmin
           []
         end
       end
-
+      
       # Returns a default set of filters for the associations
       def default_association_filters
         if resource_class.respond_to?(:reflect_on_all_associations)
@@ -132,15 +132,6 @@ module ActiveAdmin
 
           filters = poly.map(&:foreign_type) + not_poly.map(&:name)
           filters.map &:to_sym
-        else
-          []
-        end
-      end
-
-      # Returns a default set of filters for the content columns
-      def default_content_filters
-        if resource_class.respond_to? :content_columns
-          resource_class.content_columns.map{ |c| c.name.to_sym }
         else
           []
         end
--- a/lib/active_admin/resource.rb
+++ b/lib/active_admin/resource.rb
@@ -1,4 +1,5 @@
 require 'active_admin/resource/action_items'
+require 'active_admin/resource/attributes'
 require 'active_admin/resource/controllers'
 require 'active_admin/resource/menu'
 require 'active_admin/resource/page_presenters'
@@ -87,6 +88,7 @@ module ActiveAdmin
     include Sidebars
     include Routes
     include Ordering
+    include Attributes
 
     # The class this resource wraps. If you register the Post model, Resource#resource_class
     # will point to the Post class
@@ -163,6 +165,22 @@ module ActiveAdmin
       (decorator_class && resource) ? decorator_class.new(resource) : resource
     end
 
+    def resource_columns
+      resource_attributes.values
+    end
+
+    def resource_attributes
+      @resource_attributes ||= default_attributes
+    end
+
+    def association_columns
+      @association_columns ||= resource_attributes.select{ |key, value| key != value }.values
+    end
+
+    def content_columns
+      @content_columns ||= resource_attributes.select{ |key, value| key == value }.values
+    end
+
     private
 
     def method_for_find(id)
@@ -174,7 +192,7 @@ module ActiveAdmin
     end
 
     def default_csv_builder
-      @default_csv_builder ||= CSVBuilder.default_for_resource(resource_class)
+      @default_csv_builder ||= CSVBuilder.default_for_resource(self)
     end
 
   end # class Resource
--- a/lib/active_admin/resource/attributes.rb
+++ b/lib/active_admin/resource/attributes.rb
@@ -1,0 +1,44 @@
+module ActiveAdmin
+
+  class Resource
+    module Attributes
+
+      def default_attributes
+        resource_class.columns.each_with_object({}) do |c, attrs|
+          unless reject_col?(c)
+            name = c.name.to_sym
+            attrs[name] = (method_for_column(name) || name)
+          end
+        end
+      end
+
+      def method_for_column(c)
+        resource_class.respond_to?(:reflect_on_all_associations) && foreign_methods.has_key?(c) && foreign_methods[c].name.to_sym
+      end
+
+      def foreign_methods
+        @foreign_methods ||= resource_class.reflect_on_all_associations.
+          select{ |r| r.macro == :belongs_to }.
+          reject{ |r| r.chain.length > 2 && !r.options[:polymorphic] }.
+          index_by{ |r| r.foreign_key.to_sym }
+      end
+
+      def reject_col?(c)
+        primary_col?(c) || sti_col?(c) || counter_cache_col?(c)
+      end
+
+      def primary_col?(c)
+        c.name == resource_class.primary_key
+      end
+
+      def sti_col?(c)
+        c.name == resource_class.inheritance_column
+      end
+
+      def counter_cache_col?(c)
+        c.name.end_with?('_count')
+      end
+
+    end
+  end
+end
--- a/lib/active_admin/views/index_as_table.rb
+++ b/lib/active_admin/views/index_as_table.rb
@@ -233,8 +233,8 @@ module ActiveAdmin
         proc do
           selectable_column
           id_column if resource_class.primary_key
-          resource_class.content_columns.each do |col|
-            column col.name.to_sym
+          active_admin_config.resource_columns.each do |attribute|
+            column attribute
           end
           actions
         end
--- a/features/index/format_as_csv.feature
+++ b/features/index/format_as_csv.feature
@@ -12,8 +12,8 @@ Feature: Format as CSV
     When I am on the index page for posts
     And I follow "CSV"
     And I should download a CSV file for "posts" containing:
-    | Id  | Title       | Body | Published date | Position | Starred | Created at | Updated at |
-    | \d+ | Hello World |      |              |          |         | (.*)       | (.*)       |
+    | Id  | Title       | Body | Published date | Position | Starred | Foo    |Created at | Updated at |
+    | \d+ | Hello World |      |                |          |         |        |(.*)       | (.*)       |
 
   Scenario: Default with alias
     Given a configuration of:
@@ -24,7 +24,7 @@ Feature: Format as CSV
     When I am on the index page for my_articles
     And I follow "CSV"
     And I should download a CSV file for "my-articles" containing:
-    | Id  | Title       | Body | Published date | Position | Starred | Created at | Updated at |
+    | Id  | Title       | Body | Published date | Position | Starred | Foo    | Created at | Updated at |
 
   Scenario: With CSV format customization
     Given a configuration of:
--- a/features/index/index_as_table.feature
+++ b/features/index/index_as_table.feature
@@ -250,14 +250,14 @@ Feature: Index as Table
       """
     When I am on the index page for posts
     Then I should see the "index_table_posts" table:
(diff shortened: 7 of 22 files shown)


Fivell:
@timoschilling , @varyonic , @deivid-rodriguez, specs fixed, ready for review

Fivell:
@timoschilling , any update ?

lib/active_admin/resource.rb
```diff
@@ -162,6 +165,22 @@ def find_resource(id)
...
+    def resource_attributes
+      @resource_attributes ||= default_attributes
+    end
+
+    def association_columns
+      @association_columns ||= resource_attributes.select{ |key,value| key != value }.values
+    end
+
+    def content_columns
+      @content_columns ||= resource_attributes.select{ |key,value| key == value }.values
```
timoschilling: `|key, value|`

timoschilling:
Code looks ok to me, but I don't have the time to understand the functionality at the moment.

Fivell:
@timoschilling , when you have time , let's  discuss this, I have an idea about moving this to our [orm folder](https://github.com/activeadmin/activeadmin/tree/master/lib/active_admin/orm/active_record)

what do you think ?

Fivell:
@varyonic , maybe you have time to discuss this?

Fivell:
@deivid-rodriguez , this stuff with remote_const is failed again

Fivell:
@deivid-rodriguez , I do  remember you told me there were a way to fix it

lib/active_admin/resource.rb
```diff
@@ -1,4 +1,6 @@
 require 'active_admin/resource/action_items'
+require 'active_admin/resource/attributes'
+require 'active_admin/resource/ordering'
```
varyonic: Duplicate require

deivid-rodriguez:
@Fivell Yes, sorry, I forgot about it. Let me add a fix.

lib/active_admin/resource.rb
Fivell: @varyonic , attributes != ordering ... , don't understand what duplication do you mean

varyonic: ordering is included also L.15

lib/active_admin/filters/resource_extension.rb
```diff
@@ -109,19 +109,11 @@ def filter_lookup
...
       end
 
-      def custom_ransack_filters
-        if resource_class.respond_to?(:_ransackers)
-          resource_class._ransackers.keys.map(&:to_sym)
-        else
-          []
-        end
-      end
-
```
deivid-rodriguez: This method was not removed, was just moved around. Any reason? Moving it back here would simplify the diff and make reviewing this easier.

spec/support/rails_template.rb
```diff
@@ -29,6 +29,18 @@ class Post < ActiveRecord::Base
...
+      parent.table[:title]
+    end
+
+    ransacker :custom_created_at_searcher do |parent|
+      parent.table[:created_at]
+    end
+
+    ransacker :custom_searcher_numeric, type: :numeric do
+      # nothing to see here
+    end
```
deivid-rodriguez: All this stuff was moved in another PR, I think it's duplicated now.

Fivell: sure you're right

Fivell:
@activeadmin/collaborators ,  if you don't mind I will merge this in evening

deivid-rodriguez:
:+1: I'll try to review it before. If I don't make it, merge away and I'll review post-merge if I feel it's necessary!

Fivell:
@deivid-rodriguez , sure,  all previous comments were fixed and rebased

lib/active_admin/csv_builder.rb
```diff
@@ -105,7 +105,7 @@ def initialize(name, resource = nil, options = {}, block = nil)
 
       def humanize_name(name, resource, humanize_name_option)
         if humanize_name_option
-          name.is_a?(Symbol) && resource.present? ? resource.human_attribute_name(name) : name.to_s.humanize
+          name.is_a?(Symbol) && resource.respond_to?(:resource_class) ? resource.resource_class.human_attribute_name(name) : name.to_s.humanize
```
deivid-rodriguez: This is the first thing that surprised me, I would expect that we shouldn't need the `respond_to?` check because our resources always respond to `resource_class`. It turns out that the `ResourceDSL` class is initialized with an AA resource object _and_ the underlying object (raw `Post`, for example), and this second object is what's used by the `CSVBuilder`.

Since we are refactoring, I think this dependency in the `ResourceDSL` class is not necessary at all (if it needs to access the underlying resource, it can always do `config.resource_class`). So I would kill the overriden initializer in https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/resource_dsl.rb#L4-L8 and use `config` instead of `@resource` here: https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/resource_dsl.rb#L110-L112.

That way we could keep the old `present?` check and just change `resource.human_attribute_name` with `resource.resource_class.human_attribute_name` since this class now gets the actual resource and not its class.

Makes sense?

Fivell: @deivid-rodriguez , yes, let's try

Fivell:
@deivid-rodriguez , works

lib/active_admin/resource/attributes.rb
```diff
@@ -0,0 +1,31 @@
+module ActiveAdmin
+
+  class Resource
+    module Attributes
+
+      def default_attributes
+        return {} unless resource_class.respond_to?(:columns)
```
deivid-rodriguez: Can this actually happen?

lib/active_admin/views/index_as_table.rb
```diff
@@ -233,8 +233,8 @@ def default_table
         proc do
           selectable_column
           id_column if resource_class.primary_key
-          resource_class.content_columns.each do |col|
-            column col.name.to_sym
+          active_admin_config.resource_attributes.values.each do |attribute|
```
deivid-rodriguez: You can use `resource_columns` here, no?

Fivell: @deivid-rodriguez , yes

lib/active_admin/resource/attributes.rb
Fivell: @deivid-rodriguez , don't remember context, maybe for activemodel compatibility or something... looks like it is better to remove this since 
I don't see any test for this to be present

Fivell:
@deivid-rodriguez , fixed and squashed

lib/active_admin/resource/attributes.rb
```diff
@@ -0,0 +1,30 @@
+module ActiveAdmin
+
+  class Resource
+    module Attributes
+
+      def default_attributes
+        resource_class.columns.each_with_object({}) do |c, attrs|
+          #skip primary key, STI column and counter cache columns
+          unless  c.name == resource_class.primary_key || c.name == resource_class.inheritance_column || c.name =~ /_count$/
```
deivid-rodriguez: I would extract `primary_key_col?` `sti_col?` and `counter_cache_col?` methods here. That way we avoid the need for the comment since the code would be self-explanatory.

lib/active_admin/resource/attributes.rb
```diff
@@ -0,0 +1,30 @@
...
+            name = c.name.to_sym
+            attrs[name] = (method_for_column(name) || name)
+          end
+          attrs
```
deivid-rodriguez: I think you don't need to explicitly return the accumulator when you use `each_with_object`.

Fivell:
@deivid-rodriguez ,  take a look now

lib/active_admin/resource/attributes.rb
```diff
@@ -0,0 +1,44 @@
+module ActiveAdmin
+
+  class Resource
+    module Attributes
+
+      def default_attributes
+        resource_class.columns.each_with_object({}) do |c, attrs|
+          unless reject_col?(c)
```
deivid-rodriguez: Actually, do we need this check at all? I applied the following patch and all features passed:

```diff
diff --git a/lib/active_admin/csv_builder.rb b/lib/active_admin/csv_builder.rb
index ba323302..36fcd8e8 100644
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -21,7 +21,7 @@ module ActiveAdmin
     # resource's content columns
     def self.default_for_resource(resource)
       new resource: resource do
-        column :id
+        # column :id
         resource.content_columns.each { |c| column c }
       end
     end
diff --git a/lib/active_admin/resource/attributes.rb b/lib/active_admin/resource/attributes.rb
index 48063ddf..c587b72d 100644
--- a/lib/active_admin/resource/attributes.rb
+++ b/lib/active_admin/resource/attributes.rb
@@ -24,6 +24,7 @@ module ActiveAdmin
       end
 
       def reject_col?(c)
+        return false
         primary_col?(c) || sti_col?(c) || counter_cache_col?(c)
       end
 
@@ -41,4 +42,4 @@ module ActiveAdmin
 
     end
   end
-end
\ No newline at end of file
+end
diff --git a/lib/active_admin/views/index_as_table.rb b/lib/active_admin/views/index_as_table.rb
index 01323833..5da66c4f 100644
--- a/lib/active_admin/views/index_as_table.rb
+++ b/lib/active_admin/views/index_as_table.rb
@@ -232,7 +232,7 @@ module ActiveAdmin
       def default_table
         proc do
           selectable_column
-          id_column if resource_class.primary_key
+          # id_column if resource_class.primary_key
           active_admin_config.resource_columns.each do |attribute|
             column attribute
           end
```

I can see a point for skipping the inheritance column on STI models (we might want to test this), but I think the `counter_cache` columns can be useful by default, and that we don't really need to skip the primary key.

Fivell: This is out of the scope of current PR I'm not going to change functionality please read original message

deivid-rodriguez: Sorry @Fivell, I'm not sure I understand. I don't see any counter cache column detection in the previous code. Could you clarify?

Fivell: No it not changing functionality counters cache columns and sti columns were always ignored see rails content_columns method implementation

deivid-rodriguez: Oh, I didn't know rails had such method, nice. Where are we using it? In `CSVBuilder`, you seem to be using Rails `columns` method, not content columns.

deivid-rodriguez: Feel free, to not answer, I'll re-read the PR knowing what I know now. Thanks! :wink:

deivid-rodriguez: Ok, I think I get it now.

* _Columns ended with `_id`._ Rails's `content_columns` excludes these, but we don't want to do that because we want to show those in our tables and CSV exports when they are not linked to an association . So we switch from AR `content_columns` to just `columns` and don't filter out these columns. :+1: for that change in behaviour. 

* _Columns ended with `_count`_. By the same argument, I'd say we want to have these columns as well, since they are just regular columns not linked to any association. This would be a change in behavior, just like adding `_id`-ended columns. :+1: for this change in behavior as well. I think it's better behavior and it saves us from duplicating AR functionality. 

* _Primary key column_. In this case, my suggestion doesn't change the behavior but it leads to simpler code because we don't duplicate AR functionality either. So :+1: to simplifying that.

Since this PR is tagged as "refactoring", I'd suggest to try not duplicate AR functionality if we don't need to. So I would:

* Drop `reject_col?` method.
* Drop `counter_cache_col?` method.
* Drop `primary_col?` method.
* Drop https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/csv_builder.rb#L24 line.
* Drop https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/views/index_as_table.rb#L235 line.

Fivell: @deivid-rodriguez , 

Again I'm not going to change default rendering attributes.
Only one functional change is  that columns ends with _id they are not always foreign keys,  that's one of the reasons why rails content_columns is not suitable, (see foo_id in tests). Also that's why it is not about duplication of AR functionallity. 

Switching CSV export to use columns is out of scope since it is changing behaviour too, columns include also foreign keys and STI columns. 

Columns ended with _count.  They are not regular columns, they are sytem columns to cache counters, they were designed for different things, changing this behaviour anyway is not context of this PR, as well as sti column. User still can ovverride with activeadmin DSL.


Dropping id_column in index_table is bad idea, it is helper method and not just rendering primary key.

Fivell: @deivid-rodriguez , feel free to create separate PR with changing default attributes you want to render and we will discuss what kind of columns we should keep and leave and what tests and features we should add to cover this.

deivid-rodriguez: * Regarding `_count` columns. Ok, we can consider that the behavior you're adding are bug fixes since some users are complained about it, but this new behavior would be different since nobody has asked for it. So agreed, let's welcome this little extra complexity for now for the sake of keeping the current behavior.

* Regarding the `id_column` method, you're absolutely right, I completely overlooked that. I just applied the little patch and no feature specs broke so I incorrectly assumed it wasn't necessary.

So, agreed, let's not make any of the changes I suggested in https://github.com/activeadmin/activeadmin/pull/4784#discussion_r105568747. I :-1:'d my comment as well! :rofl:

deivid-rodriguez:
Thanks for your patience! :sweat_smile:

PR #4791, opened 2017-02-12
Fix unintentional usage of Bourbons' linear-gradient() function

### LINKED ISSUE #4780: Linear-gradient function from Bourbon is deprecated
Linear-gradient function from Bourbon is deprecated. This means that [this](https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/app/assets/stylesheets/active_admin/mixins/_gradients.scss#L4) will generate invalid css rules for BGs.

What options do we have:
1. backport this functions from Bourbon to AA mixins;
2. change function to mixin(```@include linear-gradient(...)```) which still works(for now);
3. update our gradient mixin to use plain css rule.

What say you?
P.S. I can make a PR for preferable method

mindhalt:
Fixes #4780

### DIFF
--- a/app/assets/stylesheets/active_admin/mixins/_gradients.scss
+++ b/app/assets/stylesheets/active_admin/mixins/_gradients.scss
@@ -3,7 +3,7 @@ $secondary-gradient-stop: #dfe1e2 !default;
 
 @mixin gradient($start, $end){
   background-color: $start;
-  background-image: linear-gradient(180deg, $start, $end);
+  background-image: unquote("linear-gradient(180deg, #{$start}, #{$end})");
 }
 
 @mixin primary-gradient {


deivid-rodriguez:
Could we add a comment somewhere with the reason for adding this? And maybe also that it can be removed once `activeadmin` supports `bourbon` >= 5.0 only.

mindhalt:
I don't think that we need it. Because `unquote()` is standart SASS function. It will helps us avoid compatibility errors with mixin/function libraries, such as Bourbon. Even if AA will support `bourbon >= 5.0`, it is not guaranteed that by that time there will be no other lib wich will couse similar errors, or some other gem in project.
P.S. where do use suggest to place such comment? Documentation, source code, README.md?

deivid-rodriguez:
Mmmm, so shoud we then unquote every other CSS function usage in ActiveAdmin just to prevent it being overriden? It feels wrong... If I understand this correctly, this is not causing any misbehavior, it's just a workaround to prevent a warning in bourbon, which does not happen if you upgrade to bourbon 5. 

I'd suggest adding the comment on the line above the code change.

mindhalt:
Not warning. Error. Bourbons' function is faulty. After gem update - I've lost all gradients on buttons and etc.

deivid-rodriguez:
What bourbon versions did you upgrade from and to? Just trying to understand the problem better.

mindhalt:
3.2.3 -> 3.2.4 it seams 🤣 , but I've lost gradients, as I mentioned before. I had 3-4 bourbon gem versions at that time, dont know is it relevant 🙄

deivid-rodriguez:
Seems like a bug introduced by bourbon then, no?

mindhalt:
Don't know. Bourbon doesn't have changes in this function between 3.2.3 -> 3.2.4. So maybe something changed elsewhere, SASS, asset-pipline, etc.

javierjulio:
@mindhalt what browser (platform and version) did you use? I was surprised you mentioned v3.x since Bourbon is on v5. Are you able to use v4 at all and if so does it fix the gradient issue?

We don't specify a version range for bourbon in the gemspec. Anyone know why?

mindhalt:
I've tried manualy set higher version of boubon in Gemfile.lock, but then my app started to crash with SASS errors. So I had to stick with what bundler suggested.

javierjulio:
@mindhalt its more important to know what browser (platform and version) you are using since they made changes on v3.x where they removed deprecated vendor prefixes and could be why you are not seeing gradients anymore.

deivid-rodriguez:
@javierjulio Weren't those changes made in 4.x?

@mindhalt From looking at the changes between 3.2.3 and 3.2.4: https://github.com/thoughtbot/bourbon/commits/v3.2.4, I guess the candidate culprit would be https://github.com/thoughtbot/bourbon/commit/0290b953f3c66ab0711c33bf100cce5f45c42bd9? What if you manually point to the commit prior to that one in your Gemfile (`gem 'bourbon', github: 'thoughtbot/bourbon', ref: '1c728257e99b15a8771dca7f720f87bc2523bf3a'`), you still get the problem?

mindhalt:
@javierjulio latest version Chome, Firefox (64bit) Win10.
@deivid-rodriguez I'll try and write back later. I think you could be right.

mindhalt:
@deivid-rodriguez Nope, `ref: '1c728257e99b15a8771dca7f720f87bc2523bf3a'` is broken, too. And button background looks like `background-image: linear, 180deg, #838a90, #414549;`, which is incorrect css rule syntax and must be `background-image: linear-gradient(180deg, #838a90, #414549);`

mindhalt:
`bourbon 3.2.3` seems broken too. May be not bourbon problem at all. Including diff of gems which were updated.
[Gemfile.lock.diff.txt](https://github.com/activeadmin/activeadmin/files/784958/Gemfile.lock.diff.txt)

deivid-rodriguez:
Thanks for the research. Maybe it could be the `sass 3.2.0` -> `sass 3.2.2` upgrade breaking it? https://github.com/thoughtbot/bourbon/issues/729? Can you check?

mindhalt:
`sass 3.2.0 -> sass 3.2.2` not this, because my `sass` version doesn't change `3.2.19 -> 3.2.19`. Not [thoughtbot/bourbon#729](https://github.com/thoughtbot/bourbon/issues/729), too. Tried to revert whole project to "before gem update", it helped. Now I'm more confused. Btw I gave wrong diff, I think.
[Gemfile.lock.diff.txt](https://github.com/activeadmin/activeadmin/files/785181/Gemfile.lock.diff.txt)

deivid-rodriguez:
Right, my bad, I misread your lock file. I'd try upgrading the dependencies one by one from the initial lock file until you hit the bug. You can start with top level dependencies (only three or four it seems). Once you isolate that one you can try updating only sub-dependencies of that direct dependency, and so on until you find the culprit.

deivid-rodriguez:
@mindhalt Did you have time to investigate more? If you could put together a sample project reproducing the problem I'd be happy to have a look.

mindhalt:
@deivid-rodriguez Sorry, I don't much free time now. I'll try to come back to this later this week. Too many dependencies to check 😢 (Rails version upgrade)

deivid-rodriguez:
Don't worry, we can wait! I'm very curious about what's triggering this... :)

deivid-rodriguez:
@mindhalt Friendly ping!

javierjulio:
@deivid-rodriguez I figured that we weren't using Bourbon to begin with. If anything maybe for gradient vendor prefixes but those are needed nowadays. Locally, on master I removed bourbon entirely and everything renders fine. Default styles are as they were before. I think the real solution here is to remove something we aren't using. What do you think, let me submit a PR to remove it from master now rather than with #3862?

deivid-rodriguez:
Closed by #5051.

Issue #4279, opened 2016-01-21
`display_name_method_for` works not as intended with Draper


faucct:
This is the breadcrumb I get while using Draper decorator with `delegate_all` option:
![2016-01-21 14 01 40](https://cloud.githubusercontent.com/assets/5202503/12480065/a5a0bf52-c050-11e5-8422-b56779c8fbd9.png)
ActiveAdmin can not look up the display method name correctly here:

``` ruby
      def display_name_method_for(resource)
        @@display_name_methods_cache ||= {}
        @@display_name_methods_cache[resource.class] ||= begin
          methods = active_admin_application.display_name_methods - association_methods_for(resource)
          method  = methods.detect{ |method| resource.respond_to? method }

          if method != :to_s || resource.method(method).source_location
            method
          else
            DISPLAY_NAME_FALLBACK
          end
        end
      end
```

My decorator instance does not respond to any other than `to_s` methods, as the model does. In this case it is supposed to use the fallback, but the problem is that decorator `to_s` method got a source location: `["~/.rvm/gems/ruby-2.1.5/gems/draper-2.1.0/lib/draper/delegation.rb", 10]`.

Fivell:
@faucct , what is your suggestion ? ActiveAdmin provides ```display_name_methods``` option for you and you are able configure it in your app and you're able to define any method from this list in your decorator.

faucct:
I would suggest removing `:to_s` from `display_name_methods` as it is already present in `DISPLAY_NAME_FALLBACK`.

Fivell:
@faucct , look like yes, ```:to_s```  can be removed from ```display_name_methods``` it can be removed, we have specs that it uses to_s after all https://github.com/activeadmin/activeadmin/commit/40e42b3be5be26c0f194bad58b235ec326fdf31a

timoschilling:
Why didn't you simply solve your problem by overriding `to_s` on your decorator or using `display_name`. You can use `display_name` as alias to `to_s` on your `ApplicationRecord`.

faucct:
Are you okay with everyone using decorators solving the problem on their own by copying the code from `DISPLAY_NAME_FALLBACK`?

timoschilling:
Are you ok with breaking existing behavior?

There are many simple ways to solve your problem without copying code or breaking AA.

faucct:
> Are you ok with breaking existing behavior?

[...]

PR #4792, opened 2017-02-12
Fix DisplayHelper working with decorators

faucct:
#4279

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -134,8 +134,7 @@ module ActiveAdmin
                                       :username,
                                       :login,
                                       :title,
-                                      :email,
-                                      :to_s ]
+                                      :email ]
 
     # To make debugging easier, by default don't stream in development
     setting :disable_streaming_in, ['development']
--- a/lib/active_admin/view_helpers/display_helper.rb
+++ b/lib/active_admin/view_helpers/display_helper.rb
@@ -25,13 +25,7 @@ module ActiveAdmin
         @@display_name_methods_cache ||= {}
         @@display_name_methods_cache[resource.class] ||= begin
           methods = active_admin_application.display_name_methods - association_methods_for(resource)
-          method  = methods.detect{ |method| resource.respond_to? method }
-
-          if method != :to_s || resource.method(method).source_location
-            method
-          else
-            DISPLAY_NAME_FALLBACK
-          end
+          methods.detect { |method| resource.respond_to?(method) } || DISPLAY_NAME_FALLBACK
         end
       end
 
--- a/spec/unit/view_helpers/display_helper_spec.rb
+++ b/spec/unit/view_helpers/display_helper_spec.rb
@@ -21,6 +21,8 @@ describe ActiveAdmin::ViewHelpers::DisplayHelper do
   end
 
   describe '#display_name' do
+    subject { display_name(resource) }
+
     ActiveAdmin::Application.new.display_name_methods.map(&:to_s).each do |m|
       it "should return #{m} when defined" do
         klass = Class.new do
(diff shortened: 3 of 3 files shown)


timoschilling:
There is a difference between `to_s` in the `display_name_methods` and in the fallback, `to_s` in the `display_name_methods` shows directly the value of `to_s`, the `to_s` in `DISPLAY_NAME_FALLBACK` only if the class hasn't a `model_name` or `primary_key`.

Why you didn't just remove `to_s` from the `display_name_methods` in your app?

timoschilling:
related: #3886

lib/active_admin/application.rb
```diff
@@ -134,8 +134,7 @@ def initialize
                                       :username,
                                       :login,
                                       :title,
-                                      :email,
-                                      :to_s ]
```
timoschilling: 👎

lib/active_admin/view_helpers/display_helper.rb
```diff
@@ -25,13 +25,7 @@ def display_name_method_for(resource)
         @@display_name_methods_cache ||= {}
         @@display_name_methods_cache[resource.class] ||= begin
           methods = active_admin_application.display_name_methods - association_methods_for(resource)
-          method  = methods.detect{ |method| resource.respond_to? method }
-
-          if method != :to_s || resource.method(method).source_location
-            method
-          else
-            DISPLAY_NAME_FALLBACK
-          end
```
timoschilling: 👍

timoschilling: I think removing this code results in the fact that `DISPLAY_NAME_FALLBACK` is never been used, because of every object (except BasicObject / BasicObject Subclasses) respond to `:to_s`.

```
irb(main):001:0> Class.new.new.respond_to? :to_s
=> true
irb(main):002:0> Class.new.new.method(:to_s).source_location
=> nil
```

lib/active_admin/view_helpers/display_helper.rb
```diff
@@ -25,13 +25,7 @@ def display_name_method_for(resource)
         @@display_name_methods_cache ||= {}
         @@display_name_methods_cache[resource.class] ||= begin
           methods = active_admin_application.display_name_methods - association_methods_for(resource)
-          method  = methods.detect{ |method| resource.respond_to? method }
-
-          if method != :to_s || resource.method(method).source_location
```
timoschilling: I think removing this code results in the fact that `DISPLAY_NAME_FALLBACK` is never been used, because of every object (except BasicObject / BasicObject Subclasses) respond to `:to_s`.

```
irb(main):001:0> Class.new.new.respond_to? :to_s
=> true
irb(main):002:0> Class.new.new.method(:to_s).source_location
=> nil
```

timoschilling: Context: `resource.method(method).source_location` is used to check if it is a native or script defined method.

faucct:
> There is a difference between to_s in the display_name_methods and in the fallback, to_s in the display_name_methods shows directly the value of to_s, the to_s in DISPLAY_NAME_FALLBACK only if the class hasn't a model_name or primary_key.
So we got two problems that cannot be solved at the same time: showing `"Post 13"` instead of custom `to_s` method and showing `Object#to_s`, e.g. "#<Post:0x0123456789>", instead of `"Post 13"`. I think the latter is worse and the first one can be easily fixed with using a `display_name` alias for the `to_s` method.

> Why you didn't just remove to_s from the display_name_methods in your app?
Because if I did then I would NEVER get inside `DISPLAY_NAME_FALLBACK`.

> I think removing this code results in the fact that DISPLAY_NAME_FALLBACK is never been used, because of every object (except BasicObject / BasicObject Subclasses) respond to :to_s.
No, it doesn't. Please, see the passed tests.
> Context: resource.method(method).source_location is used to check if it is a native or script defined method.
How is this connected to whether you want to see this method result in html? Maybe the better solution would be to check if the method owner is `Kernel`, though this does not solve the problem I solved in this PR.

lib/active_admin/application.rb
faucct: Why?

timoschilling: See the other comments

timoschilling:
Here is a example where your Code brakes the existing behavior https://gist.github.com/timoschilling/01eb123d9c732a68c14f1b17d9816e87.

faucct:
I understand that this behaviour changes and already stated it in the comment above, but I think that the problem I am trying to fix is more crucial than the existing behaviour which is not covered with tests.

timoschilling:
close while breaking existing behavior.

faucct:
It was partially covered by those tests, but they will still work if I will add `:to_s` to `display_name_methods`.

Fivell:
> Here is a example where your Code brakes the existing behavior https://gist.github.com/timoschilling/01eb123d9c732a68c14f1b17d9816e87.


todo: add this test to our specs

Issue #4100, opened 2015-08-26
Index filter as: text with custom ransacker and format


kaelumania:
I have a filter as textarea where I can paste structured text like `id1,id2,id3` (possibly 100 ids). Afterwards I want to filter all entries with these ids - this part works absolutely fine. But after the filtering, the textarea has the content `["id1,id2"]` which is not what I want. When I use `as: :string` for the filter, everything looks good. But a simple textfield is too small for my purpose.

``` ruby
Product.ransacker :retailer_product_id, formatter: proc {|ids| ids.split(',') } do |parent|
  parent.table[:retailer_product_id]
end

filter :retailer_product_id_in, as: :text
```

timoschilling:
You need to copy this file and build it with a textarea https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/inputs/filters/string_input.rb.

PR's are welcome

kaelumania:
If you might give me a little more information, I could be able to create a PR, e.g. where does the builder come from, which methods does it provide? Ideally I want to be able to define a formatter of the input, e.g. if the User pastes text like `id1, id2    , id3 \\n id4' I want it to normalize to `id1,id2,id3,id4`. I can achieve this partially with my custom ransacker for the query part, but for the render part, value would remain the same.

kaelumania:
The problem here is, that activeadmin uses the ransacker search object to determine the values of the filter inputs in the sidebar. For `text_field` it works due to ActiveView https://github.com/rails/rails/blob/9f44aa08636dfbba9261f4350ec14684425c4b7b/actionview/lib/action_view/helpers/tag_helper.rb#L179, but for `text_area` there is no special Array handling. One the one hand activeadmin shouldn't use the ransacker object (which returns an array instead of the provided value). On the other hand ActiveView might be inconsistent in their escape handling.

Fivell:
@kaelumania , ransackers with type formatters can be  not handled at the moment, 

Your case is resolved using ransackable scopes
Ransackable scopes do not modify types , so it will be always string. 


Example
```ruby
  scope :order_id_in, -> (str) { where("orders.id IN (?)", str.split(',')) }
   def self.ransackable_scopes(auth_object = nil)
      [:order_id_in]
  end

```


And then
```ruby
  filter :order_id_in, as: :text
```

jayshepherd:
[...]

PR #4794, opened 2017-02-15
Add filter for text_input to render text_area field

jayshepherd:
As loosely referred to in #4100 this pull request provides a proper rendering of a text area for filters. The current unit tests for a `:text` type pass and I'm not sure what would be reasonable tests to include with this particular addition.

### DIFF
--- a/lib/active_admin/inputs.rb
+++ b/lib/active_admin/inputs.rb
@@ -9,6 +9,7 @@ module ActiveAdmin
 
       autoload :Base
       autoload :StringInput
+      autoload :TextInput
       autoload :DatePickerInput
       autoload :DateRangeInput
       autoload :NumericInput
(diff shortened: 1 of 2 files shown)


Fivell:
hello, @jayshepherd , can you show screen how it looks like  ?

jayshepherd:
I will need to put together a sample project to be able to show what it looks like before and after. I'm pretty busy today, but will see what I can do.

The field was showing up as a required field and thus was being styled as a required field.

This is how I'm referencing it in my app:

```RUBY
filter :by_isbn13_in, label: 'ISBN List', as: :text, input_html: {cols: 20, rows: 3}
```

jayshepherd:
Before:
inventories freetextbooks 2017-02-15 08-17-57

After:
inventories freetextbooks 2017-02-15 08-18-32

Fivell:
@jayshepherd , please rebase

jayshepherd:
I don't rebase often, so can you help me out with how you would want me to do that?

timoschilling:
@jayshepherd I think in your case we can merge without a rebase

PR #4805, opened 2017-02-18
Order independent tests

deivid-rodriguez:
Just a WIP for now.

### DIFF
--- a/spec/unit/auto_link_spec.rb
+++ b/spec/unit/auto_link_spec.rb
@@ -1,20 +1,24 @@
 require 'rails_helper'
 
-RSpec.describe "auto linking resources" do
+RSpec.describe "auto linking resources", type: :view do
   include ActiveAdmin::ViewHelpers::ActiveAdminApplicationHelper
   include ActiveAdmin::ViewHelpers::AutoLinkHelper
   include ActiveAdmin::ViewHelpers::DisplayHelper
   include MethodOrProcHelper
 
   let(:active_admin_config)   { double namespace: namespace }
-  let(:active_admin_namespace){ ActiveAdmin::Namespace.new(ActiveAdmin::Application.new, :admin) }
+  let(:active_admin_namespace){ ActiveAdmin.application.namespace(:admin) }
   let(:post){ Post.create! title: "Hello World" }
 
-  def authorized?(*)
-    true
+  before do
+    allow(self).to receive(:authorized?).and_return(true)
   end
 
   context "when the resource is not registered" do
+    before do
+      load_resources {}
+    end
+
     it "should return the display name of the object" do
       expect(auto_link(post)).to eq "Hello World"
     end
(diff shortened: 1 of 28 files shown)


lib/active_admin/resource_collection.rb
```diff
@@ -19,6 +19,10 @@ def add(resource)
       end
     end
 
+    def delete(resource)
+      @collection.reject! { |k| k == resource.resource_name }
```
Fivell: @deivid-rodriguez ,  why not 

```
@collection.delete(resource.resource_name )
```
?

deivid-rodriguez: Absolutely!

However, this is the only change affecting the actual code in `lib` and it's not really in the scope of this PR. I only added it while debugging some failures but it's not really necessary. So I'd say it's best to just remove it.

deivid-rodriguez:
Ok, the specs now seem to be consistently passing with [random order enabled](https://github.com/activeadmin/activeadmin/pull/4805/commits/91f0ec089a6a0ba91d97991e84ba1c0a03caa495).

Tentatively removing the WIP (until Travis proves me wrong :smile:).

deivid-rodriguez:
Found one more failure locally, restoring WIP.

deivid-rodriguez:
Ok, it should be fixed now, let's wait for CI.

.travis.yml
```diff
@@ -6,6 +6,10 @@ sudo: false
 
 bundler_args: --without development
 
+branches:
+  only:
+    - master
+
```
timoschilling: don't do this

deivid-rodriguez: Could you please be a little more educational in your reviews? I can't learn anything from "don't do this"... :disappointed:

Anyways, it's now removed, the double CI run is now back there.

doublerun

deivid-rodriguez:
CI hurt, that's good. Will fix the remaining issues it surfaced as soon as possible.

Fivell:
@deivid-rodriguez , looks like this almost finished, 👍 , few bugs to go

deivid-rodriguez:
Did another pass. Now it's really really close to green.

deivid-rodriguez:
Two more fixes pushed... This time it _might_ pass.

deivid-rodriguez:
Travis passed! Yay!

Codecov is failing but it's unrelated to this PR. My guess is that the last commit was pushed directly to master, codecov didn't generate a report for it, and now it can't find a base report to compare this against.

If that's the case, one more reason to not commit directly to master! 😉

Anyways, sorry for the length of the PR, but this was a big project. Anyone up for reviewing?

timoschilling:
good job 👍

deivid-rodriguez:
Pushed an extra tiny fix. Travis will have one last opportunity to surface more issues :)

spec/unit/auto_link_spec.rb
```diff
@@ -1,60 +1,68 @@
...
-      it "should fallback to edit" do
-        url_path = "/admin/posts/#{post.id}/edit?locale=en"
-        expect(self).to receive(:url_options).and_return(locale: 'en')
-        expect(self).to receive(:url_for) { |url| url }
-        expect(self).to receive(:link_to).with "Hello World", url_path
-        auto_link(post)
-      end
+    it "should fallback to edit" do
+      expect(auto_link(post)).to \
+        match(%r{<a href="/admin/posts/\d+/edit">Hello World</a>})
```
deivid-rodriguez: Right, I accidentally removed that while refactoring. The test description was not related to locales, so I forgot to re-add it. I think those tests should be separate from the "normal case" because otherwise we don't test the "normal case". So I added be2d3004861bc12d9c682cb7ce2a359d44dbfe72, what do you think?

Fivell: @deivid-rodriguez , sure

Fivell:
@timoschilling , looks like ready to merge
@deivid-rodriguez , are you going to use parallel_specs after this ?

deivid-rodriguez:
@Fivell Yes, I did that work locally but tests wouldn't pass. I'll check again after merging this, hopefully this fixes that.

timoschilling:
Good work 🎉

Fivell:
@deivid-rodriguez ,  :1st_place_medal:

deivid-rodriguez:
Thanks!! :)

Issue #4104, opened 2015-09-01
Discuss Hound CI rules


seanlinsley:
Hi everyone, it's been a long time since I've had the time to work on AA. But here I am!

Shortly after I left, @timoschilling added Hound CI to the repo. That's nice in theory, because it prevents reviewers from having to go through and comment on each style issue. However, I think (at least) the rules we currently have enabled make contributing a very unpleasant experience.

This screenshot from #4093 seems actively hostile towards the submitter. They're greeted with a barrage of comments from a robot, minutes after submitting their PR. And for everyone, it becomes nearly impossible to have an actual conversation on the PR because it's so hard to scroll down to it.

<img height=600px src=https://cloud.githubusercontent.com/assets/688886/9594120/be670004-501c-11e5-8f6d-bce883e0c55a.png>

seanlinsley:
I actually prefer single quotes.

```ruby
# it signals that no interpolation is happening
'this is a static string'
# it works well for JSON
'{"foo": 42}'
```

But even if we were to reverse the setting in Hound, the opposite problem would still occur for those who submit PRs with double-quoted strings.

seanlinsley:
It seems like a more human-friendly option would be to have a test that fails if the code doesn't follow our preferred style. That way there would be a failing test that the developer could run locally before resubmitting their code for review.

seanlinsley:
Pinging some recent contributors: @cadetstar, @Fivell, @deivid-rodriguez, @scarver2, @codesnik, @juliogarciag, @dziemian007, @Shoroh, @mmustala, @bsvin33t, @javierjulio, @roshats

scarver2:
@seanlinsley Hi Sean. Thanks for the invite to this thread. 

I prefer single quotes too even if there is no performance gain. The implied "this a string with no embedded logic" is easily recognized with single quotes. Can that feature be disabled in HoundCI? The standard does need to be consistent whatever is chosen.

On the plus side, I like that Tom added HoundCI. It shows that we are taking ActiveAdmin's development seriously.

*More Preferable* Instead of the barrage of HoundCI notifications, can we instead configure a common`.rubocop.yml` so that a developer can test locally before submitting a commit?

deivid-rodriguez:
Hi! I agree sometimes HoundCI is too verbose.

[...]

PR #4810, opened 2017-02-19
Add rubocop

deivid-rodriguez:
I've noticed @timoschilling doing some code style comments during code reviews. We can do that automatically.

We used to have hound but it was disabled because it would flood PR's with automatic comments and because it would signal offenses not only on new code but on old code being modified.

This approach fixes both problems by using `RuboCop` directly and running it after the tests are run, both locally and in CI.


deivid-rodriguez:
Also, we should probably remove the `.hound.yml` file and the web hook, since we're not using Hound.

deivid-rodriguez:
Funny, I removed the `.hound.yml` file and I actually reenabled `Hound`. So it seems the only way to get rid of the "check" is to disable or remove the webhook, which can only be done by an admin.

@activeadmin/collaborators What do you think of this, by the way?

varyonic:
See my [previous comment on 4104](https://github.com/activeadmin/activeadmin/issues/4104#issuecomment-274336576)

deivid-rodriguez:
@varyonic Does that mean you agree on adding `rubocop` or that you suggest adding massive autocorrections to this PR? I think we should start with this and only consider autocorrecting at a later point.

varyonic:
I would like to see a full rubocop_todo.yml added using `--auto-gen-config` before reenabling rubocop, then selective fixes thereafter.

deivid-rodriguez:
That's exactly what I did in this PR. Only left to decide is what to do with `hound`. I see two options:

* Configure it to behave just like RuboCop. Basically pick up whatever `.rubocop.yml` file we decide (including TODO config). We wouldn't need to run `RuboCop` on `Travis` if we go this path.
* Remove it.

varyonic:
Ah, I see now.  I'd be tempted to configure Hound same as rubocop, but I'm OK either way.

deivid-rodriguez:
Yeah. Let's give hound another opportunity, I'll work on that. 

Also I assume we want to keep hound defaults over RuboCop defaults, right?

timoschilling:
👍

deivid-rodriguez:
Ok, so I implemented what I proposed, but it doesn't work. It seems like a hound issue: https://github.com/houndci/hound/issues/1253. What now? Should we just switch to the other proposal?

varyonic:
Looks like a recent problem, maybe they will fix soon.

deivid-rodriguez:
Maybe... Maybe not... From what I've seen while working on this, I'd revert to just running `rubocop` after our tests and remove `hound`. We can keep still keep `hound` defaults, that work is already done.

If we were to add a proper bot to do automated code review, I'd go directly with the "new cool way": HTTP://danger.systems

timoschilling:
HTTP://danger.systems nice

tiagotex:
Awesome work @deivid-rodriguez. I have been doing the same process in my company, so I will take the time to fix the Rubocop style guide violations after this PR is merged.

deivid-rodriguez:
I actually had another idea. We can use the `DisabledByDefault` option and enable rules and autocorrect offenses for them gradually. That way we don't need a `.rubocop_todo.yml` file and `hound` should work.

deivid-rodriguez:
Ok, it seems to be actually working, but the `StringLiterals` is special. Explanation here: https://github.com/houndci/hound/issues/1049. Figuring out what's best.

deivid-rodriguez:
I added a comment [here](https://github.com/houndci/hound/issues/1049#issuecomment-282305440), let's see what `hound` folks say. The B plan would be to just start by explicitly configuring the `StringLiterals` rule (or any other problematic one).

deivid-rodriguez:
I got no replies from `Hound` so far, what do you think about auto-correcting all string literals here and move on with the current approach?

deivid-rodriguez:
Auto-correcting string literals lead to a very big diff, which made hound analyze a lot of code, which surfaced another two more rules affected by the same problem (`DisabledByDefault` being ignored). What now?

* Remove hound.
* Make a decision now on those two rules as well (autocorrect or disabled).

If we were to choose the latter this I would just stick to `RuboCop` defaults (disable the `CollectionMethods` cop and use the `leading` style for `Style/DotPosition`). But I think I would just remove `hound` for now. With these bugs, it gets in the middle more than helping.

varyonic:
Yep, remove for now, what a pity.

deivid-rodriguez:
Yeah. I'll remove it tomorrow and just run `rubocop` after our tests. Next question, do we prefer the original `rubocop_todo.yml` approach or the current `DisabledByDefault` one? I don't have a strong preference.

varyonic:
I guess if you intend to implement only a few rules you use DisabledByDefault and if you intend to implement most you use rubbocop_todo.yml.  One way you are constantly adding, the other you are removing?

deivid-rodriguez:
Yes. Well, in the `.rubocop_todo.yml` you also have to deal with new rules on version updates. So sometimes you might "add" too. The `DisabledByDefault` approach has a less verbose config, but in the `.rubocop_todo.yml` one, potentially useful rules are more visible.

deivid-rodriguez:
Also (sorry for this many questions), should I leave current auto-corrections in place? I would leave them,  but I feel bad thinking of @javierjulio having to rebase #3862 on top of https://github.com/activeadmin/activeadmin/pull/4810/commits/2a344e7c38dc27e8da947b4fffc8f0257a832e60... :rofl:

deivid-rodriguez:
Hound removed and I ended up choosing the current `DisabledByDefault` configuration. I like the fact that once we enable a style, it's consistently enforced in the whole code base.

The PR currently enables 4 cops, the most problematic being `Style/StringLiterals` (because of the diff size) and `Style/DotPosition` (because I'm not sure which style we want).

deivid-rodriguez:
@varyonic I've deleted the potentially controversial changes, so this is ready to be reviewed again!

Fivell:
@deivid-rodriguez , maybe EnforcedHashRocketStyle can be added too ?

deivid-rodriguez:
@Fivell Absolutely, this is just a start, but I strongly believe we should use `EnforcedHashRocketStyle`. The reason I added these two rules to get started was:

* Style/TrailingWhitespace: Because it's a very easy, non-controversial rule, and there wasn't many offenses in the code. So it was a perfect candidate to experiment with hound without filling your inboxes with hundreds of notifications :)
* Style/SpaceAfterComma: Because it has appeared in a very recent code review: https://github.com/activeadmin/activeadmin/pull/4784#discussion_r101552378.

deivid-rodriguez:
Thanks for reviewing @Fivell!

I'm going to hold this off until we get #4848 in, that way we don't need to add `rubocop` conditionally on "rubies over 1.9".

javierjulio:
@deivid-rodriguez as you know I had a lot of trouble with this in #3862. 😂 My biggest problem with how we had Hound is that it would trigger on files were maybe there was a one line change but then the existing file in that state didn't even support any of the Hound rules so I would get lots of warnings for that one file when it wasn't applicable to my PR. Do these changes help in that regard?

deivid-rodriguez:
Well, this PR definitely helps in that regard since it removes hound :smiley:.

But I understand the problem you run into. I think it's not a problem with `hound` itself but with the fact that there's currently no "passing scss-lint configuration". So when you changed a scss line with offenses, the line would be sent to `hound` to be analyzed, and `hound` would complain. In `RuboCop`, there are easy ways to get an initial green configuration (what's implemented here). Not sure how that works in `scss-lint`. Getting this working is basically the task I've proposed in https://github.com/activeadmin/activeadmin/pull/3862#issuecomment-285334065 as a way to start reducing the size of that PR.

In any case, this PR surfaced other problems with `hound` so we decided to remove it for now. So your problems are "indirectly" fixed by this.

javierjulio:
@deivid-rodriguez thanks, I missed that about removing Hound haha. I came here from another PR you commented on and read that there was going to be a more polite version of Hound instead. Either way I'm open to where things are going. You and @Fivell are doing great as new additions on the maintainer team. 👏🏻 Keep it up! ❤️

deivid-rodriguez:
@javierjulio Thanks! To be fair @varyonic has been also added as a maintainer, and he's also doing a fantastic job, specially working on the documentation site, on external dependencies (`inherited_resources`, `arbre`) and on moving forward and documenting releases.

Regarding `hound`, yes, i tried to make it polite but the dog wouldn't shut up so I gave up on it.

Rebased on top of latest master, by the way, ready for re-review.

javierjulio:
Ah I thought @varyonic had been on it for awhile! Great all around. I'll work with you guys on #3862. Thanks.

varyonic:
> should I leave current auto-corrections in place?

So the remaining changes are not going to complicate rebase of #3862, right?

Gemfile
```diff
@@ -11,6 +11,9 @@ gem 'parallel_tests'
 # Debugging
 gem 'pry'                                   # Easily debug from your console with `binding.pry`
 
+# Code style
+gem 'rubocop'
```
deivid-rodriguez: @Fivell Thanks for having a look! I think we need it in both environments (Travis env is `test` by default, local env is `development` by default). That's why I included it globally.

Fivell: @deivid-rodriguez , but when I deploy to production I do not want gems like rubocop to be installed.

Fivell: or it is not production dependency ?

deivid-rodriguez: But this is not your application's `Gemfile`, this is the gem's `Gemfile`. Meaning, it doesn't introduce any runtime dependencies and it's only used for developing the gem. This is a common confusion, I've there too :)

deivid-rodriguez: Yes, better to ask when not sure! :wink:

deivid-rodriguez:
@varyonic @Fivell Any of you is ready to approve this?

I'm not in a hurry but I'd like to get https://github.com/activeadmin/activeadmin/pull/4810/commits/35b3bdb578ca579b809f07cabbd27615166d5433 in as soon as possible since it's creating problems in another PR's (my bad I suggested the change in the first place :flushed:).

I can open a separate PR for that, but I think this PR has been pretty much already reviewed and just needs a final check.

Fivell:
@deivid-rodriguez , I finished review few days ago.

deivid-rodriguez:
It wasn't clear to me since the PR changed after your approval and you raised more concerns. Thanks for the confirmation :)

Fivell:
@deivid-rodriguez , next time I will reconfirm approval in such cases

deivid-rodriguez:
Next time I will dismiss reviews if they get outdated :)

PR #4811, opened 2017-02-19
Don't trigger double CI builds

deivid-rodriguez:
I added this change in another PR, but was told "don't do this". Opening a separate PR for discussion.

As you can see in the status checks of other PR's of mine, there's two separate Travis checks: `continuous-integration/travis-ci/pr` and `continuous-integration/travis-ci/push`. One is triggered every time you push to a branch in the `activeadmin` repository, the other one is triggered every time you update a PR. I see no point on doing this. It slows down Travis builds and it's confusing to have duplicate Travis checks in our PR's. Note that if you work on your personal fork you don't experience this, since you are not pushing to a branch in `activeadmin`'s repository, so you get a single build for the PR.

@timoschilling Could you explain why we shouldn't do this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ sudo: false
 
 bundler_args: --without development
 
+branches:
+  only:
+    - master
+
 cache: bundler
 
 before_install:


timoschilling:
We can configure that in travis

deivid-rodriguez:
Why split CI configuration into a web UI and a configuration file? Also, I'm not sure whether the configurations are exactly the same. Whether this still preserves build merge commits into master, I think the web UI disables that.

In any case, please move forward any option you prefer.

timoschilling:
I have changed this configuration

deivid-rodriguez:
Thanks. So what's the rationale for not keeping it in `.travis.yml`?

timoschilling:
If you keep it in `.travis` it have some more other impacts

deivid-rodriguez:
> If you keep it in .travis it have some more other impacts.

Could you give an example?

deivid-rodriguez:
By the way, the UI option seems to behave indeed just like the `branches` configuration in `.travis.yml`. I still don't know what other impacts having it configured in `.travis.yml` could have, though.

deivid-rodriguez:
PR reopened since having the "Build branch updates" set to OFF in Travis seems to cause problems with `codecov` reports, and in any case, it seems bad to never get the master branch tested.

deivid-rodriguez:
Let's get this sorted out before 1.0.0. Just because it feels nice to see a green clean build right before and after releasing :)

deivid-rodriguez:
@timoschilling This has nothing to do with working with branches in forks. I work with branches in the main repo. This is about testing merge commits on the main master branch and not only PR's.

deivid-rodriguez:
Thanks! I re-enabled the setting in Travis CI!

varyonic:
> Note that if you work on your personal fork you don't experience this, since you are not pushing to a branch in activeadmin's repository, so you get a single build for the PR.
@deivid-rodriguez  I'm working on my personal fork but now Travis won't test my branches unless I reverse this (or start submitting myself PRs?)

deivid-rodriguez:
Any reason why you're working in your personal fork? I'm just curious, not that it changes anything since working on this repo exhibits the same behavior I think.

Normally when I need a branch built temporarily without a PR, I change `.travis.yml` like this:

```diff
  branches:
    only:
      - master
+     - my_branch_name
```

It works fine since I hardly ever want to do that, but if this is a common thing that we do, we should consider reverting back to the old way. Normally I just test locally and leave Travis for when I'm ready to open a PR, even if it's WIP.

timoschilling:
That's why I didn't want this change in the past ;)

deivid-rodriguez:
You never said so... But it's good to be given a reason even if late :stuck_out_tongue: 

Up to you folks, I prefer this config but I'm fine with reverting!

varyonic:
I'm trying to continue working the way I did before I had write access, same as a new contributor.  I want to see my branch pass before I open a pull request. I test locally with only the most recent Gemfile before pushing to Travis to test the rest. That it has taken a month to notice suggests I can live with the workaround, but at least I understand now what Timo was trying to remember. @Fivell?

PR #4813, opened 2017-02-20
Suggested README update.

varyonic:
Here's a draft based on conversations earlier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Active Admin
 
-Active Admin is a Ruby on Rails framework for creating elegant backends for website administration.
+[Active Admin](https://www.activeadmin.info) is a Ruby on Rails framework for creating elegant backends for website administration.
 
 [![Version         ](http://img.shields.io/gem/v/activeadmin.svg)                               ](https://rubygems.org/gems/activeadmin)
 [![Travis CI       ](http://img.shields.io/travis/activeadmin/activeadmin/master.svg)           ](https://travis-ci.org/activeadmin/activeadmin)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,76 +10,34 @@ Active Admin is a Ruby on Rails framework for creating elegant backends for webs
 
 ## State of the project
 
-### 1.0.0
+1.0.0.pre5 is the last release to support Rails 3.2 and Ruby 1.9.3.
```
deivid-rodriguez: Given that the new `inherited resources` realease supports ruby >= 2.1 only, I'd say we need to drop 2.0 here as well.

README.md
```diff
@@ -10,76 +10,34 @@ Active Admin is a Ruby on Rails framework for creating elegant backends for webs
...
-
-## Links
-
-* Website: <http://www.activeadmin.info> (out of date)
-* Live demo: <http://demo.activeadmin.info/admin>
-* Documentation
-  * Guides: <https://github.com/activeadmin/activeadmin/tree/master/docs>
-  * YARD: <http://rubydoc.info/gems/activeadmin>
-  * Wiki: <https://github.com/activeadmin/activeadmin/wiki>
+1.0.0 drops support for Rails 3.2 and Ruby 1.9.3.
```
deivid-rodriguez: I'd use future tense here.

README.md
timoschilling: 👍

README.md
varyonic: I'm thinking Timo will release both near the same time.

deivid-rodriguez: Well, for 1.0.0 we still need, at least, to actually drop support. Maybe also releasing an rc version and fixing last minute bugs. I'd say let's do one thing at a time ;)

README.md
```diff
@@ -10,76 +10,34 @@ Active Admin is a Ruby on Rails framework for creating elegant backends for webs
...
-
-*Keep in mind that during the time where we use `pre`-release label, things can break in each release!*
-
-### Rails 5
-
-Active Admin master has preliminary support for Rails 5. To give it a try, these Gemfile changes may be needed:
-
-```ruby
-gem 'inherited_resources', git: 'https://github.com/activeadmin/inherited_resources'
-```
```
deivid-rodriguez: I think we should leave the Rails 5 instructions there until we release 1.0.0, or at least until we bump the required `inherited_resources` in the gemspec file. Not necessary if we release everything at the same time. But doing that is going to be an (unnecessary) mess in my opinion.

deivid-rodriguez:
Any news here? Have we decided?

deivid-rodriguez:
Friendly ping @varyonic. Sorry for insisting but it's good to better communicate our roadmap in the README. When you have time, could you fix conflicts and update the PR?

deivid-rodriguez:
I wonder if we really want to remove all installation instructions. It might be fine, but maybe then we should move the `Getting Started` section up?

varyonic:
I reverted to an earlier original revision.  Still WIP, however.

deivid-rodriguez:
> I reverted to an earlier revision and fixed the reference to inherited_resournces for pre5.

It's looking good :+1:

I'm just missing the "getting started" section with the link to the installation instructions in the documentation site. Or at least some inline getting started instructions.

PR #4816, opened 2017-02-22
Update README.md

timoschilling:


deivid-rodriguez:
This and #4813 seem to be overlapping a bit. but I think it's good to split "doc changes" from "upgrade changes", so we could use both and keep @varyonic's for upgrade/version changes.

deivid-rodriguez:
Changes look good to me by the way, but the other PR include a few more/different changes. @varyonic?

PR #4823, opened 2017-02-23
Feature/jekyll redirects

varyonic:
Closes #4821. Add redirects from /docs/ links in old site cached in search engines.  Preview on www.activeadmin.org

PR #4824, opened 2017-02-23
Leftover order dependent failures

deivid-rodriguez:
Fixes the build failure in #4823. Fixing that triggered another order dependent failure that happened to be an actual bug in `ActiveAdmin`. It's a very pathological bug very unlikely to happen on any sane real application, but it doesn't hurt to fix it.

### DIFF
--- a/lib/active_admin/namespace.rb
+++ b/lib/active_admin/namespace.rb
@@ -194,7 +194,7 @@ module ActiveAdmin
       resources.each do |resource|
         parent = (module_name || 'Object').constantize
         name   = resource.controller_name.split('::').last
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Mmmm probably @varyonic restarted the build of the linked PR and it passed, so we no longer have logs of it. You can reproduce the failure though by running the commands stated in the commit messages. To reproduce the bug fixed by the third commit, though, you might need the second commit applied, not sure.

deivid-rodriguez:
@Fivell Do you plan to review this or should I move forward with @varyonic's blessing?

Fivell:
@deivid-rodriguez , move forward

PR #4828, opened 2017-02-26
fixes #2349, added option to read parent object when belongs_to

### LINKED ISSUE #2349: Need to be able to specify the association name with belongs_to
belongs_to needs to be able to take an option that specifies the parent's association name when declared.  For example:

```
ActiveAdmin.register Article do
   # ...
end

ActiveAdmin.register LinkedContent do
   belongs_to :article
   # ...
end
```

The auto_link will attempt to infer the link as @article.linked_contents.

If the association is named differently than the child model name, then an error is generated: 

```
  undefined method `linked_contents' for #<Article:0x007fd12814f950>
```

Suggested:

```
ActiveAdmin.register LinkedContent do
   belongs_to :article, as: :related_contents 
   # ...
end
```

Which would cause @article.related_contents to be inferred.

Fivell:
AA auto_link method was designed to build links by object automatically,
however in many cases this was inpossible because on activerecord side belongs_to has different options.
For instance

```
  belongs_to :author, class_name: 'User'
end
```

Then

```
 ActiveAdmin.register User
 ActiveAdmin.register Post { belongs_to :user }
```

breaks almost everything  because ActiveAdmin::Router calls post.user instead post.author for reading parent

accidentally activeadmin declares this case as supported for ages (take a look at features/belongs_to.feature)

To support this we can add just one more option to our belongs_to configuration

```
 ActiveAdmin.register Post { belongs_to :user, method_name: :author }
```

### DIFF
--- a/lib/active_admin/resource/belongs_to.rb
+++ b/lib/active_admin/resource/belongs_to.rb
@@ -13,9 +13,12 @@ module ActiveAdmin
 
       # The resource which initiated this relationship
       attr_reader :owner
+      # method_name to access parent object
+      attr_reader :method_name
 
       def initialize(owner, target_name, options = {})
         @owner, @target_name, @options = owner, target_name, options
+        @method_name = @options.delete(:method_name)
       end
 
       # Returns the target resource class or raises an exception if it doesn't exist
@@ -38,6 +41,10 @@ module ActiveAdmin
 
       def required?
         !optional?
+      end
+
+      def method_name
+        @method_name || @target_name
       end
 
       def to_param
--- a/lib/active_admin/resource/routes.rb
+++ b/lib/active_admin/resource/routes.rb
@@ -107,7 +107,7 @@ module ActiveAdmin
         # @return params to pass to instance path
         def route_instance_params(instance)
           if nested?
-            [instance.public_send(belongs_to_name).to_param, instance.to_param]
+            [instance.public_send(belongs_to_method).to_param, instance.to_param]
           else
             instance.to_param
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
@Fivell Two questions.

* Does using the same options `active_record` uses not work?

```
ActiveAdmin.register Post { belongs_to :author, class_name: 'User' }
```

* If that worked, would it be possible to do some introspection on the `ActiveRecord` associations if it can't be resolved from `ActiveAdmin`'s declaration? Namely, make registration without `class_name` work out of the box if the `author` association is defined on the `Post` model?

```
ActiveAdmin.register Post { belongs_to :author }
```

deivid-rodriguez:
However, the linked issue seems different... That would be about correctly inferring the `has_many` association name from `User` to `Post`, wouldn't it? In #2349 two potential solutions are proposed, but I still wonder if it would be possible for us to do some reflection on the target association and make it just work without any options.

Fivell:
@deivid-rodriguez ,  you're right I'm going to close this and just fix belongs_to features to use 
```
 belongs_to :author, class_name: 'User' 
```

Fivell:
@deivid-rodriguez , about linked issue, yes I did not notice it is about has_many, because auto_link doesn't accept retaliations at all.  

autolink can be used as
```
auto_link(child.parent)
```
but not as

```
auto_link(parent.childs)
```

Fivell:
Also this blocks #4771, because our belongs_to features are failing with

PR #4837, opened 2017-03-04
Up to date dependencies

deivid-rodriguez:
I had recently locked us to an old version of `thor` because the test/dev application generators displayed a lot of annoying warnings. Now all dependencies have fixed those warnings. The only one still unreleased is https://github.com/drapergem/draper/pull/779, but I can live with that :)


deivid-rodriguez:
JRuby + Rails 5 build was broken by the upgrade. Options:

* Stay on 5.0.1.
* Add something like `gem "rails", RUBY_PLATFORM == 'java' ? "5.0.1" : "5.0.2"`.
* Move back the `jruby + Rails 5` build under `allowed_failures`.

I'd vote for the last one. The `rails-5` branch we're using is totally unsupported and a WIP, so even if the build was previously passing, it still belongs in `allowed_failures`. Thoughts?

deivid-rodriguez:
I moved JRuby + Rails 5 back to `allowed_failures`. This is ready for review.

timoschilling:
@deivid-rodriguez If something is ready for review, please request a review

deivid-rodriguez:
Ok! If I don't have any specific target reviewer, should I request a review from all of you?

deivid-rodriguez:
For reference, here's the Rails issue: https://github.com/rails/rails/issues/28342

PR #4838, opened 2017-03-05
Fix random spec failure

deivid-rodriguez:
@Fivell This should fix the random test failure in #4784.


deivid-rodriguez:
@varyonic I added some comments. I really dislike this spec at the moment and would like to improve it. But for now it works to check the fix that added it doesn't regress, and I don't want to block other PR's because of this.

Fivell:
@deivid-rodriguez , rebase PR , lets see

Issue #2637, opened 2013-10-31
No ability to override new belongs_to navigation menu functionality


onyxraven:
Currently, it looks like there is no way to avoid the behavior of the navigation_menu being overridden to the belongs_to parent, as well as the menu item being automatically added and highlighted in the menu

That is, effectively when you have a resource with belongs_to, you get

``` ruby
ActiveAdmin.register Object do
  belongs_to :parent_object
  navigation_menu :parent_object
  menu "Objects"
end
```

This isn't evident until you visit that belongs_to controller, it adds all the resource that share the same belongs_to to the menu.

The old behavior was that if you had a belongs_to object that had no menu DSL statement, it would never be added to the menu (effectively menu false), and the menu would stay the default/global menu.

I've got a icky patch I've worked around to get 'back' to the old behavior.

``` ruby
module ActiveAdmin
  class ResourceDSL < DSL

    def run_registration_block(&block)
      #stuff before normal dsl that can be overridden
      navigation_menu :default

      super

      return unless block_given? and defined? config and config

      if config.belongs_to? and config.menu_item_options[:parent].blank?
        navigation_menu :default
        menu false
      end
    end

  end
end
```

So far, this seems to preserve the old behavior, but still enable the 'highlighting' of the menu item like previous.  I know this isn't a great solution but I hope it'll help get a config option started?

jackxu:
Yes, hope the guys who can fix it...

johnnagro:
:boat:

kaelumania:
come on :+1:

reneweteling:
👍  still an issue

tombeynon:
So far the closest I've got is to just use the following config against the resource. This leaves the default menu in place at all times, and doesn't throw an error if the parent ID param is missing. 
```
navigation_menu :default
menu false
```

Fivell:
@varyonic , it is our documentation issue, I'll do fix next days

Fivell:
@varyonic , or we should leave docs and fix AA to have same behaviour.

> In some cases (like Projects), there are many sub resources and you would actually like the global navigation to switch when the user navigates "into" a project. To accomplish this, Active Admin stores the belongs_to resources in a separate menu which you can use if you so wish. To use:

```
ActiveAdmin.register Ticket do
  belongs_to :project
  navigation_menu :project
end
```
```
ActiveAdmin.register Milestone do

  belongs_to :project
  navigation_menu :project
end
```

[...]

PR #4842, opened 2017-03-08
do not replace default menu with parent specific menu by default

### LINKED ISSUE #2739: Resources using belongs_to have broken menus in at least 2 ways
Howdy,

It seems that 0.6.2 menus are broken for resources using belongs_to.  There are two cases I've come across.

If the resource also includes `menu false` then AA raises a `ActiveAdmin::MenuCollection::NoMenuError`

If the the menu isn't disabled the page renders, but the menu only displays a single entry for the current resource type as can been seen below.

aa_menu_issue

Fivell:
[closes #2739, #2637]

### DIFF
--- a/features/belongs_to.feature
+++ b/features/belongs_to.feature
@@ -16,8 +16,8 @@ Feature: Belongs To
       end
     """
     When I go to the last author's posts
-    Then the "Posts" tab should be selected
-    And I should not see a menu item for "Users"
+    Then the "Users" tab should be selected
+    And I should not see a menu item for "Posts"
     And I should see "Displaying 1 Post"
     And I should see a link to "Users" in the breadcrumb
     And I should see a link to "Jane Doe" in the breadcrumb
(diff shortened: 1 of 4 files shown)


Fivell:
original commit that introduced this behaviour https://github.com/activeadmin/activeadmin/commit/5101f786f72e67a6ef25fe21eeff65b8fe37a69b

to replace navigation menu , it should be described directly

eq
```
ActiveAdmin.register User
      ActiveAdmin.register Post do
        belongs_to :user
        navigation_menu :user
      end
```

feature [test](https://github.com/activeadmin/activeadmin/blob/fd9d13145102e196fc4b67e5667e914364dc0967/features/belongs_to.feature#L113) exists


our documentation also shows that we should fix it 
take a look https://github.com/activeadmin/activeadmin/commit/644f80eca8f6e9c9361213633b50da6f2f36b541#diff-e5de019d679b6c40d5e8a6deea18c826R206

PR #4848, opened 2017-03-09
Drop ruby 1.9 and rails 3.2 support

deivid-rodriguez:
As we've discussed, now that pre5 is released, we can go ahead and start dropping old versions support for the final 1.0.0 release. I started with Ruby 1.9 and Rails 3.2.


spec/support/rails_template.rb
```diff
@@ -26,7 +26,7 @@ class Post < ActiveRecord::Base
       # nothing to see here
     end
 
-    unless Rails::VERSION::MAJOR > 3 && !defined? ProtectedAttributes
+    unless !defined? ProtectedAttributes
       attr_accessible :id, :title, :body, :starred, :author, :position, :published_date, :author_id, :custom_category_id, :category
     end
```
timoschilling: Wrong line indent and obsolete `end`, same for the next 5 `unless ...` removements

lib/active_admin/orm/active_record/comments/comment.rb
```diff
@@ -6,7 +6,7 @@ class Comment < ActiveRecord::Base
     belongs_to :resource, polymorphic: true
     belongs_to :author,   polymorphic: true
 
-    unless Rails::VERSION::MAJOR > 3 && !defined? ProtectedAttributes
+    unless !defined? ProtectedAttributes
```
timoschilling: `unless !` should be `if`

lib/active_admin/resource_dsl.rb
```diff
@@ -136,8 +134,7 @@ def action(set, name, options = {}, &block)
       title = options.delete(:title)
 
       controller do
-        callback = ActiveAdmin::Dependency.rails >= 4 ? :before_action : :before_filter
-        send(callback, only: [name]) { @page_title = title } if title
+        send(:before_action, only: [name]) { @page_title = title } if title
```
timoschilling: I think the `send` is no longer needed, `before_action(...){...}` should work

spec/support/rails_template.rb
deivid-rodriguez: I don't get it, @timoschilling. Could you elaborate?

spec/rails_helper.rb
```diff
@@ -19,10 +19,6 @@
 
 require 'rspec/rails'
 
-# Prevent Test::Unit's AutoRunner from executing during RSpec's rake task on
-# JRuby
-Test::Unit.run = true if defined?(Test::Unit) && Test::Unit.respond_to?(:run=)
-
```
timoschilling: This is not topic related or miss I something?

spec/support/rails_template.rb
```diff
@@ -26,7 +26,7 @@ class Post < ActiveRecord::Base
       # nothing to see here
     end
 
-    unless Rails::VERSION::MAJOR > 3 && !defined? ProtectedAttributes
+    unless !defined? ProtectedAttributes
```
timoschilling: `unless !defined?` should be `if`

spec/support/rails_template_with_data.rb
```diff
@@ -6,23 +6,17 @@
 
 inject_into_file 'app/admin/category.rb', <<-RUBY, after: "ActiveAdmin.register Category do\n"
 
-  if Rails::VERSION::MAJOR >= 4
-    permit_params [:name, :description]
-  end
+   permit_params [:name, :description]
```
timoschilling: looks like a wrong line indent, looks like 3 spaces

lib/active_admin/orm/active_record/comments/comment.rb
deivid-rodriguez: Yep, style change, but it reads much better so I'll add it! (I'll add the rule to `RuboCop` as well :wink:)

lib/active_admin/resource_dsl.rb
deivid-rodriguez: Yep, looks like the dymanic-ness is now gone, I'll try!

spec/rails_helper.rb
deivid-rodriguez: This was a hack needed because we were using `test-unit` for Rails 3.2. Definitely unrelated to the ruby 1.9 removal. I'll move the change to the right commit.

deivid-rodriguez:
@timoschilling Updated, ready for re-review! I also removed the `strong_parameters` checks, which are no longer necessary either.

lib/active_admin/orm/active_record/comments/comment.rb
timoschilling: Sounds bad, I would like to get a others maintainer opinion?
@Fivell @varyonic ?

deivid-rodriguez:
@varyonic I think it's fine, one is purely stylistic, the other one actually drops functionality. But I can squash. Your call, just let me know.

PR #4849, opened 2017-03-09
added features to test menu when nested to virtual menu item

Fivell:
try to reproduce bug or just to add more features.

see discussion https://github.com/activeadmin/activeadmin/commit/db43468f71650bdeea4498a117c7d37874894a00#commitcomment-21233668

### DIFF
--- a/features/menu.feature
+++ b/features/menu.feature
@@ -43,11 +43,30 @@ Feature: Menu
     """
       ActiveAdmin.register User
       ActiveAdmin.register Post do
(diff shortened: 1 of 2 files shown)


Fivell:
@adamthedeveloper, tests passing, so your issue is about something else

Fivell:
UPD, bug was not reproduced, so it's just about adding new feature tests 
@activeadmin/collaborators  , merge it ?

PR #4850, opened 2017-03-09
improve jruby ci by using  "--dev" flag

Fivell:
more details https://github.com/jruby/jruby/wiki/Improving-startup-time


deivid-rodriguez:
Since we are here, do we need the other flag? I'd try removing it, and if the build times remaining stable, just drop it.

Other than that, :+1:

PR #4852, opened 2017-03-10
Keep improving the test suite

deivid-rodriguez:
I started this to try clear @kaspth's path in #4558. But I added some other improvements in the tests while working on that.

Not a very focused PR, really, but I figure it's small enough to not need splitting.


lib/active_admin/application.rb
```diff
@@ -178,9 +178,9 @@ def namespace(name)
       name ||= :root
 
       namespace = namespaces[name] ||= begin
-        namespace = Namespace.new(self, name)
-        ActiveSupport::Notifications.publish ActiveAdmin::Namespace::RegisterEvent, namespace
-        namespace
```
timoschilling: 👎

deivid-rodriguez: Commit removed. Apparently my request for better feedback on code reviews didn't succeed... :cry:

deivid-rodriguez: Anyways, I think it's good to keep this change in `lib` (no matter if it's good or bad) out of this PR.

deivid-rodriguez:
Thanks for reviewing Timo!

Issue #4043, opened 2015-07-28
columns component in form with has_many


gfoiani:
I'm trying to display form fields on multiple columns.
I'm using columns and column components and has_many for related resources. 
It works for the main object (customer) but not for the related objects (vehicles and tyres).

``` ruby
form do |f|
    f.inputs 'Customer details' do
      columns do
        column { f.input :first_name, input_html: { size: 8 } }
        column { f.input :last_name, input_html: { size: 8 } }
      end
      columns do
        column { f.input :email, input_html: { size: 8 } }
        column { f.input :phone, input_html: { size: 8 } }
      end
    end
    f.inputs 'Vehicles' do
      f.has_many :vehicles, heading: false, allow_destroy: true do |v|
        panel 'Vehicle' do
          columns do
            column { v.input :plate, input_html: { size: 8 } }
            column { v.input :vendor, input_html: { size: 8 } }
            column { v.input :vehicle_model, input_html: { size: 8 } }
          end
        end
        v.has_many :tyres, heading: false, allow_destroy: true do |t|
          t.input :tyre_type, as: :select, collection: get_dictionary('Tyre', :tyre_type), include_blank: true
          panel 'Tyres' do
            columns do
              column { t.input :vendor, input_html: { size: 8 } }
              column { t.input :tyre_model, input_html: { size: 8 } }
            end
            columns do
              column { t.input :quantity, as: :string, input_html: { size: 8 } }
              column { t.input :wheel, input_html: { size: 8 } }
            end
          end
        end
      end
    end
    f.actions
  end
```

The related resources' fields are shown even if there aren't related objects in database (as you can see in the attachment).
Is this the right way to customize the form?

![form](https://cloud.githubusercontent.com/assets/1247078/8926726/4ce76312-3512-11e5-9328-f8450c4b3899.png)

timoschilling:
please post the complete resource file and the related 3 models

gfoiani:
Thanks @timoschilling,

Customer admin file:

```ruby
ActiveAdmin.register Customer do

  menu priority: 3

  permit_params :email, :first_name, :last_name, :phone,
    vehicles_attributes: [:id, :plate, :vendor, :vehicle_model, :_destroy,
    tyres_attributes: [:id, :tyre_type, :vendor, :tyre_model, :quantity, :wheel, :_destroy ] ]


  index do
    selectable_column
    column :full_name
    column :email
    column :phone
    actions
  end

  filter :email
  filter :first_name
[...]

PR #4854, opened 2017-03-10
Two-column layout to better utilize screen space

colorfulfool:
A small change to forms layout to better utilize the limited screen space of our laptops. I'm using a MacBook Air as my main computer, I'm sure many of you do too.

Note the difference in vertical space for the same form:

compare

Main fieldsets are laid out in two columns, nested forms — in however many fit on screen. (I guess, this functionality should be an optional and disabled by default but I don't know a good way to achieve that, is there one?)

### DIFF
--- a/app/assets/stylesheets/active_admin/_forms.scss
+++ b/app/assets/stylesheets/active_admin/_forms.scss
@@ -332,3 +332,22 @@ form.filter_form {
   a.clear_filters_btn { @include light-button; }
 }
 
+@media only screen and (min-width: 1170px) {
+  form {
+    fieldset.inputs {
+      display: inline-block;
+      vertical-align: top;
+
+      &:not(.has_many_fields) { // usual fieldsets -- in two columns
+        width: 49%;
+        margin-right: 1%;
+      }
+      
+      &.has_many_fields { // nested forms -- in three columns
+        width: 32%;
+        margin-right: 1%;
+      }
+    }
+    a.button.has_many_add { display: table; }
+  }
+}
--- a/app/assets/stylesheets/active_admin/_forms.scss
+++ b/app/assets/stylesheets/active_admin/_forms.scss
@@ -338,13 +338,12 @@ form.filter_form {
       display: inline-block;
       vertical-align: top;
 
-      &:not(.has_many_fields) { // usual fieldsets -- in two columns
+      &:not(.has_many_fields) { // usual fieldsets -- two columns
         width: 49%;
         margin-right: 1%;
       }
       
-      &.has_many_fields { // nested forms -- in three columns
-        width: 32%;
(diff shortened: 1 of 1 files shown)


varyonic:
Thanks Anton.  It doesn't work everywhere, but have you tried Arbre columns in your forms? e.g. #4043

colorfulfool:
I thought about `columns do`, of course. But that implies the need to go in and "design" every form. I.e. specify not only semantic blocks (`inputs do`) but also looks.

The value of an administrative framework is in not having to do that. You can drop in your models and get a reasonably good interface. If you want to "design" your pages, you will go for a fully custom solution, you're not an ActiveAdmin customer anymore.

deivid-rodriguez:
I would delegate this decision to @javierjulio, but it definitely does not belong in 1.0.0.

varyonic:
I'm not convinced to make this change.  I think it could be an unwanted surprise for some existing users.  The current default is easy to override if you don't like it.

deivid-rodriguez:
Me neither.

varyonic:
@colorfulfool Thanks for the suggestion. Use for yourself but we're going to pass on merging this for now.

deivid-rodriguez:
I'd still like @javierjulio's opinion.

javierjulio:
I saw this earlier before it was closed and thought I could get away. 😏 Thanks @deivid-rodriguez. 🙄😂 I do think its a nice addition but I don't think we should be adding anything unless absolutely necessary. This is something that an AA user can add easily (as was shown) so I think its a win. My thinking is we should be more concerned with the core CSS/JS in ActiveAdmin by updating and simplifying it. Having a nice set of variables for others to customize the appearance and then anything else others can write custom CSS for without having to override a lot of our styles with long selectors.

But since this is specific to forms I do have a lot to say about that. I wanted to do a lot more with forms regarding styling but its just brutal. Painful. I really despise the HTML Formtastic spits out. I know why it does it but its just very hard to style without breaking something else because of how the cascade works. Modern frameworks like Bootstrap, Foundation or the many others have solved this with a component based approach. The thing is that we'd also have to modify the Formtastic HTML. I'm not sure we'd want to do that and if we did I saw that being just as much as work as what I did in #3862 if not more so its why I just tried to clean up the styles for forms and not much else.

My thinking remains the same: to simplify and modernize the base CSS in AA so its easy and less painful for users to override because mainly there would be less of it and also not such long selectors to override (one was 3 id's and 1 class, in a single selector! 😳). With many variables thrown in, it becomes pretty easy to configure and anything else would require setting your own CSS in cases like this. But first before we think of doing this we really need to simplify the CSS we have so we can take on bigger projects with it, otherwise you're just applying small and large bandaids.

colorfulfool:
@javierjulio Thanks for such an expanded response, didn't expect that 👍

PR #4857, opened 2017-03-14
Add translation for "Per page" (en and ja only)

knu:
Some of our local clients seemed unfamiliar with the English phrase, so I think it needs translation.

Just for the record, our tentative solution was to use some JavaScript: `$(".pagination_per_page").contents().first().replaceWith("..translation..")`


deivid-rodriguez:
Thanks!

knu:
That was quick!  Thank you.

PR #4858, opened 2017-03-14
Add missing french translations

francois-ferrandis:
I chose to reorder the entries of `fr.yml` so that they match the order of the entries in `en.yml`, is that a reasonable practice?

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -75,6 +75,8 @@ fr:
         destroy: "Supprimer"
     comments:
       created_at: "Créé le"
+      resource_type: "Type de ressource"
+      author_type: "Profil de l'auteur"
       body: "Corps"
       author: "Auteur"
       title: "Commentaire"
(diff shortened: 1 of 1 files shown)


config/locales/fr.yml
```diff
@@ -73,14 +74,18 @@ fr:
...
+      author_type: "Profil de l'auteur"
       body: "Corps"
       author: "Auteur"
       title: "Commentaire"
       add: "Ajouter un commentaire"
+      delete: "Supprimer ce commentaire"
+      delete_confirmation: "Êtes-vous sûr de vouloir supprimer ce commentaire ?"
       resource: "Ressource"
-      resource_type: "Type de ressource"
-      author_type: "Profil de l'auteur"
```
timoschilling: why did you remove this keys?

deivid-rodriguez: He didn't, he just reordered them in bed7a64e1347914d27f0a742b1f047f484af0d75 to match the order in the English version. I'm :+1: for the change!

timoschilling: sorry didn't see that

deivid-rodriguez: No problem! 😉

PR #4867, opened 2017-03-17
restrict access if format was disabled with download_links

### LINKED ISSUE #3539: Default render of XML, CSV... is highly unsecure (even with download_links = false)

Fivell:
fixes #3539 
closes #3542

### DIFF
--- a/lib/active_admin/view_helpers/download_format_links_helper.rb
+++ b/lib/active_admin/view_helpers/download_format_links_helper.rb
@@ -2,6 +2,18 @@ module ActiveAdmin
   module ViewHelpers
     module DownloadFormatLinksHelper
 
+      def build_download_formats(download_links)
+        download_links = instance_exec(&download_links) if download_links.is_a?(Proc)
+
+        if download_links.is_a?(Array) && !download_links.empty?
+          download_links
+        elsif download_links == false
+          []
+        else
+          self.class.formats
+        end
+      end
+
       def build_download_format_links(formats = self.class.formats)
         params = request.query_parameters.except :format, :commit
         div class: "download_links" do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ index download_links: ->{ can?(:view_all_download_links) || [:pdf] }
 
 ### Security Fixes
 
+* Prevents access to formats that the user not permitted to see [#4867][] by [@Fivell][] and [@timoschilling][] 
 * Prevents potential DOS attack via Ruby symbols [#1926][] by [@seanlinsley][]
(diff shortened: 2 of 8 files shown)


Fivell:
one more options is remove ```download_links``` configuration if favour of ```download_formats``` and make it works globally per resource (now it works only for  links that are rendered within index action )

lib/active_admin/view_helpers/download_format_links_helper.rb
```diff
@@ -2,6 +2,18 @@ module ActiveAdmin
   module ViewHelpers
     module DownloadFormatLinksHelper
 
+      def build_download_formats(download_links)
+        download_links = instance_exec(&download_links) if download_links.is_a?(Proc)
+
+        if download_links.is_a?(Array) && !download_links.empty?
```
varyonic: download_links.present?

Fivell: @varyonic , this will not work, since download_links can be ```true```, in this case ```self.class.formats``` should be applied

varyonic: Ugh.  This code smells.  Some other time perhaps.

Fivell: @varyonic , what about ```case/when```, I think it will be the most readable in this situation

Fivell:
```ruby
case download_links
  when false
    then []
  when [], true
     then self.class.formats  
  else 
     download_links
end   
      
       
```

varyonic: Yes, that's better.  Could download_links be nil?

Fivell: @varyonic , since it can be result of custom block, technically it can be everything

deivid-rodriguez: Why do we need to support passing a block here, by the way? The [case we are testing](https://github.com/activeadmin/activeadmin/pull/4867/files#diff-21130d3a612b3d5292e9e08016a4e5e6R81) is a bit cumbersome and not dynamic.

Fivell: @deivid-rodriguez , https://github.com/activeadmin/activeadmin/pull/2040

deivid-rodriguez: :+1:, thanks!

PR #4868, opened 2017-03-18
FIX: now use resource as references at comment (start at #4734)

krim:
In base case migration used a separated resource_id and resource_type. And resource_id was a STRING type column. It's terrible.
Now we will use normal references for resource such as author.

### DIFF
--- a/features/comments/commenting.feature
+++ b/features/comments/commenting.feature
@@ -145,19 +145,6 @@ Feature: Commenting
     And I should be in the resource section for publishers
     And I should see "Hello World"
 
-  Scenario: Commenting on a class with string id
-    Given a tag with the name "coolness" exists
-    Given a configuration of:
-    """
-    ActiveAdmin.register Tag
-    """
-    Given I am logged in
-    When I am on the index page for tags
-    And I follow "View"
-    When I add a comment "Tag Comment"
-    Then I should see a flash with "Comment was successfully created"
-    And I should be in the resource section for tags
-
   Scenario: Commenting on an aliased resource with an existing non-aliased config
     Given a configuration of:
     """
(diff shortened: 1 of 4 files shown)


Fivell:
@deivid-rodriguez ,  looks like ready

deivid-rodriguez:
Mmmm, but this is not only a revert of #1063 but also a (partial) revert of #1050.

I'm not sure we want to revert #1050, the feature test it includes seems nice to have anyways.

deivid-rodriguez:
Wait, so are we breaking user's administration consoles here? Comments on resources with string ids should still work!

krim:
How this will work If resource has string id and comments has **integer** reference?

deivid-rodriguez:
I see, I mistakenly assumed the typecasting we were doing here was done automatically by Rails. So what we are breaking here is comment on classes with string ids that are "castable" to integers, right?

I think it's fine, people should use integer primary keys, even more if they plan to store integers in them!

@Fivell @varyonic @timoschilling Any thoughts?

Fivell:
@deivid-rodriguez, hard to say, when I read old discussion,   I didn't notice comments about uuid foreign keys, so  looks like yes it was about foreign keys that are numeric but stored as varchar.

Fivell:
@deivid-rodriguez, @varyonic , to summarize: 
I think if someone want to use non-integer primary keys they can use its own patch to do this, 
what we have now it our codebase is more like dirty hack that doesn't guarantee every feature of AA is working, so thats why I'm voting to merge this.

deivid-rodriguez:
@Fivell I agree with you, and I think we can do this before 1.0.0, just to cover our backs...

PR #4871, opened 2017-03-19
Run rubocop in ci

deivid-rodriguez:
I forgot to actually enable running RuboCop in CI in #4810... :flushed:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,8 @@ before_install:
   - rvm @global do gem uninstall bundler -a -x
   - rvm @global do gem install bundler -v 1.14.6 # latest version known to work
 
-before_script:
-  - bundle exec rake setup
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ before_script:
   - bundle exec rake setup
 
 script:
-  - bundle exec rake test
+  - bundle exec rake test lint
```
Fivell: since there is default task 
https://github.com/activeadmin/activeadmin/blob/master/Rakefile#L48

why not 

```
 bundle exec rake 
```

deivid-rodriguez: Sure, why not!

deivid-rodriguez:
@Fivell Travis setup simplified and tests green.

Issue #4873, opened 2017-03-20
Undefined method 'before_filter' for commits 9f83674 (3/18/17) and after


clairity:
For commits 9f83674 (committed on 3/18/17) and after, I get the following error (found by running `rails c` with the `pry` interpreter):
```
~/.rvm/gems/ruby-2.4.0@default/bundler/gems/activeadmin-9f83674d3f05/lib/active_admin/resource_controller.rb:62:in `<class:ResourceController>': undefined method `before_filter' for ActiveAdmin::ResourceController:Class
Did you mean?  before_build (NoMethodError)
  from ~/.rvm/gems/ruby-2.4.0@default/bundler/gems/activeadmin-9f83674d3f05/lib/active_admin/resource_controller.rb:12:in `<module:ActiveAdmin>'
  from ~/.rvm/gems/ruby-2.4.0@default/bundler/gems/activeadmin-9f83674d3f05/lib/active_admin/resource_controller.rb:9:in `<top (required)>'
  from ~/.rvm/gems/ruby-2.4.0@default/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:292:in `require'
  from ~/.rvm/gems/ruby-2.4.0@default/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:292:in `block in require'
  from ~/.rvm/gems/ruby-2.4.0@default/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:258:in `load_dependency'
  from ~/.rvm/gems/ruby-2.4.0@default/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:292:in `require'
  from ~/app/config/initializers/active_admin.rb:118:in `block in <top (required)>'
  from ~/.rvm/gems/ruby-2.4.0@default/bundler/gems/activeadmin-9f83674d3f05/lib/active_admin.rb:72:in `setup'
  from ~/app/config/initializers/active_admin.rb:1:in `<top (required)>'
  from ~/.rvm/gems/ruby-2.4.0@default/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:286:in `load'
  from ~/.rvm/gems/ruby-2.4.0@default/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:286:in `block in load'
[...]

PR #4875, opened 2017-03-20
use before_action instead of before_filter

Fivell:
refs #4873

### DIFF
--- a/lib/active_admin/resource_dsl.rb
+++ b/lib/active_admin/resource_dsl.rb
@@ -180,14 +180,20 @@ module ActiveAdmin
     delegate :before_save,    :after_save,    to: :controller
     delegate :before_destroy, :after_destroy, to: :controller
 
-    # This code defines both *_filter and *_action for Rails 4.0 to Rails 5.
-    actions = [
+    # This code defines both *_filter and *_action for Rails 4.0 to Rails 5 and  *_action for Rails >= 5.1
+    phases = [
       :before, :skip_before,
       :after,  :skip_after,
       :around, :skip
     ]
-    [:action, :filter].each do |name|
-      actions.each do |action|
+    keywords = if Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 1
+                       [:action]
+                     else
+                       [:action, :filter]
+                     end
+
+    keywords.each do |name|
+       phases.each do |action|
         define_method "#{action}_#{name}" do |*args, &block|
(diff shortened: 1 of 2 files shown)


Fivell:
@deivid-rodriguez , no this is a backwards compatible.

before_filter is old rails 3 syntax.  Is alias for before_action in rails 4.0+, 
and it is deprecated in rails 5, and dropped in rails 5.1.   no more reason to have it since we stopped support for rails 3.

Fivell:
@deivid-rodriguez , but you're right we can only change this line https://github.com/activeadmin/activeadmin/pull/4875/commits/248cbf31a96f4f074793c74e9e77e7d39cf45d11#diff-aefa2d67d2ace99bd7a17454959feabdR62

deivid-rodriguez:
Yees, that line is good, we should always use the new syntax internally.

lib/active_admin/resource_dsl.rb
```diff
@@ -180,19 +180,10 @@ def decorate_with(decorator_class)
...
-      actions.each do |action|
-        define_method "#{action}_#{name}" do |*args, &block|
-          controller.public_send "#{action}_action", *args, &block
-        end
-      end
-    end
+    # Standard rails filters
+    delegate :before_action,  :skip_before_action, to: :controller
+    delegate :after_action,   :skip_after_action,  to: :controller
+    delegate :around_action,  :skip_action,        to: :controller
```
varyonic: Keep delegating *_filter in DSL for now, right?

deivid-rodriguez:
```diff
     delegate :before_save,    :after_save,    to: :controller
     delegate :before_destroy, :after_destroy, to: :controller
 
-    # This code defines both *_filter and *_action for Rails 4.0 to Rails 5.
-    actions = [
+    phases = [
       :before, :skip_before,
       :after,  :skip_after,
       :around, :skip
     ]
-    [:action, :filter].each do |name|
-      actions.each do |action|
+
+    keywords = if Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 1
+                 [:action]
+               else
+                 [:action, :filter]
+               end
+
+    keywords.each do |name|
+      phases.each do |action|
         define_method "#{action}_#{name}" do |*args, &block|
           controller.public_send "#{action}_action", *args, &block
         end
```

Fivell: @deivid-rodriguez , nice solution!

PR #4882, opened 2017-03-25
Test support for Rails 5.1

varyonic:
Waiting for database_cleaner 1.6.0 with deprecation warning fixed.


features/support/env.rb
```diff
@@ -103,7 +103,7 @@
 end
 
 # Force deprecations to raise an exception.
-ActiveSupport::Deprecation.behavior = :raise
+ActiveSupport::Deprecation.behavior = ActiveAdmin::Dependency.rails?('< 5.1.x') ? :raise : :stderr # FIXME: need database_cleaner 1.6.0.
```
Fivell: @varyonic ,  can we change Gemfile to use  database_cleaner 1.6.0. from github ?

features/registering_assets.feature
```diff
@@ -12,8 +12,8 @@ Feature: Registering Assets
 
   Scenario: Viewing default asset files
     When I am on the index page for posts
-    Then I should see the css file "active_admin.css"
-    Then I should see the js file "active_admin.js"
+    Then I should see the css file "active_admin"
```
Fivell: @varyonic , would you please explain why we remove file extension here?

activeadmin.gemspec
```diff
@@ -30,5 +30,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'railties',            '>= 4.0', '< 5.1'
   s.add_dependency 'ransack',             '~> 1.3'
   s.add_dependency 'sass-rails'
+  s.add_dependency 'thor',                '0.19.1', '!=0.19.2', '!=0.19.3', '!=0.19.4' # http://stackoverflow.com/questions/41207432/
```
timoschilling: `'0.19.1', '!=0.19.2', '!=0.19.3', '!=0.19.4'` doesn't make sense.
`'~>0.19.1', '!=0.19.2', '!=0.19.3', '!=0.19.4'`, would make sense.
Be in general I think we should not fix third party bugs. We only should limiting version if our code has a problem with this version, not if JBuilder has a problem with this version.

features/support/env.rb
varyonic: 1.6.0 is not released yet. We can revert Gemfile to `gem 'database_cleaner',  github: 'DatabaseCleaner/database_cleaner'` as it was last summer after Rails 5.0.

features/registering_assets.feature
varyonic: Because I removed asset.digest = false, so the css file is now something like active_admin.87a9...98ef.css

activeadmin.gemspec
varyonic: Is this better?
```
  s.add_dependency 'jbuilder',            '>= 2.6.3' # jbuilder#372
```

deivid-rodriguez: Is jbuilder a direct runtime dependency of AA or does this fix only a problem in our tests? If the latter, the dependency should be added to the Gemfile, or even better, to lock files committed to version control.

varyonic: Only impacts rake:setup and is an unclear warning rather than an error.  I'm inclined to omit this.

features/support/env.rb
Fivell: @varyonic , basically  I'm ok with both options, but we should somehow not forget to return  ```ActiveSupport::Deprecation.behavior```

activeadmin.gemspec
deivid-rodriguez: Yeah, let's do that. I need to find time to propose the addition of lock files to the repo, but it definitely doesn't belong here.

varyonic:
Rebased and updated for review.

varyonic:
@timoschilling Shall I add to CHANGELOG?

.travis.yml
```diff
@@ -9,6 +9,8 @@ bundler_args: --without development
 cache: bundler
 
 before_install:
+  - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install 4
+  - npm install -g yarn
```
deivid-rodriguez: Do we need `nvm` and `yarn` in CI now?

varyonic: Yes, asset compilation will fail without it as they are prerequistes for [webpacker](https://github.com/rails/webpacker).

deivid-rodriguez: But I thought this was an opt-in feature and I don't see the `webpacker` gem specified anywhere...

In any case, I still don't understand why we need `nvm` and why we use it to install... nodejs v4. `npm` is available in Travis already, and I'd say that's the only thing you need to install `yarn`.

varyonic: Feel free to offer a better solution.

deivid-rodriguez: My try would be to remove those two lines, since I don't think they are needed. It's very surprising to me that the build fails without them.

varyonic: The build did not work before those two lines.  Try for yourself.

deivid-rodriguez: I believe you, I just don't understand why. Anybody?

varyonic: Appears I was wrong.  They appeared necessary before but no longer.  Will push an update.

.gitignore
```diff
@@ -47,3 +47,4 @@ public
 .rbenv-version
 .localeapp/*
 lib/bug_report_templates/tmp
+gemfiles/bin
```
deivid-rodriguez: What about this? I wouldn't expect any binstubs generated here. Maybe it's related to the old setup with `yarn` and `nvm`?

features/registering_assets.feature
deivid-rodriguez: Since our rough conclusion was that `webpacker` is opt-in and we're not using it, I have the feeling that the old setup might actually work. Should we double check this?

PR #4883, opened 2017-03-25
fixes Issue #4424

### LINKED ISSUE #4424: new_with_cast no longer works (regression)

Fivell:
fixes #4424

adds tests for rebased #4680


faucct:
This change has broken this feature - https://github.com/activeadmin/activeadmin/pull/4166. What do you think about leaving the assign_attributes piece and limiting the resource params to STI column name so the virtual attributes can be introduced in a decorated model?

Fivell:
@faucct unfortunately #4166 was merged without any spec, so sooner or later such features become broken, it would be great help if you could provide failing test for what you are talking about

PR #4889, opened 2017-03-29
Remove empty rows generated in migration

Neodelf:


deivid-rodriguez:
Thanks! Do you think we can use `<%-` so we can keep our template readable as well?

lib/generators/active_admin/install/templates/migrations/create_active_admin_comments.rb.erb
```diff
@@ -1,9 +1,6 @@
-<%
-parent_class = ActiveRecord::Migration
+<%parent_class = ActiveRecord::Migration
 parent_class = parent_class[5.0] if Rails::VERSION::MAJOR >= 5
-%>
-
-class CreateActiveAdminComments < <%= parent_class.to_s %>
+%>class CreateActiveAdminComments < <%= parent_class.to_s %>
```
timoschilling: use this style:
```erb
<%
parent_class = ActiveRecord::Migration
parent_class = parent_class[5.0] if Rails::VERSION::MAJOR >= 5
-%>
class CreateActiveAdminComments < <%= parent_class.to_s %>
```
This should produce the same result.

Neodelf:
Thanks for the quick feedback

deivid-rodriguez:
I'm not able to actually try this myself, but looks good. If either of you have confirmed that this works, I'm happy to see this merged!

PR #4900, opened 2017-04-07
Add missing translation for Password Confirmation field

danielricecodes:
Add missing translation for Password Confirmation field.

When using Devise's Registerable module, you'll see this on the login page right now.

```html
<span class="translation_missing" title="translation missing: en.active_admin.devise.password_confirmation.title">Title</span>
```

Adding the missing English translation here!  I am not a translator so I'll leave the additional language translation to others.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -98,6 +98,8 @@ en:
         title: "Subdomain"
       password:
         title: "Password"
+      password_confirmation:
+        title: "Confirm Password"
       sign_up:
         title: "Sign up"
         submit: "Sign up"


Fivell:
@timoschilling , should we include localisation for each device module? maybe there is another way to do this?

deivid-rodriguez:
I guess we could just not provide these and advice users who want i18n to install the `devise-i18n` gem...

timoschilling:
I think we should not merge this and should start using the devise translations

Fivell:
@timoschilling sort of [this](https://github.com/tigrish/devise-i18n) ?

timoschilling:
@Fivell yes

deivid-rodriguez:
I kind of changed my mind regarding this. We're already translating devise inside AA, so I think this is a good change.

I'm in favor of dropping these translations and delegating them to `devise-i18n`, but this would be a backwards incompatible change, so I'd merge this and them remove all devise translations and recommend `devise-i18n` at a later stage (post 1.0.0).

danielricecodes:
Thanks for merging this!

deivid-rodriguez:
Hei! Any more opinions on this? I think we should merge for now and think about the process of dropping `devise` translations later.

timoschilling:
@deivid-rodriguez I agree with you

PR #4906, opened 2017-04-11
simplify ActiveAdmin::Filters::Active

Fivell:
since we removed support for old rails versions we can simplify this since to_unsafe_h returns
Hash for rails <= 4.2 and HashWithIndifferentAccess for rails > 4.2

### DIFF
--- a/lib/active_admin/filters/active.rb
+++ b/lib/active_admin/filters/active.rb
@@ -8,23 +8,15 @@ module ActiveAdmin
 
       def initialize(resource_class, params)
         @resource_class = resource_class
-        @params = normalize_params(params)
+        @params = params.to_unsafe_h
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Maybe relevant here? https://github.com/rails/rails/pull/28734

Fivell:
@deivid-rodriguez , don't think so ,  this is more  just about removing ```normalize_params``` method , we don't need it anymore, and we can work with simple hash now in this class internally

deivid-rodriguez:
Ok. So using this `to_unsafe_h` here is cool, since this is just for GET actions and we don't need explicit whitelisting here, right?

I know the method is already there and this is just a straightforward refactoring. Just wondering about this.

PR #4919, opened 2017-04-18
Add tabs documentation

jstumbaugh:
On the [Arbre Components](https://activeadmin.info/12-arbre-components.html) page, there was no documentation for [tabs](https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/lib/active_admin/views/components/tabs.rb), so I added some.

EDIT: It was [pointed out](https://github.com/activeadmin/activeadmin/pull/4919#issuecomment-295824610) that there is [existing documentation](https://github.com/activeadmin/activeadmin/blob/6fe4013d3463135183b50f73fd8d5113c4ca89c4/docs/6-show-pages.md#tabs) for Tabs. I removed it in favor of having it on the Arbre Components page. I also added a link for the Table of Contents to the new Tabs documentation.

### DIFF
--- a/docs/12-arbre-components.md
+++ b/docs/12-arbre-components.md
@@ -151,3 +151,26 @@ status_tag 'active', :ok
 status_tag 'active', :ok, class: 'important', id: 'status_123', label: 'on'
 # => <span class='status_tag active ok important' id='status_123'>on</span>
 ```
+
+## Tabs
+
+The Tabs component is helpful for saving page real estate. The first tab will be
+the one open when the page initially loads and the rest hidden. You can click
+each tab to toggle back and forth between them. Arbre supports unlimited number
+of tabs.
+
+```ruby
+tabs do
+  tab :active do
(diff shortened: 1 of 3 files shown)


Fivell:
@jstumbaugh , there is already documentation for tabs here https://github.com/activeadmin/activeadmin/blob/6fe4013d3463135183b50f73fd8d5113c4ca89c4/docs/6-show-pages.md#tabs. Any thoughts about what is better to put it in different pages or add link from one page to another?

jstumbaugh:
@Fivell Considering it is an Arbre Component, I think its documentation is best served on that page. I think it would be better to remove it from [Customizing the Show Page](https://github.com/activeadmin/activeadmin/blob/6fe4013d3463135183b50f73fd8d5113c4ca89c4/docs/6-show-pages.md) altogether, as it really has nothing to do with that and doesn't say much about it.

PR #4922, opened 2017-04-21
Avoid method missing

### LINKED ISSUE #4897: Crash on new resource page
This seems to have been introduced by #4477.

Posting here reproduction instructions for @Fivell :wink: 

* git clone https://github.com/EnComu/participa
* cd participa
* Follow development setup instructions [here](https://github.com/EnComu/participa/#instalaci%C3%B3n). I'm sorry we don't have something fully automatized. There's also a docker setup, but it hasn't been used in a while, you might want to try that as well. This is a standard Rails application.
* After setup, edit Gemfile to stop pinning activeadmin to the revision prior to #4477, and rebundle.
* Start the application, login with the test admin account (user: admin@example.com, pass: 111111), and visit http://localhost:3000/admin/verification_centers/new.

The app should crash.

I'm sorry I don't have the time to provide a more isolated example at the moment...

deivid-rodriguez:
Closes #4897.

I only enabled the `Style/ExtraSpacing` RuboCop cop, because there was a double whitespace in that same line that was bothering me :sweat_smile:.

I figured the change was small and simple enough (just whitespace) to be shipped in the same PR, but let me know if you want commits splitted into different PR's.

### DIFF
--- a/lib/active_admin/views/components/active_admin_form.rb
+++ b/lib/active_admin/views/components/active_admin_form.rb
@@ -78,7 +78,7 @@ module ActiveAdmin
       end
 
       def add_create_another_checkbox
-        if %w(new create).include?(action_name) &&  active_admin_config && active_admin_config.create_another
+        if %w(new create).include?(helpers.action_name) &&  active_admin_config && active_admin_config.create_another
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Done!

PR #4923, opened 2017-04-21
Commit lock files

deivid-rodriguez:
After CI failed in #4922 because CI environment does not match my local environment, I decided to address @timoschilling's request in https://github.com/activeadmin/activeadmin/pull/4624#discussion_r98179872.

This is just another example of environment mismatch getting in the middle.

Another one could be https://github.com/activeadmin/activeadmin/pull/4882#discussion_r108101750. In that PR @varyonic lost time fighting with some warnings I had already dealt with in https://github.com/activeadmin/activeadmin/commit/927f188da2bfa40917213ecb0ed88c35109f1886. If we had have lock files in version control, I would've been able to communicate "the fix" by updating the gems in the lockfiles with the warnings fixed, and @varyonic might not have even seen them at all.

In general, I feel this keeps getting in the middle, so we should maintain our lock files manually.


deivid-rodriguez:
Fair enough, done in #4929.

Fivell:
related 
https://github.com/rails/rails/commit/b1edc3789a39913b889b3ab18039ecca17835df3#diff-e79a60dc6b85309ae70a6ea8261eaf95 


>> In our CI server we delete the lock file to make sure we are always
testing against the newest release of our dependencies.

@deivid-rodriguez . any idea why they need this and we don't? 


also nice one http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

deivid-rodriguez:
> @deivid-rodriguez . any idea why rails needs this and we don't?

They don't "need" it, but they want to make sure the test against the latest versions. I prefer to proactively maintain to `Gemfile.lock`.

> also nice one http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

That article has been mentioned several times. I don't fully agree with it. I think the predictability and consistency between CI and all developers is more important that "exploring" new dependencies in search for breaking changes. And if your dependency is important enough, you should keep separate gemfiles and test against different versions.

deivid-rodriguez:
> I'm undecided on this. Yes, locking these down means less surprises, but they will churn a lot or get outdated.

I understand. I'll close for now and maybe I'll revisit if I manage to convince the bundler team to recommend this workflow :stuck_out_tongue:

PR #4926, opened 2017-04-21
Officially support MRI 2.1 by testing against it in Travis CI

deivid-rodriguez:
Before 1.0.0, we should decide what to do with MRI 2.1 support. There's 2 options:

* Drop it. I vote for this since [it's been EOL'd](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/) and dropping it gets us in sync with Rails support range.

* Support it. In that case, we should start testing it in CI to ensure AA actually works against this version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ script:
 
 rvm:
   - jruby-9.1.8.0
+  - 2.1.10
   - 2.2.7
   - 2.3.4
   - 2.4.1
@@ -35,6 +36,12 @@ env:
 
 matrix:
   fast_finish: true
+
+  exclude:
+    - rvm: 2.1.10
+      gemfile: gemfiles/rails_50.gemfile
+    - rvm: 2.1.10
+      gemfile: gemfiles/rails_51.gemfile
 
   allow_failures:
     - rvm: jruby-9.1.8.0


Fivell:
@deivid-rodriguez , I didn't expected this, using 1.0.0. pre versions on many projects, some of them are on ruby 2.1 and are in support phase, would be great if I can update them with AA 1.0.0 :)

Fivell:
@deivid-rodriguez , also we merged before to our readme
```
1.0.0 will drop support for Rails 3.2 and Ruby 2.0 or earlier.
```

nothing about 2.1 .


so my suggestion is to drop it after 1.0.0 relase

deivid-rodriguez:
> @deivid-rodriguez , I didn't expected this, using 1.0.0. pre versions on many projects, some of them are on ruby 2.1 and are in support phase, would be great if I can update them with AA 1.0.0 :)

In my opinion, once a Ruby version stops receiving severe security fixes and is EOL'd, it's always more urgent to upgrade that than to upgrade AA.

> @deivid-rodriguez , also we merged before to our readme
> ```
> 1.0.0 will drop support for Rails 3.2 and Ruby 2.0 or earlier.
> ```
> nothing about 2.1 .

Of course, we need to update the README if we go this path.

Let's hear what @timoschilling and @varyonic say.

varyonic:
I'm with @Fivell on this.

deivid-rodriguez:
Ok! PR updated to start testing it in CI.

PR #4918, opened 2017-04-18
Add Chinese translations of unsupported broswer

zhaqiang:

PR #4931, opened 2017-04-22
Enable mention-bot for translations

deivid-rodriguez:
I just had a look at #4918.

Obviously we don't speak all the languages `activeadmin` supports. So I thought we could use [mention-bot](https://github.com/facebook/mention-bot) for this.

With this configuration, once we label a PR with the `i18n` label, the mention bot will scan previous contributors to the translation and ping them. Theoretically, since I've never actually used it.

Do you want to give it a try?

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -1,0 +1,9 @@
+{
+  "message": "@pullRequester, thanks for contributing an activeadmin
+              translation! @reviewers, are you able to review this?",
+  "findPotentialReviewers": true,
+  "actions": ["labeled"],
+  "createComment": true,
+  "withLabel": "i18n",
+  "delayed": false
+}


deivid-rodriguez:
Ok. I added this because I thought your manual effort was a good idea and might actually work well. And keeping translations under a label does no harm anyways.

Should I close this then?

deivid-rodriguez:
@Fivell @timoschilling What do you think? Shall I just close this?

deivid-rodriguez:
I'm going to merge this with @timoschilling's approval. If we find it useless, we can remove it at a later stage.

@activeadmin/collaborators From now on, make sure you tag translation PR's with the `i18n` label. If this PR works, previous translators should be automatically notified by the "mention-bot".

PR #4932, opened 2017-04-22
I18n tasks

deivid-rodriguez:
Supersedes #4310.

I added a separate PR since #4310 was very outdated and the branch seemed deleted.

* Didn't enable Google Translate since I think "no translations" is better than "bot translations".
* Didn't enable "missing translation" checks since we have a lot of missing translations.
* Did enable "unused translation" checks and fixed all "offenses". This needs careful reviewing since guessing this is complicated and there might be false positives.
* Had to disable `simplecov` in local runs since due to a bug, the spec running `i18n-tasks` is extremely slow (see https://github.com/colszowka/simplecov/issues/404, https://github.com/glebm/i18n-tasks/issues/221).


deivid-rodriguez:
Ping! @varyonic?

PR #4933, opened 2017-04-22
Remove unnecessary css class

### LINKED ISSUE #2930: Rename class name of resource
Hello,

ActiveAdmin automatically adds the classname of an resource to all elements, for example: class="attributes_table section".

My Model has the name "Section" and a class of ActiveAdmin also so i get a lot of formatting problems with this resource. Its possible to rename a class name or to prevent ActiveAdmin of adding this classes?

bildschirmfoto 2014-02-05 um 09 01 33

i also tried to rename the resource with as: "", but the classnames always take the name of the models.

deivid-rodriguez:
Should fix #2930?

In a quick look didn't see any usage of this class. If there's any, we should just use the "panel" class instead... I guess.

Post 1.0.0, of course.

### DIFF
--- a/app/assets/stylesheets/active_admin/components/_panels.scss
+++ b/app/assets/stylesheets/active_admin/components/_panels.scss
@@ -1,5 +1,5 @@
 // ----------------------------------- Helper class to apply to elements to make them sections
-.section, .panel{ @include section; }
+.panel{ @include section; }
 
 // ----------------------------------- Sidebar Sections
 


deivid-rodriguez:
I just show [this comment](https://github.com/activeadmin/activeadmin/issues/2930#issuecomment-62893673) by @timoschilling... Closing.

@timoschilling Did you end up opening that issue in Arbre?

varyonic:
I could not find the Arbre issue, but this CSS class is still redundant and can be removed, no?

deivid-rodriguez:
Mmmm I don't think so, unless all the elements with the 'section' class also have the 'panel' class... Right?

PR #4940, opened 2017-04-25
Attributes table title

ajw725:
Allow a custom title given with `attributes_table` for the panel wrapping the actual attributes table (replaces the default "Object Details" title). Accepts options hash with either `:title` or `:header` key, and takes string, method name, or proc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,13 @@
 # Changelog
+
+## Master (unreleased)
+
+### Enhancements
+
+##### Minor
+
+* Allow custom panel title given with `attributes_table` [#4940][] by [@ajw725][]
+
 ## 1.0.0 [☰](https://github.com/activeadmin/activeadmin/compare/v0.6.3...master)
 
 ### Breaking Changes
--- a/lib/active_admin/views/pages/show.rb
+++ b/lib/active_admin/views/pages/show.rb
@@ -25,7 +25,12 @@ module ActiveAdmin
         end
 
         def attributes_table(*args, &block)
-          table_title = ActiveAdmin::Localizers.resource(active_admin_config).t(:details)
+          opts = args.extract_options!
+          table_title = if opts.has_key?(:title)
+            render_or_call_method_or_proc_on(resource, opts[:title])
+          else
+            ActiveAdmin::Localizers.resource(active_admin_config).t(:details)
+          end
           panel(table_title) do
             attributes_table_for resource, *args, &block
           end
--- a/features/show/attributes_table_title.feature
+++ b/features/show/attributes_table_title.feature
@@ -1,0 +1,54 @@
+Feature: Show - Attributes Table Title
+  
+  Modifying the title of the panel wrapping the attributes table
+  
+  Background:
+    Given a post with the title "Hello World" written by "Jane Doe" exists
+  
+  Scenario: Set a string as the title
+    Given a show configuration of:
+    """
+      ActiveAdmin.register Post do
+        show do
+          attributes_table title: "Title From String"
+        end
+      end
+    """
+    Then I should see the panel title "Title From String"
+    And I should see the attributes table
+  
+  Scenario: Set a method to be called on the resource as the title
+    Given a show configuration of:
+    """
+      ActiveAdmin.register Post do
+        show do
+          attributes_table title: :title
+        end
+      end
+    """
+    Then I should see the panel title "Hello World"
+    And I should see the attributes table
+  
+  Scenario: Set a proc as the title
+    Given a show configuration of:
+    """
+      ActiveAdmin.register Post do
+        show do
+          attributes_table title: proc{ |post| "Title: #{post.title}" }
+        end
+      end
+    """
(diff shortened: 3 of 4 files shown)


lib/active_admin/views/pages/show.rb
```diff
@@ -25,7 +25,11 @@ def main_content
         end
 
         def attributes_table(*args, &block)
+          opts = args.extract_options!
           table_title = ActiveAdmin::Localizers.resource(active_admin_config).t(:details)
+          if given_title = opts[:title] || given_title = opts[:header]
```
Fivell: @ajw725 , why not to keep only one possible option ?

Fivell:
@ajw725 too much commits for adding and option, can you squash them to single one please

lib/active_admin/views/pages/show.rb
```diff
@@ -25,7 +25,11 @@ def main_content
         end
 
         def attributes_table(*args, &block)
+          opts = args.extract_options!
           table_title = ActiveAdmin::Localizers.resource(active_admin_config).t(:details)
```
Fivell: I think ActiveAdmin::Localizers.resource(active_admin_config).t(:details) should work as fallback if ```: title``` options was not passed

lib/active_admin/views/pages/show.rb
ajw725: No good reason, I guess. I'd have been inclined to use only `:title`, but I think I saw one or two people mention `:header`, so I figured I'd have both. Do you have a preference for one or the other?

Fivell: @ajw725 ,  looks like title is more preferable https://github.com/activeadmin/activeadmin/blob/9739ac47dc4dd7f821cdf84adf2dd3afed536942/lib/active_admin/views/pages/show.rb#L11

ajw725:
@Fivell , will squash commits. Also realized one of my tests is not really written properly, so going to fix that and update branch

lib/active_admin/views/pages/show.rb
ajw725: Are you suggesting a change to this, or just saying you think it's ok?

Fivell: @ajw725 , I think it should be conditional something like
```
table_title = if opts.has_key?(:title)
  render_or_call_method_or_proc_on(resource, opts[:title])
else
  ActiveAdmin::Localizers.resource(active_admin_config).t(:details)
end
```

ajw725:
@Fivell , I've updated the conditional assignment of `table_title` as per your suggestion, and also squashed the commits into one.

CHANGELOG.md
```diff
@@ -56,6 +56,7 @@ index download_links: ->{ can?(:view_all_download_links) || [:pdf] }
 ```
 * Comments menu can be customized via configuration passed to `config.comments_menu` [#4187][] by [@drn][]
 * Added `config.route_options` to namespace to customize routes [#4467][] by [@stereoscott[]]
+* Allow custom panel title given with `attributes_table` [#4940][] by [@ajw725][]
```
deivid-rodriguez: You need to add it at the top in a new `## Master (unreleased)` section since this is the first change in master after 1.0.0 release. Sorry for not mentioning it!

lib/active_admin/views/pages/show.rb
```diff
@@ -25,7 +25,12 @@ def main_content
         end
 
         def attributes_table(*args, &block)
-          table_title = ActiveAdmin::Localizers.resource(active_admin_config).t(:details)
+          opts = args.extract_options!
+          table_title = if opts.has_key?( :title )
```
deivid-rodriguez: The rubocop `Style/PredicateName` cop would complain about this if we enabled it (expecting us to use `key?` instead), but I actually like how `has_` predicate methods read.

@varyonic @Fivell Thoughts? I could enable the cop it for `is_` and `have_` methods only,

Fivell: @deivid-rodriguez , we can fix it with additional commit while adding cop.

deivid-rodriguez: Yes, I'm not asking for any changes in this PR, just taking the chance to discuss the style we like. :)

Fivell: @deivid-rodriguez  
>> could enable the cop it for is_ and have_ methods only,

I agree

CHANGELOG.md
ajw725: Ah, of course, sorry. Makes sense. Fixed

deivid-rodriguez:
@ajw725 Could you rebase on top of latest master and fix the tiny style issue that will be surfaced?

lib/active_admin/views/pages/show.rb
varyonic: Fine by me.

deivid-rodriguez: I added it in #4942.

ajw725:
@deivid-rodriguez all set now.

deivid-rodriguez:
Thanks!!

Issue #4132, opened 2015-09-22
Speed up scope counts


seanlinsley:
In the case that you have a complex query applied to your index page (perhaps through a filter) that takes a long time, say 150 ms, you can end up losing around a second to scope counts. This happens because for each scope (the page in question had five), we perform a database request.

Postgres 9.4 supports an easy way to request these counts without performing multiple database requests: [filters](http://blog.2ndquadrant.com/the-within-group-and-filter-sql-clauses-of-postgresql-9-4/).

For example, given 5 users who are enabled and 7 users are disabled:
```sql
select
  count(*) filter (where status = 'enabled') as enabled_count,
  count(*) filter (where status = 'disabled') as disabled_count
from users;
 enabled_count | disabled_count 
---------------+----------------
             5 |              7
```

The downside is that no other database yet supports it (it is part of the SQL standard though).

Questions:
- how do we reliably detect that Postgres 9.4 is being used?
- given an Active Record scope, is there any way to extract its SQL so we can inject it into a `filter`?

timoschilling:
Can you provide some benchmarks for a possible improvement?

varyonic:
Would scope counts be a good place to use AJAX?  Perhaps a plugin like @holyketzer's async_panel?

holyketzer:
I suppose this optimized query doesn't give meaningful performance boost. It's complexity should be a sum of complexity of it's sub-queries, so your profit is only twice decreased network latency to the database, it's a very small profit.

seanlinsley:
@holyketzer while doing all the calculations as one query does incur the cost of the calculations, it's still faster than doing them separately because the database only has to load the records once, then can do the calculations while they're in memory. The gains multiply as you add more and more scopes.

In this example, 1 combined query is twice as fast as 6 sepearate queries: (1 second vs 2 seconds)

```sql
-- a table with 10 million rows
create table foo as select id from generate_series(1, 10000000) id;

SELECT 10000000
Time: 4908.401 ms

select count(*) from foo;
  count   
----------
 10000000
(1 row)

Time: 329.191 ms

select count(*) from foo where id % 2 = 0;
  count  
---------
 5000000
(1 row)

Time: 382.320 ms

select count(*) from foo where id % 2 = 1;
  count  
---------
 5000000
(1 row)

Time: 357.944 ms

select count(*) from foo where id % 3 = 0;
  count  
---------
 3333333
(1 row)

Time: 345.678 ms

[...]

PR #4948, opened 2017-04-28
Enable caller to pass a scope count to `scope` via `show_count`

Empact:
This is useful in situations where there are multiple scopes and the count
can be generated for all of them in a single query.


Empact:
I'm closing this as it's probably not broadly useful enough to justify being in the library.

Here's a reference of how this came up: https://stackoverflow.com/questions/26530016/get-around-the-activeadmindatabasehitduringload

varyonic:
Hmm, see also #4132

PR #4970, opened 2017-05-22
Add issue template.

varyonic:
See e.g. https://github.com/electron/electron/issues/new

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,26 @@
+<!--
+Thanks for opening an issue! A few things to keep in mind:
+
+- The issue tracker is only for bugs and feature requests.
+- If you need general help please post to StackOverflow http://stackoverflow.com/questions/tagged/activeadmin
+- Before reporting a bug, please search all issues, closed and open, and try reproducing your issue against
+  the latest version.
+-->
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+
+<!-- What actually happens? -->
+
+### How to reproduce
+
+<!-- 
+
+Your best chance of getting this bug looked at quickly is to provide a **code sample** or an **executable test case** demonstrating the expected behavior that is not occurring.
+
+See CONTRIBUTING for more details.
```
deivid-rodriguez: Maybe we could add direct links to the CONTRIBUTING file and the bug report template script.

varyonic: GitHub already adds a link to CONTRIBUTING ('the guidelines') in the flash message displayed when you select 'New Issue'  I'm not sure how useful non-clickable links are when defaulted in the text input.  I guess you can click on them if you switch to 'Preview', is that the idea?

deivid-rodriguez: @varyonic You're absolutely right! This is indeed viewable in the preview screen but I'm not sure either of the usefulness. And yes, the pop up contributing link seems enough!

PR #4978, opened 2017-06-01
Add filter predicates translations

francois-ferrandis:
This adds two translations for datetime filter predicates


Fivell:
@francois-ferrandis , thanks

francois-ferrandis:
@Fivell Well, that was easy! Thank you for reviewing.

PR #4983, opened 2017-06-03
Remove unnecessary br tags

javierjulio:
These `<br>` tags cause excessive bottom padding in the login panel (logged_out view). We want to be able to control spacing through CSS. This is pulled out from #3862.


deivid-rodriguez:
Happy to see this started!

PR #4987, opened 2017-06-03
Wrap resource_selection_toggle_cell in a label tag for greater hit area

javierjulio:
This way the text itself is also clickable and toggles the checkbox. No need for additional div wrappers either. This is being pulled out of #3862.

### DIFF
--- a/lib/active_admin/batch_actions/views/selection_cells.rb
+++ b/lib/active_admin/batch_actions/views/selection_cells.rb
@@ -7,7 +7,7 @@ module ActiveAdmin
     class ResourceSelectionToggleCell < ActiveAdmin::Component
       builder_method :resource_selection_toggle_cell
 
-      def build(label_text='')
+      def build(label_text = '')
         label do
           input type: "checkbox", id: "collection_selection_toggle_all", name: "collection_selection_toggle_all", class: "toggle_all"
           text_node label_text if label_text.present?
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
We should get rid of the styles as well: https://github.com/activeadmin/activeadmin/blob/5390b88ed0d26e0fbffde3bd699052e2b8a648a1/app/assets/stylesheets/active_admin/components/_batch_actions.scss#L6-L11. Right?

javierjulio:
@deivid-rodriguez sure, I just thought I'd let the css branch handle that but I made the change anyway since it failed and just force pushed. I misread the code and didn't see how the label text was being used. I should have known it was always there for good reason. The build should pass now.

lib/active_admin/batch_actions/views/selection_cells.rb
```diff
@@ -7,8 +7,11 @@ module BatchActions
     class ResourceSelectionToggleCell < ActiveAdmin::Component
       builder_method :resource_selection_toggle_cell
 
-      def build
-        input type: "checkbox", id: "collection_selection_toggle_all", name: "collection_selection_toggle_all", class: "toggle_all"
+      def build(label_text='')
```
deivid-rodriguez: I want to normalize the style here and use spaces around the equal sign here (https://github.com/activeadmin/activeadmin/pull/4990). If the others agree, could you fix the tiny issue here?

javierjulio: Sounds good! I actually setup Rubocop recently at my work so all for it. 👍🏻

javierjulio:
@deivid-rodriguez style issue fixed. Once you merge in #4990, do you want me to rebase master here and force push?

deivid-rodriguez:
If you've fixed the issue here, I think it's not necessary. We used to require rebased PR as a condition for merging, but I think we dropped that requirement.

PR #4988, opened 2017-06-03
Add descriptive placeholder to date range filter fields

javierjulio:
This removes the separator and icon for date range filter fields and in their place uses a descriptive placeholder so its clear what each field is for since they appear side by side, in their current state its hard to tell what each does other than take a date. Now the first field has a placeholder of "Starts with" and the second has "Ends with" which comes from existing locale text. This is being pulled out of #3862.


deivid-rodriguez:
Hei @javierjulio, I like the new little placeholders!

However, I'm not sure I like the "Starts with"/"Ends with" text, wouldn't it be better something like "From" and "To"?

Also, I don't dislike the old little calendar icon? Can we keep both? If it doesn't fit in your overall conception of the new design I'm fine with dropping it, though.

javierjulio:
@deivid-rodriguez yes that's fine. I updated the placeholders and added the locale text in the same section. I didn't think it would need to go elsewhere but I can move it if necessary.

The icons don't fit. The whole notion was to streamline the design to the bare essentials. To clean up the CSS so future changes and enhancements like that don't get bogged down by outdated styles. That doesn't mean that there shouldn't be icons, just that it should be stripped of anything outdated and to simplify. My thinking was that something like icons/theme could be added in another PR after the initial CSS update. The core was in real need of improving in my opinion. It should make some custom styling a lot less painful now. Also SVG support is great. A new PR can decide some icons we like as a default theme and make sure they are easy to configure/change.

PR #3677, opened 2014-12-06
#2753 - status_tag supports boolean status (fix)

vanekjar:
Hi guys,

I fixed this issue little bit. Previous version didn't worked for me. It had been accepting only boolean value in a string which I think made no sense.

PR #4989, opened 2017-06-03
Remove status_tag type parameter

javierjulio:
This is a change that was requested by @timoschilling in https://github.com/activeadmin/activeadmin/pull/3677#issuecomment-65922117 (look at item 2) and was pulled from #3862 there since we want to incorporate it earlier and minimize the amount of changes needed for review. Also we did want to remove the additional classes as users should be able to add those on their own rather than defining ones and styles they might not want.

Before providing any options, a second argument called `type` was required. So you would write something like the following:

```ruby
status_tag('active', :ok, class: 'abc', label: 'on')
status_tag('active', nil, class: 'abc', label: 'on')
```

This didn't make much sense as type was just appending to class but its just easier for users to provide their own classes for styling while ActiveAdmin just focuses on providing the defaults it handles which are true/false. 

**Now with `type` removed, the usage becomes**:

```ruby
status_tag('active', class: 'abc', label: 'on')
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## Master (unreleased)
 
+### Deprecations
+
+* Deprecated `type` param from `status_tag` and related CSS classes [#4989][] by [@javierjulio][]
+  * The method signature has changed from:
+    ```ruby
+    status_tag(status, :ok, class: 'completed', label: 'on')
+    # now changes to:
+    status_tag(status, class: 'completed ok', label: 'on')
+    ```
+  * The following CSS classes have been deprecated and will be removed in the future:
+    ```css
+    .status_tag {
+      &.ok, &.published, &.complete, &.completed, &.green { background: #8daa92; }
+      &.warn, &.warning, &.orange { background: #e29b20; }
+      &.error, &.errored, &.red { background: #d45f53; }
+    }
+    ```
+
+
 ### Enhancements
 
 ##### Minor
@@ -69,7 +88,7 @@ index download_links: ->{ can?(:view_all_download_links) || [:pdf] }
 
 ### Security Fixes
 
-* Prevents access to formats that the user not permitted to see [#4867][] by [@Fivell][] and [@timoschilling][] 
+* Prevents access to formats that the user not permitted to see [#4867][] by [@Fivell][] and [@timoschilling][]
 * Prevents potential DOS attack via Ruby symbols [#1926][] by [@seanlinsley][]
   * [this isn't an issue for those using Ruby >= 2.2](http://rubykaigi.org/2014/presentation/S-NarihiroNakamura)
 
--- a/app/assets/stylesheets/active_admin/components/_status_tags.scss
+++ b/app/assets/stylesheets/active_admin/components/_status_tags.scss
@@ -6,6 +6,7 @@
   padding: 3px 5px 2px 5px;
   font-size: 0.8em;
 
+  // TODO: deprecated, remove in a v2 release
   &.ok, &.published, &.complete, &.completed, &.green { background: #8daa92; }
   &.warn, &.warning, &.orange { background: #e29b20; }
   &.error, &.errored, &.red { background: #d45f53; }
--- a/lib/active_admin/views/components/status_tag.rb
+++ b/lib/active_admin/views/components/status_tag.rb
@@ -12,9 +12,8 @@ module ActiveAdmin
         'status_tag'
       end
 
-      # @overload status_tag(status, type = nil, options = {})
+      # @overload status_tag(status, options = {})
       #   @param [String] status the status to display. One of the span classes will be an underscored version of the status.
-      #   @param [Symbol] type type of status. Will become a class of the span. ActiveAdmin provide style for :ok, :warning and :error.
       #   @param [Hash] options
       #   @option options [String] :class to override the default class
       #   @option options [String] :id to override the default id
@@ -26,17 +25,14 @@ module ActiveAdmin
       #   # => <span class='status_tag in_progress'>In Progress</span>
       #
       # @example
-      #   status_tag('active', :ok)
-      #   # => <span class='status_tag active ok'>Active</span>
-      #
-      # @example
-      #   status_tag('active', :ok, class: 'important', id: 'status_123', label: 'on')
-      #   # => <span class='status_tag active ok important' id='status_123'>on</span>
+      #   status_tag('active', class: 'important', id: 'status_123', label: 'on')
+      #   # => <span class='status_tag active important' id='status_123'>on</span>
       #
       def build(*args)
         options = args.extract_options!
         status = args[0]
         type = args[1]
+
         label = options.delete(:label)
         classes = options.delete(:class)
         status = convert_to_boolean_status(status)
@@ -50,7 +46,19 @@ module ActiveAdmin
         super(content, options)
 
         add_class(status_to_class(status)) if status
-        add_class(type.to_s) if type
+
+        if type
+          Deprecation.warn <<-MSG.strip_heredoc
+            The `type` parameter has been deprecated. Provide the "#{type}" type as
+            a class instead. For example, `status_tag(status, :#{type}, class: "abc")`
+            would change to `status_tag(status, class: "#{type} abc")`. Also note that
+            the "#{type}" CSS rule will be removed too, so you'll have to provide
+            the styles yourself. See https://github.com/activeadmin/activeadmin/blob/master/CHANGELOG.md#110-
+            for more information.
+          MSG
+          add_class(type.to_s)
+        end
+
         add_class(classes) if classes
       end
 
--- a/docs/12-arbre-components.md
+++ b/docs/12-arbre-components.md
@@ -138,18 +138,15 @@ takes a block.
 Status tags provide convenient syntactic sugar for styling items that have
 status. A common example of where the status tag could be useful is for orders
 that are complete or in progress. `status_tag` takes a status, like
-"In Progress", a type, which defaults to nil, and a hash of options. The
-status_tag will generate html markup that Active Admin css uses in styling.
+"In Progress", and a hash of options. The status_tag will generate HTML markup
+that Active Admin CSS uses in styling.
 
 ```ruby
 status_tag 'In Progress'
 # => <span class='status_tag in_progress'>In Progress</span>
 
-status_tag 'active', :ok
-# => <span class='status_tag active ok'>Active</span>
-
-status_tag 'active', :ok, class: 'important', id: 'status_123', label: 'on'
-# => <span class='status_tag active ok important' id='status_123'>on</span>
+status_tag 'active', class: 'important', id: 'status_123', label: 'on'
+# => <span class='status_tag active important' id='status_123'>on</span>
 ```
 
 ## Tabs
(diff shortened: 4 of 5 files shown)


javierjulio:
@varyonic sounds good. Just pushed updates for that. Let me know how they look. Thanks!

javierjulio:
@deivid-rodriguez if you have the time would appreciate your input here too. Thanks!

lib/active_admin/views/components/status_tag.rb
```diff
@@ -60,7 +53,7 @@ def convert_to_boolean_status(status)
         case status
         when true, 'true'
           'Yes'
-        when false, 'false', nil
+        when *FALSE_VALUES
```
deivid-rodriguez: Why is the "false values" list incremented? And why is the "true values" list not incremented accordingly?

javierjulio: @deivid-rodriguez I don't know. I tried to see if I made the change on my own or if there was a comment in #3862 by someone to suggest otherwise but doesn't seem to be. I must have done that on my own, but don't remember. I can reset it back to what it was. Should I do that?

deivid-rodriguez: Yes, let's leave it as it was if we can't remember the reason for the change.

javierjulio: @deivid-rodriguez code updated.

javierjulio:
Let me know what you end up preferring here. I realize now that the original code worked with how I was using it because of the splat. I could either revert it back to that or do something like:

```ruby
def build(status, type = nil, options = {})
  if !type.is_a?(Hash)
    Deprecation.warn "the `type` parameter is deprecated! Use `status_tag(status, class: 'type')`."
  end

  options = type if type.is_a?(Hash)
  label = options.delete(:label)
  classes = options.delete(:class)
  status = convert_to_boolean_status(status)

  # ...
  
  add_class(status_to_class(status)) if status
  add_class(type.to_s) if type && !type.is_a?(Hash)
  add_class(classes) if classes
end
```

And from there we would have to update our tests as they now fail with the deprecation warning. If we go this route, do you just want the existing tests in place but just stub out the deprecation calls?

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
 ## Master (unreleased)
 
+### Breaking Changes
+
+* Remove `type` param from `status_tag` [#4989][] by [@javierjulio][]
+
+    ```ruby
+    status_tag(status, :ok, class: 'completed', label: 'on')
+    # now changes to:
+    status_tag(status, class: 'completed ok', label: 'on')
```
varyonic: But the provided .completed and .ok classes have been removed also, no?

javierjulio: They have. But this just demonstrates the expected output. I could add a note that previous classes were also removed too.

javierjulio:
The build failure has to do with phantomjs. Although I see in the response log outputted about a GitHub request and it had some outage in the past hour so that's most likely the culprit.

CHANGELOG.md
varyonic: Yes, as someone who has to make this change I now have to go define these styles as part of the upgrade.

javierjulio: @varyonic changelog updated.

deivid-rodriguez:
@javierjulio True, the `type` parameter was actually optional... I'm not really sure now. With "reverting back" you mean ditching this whole PR? What are the reasons for removing the `type` parameter anyways?

In any case, if we're going to make the breaking change, let's add the deprecation since it seems easy enough as you just proved and can help users get ready for the new way.

Regarding tests, let's take care of that once we make a decision.

javierjulio:
@deivid-rodriguez by reverting meaning going back to using splat. I removed the splat based on comments made in #3862. The `type` param was removed because that just set the CSS class when users can set that themselves and makes the method signature simpler. The param was requested to be removed by @timoschilling, the comment is linked in the PR description. I'll make the necessary updates and stub the deprecation calls in the tests.

deivid-rodriguez:
@javierjulio Sorry I didn't properly read through the link. Actually that comment also answers the previous question about boolean list (bullet point 3). Let's leave that for now anyways.

Your plan sounds good to me, thanks for your work!

javierjulio:
@deivid-rodriguez fixing the unit tests isn't bad but since I'm not familiar working with cucumber much, I'm not sure the best way to stub out the deprecation warnings as that is now causing many cucumber failures. Any ideas?

lib/active_admin/views/components/status_tag.rb
```diff
@@ -50,7 +47,7 @@ def build(*args)
         super(content, options)
 
         add_class(status_to_class(status)) if status
-        add_class(type.to_s) if type
+        add_class(type.to_s) if type && !type.is_a?(Hash)
```
deivid-rodriguez: Regarding the implementation, after a second read I think it's better to keep the splat. I think we can actually leave this method exactly as it was, but just changing this:

```diff
- add_class(type.to_s) if type
+ if type
+   Deprecation.warn "the `type` parameter is deprecated! Provide as a class: `status_tag(status, class: 'abc')`."
+   add_class(type.to_s)
+  end
```

deivid-rodriguez:
Not not a lot about cucumber either, but will try to help!

lib/active_admin/views/components/status_tag.rb
javierjulio: Yes it will be better and less work. Cucumber tests are passing with it going back to being a splat. 👊🏻

javierjulio:
@deivid-rodriguez you won't need too! As I thought when you suggested going back to the splat version, the cucumber tests wouldn't be an issue anymore. And they aren't. 😏 Tests are all green locally. I think with this we've addressed everything. Let me know what you and @varyonic think. Its coming together.

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
+    status_tag(status, class: 'completed', label: 'on')
+    ```
+  * The following CSS classes have been **removed**:
+    ```css
+    .status_tag {
+      &.ok, &.published, &.complete, &.completed, &.green { background: #8daa92; }
+      &.warn, &.warning, &.orange { background: #e29b20; }
+      &.error, &.errored, &.red { background: #d45f53; }
+    }
+    ```
```
deivid-rodriguez: This is no longer a breaking change (not yet). I guess we should move it to a "deprecations" section or something like that.

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
 ## Master (unreleased)
 
+### Breaking Changes
+
+* Removed `type` param from `status_tag` and some CSS classes [#4989][] by [@javierjulio][]
+  * The method signature has changed from:
+    ```ruby
+    status_tag(status, :ok, class: 'completed', label: 'on')
+    # now changes to:
+    status_tag(status, class: 'completed', label: 'on')
```
deivid-rodriguez: Does the new code not need the "ok" class as well?

javierjulio: Yes it can be added to demonstrate that's what it was originally doing and to maintain the same behavior it would have to be added if they were still using that CSS class.

CHANGELOG.md
javierjulio: I'll move it to a Deprecations section instead.

javierjulio:
@deivid-rodriguez renamed section as Deprecations and rewrote that that `type` param is now deprecated.

app/assets/stylesheets/active_admin/components/_status_tags.scss
```diff
@@ -6,10 +6,6 @@
   padding: 3px 5px 2px 5px;
   font-size: 0.8em;
 
-  &.ok, &.published, &.complete, &.completed, &.green { background: #8daa92; }
-  &.warn, &.warning, &.orange { background: #e29b20; }
-  &.error, &.errored, &.red { background: #d45f53; }
-
```
deivid-rodriguez: I don't think it makes sense to keep compatibility with a deprecation, but remove the styles, since people's designs would break anyways. I think we should add TODO note here to "schedule removal" and we actually remove the feature. Would that get in the middle somehow of the rest of the css overhaul?

deivid-rodriguez:
@javierjulio I added yet another comment. Sorry for the extra pickyness here, but I think the better work we do now, the less work we'll have later...

CHANGELOG.md
varyonic: Hmm, removal of the CSS classes still makes it a breaking change IMHO.

app/assets/stylesheets/active_admin/components/_status_tags.scss
javierjulio: Yes that's fine. We can deprecate the CSS. I'll have to be careful when rebasing on the css overhaul that it doesn't get lost. I will still keep the CSS in the changelog though and note that its deprecated so if users want it they should copy it themselves.

lib/active_admin/views/components/status_tag.rb
```diff
@@ -50,7 +46,12 @@ def build(*args)
         super(content, options)
 
         add_class(status_to_class(status)) if status
-        add_class(type.to_s) if type
+
+        if type
+          Deprecation.warn "the `type` parameter is deprecated! Provide as a class: `status_tag(status, class: 'abc')`."
```
deivid-rodriguez: I think we can still improve the message a bit. What about this:

The `type` parameter has been deprecated. Provide the "#{type}" type as a class option instead. For example, `status_tag(status, :ok, class: "abc")` would change to `status_tag(status, class: "ok abc")`. Also note that the #{type} CSS class will be removed too, so you'll have to provide the CSS rules yourself. See https://github.com/activeadmin/activeadmin/blob/master/CHANGELOG.md#110- for more information.

javierjulio: Can't be that detail so it looks great to me!

javierjulio: I just tweaked it a little bit:

```
The `type` parameter has been deprecated. Provide the "#{type}" type as a class instead. For example, `status_tag(status, :ok, class: "abc")` would change to `status_tag(status, class: "ok abc")`. Also note that the "#{type}" CSS rule will be removed too, so you'll have to provide the styles yourself. See https://github.com/activeadmin/activeadmin/blob/master/CHANGELOG.md#110- for more information.
```

javierjulio:
@deivid-rodriguez updated the message. Also made sure the full status_tag example in the message evaluated the type param so it output for that users' specific usage.

deivid-rodriguez:
It looks great, thanks!

javierjulio:
@varyonic not sure why you dismissed your review but in case you wanted to review again, I addressed the latest issues regarding the CSS and deprecation message. Let me know what you think.

deivid-rodriguez:
It was when we noticed that removing the CSS was actually a breaking change. He'll probably reapprove when he sees this! :wink:

javierjulio:
@varyonic I wasn't sure whether to squash related commits or just down to one but did the latter. Kept the meaningful messages intact that mention deprecation, not removing. Thanks. Let me know how it is. Thanks for the help @deivid-rodriguez!

deivid-rodriguez:
Yeah, both were fine! Thanks, @javierjulio!!

seanlinsley:
If we're going to remove the default `status_tag` colors, I think it'd make sense to move them into the documentation as a suggested starting point.

Developers aren't the best designers, so having a color scheme to start from can help them build a better information architecture.

PR #4990, opened 2017-06-04
Enforce spaces around optional parameter values

deivid-rodriguez:


varyonic:
Rebased, resolved conflict with 4994.

PR #4991, opened 2017-06-04
Fix doc generation

deivid-rodriguez:
I noticed that the manual pages for the different index are automatically generated from the comments in the source files.

This PR makes sure the generated docs are always in sync (they were not before this PR).


deivid-rodriguez:
Rebased and improved a bit. Kudos to rubocop since this patch is inspired in what they do there.

PR #4992, opened 2017-06-06
Add more features to generated test app

javierjulio:
For contributing and development, its essential the generated app has more features enabled to make testing and editing easier. For example, for the work being done in #3862 it was critical to use the many available components in all sorts of ways to test appearance, functionality and breakage. Having this out of the box from a generated test app saves a lot of time by not losing that effort spent setting it up.

## Todo list

- [x] sample `has_many` form
- [x] update a form to [use all sorts of components](https://github.com/activeadmin/activeadmin/blob/master/docs/5-forms.md#forms)
- [x] a custom batch actions form (posts page to set starred? set author from drop down?)
- [x] update a page to [use a different index page type](https://github.com/activeadmin/activeadmin/blob/master/docs/3-index-pages.md#customizing-the-index-page)
- [x] create a custom page and call it Kitchen Sink
  - [x] add `tabs` with dummy content
  - [x] add basic HTML elements
- [x] add `columns` to post show page, use `attributes_table_for` inside panels
- [x] add `attributes_table_for` to sidebar in a show page
- [x] add a custom action item and member action
- [x] add menu to navigation bar
- [x] add eager loading to minimize noise in logs

### DIFF
--- a/spec/support/rails_template_with_data.rb
+++ b/spec/support/rails_template_with_data.rb
@@ -39,7 +39,7 @@ inject_into_file 'app/admin/user.rb', <<-RUBY, after: "ActiveAdmin.register User
     end
 
     panel 'Posts' do
-      table_for(user.posts.order(:updated_at).limit(10)) do
+      table_for(user.posts.includes(:category).order(:updated_at).limit(10)) do
         column :id do |post|
           link_to post.id, admin_post_path(post)
         end
@@ -59,6 +59,8 @@ RUBY
 inject_into_file 'app/admin/post.rb', <<-RUBY, after: "ActiveAdmin.register Post do\n"
 
   permit_params :custom_category_id, :author_id, :title, :body, :published_date, :position, :starred, taggings_attributes: [ :id, :tag_id, :name, :position, :_destroy ]
+
+  includes :author, :category, :taggings
 
   scope :all, default: true
 
--- a/spec/support/rails_template_with_data.rb
+++ b/spec/support/rails_template_with_data.rb
@@ -1,6 +1,6 @@
 apply File.expand_path("../rails_template.rb", __FILE__)
 
-%w{Post User Category}.each do |type|
+%w{Post User Category Tag}.each do |type|
   generate :'active_admin:resource', type
 end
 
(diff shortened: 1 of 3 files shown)


spec/support/rails_template_with_data.rb
```diff
@@ -35,6 +156,137 @@
   scope :my_posts do |posts|
     posts.where(author_id: current_admin_user.id)
   end
+
+  controller do
+
+    def scoped_collection
+      super.includes(:author, :category, :taggings)
```
Fivell: @javierjulio  it can be simplified using includes on resource level. take a look 
https://github.com/activeadmin/activeadmin/pull/3464#issue-44728695

javierjulio:
@Fivell @deivid-rodriguez any ideas what I should do about failing codecov/project? It has the X status but when I click the details link it reports no flags found. Unclear if everything is ok and what I need to do.

spec/support/rails_template_with_data.rb
javierjulio: Neither did I. I've been doing it the other way and typing so much more code. 🤣

Fivell:
@javierjulio , as I know  you can ignore this, it is not a blocker for merging

deivid-rodriguez:
@javierjulio It passed now! :)

javierjulio:
@Fivell updated to use root level `includes` DSL method and tabs within the form for the Post model.

spec/support/rails_template_with_data.rb
```diff
@@ -1,22 +1,145 @@
...
+
+%w{Post User Category Tag}.each do |type|
   generate :'active_admin:resource', type
 end
 
 inject_into_file 'app/admin/category.rb', <<-RUBY, after: "ActiveAdmin.register Category do\n"
 
+  config.create_another = true
+
   permit_params [:name, :description]
```
Fivell: @javierjulio , I think ```permit_params``` are declared
like
```
permit_params :name, :description
```

however this works with array too, so not a blocker

javierjulio:
Everything is passing except for the jruby rails 4.2 build as its timing out:

> The job exceeded the maximum time limit for jobs, and has been terminated.

If it fails again would you be ok with me merging in?

PR #4996, opened 2017-06-07
allow scope to paginate collection

### LINKED ISSUE #4977: Active admin ignoring .limit() in scopes
Say I have a scope like this on my Active Record model
  
    scope :test, -> {  where(arel_table[:created_at].gteq 30.days.ago) }

which returns 5 records

and one like this

     scope :test_limit, -> {  where(arel_table[:created_at].gteq 30.days.ago).limit(1) }

And I change the default scope on my Active Admin record  from

     scope :test, default: true

to 

    scope :test_limit, default: true

### Expected behavior

Active Admin should go from displaying 5 records to only displaying 1 record when I change the scope from :test to :test_limit

### Actual behavior

Active admin keeps on displaying 5 records even with the explicit scope limiting it to 1 but, the count in the scope tab displays 1

### How to reproduce

See above. Also just add `.limit(1)` to any of your Model scopes that you are referencing in Active admin. The tab will say (1) but it will display all the records up to the page limit

Fivell:
PaginatedCollection works only with paginated scope, so
skip apply_pagination if pagination was applied by scope logic
to make this work scope can be defined as
```
 ActiveAdmin.register Post do
    scope :all
    scope :random do |posts|
      posts.random.page(1).per(3)
    end
 end
```

[fixes #4977]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 
 ##### Minor
 
+* Allow AA scopes to return paginated collections [#4996][] by [@Fivell][]
 * Added `scopes_show_count` configuration to  setup show_count attribute for scopes globally [#4950][] by [@Fivell][]
 * Allow custom panel title given with `attributes_table` [#4940][] by [@ajw725][]
 
@@ -151,6 +152,7 @@ Please check [0-6-stable](https://github.com/activeadmin/activeadmin/blob/0-6-st
 [#4867]: https://github.com/activeadmin/activeadmin/pull/4867
 [#4882]: https://github.com/activeadmin/activeadmin/pull/4882
 [#4950]: https://github.com/activeadmin/activeadmin/pull/4950
+[#4996]: https://github.com/activeadmin/activeadmin/pull/4996
 
 
 [@bolshakov]: https://github.com/bolshakov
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
On an unrelated note, when I tried this I don't see a cursor when hovering over the scope. If I remove the generated `table_tools_segmented_control` CSS class, it works fine. Does it happen to you too? @Fivell? @javierjulio?

Fivell:
@deivid-rodriguez , +1 for "Single", don't know about cursor, hope @javierjulio , will take care

javierjulio:
@deivid-rodriguez @Fivell I'm not sure but either way I've redone all that so the segmented control is just made up of button components using the button class which has the cursor set.

concept47:
@Fivell so essentially you have to manually page a scope if you want something like .limit() to work from an already defined AR scope?

Fivell:
@concept47, in general yes.
 Only paginated collection can be used within index view, so limit is not enough
take a look https://github.com/activeadmin/activeadmin/blame/7461a55e8c367cb6546c96ed7bfb90c952da2eb5/lib/active_admin/views/components/paginated_collection.rb#L49

I didn't want to make breaking changes , so thats why I just add skipping of applying pagination if for such case.

concept47:
Can't really test it out because master seems to force me to require activemodel 5.1.1. Project is on Rails 4.1 😬

PR #4998, opened 2017-06-08
Replace reset with normalize.css

javierjulio:
Best to normalize CSS, not reset it. When using a reset it adds significant amount of noise to the element inspector. Plus after resetting all elements, you have to respecify all values you want. Some that can't even be set back easily once removed. I've long thought it best to target only the browser defaults that are inconsistent. Normalize.css aims to do this.


deivid-rodriguez:
@javierjulio Thanks! So what's the plan here? Get #4992 in before this one to be able to better test it? Or you've already tried this and you are already happy?

deivid-rodriguez:
Merging after seeing @javierjulio's message on slack!

PR #5001, opened 2017-06-09
Tweak test coverage setup

deivid-rodriguez:
This improves our code coverage config.

The biggest improvement is that we now track test files too. The new configuration detected quite a lot of dead code in the `features` folder. We should review it and either remove what we don't want, or start using what we want.

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+# @todo: Always run simplecov again once
+# https://github.com/colszowka/simplecov/issues/404,
+# https://github.com/glebm/i18n-tasks/issues/221 are fixed
+if ENV['COVERAGE'] == 'true'
+  SimpleCov.start do
+    add_filter 'spec/'
+    add_filter 'features/'
+  end
+end
+
+if ENV['CI'] == 'true'
+  require 'codecov'
+  SimpleCov.formatter = SimpleCov::Formatter::Codecov
+end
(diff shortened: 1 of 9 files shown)


.simplecov
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# @todo: Always run simplecov again once
+# https://github.com/colszowka/simplecov/issues/404,
+# https://github.com/glebm/i18n-tasks/issues/221 are fixed
+if ENV['COVERAGE'] == 'true'
+  SimpleCov.start do
+    add_filter 'spec/rails/'
+  end
```
Fivell: @deivid-rodriguez , can please explain in more details why  adding filter with 'spec/rails/' needed?

deivid-rodriguez: Yes, it's generated code! We don't want to track coverage of generated code!

deivid-rodriguez: This was already ignored in the old config but in the new one we need to make it explicit. I can expand the commit message because I don't think I made this clear.

Fivell: @deivid-rodriguez , I forgot simplecov syntax,  thought add_filter means add_group

deivid-rodriguez:
A test started failing against Ruby 2.1... Should we drop MRI 2.1 support for 1.1.0 release? :sunglasses:. I vote yes, but if not, I'll look into it!

Fivell:
@deivid-rodriguez , As I remember we were going to make it after 1.0.0 release so lets do it

deivid-rodriguez:
Yes! Since installation will be prevented, this is not really a backwards incompatible change as in "it will break production apps". So I think it's fine to do this for 1.1.0.

PR #5006, opened 2017-06-13
added missing lt translations

vytenis-s:


varyonic:
Great.  Good to see you figured it out.  We just need to get our jruby tests fixed...

PR #5007, opened 2017-06-14
Travis upgrades

deivid-rodriguez:
These are some changes in the process of investigating why our jruby build got so slow. Note that with the upgrade to trusty, we also go from jdk 7 to jdk 8.

Let's see what Travis says.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,15 +46,3 @@ matrix:
       gemfile: gemfiles/rails_50.gemfile
     - rvm: jruby-9.1.10.0
       gemfile: gemfiles/rails_51.gemfile
-
-notifications:
-  irc:
-    channels:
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Seems to do the trick... maybe? I'd say let's merge and see if the build times stay sane or it was just luck.

deivid-rodriguez:
By the way: https://travis-ci.org/activeadmin/activeadmin/jobs/242896409. Cucumber segfaulting... :S Hope it's not something very frequent or we'll have to deal somehow (maybe reverting to 1.x or trying out 3.0.0.pre.1)...

deivid-rodriguez:
I removed the bundler upgrade until we make a decision in #5009. Jruby build was fast enough again, so this seems to do the trick indeed.

PR #5009, opened 2017-06-15
use bundler version from gemfile.lock on travis

timoschilling:


timoschilling:
sure, I don't know why someone added this lines, remove them if you want.

deivid-rodriguez:
We added them because CI would break because of bugs in the most recent bundler releases. So we decided to lock it.

deivid-rodriguez:
Apparently bundler 2 will automatically use the version in `Gemfile.lock` by default. And you can opt in to this feature via `ENV['BUNDLE_TRAMPOLINE_FORCE']`. But we would have to commit `Gemfile.lock` for it to have a meaningful effect, I think.

timoschilling:
I close this PR, add to your PR what do you think it's the best

deivid-rodriguez:
@timoschilling To be honest, I'm not really sure what to do.

I guess we could stop locking bundler. Since we dropped support for old rubies, our dependency management is much less of a nightmare and less vulnerable to hit bugs in bundler. And since our current strategy is to not lock dependencies (no lock files committed, no specific rubygems version enforced), it probably makes sense to let bundler free as well.

@Fivell, @varyonic, @javierjulio Thoughts?

varyonic:
Locking bundler seemed a good idea at the time (#4744) but I've since gone off it.  I'm fine with freeing the dependency.

javierjulio:
@deivid-rodriguez I'm not sure what is best, or much much of what to do here, sorry. 😞 I trust your input though so if you feel that's best then lets do it.

deivid-rodriguez:
Ok, let's free the dependency then. I'll do it when I get some of the other PR's in, so they don't step on each other.

PR #5011, opened 2017-06-15
do not pass humanize_name option to CSV when exporting

### LINKED ISSUE #5010: humanize_name option for csv DSL columns
feature requests, I think. I would like the _humanize_name_ on the *csv* DSL (for all columns on a resource). Tried this:
`csv humanize_name: false do`
but it is throwing an "ArgumentError at /admin/processing_orders.csv Unknown options:  humanize_name" error.
I thought that pull #3145 added that.
Using 607054214d4eaba5b59e2eab736c6a41edfc7795 revision, btw.

Fivell:
[fixes #5010]

### DIFF
--- a/features/index/format_as_csv.feature
+++ b/features/index/format_as_csv.feature
@@ -61,6 +61,23 @@ Feature: Format as CSV
       | Title        | Body |
       | Hello, World | (.*) |
 
+  Scenario: With humanize_name option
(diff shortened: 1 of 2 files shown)


Fivell:
@deivid-rodriguez , healthy after jruby's 37 min build

PR #5015, opened 2017-06-17
Remove mention_bot

deivid-rodriguez:
This reverts commit fee94b9ad170717ac5917534154f2202675cc151, reversing
changes made to c5b7746e15071d7a2e0a0d4cfbf19e8c1e699583.

The configuration added is not working, so I'll just remove it since the benefit would be minor anyways (and might annoy some people?). The right thing to do would be to get an official translation maintainer for each language, I guess.

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -1,9 +1,0 @@
-{
-  "message": "@pullRequester, thanks for contributing an activeadmin
-              translation! @reviewers, are you able to review this?",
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Funny, I was deleting the mention-bot webhook and accidentally found out why it was not working. I forgot to check the events to be sent on "Labelling". Oh well.

mentionbot

PR #5016, opened 2017-06-17
Remove jslint

deivid-rodriguez:
It's unmantained, it was skipped unless java was installed and it was
not analyzing any files!


deivid-rodriguez:
Oops, yeah, nice catch. Let me remove that!

PR #5019, opened 2017-06-17
Missing es keys, minor fixes

mauriciopasquier:
As it says in the title, there were some keys missing and some content not translated correctly.

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -8,6 +8,7 @@ es:
     edit: "Editar"
     delete: "Eliminar"
     delete_confirmation: "¿Está seguro de que quiere eliminar esto?"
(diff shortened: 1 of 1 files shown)


mauriciopasquier:
Also, to fix the same locale but against 1.0.0 stable, so it can be released freely, which branch should I use?

config/locales/es.yml
```diff
@@ -8,6 +8,7 @@ es:
     edit: "Editar"
     delete: "Eliminar"
     delete_confirmation: "¿Está seguro de que quiere eliminar esto?"
+    create_another: "Crear otro %{model}"
```
deivid-rodriguez: Maybe use `otro/a` to play nice with feminine nouns?

PR #5021, opened 2017-06-20
Missing internationalized string

mauriciopasquier:
I know this is really a minor thing but for completeness sake, the "admin user" generator was missing this string for interpolation. The keys weren't alphabetized so I added it near the dashboard ones.

I took the liberty of translating the addition for spanish, in case it's accepted.

### DIFF
--- a/lib/generators/active_admin/install/templates/admin_user.rb.erb
+++ b/lib/generators/active_admin/install/templates/admin_user.rb.erb
@@ -17,7 +17,7 @@ ActiveAdmin.register <%= @user_class %> do
   filter :created_at
 
   form do |f|
-    f.inputs "Admin Details" do
+    f.inputs do
       f.input :email
       f.input :password
       f.input :password_confirmation


config/locales/en.yml
```diff
@@ -4,6 +4,8 @@ en:
     dashboard_welcome:
       welcome: "Welcome to Active Admin. This is the default dashboard page."
       call_to_action: "To add dashboard sections, checkout 'app/admin/dashboard.rb'"
+    admin_user:
+      details: "Admin Details"
```
deivid-rodriguez: I think I'd go with a different key name. Otherwise it could be interpreted that there's a `details` column on the `AdminUser` model.

config/locales/es-MX.yml
```diff
@@ -4,6 +4,8 @@ es-MX:
     dashboard_welcome:
       welcome: "Bienvenido a Active Admin. Esta es la página de inicio predeterminada."
       call_to_action: "Para agregar secciones edite 'app/admin/dashboard.rb'"
+    admin_user:
+      details: "Detalles de administración"
```
deivid-rodriguez: I think it'd be better "Detalles de administrador", since it refers to the admin user you are creating.

mauriciopasquier:
I'm ok not providing any label, I just wanted the consistency of not having an untranslated string even when setting up activeadmin.

I agree with your comments as well, so whatever you decide I'll fix the PR.

deivid-rodriguez:
> I'm ok not providing any label, I just wanted the consistency of not having an untranslated string even when setting up activeadmin.

That makes sense, we'll fix that in this PR one way or another :)

> I agree with your comments as well, so whatever you decide I'll fix the PR.

Ok, @Fivell are you ok with my proposal?

Fivell:
@deivid-rodriguez , I'm ok with removing this key, since it is template

deivid-rodriguez:
Let's do it then. @mauriciopasquier can you update the PR to remove the key? Thanks!

mauriciopasquier:
Done!

deivid-rodriguez:
Thank you! :heart:

PR #5024, opened 2017-06-21
Fix the interaction between batch actions

sasha-bichkov:
After the first usage of a batch action, reuse of another one that uses **ActiveAdmin.modal_dialog** is impossible because of **batch_action_inputs** form element is disabled and the server doesn't get its required parameters.

### DIFF
--- a/app/assets/javascripts/active_admin/lib/batch_actions.js.coffee
+++ b/app/assets/javascripts/active_admin/lib/batch_actions.js.coffee
@@ -14,7 +14,7 @@ $(document).on 'ready page:load turbolinks:load', ->
 
   $('.batch_actions_selector li a').on 'confirm:complete', (e, inputs)->
     if val = JSON.stringify inputs
-      $('#batch_action_inputs').val val
+      $('#batch_action_inputs').removeAttr('disabled').val val
     else
       $('#batch_action_inputs').attr 'disabled', 'disabled'
 


Fivell:
@sasha-bichkov , can you please explain what do you mean by reuse? 
each batch action performs page reload so each usage of batch action seems to be first

sasha-bichkov:
I'm using both **send_file** and **send_data** in my batch actions in order to generate pdf/xlsx/zip files for selected items. So, there's no redirecting as well as a page reloading.

Steps to reproduce that bug:
1. Click on the batch_action that sents a file (instead of redirecting and reloading a page):
```
batch_action :action1 do |ids|
  table = TableService.create ids
  send_file table.path, filename: 'table.xlsx' 
end
```
2. Click on the batch_action with a custom form. For example, like this one:
```
batch_action :action2, form: {
  status: ['status1', 'status2']
} do |ids, inputs|
  # do something
end
```
3. Select an option or leave it with the default value and then press "OK"

The bug will be occurred with the following message: "Status couldn't be empty".

Fivell:
@sasha-bichkov , makes sense

app/assets/javascripts/active_admin/lib/batch_actions.js.coffee
```diff
@@ -14,6 +14,7 @@ $(document).on 'ready page:load turbolinks:load', ->
 
   $('.batch_actions_selector li a').on 'confirm:complete', (e, inputs)->
     if val = JSON.stringify inputs
+      $('#batch_action_inputs').removeAttr 'disabled'
```
Fivell: maybe we can you chain here like 
```
$('#batch_action_inputs').removeAttr('disabled').val val
```

sasha-bichkov: I'm agree with you. I've just made this fix.

Fivell:
@sasha-bichkov , squash commits please

Fivell:
@sasha-bichkov 🚀 ,thanks

PR #5029, opened 2017-06-23
allow to pass proc to filters input_html option

### LINKED ISSUE #5026: Passing rails url helper to input_html attribute in filter
### Expected behavior
Setting rails url helper in input_html should be parsed and correctly set to the data attribute:

```   filter :user, collection: -> { User.all }, input_html: {class: "select-remote-data", "data-ajax-url": -> { users_api_url } }```

### Actual behavior

No matter in which form I will pass the rails url helper it fails to parse it with error `undefined method for`,
in the case of passing it directly ``` "data-ajax-url": root_url``` or it is adding whole proc object into data attribute if I will pass as above the lambda.

### How to reproduce

Simply add `input_html: { "data-url": root_url } to any filter in the AA model.

The idea is to use `select2` with a remote call for fetching big data set, for this one I need proper extra attribue which `select2` will parse while initializing select. 

Update:
Just to mention that for `f.input` it works correctly:
```
 f.input :user, as: :select,
   59               collection: User.where(id: object.user_id), include_blank: false,
   60               input_html: {class: "select-remote-data", "data-ajax-url": users_api_url}
```

Thanks!

### PR DESCRIPTION
**Pull request by Fivell**
[fixes #5026]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 
 ##### Minor
 
+* Support proc as an input_html option value when declaring filters [#5029][] by [@Fivell][]
 * Base localization support, better associations handling for active filters sidebar [#4951][] by [@Fivell][]
 * Allow AA scopes to return paginated collections [#4996][] by [@Fivell][]
 * Added `scopes_show_count` configuration to  setup show_count attribute for scopes globally [#4950][] by [@Fivell][]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Fivell** [comment]
rubocop check fails  with
```
unexpected token tCOLON
(Using Ruby 2.1 parser; configure using TargetRubyVersion parameter, under AllCops)
```
so let's wait for #5030

**Comment by Fivell** [comment]
```The job exceeded the maximum time limit for jobs, and has been terminated.``` for jruby again

**Comment by Fivell** [comment]
restart build again

File: CHANGELOG.md [hunk 1]
```diff
@@ -29,6 +29,7 @@
 
 ##### Minor
 
+* Allow to use proc for `:input_html` option when declaring filters [#5029][] by [@Fivell][]
```
javierjulio: Change to: "Support `proc` as an `input_html` option value when declaring filters"

PR #5033, opened 2017-06-24
Workaround slow JRuby build

deivid-rodriguez:
0d564c7 seems to workaround the slow jruby builds for me.

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -3,10 +3,8 @@
 # TODO: Always run simplecov again once
 # https://github.com/colszowka/simplecov/issues/404,
 # https://github.com/glebm/i18n-tasks/issues/221 are fixed
-if ENV['COVERAGE'] == 'true'
-  SimpleCov.start do
-    add_filter 'spec/rails/'
-  end
+SimpleCov.start do
+  add_filter 'spec/rails/'
 end
 
 if ENV['CI'] == 'true'
--- a/.simplecov
+++ b/.simplecov
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# @todo: Always run simplecov again once
+# TODO: Always run simplecov again once
 # https://github.com/colszowka/simplecov/issues/404,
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
Around 1h less that the our average build time!

deivid-rodriguez:
Nope, there's none (at least when I checked). We would have to further isolate this and open one.

deivid-rodriguez:
That's it! Setting `JAVA_OPTS="-Djava.security.egd=file:/dev/./urandom"` did the tricky! Kudos to @varyonic who got me on the right track!

javierjulio:
@deivid-rodriguez I'm not familiar with jruby but @Fivell has already approved. Lets see what happens. I'm confident this will do it. You've worked hard on resolving this so thanks!

deivid-rodriguez:
@Fivell approved an old version of this PR, but he approved based on build times, and they're still fast, and with a simpler patch. So it should still be fine! :)

deivid-rodriguez:
Jruby is having the same problems in jruby/jruby#4696, so I'm no longer sure this was it... Let's wait and see.

headius:
FWIW, the latest jruby-launcher gem (a native executable you can install) and the upcoming JRuby 9.1.15.0 bash script equivalent both have the urandom tweak shown here as a standard startup flag. Installing jruby-launcher in any *nix environment should be sufficient to get the tweaked property.

headius:
See jruby/jruby#4696 and jruby/jruby#4857.

deivid-rodriguez:
Thanks @headius. We're still adding `-J-Djava.security.egd=file:/dev/./urandom`, but I'll make sure we remove it once `9.1.15.0` is released!

By the way, we added the latest `activerecord-jdbc-releases` to our build [recently](https://github.com/activeadmin/activeadmin/pull/5248), and the full jruby matrix is passing `activeadmin`'s tests now!

PR #4951, opened 2017-05-02
rework displaying current scope and filters in active filters section

Fivell:
- use ```DataAccess#current_scope``` method instead of ```ActiveAdmin::Filters::Active#scope``` since ```current_scope``` returns scope object which  supports localisations, dynamic names and everything else
 - display current_scope if at least 1 was declared
- added css classes for current filters and scopes for easy customisation and testing
- refactor ```ActiveAdmin::Filters::Active```
- use ransack condition object instead of request params to better handling current filters
- use translations for active filters, see ```ActiveAdmin::Filters::ActiveFilter#label```
- display related model names by ids instead of ids  see ```ActiveAdmin::Filters::ActiveFilter#values``` +  rendering current filters using pretty_format helper 

[closes #4899, #4859, #4943, #3902, #4214]

PR #5035, opened 2017-06-25
improve predicates translations for active filters

Fivell:
since we have aliases for ransack predicates
and localization strings for them
we should find translation in 'active_admin.filters.predicates' scope
and in case when not found use ransack localization

source: https://github.com/activeadmin/activeadmin/pull/4951#issuecomment-304500117

### DIFF
--- a/lib/active_admin/filters/active_filter.rb
+++ b/lib/active_admin/filters/active_filter.rb
@@ -26,11 +26,18 @@ module ActiveAdmin
       end
 
       def label
+        # TODO: to remind us to go back to the simpler str.downcase once we support ruby >= 2.4 only.
+        translated_predicate = predicate_name.mb_chars.downcase.to_s
         if related_class
-          "#{related_class.model_name.human} #{predicate_name}".strip
+          "#{related_class.model_name.human} #{translated_predicate}".strip
         else
-          "#{attribute_name} #{predicate_name}".strip
+          "#{attribute_name} #{translated_predicate}".strip
         end
+      end
+
+      def predicate_name
+        I18n.t("active_admin.filters.predicates.#{condition.predicate.name}",
+               default: ransack_predicate_name)
       end
 
       def html_options
@@ -56,8 +63,7 @@ module ActiveAdmin
         condition_attribute.attr_name
       end
 
-      # translated predicated (equals, contains, etc)
-      def predicate_name
+      def ransack_predicate_name
         Ransack::Translate.predicate(condition.predicate.name)
       end
 
--- a/spec/unit/filters/active_filter_spec.rb
+++ b/spec/unit/filters/active_filter_spec.rb
@@ -34,6 +34,10 @@ RSpec.describe ActiveAdmin::Filters::ActiveFilter do
     expect(subject.label).to eq("Title equals")
   end
 
+  it 'should pick predicate name translation' do
+    expect(subject.predicate_name).to eq(I18n.t("active_admin.filters.predicates.equals"))
+  end
+
   context 'search by belongs_to association' do
     let(:search) do
       Post.ransack(custom_category_id_eq: category.id)
@@ -45,6 +49,10 @@ RSpec.describe ActiveAdmin::Filters::ActiveFilter do
 
     it 'should have valid label' do
       expect(subject.label).to eq("Category equals")
+    end
+
+    it 'should pick predicate name translation' do
+      expect(subject.predicate_name).to eq(Ransack::Translate.predicate('eq'))
     end
 
   end


javierjulio:
@Fivell I could be wrong, but does this fix #4622? They seem very related and since I've been going through PR's I was wondering if they were.

lib/active_admin/filters/active_filter.rb
```diff
@@ -26,13 +26,21 @@ def values
...
+          "#{attribute_name} #{translated_predicate}".strip
         end
       end
 
+      # translated active admins predicated equals, contains, starts_with ..
+      # with fallback to ransack eq, cont, start ...
+      def predicate_name
+        key = condition.predicate.name
+        I18n.t("active_admin.filters.predicates.#{key}", default: Ransack::Translate.predicate(key))
+      end
```
javierjulio: @Fivell I've learned over time that while comments aren't a bad thing they are failing us here because its telling us the code isn't as descriptive/readable as it could be. We can avoid the comments by just breaking this out with descriptive methods that make it clear what we want to do. For example, something like:

```ruby
def translated_predicate_name
  I18n.t("active_admin.filters.predicates.#{condition.predicate.name}", 
    default: fallback_to_ransack_predicate_name)
end

def fallback_to_ransack_predicate_name
  Ransack::Translate.predicate(condition.predicate.name)
end
```

Which also has the nice benefit of avoiding the local variable too.

Fivell: @javierjulio , sure , this is much better

Fivell:
@javierjulio , not sure it fixes #4622 since it is about missed locales after https://github.com/activeadmin/activeadmin/commit/9c3db9f486623962ee4e85cd15bf4ea0f2b4e896  I think

lib/active_admin/filters/active_filter.rb
deivid-rodriguez:
```
def fallback_to_ransack_predicate_name
  Ransack::Translate.predicate(condition.predicate.name)
end
```

Now we are back to the original method, just named differently and in another position in the file. I think the old name is good enough, `predicate_name` (vs the new `translated_predicate_name`).

lib/active_admin/filters/active_filter.rb
```diff
@@ -26,19 +26,29 @@ def values
       end
 
       def label
+        translated_predicate = translated_predicate_name.mb_chars.downcase.to_s
```
deivid-rodriguez: Is the `to_s` needed?

Fivell:
```
"àáâãäå".mb_chars.downcase.class
 => ActiveSupport::Multibyte::Chars 
```

deivid-rodriguez: What's special about this situation? We're not handling multibyte chars specially anywhere else, right? And this works out of the box in ruby 2.4

```shell
$ ruby --version
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
irb
irb(main):001:0> "ÀÄáâãäå".downcase
=> "àäáâãäå
```

javierjulio:
@deivid-rodriguez I've approved but did you have any concerns?

lib/active_admin/filters/active_filter.rb
javierjulio: @deivid-rodriguez we support down to Ruby 2.2. Are there any issues in 2.2 and 2.3?

deivid-rodriguez: It fallsback to not downcasing.

```shell
$ ruby --version
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
$ irb
irb(main):001:0> "ÀÄáâãäå".downcase
=> "àäáâãäå
```

deivid-rodriguez: It fallsback ~ It's buggy, you name it :)

deivid-rodriguez:
@javierjulio  No concerns, but the current diff feels a little ackward. We are renaming a method and moving it around (without changing implementation) for no apparent reason, I think.

captura de pantalla 2017-07-05 17 00 20

lib/active_admin/filters/active_filter.rb
Fivell: @deivid-rodriguez , but for cyrillic it works

str = "Привет"
 => "Привет" 
2.2.3 :002 > str.mb_chars.downcase.to_s
 => "привет" 
2.2.3 :003 > str = "Привет"
 => "Привет" 
2.2.3 :003 >str.downcase
 => "Привет"

Fivell:
@deivid-rodriguez , will change this to make cleaner diff later

lib/active_admin/filters/active_filter.rb
deivid-rodriguez: Yes, your solution works in every ruby we support.

I'm fine with it, I would maybe add a `TODO:` to remind us to go back to the simpler `str.downcase` once we support `ruby >= 2.4` only.

deivid-rodriguez:
@Fivell Force-pushed a tiny change in the commit message to retrigger the Travis build, now that it's sorted out.

PR #5036, opened 2017-06-26
Remove remaining deleted task reference

javierjulio:
We had one usage of `enforce_version` as a task dependency but it has since been removed.


deivid-rodriguez:
Nice! Will merge once CI is done!

Issue #5031, opened 2017-06-24
Support sassc-rails


swrobel:
The fact that activeadmin depends on sass-rails causes [conflicts](https://github.com/sass/sassc-rails#installing-alongside-a-gem-that-depends-on-sass-rails) with apps that try to use sassc-rails leading to [complicated](https://github.com/sass/sassc-rails/issues/6#issuecomment-277471795) [workarounds](https://github.com/sass/sassc-rails/issues/68)

Fivell:
possible solution is to require sassc-rails and sass-rails as fallback 

like 
```ruby
begin
  require 'sassc-rails'
rescue LoadError
  require 'sass-rails.'
end
```

Fivell:
@javierjulio , what do you think about it?

javierjulio:
@Fivell I'm not familiar with sassc-rails so I'm not sure what's best. First I've heard of it. What I realized though with this is why do we even require sass-rails as a dependency. I've looked at some other gems like foundation-rails which I use and am thinking perhaps its better for us to require sass as dependency instead? Since ActiveAdmin is only meant for Rails, there won't be any concern since sass-rails is the default and users can decide if they rather use something else. Wouldn't it be better to just replace the dependency with the minimum required and thus we avoid having to require various sass rails based gems to accommodate cases like this?

javierjulio:
@Fivell we require sass-rails here https://github.com/activeadmin/activeadmin/blob/11c1253d2d228cdae27537fff16d75eb091dca69/lib/active_admin.rb#L10 but I wonder why since we don't use it directly. We do have a request spec for stylesheets but I think that's unnecessary. We shouldn't be testing what Rails already tests.

Further I'm not sure why we add our stylesheets and javascripts to the precompile path since we have them in the conventional folders that Rails will just add and compile them for us (from what I understand). For example, reviewing more foundation-rails it [requires smaller dependencies despite using sass based stylesheets](https://github.com/zurb/foundation-rails/blob/master/foundation-rails.gemspec) and also [doesn't do any custom engine stuff for asset precompilation](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/engine.rb#L3-L10).

javierjulio:
I realize now we precompile as users can add their own stylesheets and javascripts too. But even then I wonder if its really necessary? Users can just import/require in the active_admin.scss and active_admin.js file generated.

javierjulio:
[...]

PR #5037, opened 2017-06-26
Require only minimum of sass as dependency

javierjulio:
This should help with issues like #5031 where users want to use sassc-rails but were not able to since we were requiring sass-rails. We could probably still have it as a dependency but simply **not** require it in Ruby code. This change was to mimic what I've seen in other gems for Rails apps that include SASS like [foundation-rails where they specify the minimum dependencies](https://github.com/zurb/foundation-rails/blob/master/foundation-rails.gemspec#L21-L23) rather than require sass-rails directly.


javierjulio:
@deivid-rodriguez another codecov/project failure. Is it ok to ignore?

Fivell:
@javierjulio , maybe this worth change-log line

deivid-rodriguez:
@javierjulio The codecov/project failure is fine, yes!

The changes here look good to me, let's wait for @swrobel's feedback!

swrobel:
Seems to be working great on my staging server! 👍

deivid-rodriguez:
Nice! @javierjulio `sass-rails` depends on `sass ~> 3.1`, should we add that here too, just in case?

PR #5038, opened 2017-06-27
Render meta tags before javascripts so they are available when JS loads.

bcavileer:
The activeadmin tests passed using `export BUNDLE_GEMFILE=gemfiles/rails_51.gemfile`

I have moved the meta tags up in the header.  This way meta tags are in the head when the JS code loads. JS code currently loads before meta tags are rendered.

### DIFF
--- a/lib/active_admin/views/pages/base.rb
+++ b/lib/active_admin/views/pages/base.rb
@@ -29,16 +29,16 @@ module ActiveAdmin
               text_node stylesheet_link_tag(style, options).html_safe
             end
 
+            active_admin_namespace.meta_tags.each do |name, content|
+              text_node(tag(:meta, name: name, content: content))
+            end
+
             active_admin_application.javascripts.each do |path|
               text_node(javascript_include_tag(path))
             end
 
             if active_admin_namespace.favicon
               text_node(favicon_link_tag(active_admin_namespace.favicon))
-            end
-
-            active_admin_namespace.meta_tags.each do |name, content|
-              text_node(tag(:meta, name: name, content: content))
             end
 
             text_node csrf_meta_tag


deivid-rodriguez:
Seems harmless towards ActiveAdmin but I wonder about the use case. Something like [this](https://stackoverflow.com/questions/7234601/is-it-possible-to-edit-meta-tag-content-via-javascript-before-the-page-load-even)?

Fivell:
@deivid-rodriguez , I think according to w3 meta should be before js and css 
https://www.w3.org/wiki/The_HTML_head_element#Stop_right_there.21_Inline_CSS_and_JavaScript_is_not_too_clever.21

deivid-rodriguez:
Mmmm, that's just a sample code to show that inline JS and CSS is bad, right? I don't see any mentions to the order of things in `<head>`...

Fivell:
@deivid-rodriguez , yes looks like no mentioning about any recommendation

Fivell:
@bcavileer explain why this needed please

javierjulio:
@Fivell because the JS is set before the meta tags and sometimes you want to use the meta tags from JS.

I've just followed the [order set on the HTML5 Boilerplate](https://github.com/h5bp/html5-boilerplate/blob/master/src/index.html#L3-L15) since I respect the backing it has in the community. I'd suggest following what's there.

deivid-rodriguez:
But you would need to use the meta tags from js _before the page is loaded_, right? Otherwise this doesn't make a difference.

I'm cool with this anyways, I'm just curious.

bcavileer:
Hi everyone, thank you for reviewing and accepting this PR.   Just to answer a few questions from above, I'm using action cable with active admin.

Action Cable can set it's path using a meta tag

https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/actioncable/app/assets/javascripts/action_cable.coffee.erb#L15

So the meta tag has to be present when the JS loads.  In my case the JS was loading before the meta tag was present, so my meta tag wasn't working and action cable was defaulting to the /cable path when I wanted to specify /admin/cable

It is possible to configure the `ActionCable.createConsumer()` with a url but I wanted to use the meta tag =)

in my config/active_admin.rb
```
config.register_javascript 'admin_cable.js'

config.meta_tags = {
  :'action-cable-url' => '/admin/cable/'
}
```
admin_cable.js contains
```
//= require action_cable
//= require_self
//= require_tree ./admin/channels

(function() {
    this.App || (this.App = {});

    App.cable = ActionCable.createConsumer();

}).call(this);
```

PR #5039, opened 2017-06-27
Update bug report script to focus on ActiveAdmin

javierjulio:
The bug report script we have is awesome but it should be focused on ActiveAdmin, not Rails. Its confusing because seeing it you think you get a pass because you wouldn't be running into a Rails issue, but an ActiveAdmin one. We want to push users to start using this executable script for bugs as its very useful and much faster to setup then a sample app. This also does some general clean up and improvements on contributing guidelines so the issue template focuses on directing users there and to use that script template.


lib/bug_report_templates/rails_5_master.rb
```diff
@@ -51,32 +49,28 @@ class ApplicationController < ActionController::Base
 
 class TestApp < Rails::Application
   config.root = File.dirname(__FILE__)
```
deivid-rodriguez: Rails uses `__dir__` now here, which is simpler.

lib/bug_report_templates/rails_5_master.rb
```diff
@@ -51,32 +49,28 @@ class ApplicationController < ActionController::Base
...
 end
 
-# create models
 class YourModel < ActiveRecord::Base
 end
 
-# configure active_admin
 ActiveAdmin.setup do |config|
-  # Authentication disabled by default. Override if necessary
+  # Authentication disabled by default. Override if necessary.
```
deivid-rodriguez: You removed all comments from the template except for this one. Any reason?

lib/bug_report_templates/active_admin_master.rb
```diff
@@ -8,22 +8,23 @@
 gemfile(true) do
   source 'https://rubygems.org'
 
-  gem 'rails', require: false
-  gem 'sqlite3', platform: :mri
+  # Use local changes or ActiveAdmin master.
+  if ENV['ACTIVE_ADMIN_PATH']
+    gem 'activeadmin', path: ENV['ACTIVE_ADMIN_PATH'], require: false
```
timoschilling: did you know this
`bundle config local.rack ~/Work/git/rack`
http://bundler.io/v1.5/man/bundle-config.1.html#LOCAL-GIT-REPOS

lib/bug_report_templates/rails_5_master.rb
```diff
@@ -58,7 +54,11 @@ class TestApp < Rails::Application
   Rails.logger  = config.logger
 end
 
-class YourModel < ActiveRecord::Base
+class ApplicationController < ActionController::Base
+  include Rails.application.routes.url_helpers
```
deivid-rodriguez: I guess this change belongs in the previous commit... :) I'm a git history freak, I know.

CONTRIBUTING.md
```diff
@@ -68,7 +68,7 @@ and a **code sample** or an **executable test case** demonstrating the expected
 * If possible, use the relevant bug report templates to create the issue.
 Simply copy the content of the appropriate template into a .rb file, make the necessary changes to demonstrate the issue,
 and **paste the content into the issue description**:
-  * [ActiveAdmin master](https://github.com/activeadmin/activeadmin/blob/master/lib/bug_report_templates/active_admin_master.rb)
+  * [**ActiveAdmin** master issues](https://github.com/activeadmin/activeadmin/blob/master/lib/bug_report_templates/active_admin_master.rb)
```
deivid-rodriguez: Squashable with 95774f09247eb9de8ee9a3e7277da7dca2e0d1e7. :)

lib/bug_report_templates/rails_5_master.rb
javierjulio: Because the others are self explanatory, ActiveAdmin.setup, Rails.app.initialize, etc. I liked having this here because its common to have authentication for AA so its a nice heads up that as a convenience its disabled.

lib/bug_report_templates/active_admin_master.rb
javierjulio: No I didn't know about, neat! I didn't want to change this because its how it works with the spec and also because of time haha. Is that ok? Or you prefer we use that? We can always address separately too.

javierjulio:
@deivid-rodriguez squashes several related commits and addressed all your changes. Thanks!

javierjulio:
@deivid-rodriguez the reason they are wrapped in HTML comments is so they aren't show in Preview but I think that's worse. Its helpful to view in preview since it is markdown and thus you are able to click on links like the contributing guide link to quickly get to it. I've noticed the same approach with other major projects like Rails and Sidekiq. They keep it short and sweet, directing users to the contributing guidelines.

Since I have your approval, I'm good with the changes here so I will merge in soon. I can revert that issue template change if its really necessary but I think its best to leave out the HTML comments. As admins we can update any issue description so if the help text is that bothersome it could be removed but I have faith most users will know to replace it.

deivid-rodriguez:
Agreed, merge away!

deivid-rodriguez:
Oh, I understood now, after this change, users need to remove the text themselves. Before it would be hidden without user intervention.

I can see both points, and it seems that we can't get the best of both worlds, so let's see how this works out.

javierjulio:
@deivid-rodriguez after merging this into master, that build failed due to jruby again. 😞 The build is https://travis-ci.org/activeadmin/activeadmin/jobs/248076568 and failed with a message of "The job exceeded the maximum time limit for jobs, and has been terminated." which appears at the end of the build log.

deivid-rodriguez:
#5033 should fix that, let's merge it.

PR #5042, opened 2017-06-28
Update 2-resource-customization.md

faberge-eggs:
Otherwise scope will be trying to add the posts method as association_method

### DIFF
--- a/docs/2-resource-customization.md
+++ b/docs/2-resource-customization.md
@@ -303,7 +303,7 @@ ActiveAdmin.register Post do
   scope_to :current_user, association_method: :blog_posts
 
   # Finally, you can pass a block to be called:
-  scope_to do
+  scope_to association_method: :itself do
     User.most_popular_posts
   end
 end


docs/2-resource-customization.md
```diff
@@ -303,7 +303,7 @@ ActiveAdmin.register Post do
   scope_to :current_user, association_method: :blog_posts
 
   # Finally, you can pass a block to be called:
-  scope_to do
+  scope_to association_method: :itself do
```
Fivell: @faberge-eggs , what `:itself` means?

Fivell: @faberge-eggs , please add more details I never used this syntax, I just can see that 
https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/lib/active_admin/resource_controller/scoping.rb#L26 it used here , do not understand 
how `:itself` helps

faberge-eggs:
Sorry for with hasty PR. The issue is more complicated and I've created a new pull request with fix 
https://github.com/activeadmin/activeadmin/pull/5058
We could discuss there

Issue #5043, opened 2017-06-28
When custom scopes are defined, but none are specified as "default", an error happens when searching


andey:
` revision: 417ff4be6d44ef2ef4b35e24f25ac686478f9e3c`

### Expected behavior

specifying a "default" scope, should not be necessary

### Actual behavior

ActionView::Template::Error: undefined method `name' for nil:NilClass

```````

File "/home/rof/cache/bundler/ruby/2.3.0/bundler/gems/activeadmin-417ff4be6d44/lib/active_admin/view_helpers/scope_name_helper.rb" line 6 in scope_name
File "/home/rof/cache/bundler/ruby/2.3.0/gems/arbre-1.1.1/lib/arbre/element.rb" line 180 in method_missing
File "/home/rof/cache/bundler/ruby/2.3.0/bundler/gems/activeadmin-417ff4be6d44/lib/active_admin/filters/active_sidebar.rb" line 17 in block (2 levels) in block
File "/home/rof/cache/bundler/ruby/2.3.0/gems/arbre-1.1.1/lib/arbre/element/builder_methods.rb" line 31 in block in build_tag
File "/home/rof/cache/bundler/ruby/2.3.0/gems/arbre-1.1.1/lib/arbre/context.rb" line 92 in with_current_arbre_element
File "/home/rof/cache/bundler/ruby/2.3.0/gems/arbre-1.1.1/lib/arbre/element/builder_methods.rb" line 49 in with_current_arbre_element
File "/home/rof/cache/bundler/ruby/2.3.0/gems/arbre-1.1.1/lib/arbre/element/builder_methods.rb" line 26 in build_tag
File "/home/rof/cache/bundler/ruby/2.3.0/gems/arbre-1.1.1/lib/arbre/element/builder_methods.rb" line 39 in insert_tag
File "/home/rof/cache/bundler/ruby/2.3.0/gems/arbre-1.1.1/lib/arbre/element/builder_methods.rb" line 14 in span
```````

### How to reproduce

1. Define a scope in the model

`````
class User < ActiveRecord::Base
  scope :admins, -> { where(admin: true) }
end
`````

2. Define the matching scope in the active_admin level, do not specify it being the "default"

`````
ActiveAdmin.register User do
  scope :all
  scope :admins
end
`````

3. Search anything using a filter

### How to temporarily fix

`````
ActiveAdmin.register User do
  scope :all, default: true
  scope :admins
end
`````

PR #5044, opened 2017-06-28
fixed current_scope rendering in ActiveSidebar

Fivell:
refs #5043


javierjulio:
I ran into this issue last week or two on master as well. I solved it though because I realized in some of my admin pages I didn't specify what scope was the default when the documentation demonstrates it to be required. I figured it was intended. Best to fix #5043. Change looks good!

PR #5045, opened 2017-06-29
handle visibility of html elements in tests

Fivell:


features/show/tabs.feature
```diff
@@ -23,5 +24,8 @@ Feature: Show - Tabs
       end
     """
     Then I should see two tabs "Overview" and "Details"
-    And I should see "tab 1"
-    And I should see "tab 2"
+    And I should see the element "#overview span"
+    And I should not see the element "#details span"
+    Then I follow "Details"
+    And I should not see the element "#overview span"
+    And I should see the element "#details span"
```
deivid-rodriguez: I see how the test is more complete now, but why not checking the tab content like before? Also (only partially related), do activeadmin tabs work without javascript?

Fivell: @deivid-rodriguez , suddenly when I use
```And I should see "tab 1"``` it always return true  because it just check content of page (no matter visible or not) so I changed it to css selectors to make it at least display difference between tabs. 
and yes it doesn't work without js (do not change visibility of content)

Fivell: basically I need this to be merged to complete #5046

deivid-rodriguez: > basically I need this to be merged to complete #5046

It's cool, we can improve it later if needed!

> and yes it doesn't work without js (do not change visibility of content)

Noted!

> And I should see "tab 1" it always return true because it just check content of page (no matter visible or not) so I changed it to css selectors to make it at least display difference between tabs.

What's this "I should see" step? I grepped for it, but couldn't find where it's defined. Is it a `cucumber` built-in or something? I wonder why it does not check just for visible elements, seems like it should be the default, like in `capybara`.

deivid-rodriguez: Also found this one: https://github.com/activeadmin/activeadmin/blob/master/features/step_definitions/tab_steps.rb#L1-L3, which could be useful here.

Fivell: @deivid-rodriguez , this was added by @gregbell  before tabs was added to AA, it is about navigation tabs, so this is naming gap,  I think it should be  separated to different step files

PR #5051, opened 2017-06-29
Remove unused bourbon gem

### LINKED ISSUE #4780: Linear-gradient function from Bourbon is deprecated

javierjulio:
I thought we were using bourbon but realized we weren’t using it to begin with so there is no need for the dependency. We have our own SASS mixins that we use but none from Bourbon. Testing locally against master, our components appear just fine, including things like gradients so this is safe removal now. No need to wait until v2.

I misread #4780 entirely since Bourbon only works with the `@include` syntax and but in our gradient mixin, we provide the explicit value of `linear-gradient` for `background-image` so we weren't even using Bourbon.

Fixes #4780 and #4791

### DIFF
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -19,7 +19,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.2'
 
   s.add_dependency 'arbre', '>= 1.1.1'
-  s.add_dependency 'bourbon'
   s.add_dependency 'coffee-rails'
   s.add_dependency 'formtastic', '~> 3.1'
   s.add_dependency 'formtastic_i18n'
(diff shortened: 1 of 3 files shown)


javierjulio:
Removing dependencies.. its the best kind of party. 😎

PR #5052, opened 2017-06-29
Use jQuery v3 as default

### LINKED ISSUE #5012: Javascript Errors
I can't seem to get the included javascript file to run without any errors. Basically, it seems like jquery UI dialog isn't loaded before dialog.

```
Uncaught TypeError: $.widget is not a function
    at dialog.self-8b9019c….js?body=1:46
    at dialog.self-8b9019c….js?body=1:42
    at dialog.self-8b9019c….js?body=1:44

Uncaught TypeError: Cannot read property 'mouse' of undefined
    at sortable.self-e213a88….js?body=1:37
    at sortable.self-e213a88….js?body=1:33
    at sortable.self-e213a88….js?body=1:35

Uncaught TypeError: $.widget is not a function
    at tabs.self-3c3b343….js?body=1:39
    at tabs.self-3c3b343….js?body=1:35
    at tabs.self-3c3b343….js?body=1:37
```

I tried to resolve this by updating the jquery_ui file (https://github.com/activeadmin/activeadmin/blob/master/app/assets/javascripts/active_admin/jquery_ui.js.erb) by placing widget above dialog, but this just resulted in `base is not a constructor`error.

Have you seen anything like this? I am not sure what is unique about my environment that it's messing with the JS loading. 

Relevant gems:

```
gem "rails", "~> 5.0.0"
gem 'jquery-rails'                                                                                                                                              
gem 'jquery-ui-rails', '~> 6.0.1'
```

*/assets/javascripts/active_admin.js.coffee*:
```#= require active_admin/base```

### LINKED ISSUE #5047: Upgrade default to jQuery v3
We are still running the admin on jQuery v1 so lets upgrade to v3 and see what happens.

### PR DESCRIPTION
**Pull request by javierjulio**
Going to let tests run to see how things work.

Fixes #5047 
Fixes #5012

### CODE REVIEW

**Comment by javierjulio** [comment]
As far as the test failures, its happening just with batch actions but the [error is in jQuery UI](https://github.com/jquery/jquery/issues/3181) so it affects any widget. Its not going to [work with jQuery 3 unless you include the jQuery migrate plugin](https://github.com/jquery/jquery/issues/3181#issuecomment-226997483). We can fix this a few ways:

* use jQuery v3 and include the jQuery migrate plugin in our repo and require in `base.js.coffee`
* use jQuery v2 instead

There are production builds of the jQuery migrate plugin so keep in mind that option isn't abnormal and if anything common. Let me know what you think.

I'd very much like want to drop jQuery UI to begin with and use smaller component focused libraries so this ideally will be solved that way but we can still upgrade jQuery in the meantime. Dropping jQuery UI would be a much bigger change and we are looking into that in #5048.

**Comment by javierjulio** [comment]
@deivid-rodriguez @Fivell I'm wondering if its just better to hold off on upgrading jQuery (although its just fine with this or jQuery v2) and just tackle this as part of AA v2 where one objective is to remove jQuery UI and thus we can safely upgrade to jQuery v3?

**Comment by deivid-rodriguez** [comment]
I had a quick look at this. The incompatibility you linked to is with `jquery-ui 1.11`, but it shouldn't happen with `1.12` (which is shipped with `jquery-ui-rails 6.0`). The problem is that our tests are using `jquery-ui-rails 5.0`, that's why they failed. We currently support both versions though, so it doesn't really fix anything, just explains the failures.

I think an "intermediate solution" as a middle step towards removing the `jquery-ui` dependency would be to vendor it (just what we are using). That way we can drop the dependency from the gemspec, and we don't need to handle compatibility with [multiple incompatible versions](https://github.com/activeadmin/activeadmin/blob/5b8ded8bcc1cd3ed09ea5c10fc992e47614d4bd1/app/assets/javascripts/active_admin/jquery_ui.js.erb). And we are more aware of what we need to reimplement or replace somehow.

**Comment by javierjulio** [comment]
@deivid-rodriguez great! I'll give that a shot. Essentially same thing as here but instead of vendoring the migrate plugin we should just include the files directly that we are using from jQuery UI (jquery-ui-rails v6).

**Comment by deivid-rodriguez** [comment]
Exactly, and we no longer need the compatibility tricks nor the gemspec dependency.

**Comment by javierjulio** [comment]
@deivid-rodriguez the trade off is that we will have to add a lot of files to the vendor directory for jQuery UI. Each of the files we require in jquery_ui.js.erb in turn require a lot more files (e.g. dialog has 13 other dependent files). You ok with that?

**Comment by deivid-rodriguez** [comment]
I'm fine with it. @varyonic?

**Comment by javierjulio** [comment]
When you see the amount of JS files added don’t freak out. The dialog.js has the most amount of dependencies by far! Its dependencies have other dependencies too that aren't used elsewhere. 😱 Biggest win we’ll be first removing dialog usage. The harder part though will be finding a suitable sortable JS solution.

**Comment by javierjulio** [comment]
For the jQuery UI files in the vendor directory, I used code from the latest stable tagged release of jquery-ui-rails which is v6.0.1. I've tried it locally and it works great. Since I'm using the same files from the gem, all the requires work fine since its no different then from using the gem. With this approach we can kill off components as we replace them. The biggest win should be replacing the dialog component since that had the most dependencies (its deps also had several one off deps) across the board.

**Comment by deivid-rodriguez** [comment]
Do calendars, for example, still display ok? In other words, do we need to vendor some css here as well?

**Comment by javierjulio** [comment]
@deivid-rodriguez oddly enough no because we weren't even requiring their CSS to begin with. 😳🤣 I double checked the jquery-ui-rails docs and it states we'd have to [require their styles](https://github.com/jquery-ui-rails/jquery-ui-rails#require-everything) for it to work. It makes sense since I remember when working on #3862, for example, that the date picker styles were included directly (customized) rather than importing from jquery-ui-rails. 

I've done testing with the updated test app template and it works just fine. The dragging and sorting for "has many" forms works great, preserves updated order. The date picker works too. Everything should appear the same as before on master since again no CSS was imported from the jquery-ui-rails gems.

**Comment by deivid-rodriguez** [comment]
Nice, that's what I wanted to know :)

One last super tiny thing. I grepped for `jquery-ui` and found one leftover documentation reference to the gem.

```
lib/active_admin/dependency.rb
41:    # ActiveAdmin::Dependency['jquery-ui-rails'] < 5
```

We could replace that with `jquery-rails` which we still depend on.

javierjulio:
Yes I noticed that too but didn't change because it was a helpful example with dashes. I forgot though that jquery-rails fits that. I'll do the update now.

deivid-rodriguez:
Approved. I would wait for confirmation from @varyonic. Just because it's ugly to commit 11K lines and then revert it :laughing: .

javierjulio:
@deivid-rodriguez ok cool. @varyonic we are in agreement with this approach. Let us know if you have concerns and we will help address them. This ideally is better than the migrate plugin approach since the real issue is that we were allowing multiple versions of jquery-ui through the gem. Also we don't want to force a gem version as we want to make it painless for users to update to v1.1. With this PR nothing changes for them and they get the benefit of the latest jQuery and jQuery UI. Further this allows us to replace each jQuery UI component with our own and we can get a good comparison at how much we are saving in bloat.

javierjulio:
@deivid-rodriguez confirmed in Slack that @varyonic still approves. Any other concerns before I merge this? I'm under the impression we get this out for v1.1.

deivid-rodriguez:
Nope, no concerns! Hit the button!

PR #5053, opened 2017-06-29
Enhance codecov config a bit

deivid-rodriguez:
When PRs remove code, the global coverage percentage decreases. Codecov is too strict about this (see #5037 or #5051).

I think we can get around this by reducing a bit the precision of the bot's numbers. These checks are just to give an idea about the PR's coverage and how it affects global coverage, but nobody expects it to be very precise.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,13 +1,20 @@
+---
+
 codecov:
   notify:
     require_ci_to_pass: true
+
 comment: off
+
 coverage:
   precision: 2
+
   range:
-  - 70.0
-  - 100.0
+    - 70.0
+    - 100.0
+
   round: down
+
   status:
     changes: false
     patch: true
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,13 +7,13 @@ codecov:
 comment: off
 
 coverage:
-  precision: 2
+  precision: 1
 
   range:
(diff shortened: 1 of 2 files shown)


javierjulio:
Why not remove it from the github status feature of the codecov checks? That way it doesn't show up with a red X that the PR failed since its just for information. Not necessarily merge blocking. Or you want it to be?

deivid-rodriguez:
I think it's not a required check at the moment. Only Travis and a human approved are required, I think.  The red cross is just annoying...

deivid-rodriguez:
It's also weird that coverage was around 98% two days ago, and now it's around 93%...

deivid-rodriguez:
The decrease in coverage is now fixed via https://github.com/activeadmin/activeadmin/pull/5053/commits/73423bd8cc2b9785f77ebde410357d8ca5d590b1. It was a regression I recently introduced.

Back to 98.1% now!

deivid-rodriguez:
@javierjulio Funny I just discovered you are a contributor to https://github.com/eval/envied. I guess `envied` helps dumb people like me to not being fooled by ruby's `ENV` like in here? :)

deivid-rodriguez:
Thanks @javierjulio! :)

PR #5057, opened 2017-07-01
"Forgot your password?" and "Sign up" links are overlapping

bryndyment:
image


javierjulio:
@bryndyment thanks for the fix! I approved initially but its not the right fix since it keeps the links inline and there can be many depending on how many features you have enabled. 

I decided its just easiest to bring back the br tags that I removed initially. Easiest solution for now as I wanted to update the layouts/styles but that's best tackled in #3862/v2.

bryndyment:
👍

PR #5058, opened 2017-07-03
Fix the default scope error when scope_to is used with a block

faberge-eggs:
In the docs
https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#scoping-the-queries
there is the case
```
  scope_to do
    User.most_popular_posts
  end
```
But in this case, AA will use inherited_resource association method 
https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/resource_controller/scoping.rb#L25

And it will be error, because AA is trying to execute a request like this 
`User.most_popular_posts.posts`

In my case, I try to use Pundit as authorisation(Admin can view all records, users restricts by some logic)
```
  scope_to do
    Pundit.policy_scope(current_manager, TradePoint)
  end
```
Also AA tries to use `build` method except `new` for new records if score exists

### DIFF
--- a/features/index/index_scope_to.feature
+++ b/features/index/index_scope_to.feature
@@ -11,7 +11,7 @@ Feature: Index Scope To
       ActiveAdmin.register Post do
         # Scope section to a specific author
         scope_to do
-          User.find_by_first_name_and_last_name "John", "Doe"
+          User.find_by_first_name_and_last_name("John", "Doe").posts
         end
 
         # Set up some scopes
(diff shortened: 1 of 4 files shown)


timoschilling:
For me it doesn't make sence to use `scope` with a `block`, your example should be resolved with `scoped_collection`, `default_scope` or the AuthorisationAdapter.

faberge-eggs:
In this case may be we should fix docs. For now, an association method is implicitly used for a block. I think this is not expected behaviour

javierjulio:
Since it's been so long it seems this may not be much of a need now. Last note was about updating the docs which I'd be happy to accept any improvements there. Feel free to open a new PR for that. Thank you.

PR #5060, opened 2017-07-05
Deprecate old asset registration with upgrade messages

javierjulio:
No need to keep this around since its predates the asset pipeline! 😳 Which is the preferred and best way to do this since it allows users to import what they like through the generated stylesheet and javascript files. In case anyone is using this config based approach though lets give them a heads up so they can migrate before v2 where it will be removed.

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -271,10 +271,12 @@ module ActiveAdmin
   private
 
     def register_default_assets
-      register_stylesheet 'active_admin.css',       media: 'screen'
-      register_stylesheet 'active_admin/print.css', media: 'print'
+      ActiveSupport::Deprecation.silence do
+        register_stylesheet 'active_admin.css',       media: 'screen'
+        register_stylesheet 'active_admin/print.css', media: 'print'
 
-      register_javascript 'active_admin.js'
+        register_javascript 'active_admin.js'
+      end
     end
 
     # Since app/admin is alphabetically before app/models, we have to remove it
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -271,12 +271,9 @@ module ActiveAdmin
   private
 
     def register_default_assets
-      ActiveSupport::Deprecation.silence do
-        register_stylesheet 'active_admin.css',       media: 'screen'
-        register_stylesheet 'active_admin/print.css', media: 'print'
-
-        register_javascript 'active_admin.js'
-      end
+      stylesheets['active_admin.css'] = { media: 'screen' }
+      stylesheets['active_admin/print.css'] = { media: 'print' }
+      javascripts.add 'active_admin.js'
     end
 
     # Since app/admin is alphabetically before app/models, we have to remove it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Deprecations
 
+* Deprecated `config.register_stylesheet` and `config.register_javascript`. Import
+  your CSS and JS files in `active_admin.scss` or `active_admin.js`. [#5060][] by [@javierjulio][]
 * Deprecated `type` param from `status_tag` and related CSS classes [#4989][] by [@javierjulio][]
   * The method signature has changed from:
     ```ruby
--- a/features/registering_assets.feature
+++ b/features/registering_assets.feature
@@ -18,8 +18,10 @@ Feature: Registering Assets
   Scenario: Registering a CSS file
     Given a configuration of:
     """
-      ActiveAdmin.application.register_stylesheet "some-random-css.css", media: :print
-      ActiveAdmin.register Post
+      ActiveSupport::Deprecation.silence do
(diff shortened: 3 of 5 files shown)


javierjulio:
@deivid-rodriguez codecov failures again but all the tests passed.

javierjulio:
@deivid-rodriguez if you merge in #5066, should I push an update here to get codecov to pass?

deivid-rodriguez:
Just merged it. Give it a try!

lib/active_admin/application.rb
```diff
@@ -271,10 +271,12 @@ def controllers_for_filters
   private
 
     def register_default_assets
-      register_stylesheet 'active_admin.css',       media: 'screen'
-      register_stylesheet 'active_admin/print.css', media: 'print'
+      ActiveSupport::Deprecation.silence do
```
deivid-rodriguez: Maybe we can skip calling `register_{stylesheet,javascript}` internally so we don't have to silence the deprecations.

javierjulio: This was the easiest solution but we can skip it. We'd have to delete the code here though and add these as defaults in AssetRegistration. Do you prefer that?

deivid-rodriguez: I mean doing `stylesheets["active_admin.css"] = { media: "screen"}` and so on directly, so our own code raises no deprecations.

javierjulio: Ah right we have access to that. Didn't think of that! I'll get that change in. Thanks.

deivid-rodriguez: Although after a second though, the current approach looks fine too! Do what you think it's best...

javierjulio: Haha now you tell me! After I make the change! 🙄😂 Its done and if anything better change for when we remove this as we can just either remove asset registration or just use the module to have the javascripts and stylesheets methods.

deivid-rodriguez: Yes, sorry, I realized too late that there's little gain in what I proposed, I think.

javierjulio: No problem at all buddy. I pushed the changes. Let me know what you think. Tests passed for me locally. Should be green.

deivid-rodriguez:
I have another thought here. Again, sorry for bringing this up just now.

Doesn't it make more sense to bring these deprecations once (or at the same time) we actually remove the feature? You normally start doing something in a new different way, but keep support for the old way of doing it for a period, to let them users know about the new way without breaking their apps.

In this case, nothing is changing internally, we are just kind of "promising" users we'll drop the feature. Doesn't seem worth it. What if we end up not dropping this for some reason? Because it's too difficult, or because we change our minds?

javierjulio:
@deivid-rodriguez haha I will definitely be dropping this so if you are worried about who's going to submit that PR its 100% will be me. 😏 Personally, I don't see any reason to have this. You can import what you need through the generated AA asset files. This is just more code that we have to support and maintain. 👎🏻 The asset pipeline which we use solves this issue. 👍🏻 Not only do I not see a case for using this, I'd rather have a clean slate if we were to come up with a similar solution. I'd rather not account for a hypothetical. In fact I think its better to deprecate. We might even get messages because someone is using this. Either way I feel strongly we should remove cruft like this.

deivid-rodriguez:
@javierjulio Yes, I'm overthinking this. You definitely know this stuff better than me! :)

phillc:
Hello,

I'm not 100% sure I am posting this to the correct location, however this is where the changelog directs to for `config.register_javascript` being deprecated.

I'm not sure that it is correct to remove that feature. I agree it does not make sense to use register_javascript to include multiple scripts, given that the asset pipeline is a better solution. However, without register_javascript, it becomes much more difficult to include an external javascript library, that might be dynamic.

Examples might be google analytics or google maps... dependencies that really should not be brought into the asset pipeline, as they can change.

The only way I can think of to bring those in without using register_javascript, would be to figure out how to inject them into the actual layout of active admin, but that is a quite difficult task.

-Thanks

introvert:
Completely agree with what phillc wrote, this was unnecessary.

PragTob:
Fun that I get here 18 minutes later, but I'm facing exactly the same problem as @phillc - might wanna open a separate issue though :)

yaaase:
+1 to what @phillc said

ivanyv:
+1 to bringing this back, unless there's another way to insert external scripts?

esb:
+1 to restore this as well. Of course the Asset Pipeline is the way to go, but I can also see the need to be able to automatically inject assets dynamically. Been looking at the sortable tree add-on and this uses register_javascript to inject the correct script based on the version of jQuery that's installed.

faucct:
One more reason to keep the methods: it is a good place for pluggable engines to add their assets.

an-nasir:
+1 to what @phillc said

deivid-rodriguez:
I plan to bring back this feature for ActiveAdmin 2.0 as [stated in this thread](https://github.com/activeadmin/activeadmin/issues/5170#issuecomment-445549181). After AA 2.0 we can discuss what's the best replacement for it (if we can find any).

PR #5063, opened 2017-07-05
add new asset inheritable_setting's

timoschilling:
This is just a Proposal, maybe @deivid-rodriguez like to continue the work.


javierjulio:
@timoschilling I'm tackling this for v2. I've already deprecated this in #5060. In v2 we won't have a separate print stylesheet because its unnecessary. We can include it in the main stylesheet using print media. This way we avoid an extra request. Its long been supported by browsers (d1ab92a23470618c3585b0e25205f3b81294282e).

timoschilling:
>  In v2 we won't have a separate print stylesheet because its unnecessary. We can include it in the main stylesheet using print media.

👍 

As I already said in the chat, we should keep a option to inject different javascript / css for different namespaces. I think

javierjulio:
@timoschilling haha I wasn't expecting you to reply so quickly. I didn't even have Slack open so I missed that you replied. My bad. I'm still a 👎🏻 on the css/js setting even for namespaces. I'd be more concerned about how you can configure namespaces but from what I read in the docs I don't think its fleshed out enough. If anything I say lets tackle namespaces in a future update.

javierjulio:
Going to close out as I will address this in a new PR since we just merged #5060 where we add deprecation warnings.

PR #5069, opened 2017-07-06
Add default templates for Kaminari

### LINKED ISSUE #4173: Pagination looks broken

javierjulio:
We should have Kaminari's default templates at a minimum just like we do for Devise regardless if we change them or not (although we will be changing them in #3862). We should include the defaults now to fix this longstanding issue where a host app defines its own Kaminari templates but because ActiveAdmin doesn't, it inherits those and breaks the pagination display. 

By including these defaults we don't break the host app that has custom templates since AA will always use what we have set. Plus, if the host app **does not** have any templates set, it will use the Kaminari default as expected and not what AA has set. All around we fix any possible issue of one overriding the other so we are safer by including our own templates for Kaminari than not.

Fixes #4173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (unreleased)
 
+* Fixed [#4173](../../issues/4173) by including the default Kaminari templates [#5069](../../pull/5069) by [@javierjulio][]
+
 ### Removals
 
 * Ruby 2.1 support has been dropped [#5002][] by [@deivid-rodriguez][]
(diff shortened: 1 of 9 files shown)


javierjulio:
@deivid-rodriguez great thanks! Just remembered, let me add a changelog notice so others see this coming. Its a very helpful fix.

javierjulio:
@deivid-rodriguez for some reason this was marked stale despite all being done today. I'm going to merge since only change was for the changelog. I'm happy though to revisit if you have more concerns. Thanks.

Issue #2677, opened 2013-11-08
Circular dependency detected


tibo:
Hello,

I'm working on a rails 4 upgrade and I'm experiencing a runtime error:
Circular dependency detected while autoloading constant AdminUser (RuntimeError)

any idea how can I solve this?

Thanks

aug-riedinger:
I still having the issue with the following gem versions:

```
    activeadmin (1.0.0.pre1)
    rails (4.2.0)
```

Trying to run `rake db:migrate` to set up my environment locally. Here's my stacktrace:

```
rake aborted!
Circular dependency detected while autoloading constant FeaturedPartner
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:492:in `load_missing_constant'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:184:in `const_missing'
/home/augustin/Workspace/gavoune/app/admin/featured_partner.rb:1:in `<top (required)>'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:457:in `load'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:457:in `block in load_file'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:647:in `new_constants_in'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:456:in `load_file'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:354:in `require_or_load'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:494:in `load_missing_constant'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:184:in `const_missing'
/home/augustin/Workspace/gavoune/app/admin/featured_partner.rb:1:in `<top (required)>'
/home/augustin/.rvm/gems/ruby-2.2.1/gems/activesupport-4.2.0/lib/active_support/dependencies.rb:268:in `load'
[...]

Issue #2334, opened 2013-07-10
ArgumentError: A copy of ApplicationController has been removed from the module tree but is still active


subhashb:
I am using the latest master, with Rails: 4.0.0.rc2
I experience the above exception, only when I change something in admin/*.rb files, and try to refresh the view.

The entire gist of the problem is here: [Stackoverflow 17561697](http://stackoverflow.com/questions/17561697/argumenterror-a-copy-of-applicationcontroller-has-been-removed-from-the-module)

I think this is related to ActiveAdmin + InheritedResources generated controllers "not being" in the autoload path of Rails, but I might be wrong.

Not sure if this is related to #697 either.

Is there a way to circumvent the problem?

monteirobrena:
@mindhalt you are my hero https://github.com/activeadmin/activeadmin/issues/2334#issuecomment-40253691

prognostikos:
Fully qualifying all constants in our app's ApplicationController fixed this issue for us.

Rails 4.2 with ActiveAdmin 1.0.0.pre1.

JonRowe:
I get this on Rails 4.2 with ActiveAdmin 1.0.0.pre1 but with `ApplicationHelper` rather than `ApplicationController`

ChristianPeters:
> fully qualifying the constant name, like so ::CommonPresenter.new

Thanks @subhashb for the workaround. Would be great to have this fixed – we restarted the server hundreds of times before starting looking for this. x)

jayshepherd:
We also had this problem and fully qualifying our objects in the ApplicationHelper was the trick.

So, to be clear, I looked for any of our specific models and preceded with `::`

Example:
A reference to `UserSession` was updated to `::UserSession`

westonplatter:
fully qualifying ALL constants solved the issue for: Rails 4.1.15 // ActiveAdmin 1.0.0.pre2

Example:
A reference to UserSession was updated to ::UserSession

mklnz:
I'm still getting the circular dependency issue with Rails 4.2.7 and latest ActiveAdmin from git
After doing a fresh install, the generator adds app/admin/admin_user.rb, but seems to conflict with app/models/admin_user.rb

Coming from an older version I recall all models were plural, but now it seems they are singular?

```
RuntimeError: Circular dependency detected while autoloading constant AdminUser
[...]

PR #5076, opened 2017-07-08
Fix sometimes broken ActiveAdmin.application.load_paths

mluts:
Rails.root is always `nil` when active_admin/application.rb is loaded.

Initialize this setting at it's first-time use, so `Rails.root` will
be already there.

My application don't start when `CWD` is `$HOME`, `load_paths` don't clean-up and my application crashes because of circular dependency that is caused by same file names in `app/admin` and `app/models` directories.

After these changes tests are green without setting `load_paths` in `application_spec.rb`

### DIFF
--- a/lib/active_admin/helpers/settings.rb
+++ b/lib/active_admin/helpers/settings.rb
@@ -22,6 +22,10 @@ module ActiveAdmin
     end
 
     def read_default_setting(name)
+      if (value = default_settings[name]).respond_to?(:call)
+        default_settings[name] = value.call
+      end
+
       default_settings[name]
     end
 
--- a/lib/active_admin/helpers/settings.rb
+++ b/lib/active_admin/helpers/settings.rb
@@ -22,11 +22,7 @@ module ActiveAdmin
     end
 
     def read_default_setting(name)
-      if (value = default_settings[name]).respond_to?(:call)
-        default_settings[name] = value.call
-      end
-
-      default_settings[name]
+      self.class.unwrap_value(default_settings[name])
     end
 
     private
@@ -36,6 +32,19 @@ module ActiveAdmin
     end
 
     module ClassMethods
+      WrappedValue = Struct.new(:block) do
+        def unwrap
+          block.call
+        end
+      end
+
+      def wrap_value(&block)
+        WrappedValue.new(block)
+      end
+
+      def unwrap_value(value)
+        value.is_a?(WrappedValue) ? value.unwrap : value
+      end
 
       def setting(name, default)
         default_settings[name] = default
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -24,7 +24,7 @@ module ActiveAdmin
     # Load paths for admin configurations. Add folders to this load path
     # to load up other resources for administration. External gems can
     # include their paths in this load path to provide active_admin UIs
-    setting :load_paths, [File.expand_path('app/admin', Rails.root)]
+    setting :load_paths, proc { [File.expand_path('app/admin', Rails.root)] }
 
     # The default number of resources to display on index pages
     inheritable_setting :default_per_page, 30
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -24,7 +24,7 @@ module ActiveAdmin
     # Load paths for admin configurations. Add folders to this load path
     # to load up other resources for administration. External gems can
     # include their paths in this load path to provide active_admin UIs
-    setting :load_paths, proc { [File.expand_path('app/admin', Rails.root)] }
+    setting :load_paths, wrap_value { [File.expand_path('app/admin', Rails.root)] }
 
     # The default number of resources to display on index pages
     inheritable_setting :default_per_page, 30
--- a/spec/unit/application_spec.rb
+++ b/spec/unit/application_spec.rb
@@ -5,16 +5,6 @@ RSpec.describe ActiveAdmin::Application do
 
   let(:application) { ActiveAdmin::Application.new }
 
-  around do |example|
-    old_load_paths = application.load_paths
-    # TODO: Figure out why load paths need to be overriden
-    application.load_paths = [File.expand_path('app/admin', Rails.root)]
-
-    example.call
-
-    application.load_paths = old_load_paths
-  end
-
   it "should have a default load path of ['app/admin']" do
     expect(application.load_paths).to eq [File.expand_path('app/admin', Rails.root)]
   end


deivid-rodriguez:
This looks reasonable to me.

deivid-rodriguez:
@varyonic Looks reasonable too, and we wouldn't have to start supporting "callable settings".

@mluts What do you think about the alternative approach of explicitly setting the root in an initializer? Also, what's the easiest way to reproduce the problem in a real world application? I know we can do it by removing the "hack" from the unit test, by I wonder how to reproduce it "for real".

mluts:
@deivid-rodriguez it's pretty simple to reproduce, just have same file name in `app/admin` and `app/models`, and boot application from directory other than `Rails.root`, for example run `~/my_application/bin/rails runner "puts MyModel"` from `~`.

Just run this script to see the bug in action: https://gist.github.com/mluts/7f6fb56574d92b30d17f2f213d208900

Personally i met this problem after writing such script

```
#!/home/user/my_application/bin/rails runner
run_script_within_the_application
```

@varyonic @deivid-rodriguez  because `ActiveAdmin::Engine.root` is activeadmin _library_ root, not _application_ root - this is not what we want, we should wait for `Rails.root` to become available. It means that we always wait and assign `Rails.root` somewhere, we should only make a choice - where.

I think, that the best choice for assigning default value for "load_paths" is the place where this setting is declared (like it's now), so code reader should not navigate between files examining default settings.

Fivell:
>> just have same file name in app/admin and app/models


@mluts  convention is next
 - filename in model path has singular name (ex store.rb)
- filename in admin path has plural name  (ex stores.rb)

mluts:
@Fivell i think file generators should follow the convention

The application is broken after using generators only. https://gist.github.com/mluts/7f6fb56574d92b30d17f2f213d208900

Fivell:
@mluts , they should follow need to check, however by documentation 
```bundle exec rails generate active_admin:resource my_model```
should be replaced with
```bundle exec rails generate active_admin:resource MyModel```

**Comment by varyonic** [comment]
I was referring above to the accepted answer on StackOverflow, not `MyEngine::Engine.root`, which as you say, is incorrect.

**Comment by mluts** [comment]
@varyonic sorry, my answer turned out to be too focused on incorrect `Engine#root`, anyway i meant that wherever `loads_paths` may be assigned, the best place in my opinion will be near `setting :load_paths`, because i can see the default value without extra navigation

**Comment by deivid-rodriguez** [comment]
What I meant is to set a new `application_root` setting in an initializer (where `Rails.root` should be already available), and then use that as a base for `load_paths`.

**Comment by mluts** [comment]
@deivid-rodriguez sure, i understand your point, but putting `load_paths` into `Engine` looks for me like like adding unnecessary level of indirection, apart from this complexity - tests are failing strangely and unexpectedly way when i did this:

```diff
  lib/active_admin/application.rb |  2 +-
 lib/active_admin/engine.rb      |  4 ++++
 spec/unit/application_spec.rb   | 10 ----------
 3 files changed, 5 insertions(+), 11 deletions(-)

diff --git a/lib/active_admin/application.rb b/lib/active_admin/application.rb
index 3ffc721..882b784 100644
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -24,7 +24,7 @@ module ActiveAdmin
     # Load paths for admin configurations. Add folders to this load path
     # to load up other resources for administration. External gems can
     # include their paths in this load path to provide active_admin UIs
-    setting :load_paths, [File.expand_path('app/admin', Rails.root)]
+    setting :load_paths, []
 
     # The default number of resources to display on index pages
     inheritable_setting :default_per_page, 30
diff --git a/lib/active_admin/engine.rb b/lib/active_admin/engine.rb
index 725a961..85cf017 100644
--- a/lib/active_admin/engine.rb
+++ b/lib/active_admin/engine.rb
@@ -12,5 +12,9 @@ module ActiveAdmin
     initializer 'active_admin.routes' do
       require 'active_admin/helpers/routes/url_helpers'
     end
+
+    initializer 'active_admin.load_paths', before: :load_config_initializers do
+      ActiveAdmin.application.load_paths = [Rails.root.join('app', 'admin').to_s]
+    end
   end
 end
diff --git a/spec/unit/application_spec.rb b/spec/unit/application_spec.rb
index 7a2bb1a..0929692 100644
--- a/spec/unit/application_spec.rb
+++ b/spec/unit/application_spec.rb
@@ -5,16 +5,6 @@ RSpec.describe ActiveAdmin::Application do
 
   let(:application) { ActiveAdmin::Application.new }
 
-  around do |example|
-    old_load_paths = application.load_paths
-    # TODO: Figure out why load paths need to be overriden
-    application.load_paths = [File.expand_path('app/admin', Rails.root)]
-
-    example.call
-
-    application.load_paths = old_load_paths
-  end
-
   it "should have a default load path of ['app/admin']" do
     expect(application.load_paths).to eq [File.expand_path('app/admin', Rails.root)]
   end
```

deivid-rodriguez:
Makes sense to me. Not sure about @Fivell's comments about file naming conventions, but the patch looks good, I think.

javierjulio:
@deivid-rodriguez @varyonic @Fivell I don't really know about how this works in Ruby so you all are best to determine if this is good to go in or not. Has your approval?

lib/active_admin/helpers/settings.rb
```diff
@@ -22,6 +22,10 @@ def self.included(base)
     end
 
     def read_default_setting(name)
+      if (value = default_settings[name]).respond_to?(:call)
```
varyonic: Meh.  The more I look at the implementation of application settings the less I like it.  I would like a little time to submit a counter proposal.

javierjulio: @varyonic sounds good to me. I'm indifferent to the change here but thought this was a fix we could get in so its why I pushed it. I do think the app settings is over engineered. No need for all it offers. I'll leave it up to you then. Thanks!

javierjulio:
@mluts in #5105 @varyonic took another approach to addressing your issue. Can you try that out and see how it works? The description contains the reason as to why he went with that solution.

mluts:
@javierjulio my application crashes on this branch https://github.com/activeadmin/activeadmin/pull/5105, but still works on mine

mluts:
I run this command from my `$HOME` directory
`michael@u1:~$ ~/project-name/bin/rails runner 'puts 123'`

and see the same error as it was before the fix:
```
/home/michael/.rbenv/versions/2.3.4/lib/ruby/gems/2.3.0/gems/activesupport-5.0.4/lib/active_support/dependencies.rb:509:in `load_missing
_constant': Circular dependency detected while autoloading constant AdminUser (RuntimeError)
```

on my branch with same command i see just an output:
```
123
```

javierjulio:
@mluts it could be that its still in development, if so my apologies. Can you please post on #5105 about issues regarding that branch there and work with @varyonic on it? It could be that some more needs to be done. We appreciate the follow up! 🙏🏻❤️

mluts:
@varyonic Hi, could you review my changes? I just have updated my pull request, maybe you will like it more. Now i don't change anything in reader method

varyonic:
Having reviewed #2677 and #2334 I'm inclined to agree with @Fivell that this may be a file naming issue.  I fixed the application_spec.rb TODO but that's as far as I'm willing to go on this.  Please confirm whether you still have the `Circular dependency detected while autoloading` after making @Fivell's suggested change.  As to tweaking helpers/settings.rb I have now submitted #5106 proposing to completely rip and replace it.

mluts:
@varyonic yes, it works after renaming all files in `app/admin` directory to not match files from `app/models`, but

* their name was picked by activeadmin generator, not by hand. If there is some convention, i'm sure it should be expressed in activeadmin codebase, currently there is no such convention and application crashes on activeadmin defaults.
* you should not use value returned by `Rails.root` before rails application was booted, because it returns `nil`, this is wrong, this causes weird exceptions during boot.

varyonic:
> you should not use value returned by Rails.root before rails application was booted

Agreed, both this PR and #5105 address this.

> If there is some convention, i'm sure it should be expressed in activeadmin codebase

I agree that there may be a [bug or usability issue](https://github.com/activeadmin/activeadmin/issues/2334#issuecomment-53041786) with the generator.  I think that is a separate discussion.

mluts:
seems like this pull request is not relevant now, thank you

varyonic:
Found [this comment](https://github.com/activeadmin/activeadmin/issues/2334#issuecomment-249135360) very similar to your issue.

PR #5078, opened 2017-07-11
Azerbaijani locale file

orkhan:


config/locales/az.yml
```diff
@@ -0,0 +1,138 @@
...
+      resend_confirmation_instructions:
+        title: "Выслать повторно письмо с активацией"
+        submit: "Выслать повторно письмо с активацией"
+      links:
+        sign_up: "Qeydiyyat"
+        sign_in: "Giriş"
+        forgot_your_password: "Şifrəni unutmusunuz?"
+        sign_in_with_omniauth_provider: "%{provider} vasitəsilə daxil ol"
+        resend_unlock_instructions: "Повторная отправка инструкций разблокировки"
+        resend_confirmation_instructions: "Повторная отправка инструкций подтверждения"
```
Fivell: @orkhan , why part of localisation  strings are russian ?

seanfcarroll:
@orkhan are you able to respond to the Russian question? If this is going to go ahead, I wonder if you would consider localising Ransack also, as it is used by ActiveAdmin. I have opened an issue [here](https://github.com/activerecord-hackery/ransack/issues/905).

deivid-rodriguez:
I rebased this PR and added a changelog entry. It should be good to go now.

Only concern is whether https://github.com/activeadmin/activeadmin/pull/5078/commits/93cd894da9fc496edcf8ef6c9f23bca7222e595d makes sense at all. I added it so that the `bin/i18n-tasks check-consistent-interpolations` would pass.

deivid-rodriguez:
I think we can go ahead a merge this. If there's a problem with my changes, Azerbaijani speaking users can find it out after the next release :). Settings this as ready.

deivid-rodriguez:
@javierjulio I'm going to merge this and release 2.5.0, if that sounds good to you.

javierjulio:
Oh wow sorry not sure how I missed this. Yes I think that’s a great idea. I’ll approve so you can merge. Thanks!

PR #5081, opened 2017-07-14
Rewrite CoffeeScript components in ES6

### LINKED ISSUE #5050: Look into rewriting JS as ES6 and drop CoffeeScript
JS has come a long way now and other major gems like foundation-rails use ES6 code via sprockets-es6 gem. Something to give a try and see how it works out.

javierjulio:
This adds the sprockets-es6 gem so we can write our JS as ES6 and remove the coffeescript dependency. While converting the CoffeeScript to ES6 is easy enough, I wanted to add a quick proof of concept for replacing jQuery UI widgets with a standalone ones where possible. I've started with tabs as that was the easiest but also a big win. The tabs in jQuery UI are well over 900+ lines of code whereas this is just 200 and the same functionality. I challenged myself today to get this done in an hour and was successful.

I've since removed the 2 commits that had converted the tabs from jQuery UI to something that is standalone and working from this PR. I created a new branch from this one called [add-tabs-from-bootstrap-v4](https://github.com/javierjulio/activeadmin/tree/add-tabs-from-bootstrap-v4) where I left that change in place to get it later. I will continue converting to ES6 in this PR.

Closes #5050


javierjulio:
My plan here is to just rewrite to ES6 but leave CoffeeScript in place. In a separate PR I will remove CoffeeScript dependency. The reason I want to do this is I'm going to leave `initializers/batch_actions.js.coffee` as is since I remove it in #3862 as its unnecessary (addressed in CSS) and tired of dealing with the conflicts on that one. Let me know if any concerns with that approach.

javierjulio:
After merging this in, I will start replacing the `$.widget.bridge` calls that way we rely less on jQuery UI. This is easy to implement internally as I've done jQuery plugins before. Slowly, we'll be removing the jQuery UI dependency.

javierjulio:
All I have left is the `has_many` code to convert.

javierjulio:
Scratch that, I'm just going to remove `initializers/batch_actions.js.coffee` from here anyway since its only for an extreme case and has no need. This can and should be handled in CSS which is addressed in #3862. Easier to just deal with it now since it doesn't break functionality.

javierjulio:
This is ready for review if you have the time. I've kept the commits focused so open each if you want to see only the changes for that. Should be easier to review and see what's changing. Thanks!

After merging this in, I will start replacing the `$.widget.bridge` calls that way we rely less on jQuery UI. This is easy to implement internally as I've done jQuery plugins before. And from there further changes to slowly start removing the jQuery UI dependency.

javierjulio:
The changes are fine to review as my build passes locally. I do have an error though when trying to run a local server for the test app. Not sure why since I've removed all instances of CoffeeScript, even reset the test app, also deleted the `tmp` directories. Not sure what I've missed... If you have any ideas would appreciate the help!

javierjulio:
At first running the test app in a browser after starting the local server wasn't working, but it seems that my test app stuck around even after clearing out the test app with `rm -rf spec/rails && bundle update`. Either way since it was still there it had the original `active_admin.js.coffee` file so all I had to do was update that and it works fine. The tests are passing.

javierjulio:
@varyonic thanks! You reviewed, but not approved. Are you ok with merging in? Anyone else feel like putting it in a review. I would like to get this in sooner as I've already started updating components. Its going really well! I'm updating the PerPage component to demonstrate what it would be like without the jQuery UI widget bridge and also not needing to attach Turbolinks events either (we'll do event delegation through document instead). Working great so far!

javierjulio:
The build is green but for some reason codecov failed. Its not able to compare commits. I had rebased and edited a commit so that could be why.

faucct:
This change makes the `phantomjs` features testing modal windows crash with `ReferenceError: Can't find variable: Symbol`. Maybe you should include some polyfills with it.

faucct:
Simply requiring the `babel/polyfill` from https://github.com/babel/ruby-babel-transpiler/blob/6878f6edb4e717066bbe344c25b05055cd2d914f/babel-source.gemspec.erb#L14 does the thing:

```
# active_admin.js.coffee
#= require babel/polyfill

javierjulio:
@faucct since you mentioned phantomjs, I know it was [replaced recently with headless chrome](https://github.com/activeadmin/activeadmin/pull/5246/files) so it could be you are running tests on an older version. Best we moved off of phantomjs.

faucct:
I don't mean `activeadmin` repository feature tests, but our own. Still, the polyfills should be included to avoid breaking compatibility with browsers not having es6 features.

javierjulio:
So you're running the tests against the ES6 code rather than what it compiles too? When it tested against CoffeeScript, did it not first convert it to JS or were the tests running against the CoffeeScript code itself?

faucct:
Of course the tests are running against the compiled code, but the compiled code expects your browser have some es6 constants defined (Symbol in this case). That's what are polyfills for. You can see the compiled code for yourself and find that [this line](https://github.com/activeadmin/activeadmin/blob/master/app/assets/javascripts/active_admin/lib/modal_dialog.es6#L23) has transformed into something that uses global `Symbol` variable which is not present at all browsers - http://kangax.github.io/compat-table/es6/#test-Symbol

javierjulio:
I ask because I'm not aware if they are so its not obvious to me. Do you know why its expecting to have an ES6 constant defined? I thought by having this here it would convert but keep the code in ES5 and not rely on something that is ES6 based in the compiled code. Do you know why its using the Symbol variable? Do you have a suggestion on rewriting that line so it doesn't use the Symbol variable perhaps? Yes I know what polyfills are for. This is unexpected so its not something you can expect me or others to be aware of since it wasn't even clear what it was when you described the issue. There is no need for a condescending tone. I'm trying to help but first I need to understand what's wrong.

faucct:
Sorry, didn't mean to be offensive: the ES6 is a new thing to me too and I don't understand the reasons for a lot of things there. You should use [`for ... in ...`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) instead of [`for ... of ...`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) for iterating over object properties. Then you won't need the `Symbol` constant defined.

javierjulio:
I'm sorry as ES6 is also new to me. I love some of the improvements but overall I don't know ES6 very well. 

I'm glad you pointed that out, because why I asked several questions before is that I used a CoffeeScript to ES6 converter which now addresses your concern as to why it might seem odd it was one way rather than another. 

Now we are getting somewhere and are understanding each other. 👍🏻

A fair amount of JS I rewrote because it was simple enough. The portion you linked too though is so gnarly. 😱 I dug through many commits to understand why it was written the way it was. I gave it a shot but had trouble wrapping my head around it from what I can last remember. I did test the change though so I know it works. I asked if you have thoughts on how that line can be rewritten because I passed it through a converter.

It sounds like we should be using `for in` and hopefully that will solve both issues, no need for a polyfill which I feel is ideal and those feature tests should run successfully. Can you make that ES6 change, test it locally and let me know how it works for you? I want to be sure the change fixes the feature tests too. I can give it a try (I'm sure other contributors will comment) if you submit a PR with that change. Thanks!

faucct:
Okay, I will have the PR tomorrow.

PR #5084, opened 2017-07-16
Remove MethodOrProcHelper from FormBuilder

varyonic:
This module is included in half-a-dozen locations but contains a couple of code paths that are only ever used once.  The first of these is in form_builder, where I replaced it with a protected method, ~~and the second is dispatching an Active Admin access denied exception where call_method_or_extract_proc was extracted but has never been reused~~.

### DIFF
--- a/lib/active_admin/form_builder.rb
+++ b/lib/active_admin/form_builder.rb
@@ -13,7 +13,6 @@ end
 
 module ActiveAdmin
   class FormBuilder < ::Formtastic::FormBuilder
-    include MethodOrProcHelper
 
     self.input_namespaces = [::Object, ::ActiveAdmin::Inputs, ::Formtastic::Inputs]
 
@@ -87,10 +86,7 @@ module ActiveAdmin
         contents << template.content_tag(:li) do
           template.link_to I18n.t('active_admin.has_many_remove'), "#", class: 'button has_many_remove'
         end
-      elsif call_method_or_proc_on(has_many_form.object,
-                                   builder_options[:allow_destroy],
-                                   exec: false)
-
+      elsif allow_destroy?(has_many_form.object, builder_options[:allow_destroy])
         has_many_form.input(:_destroy, as: :boolean,
                             wrapper_html: {class: 'has_many_delete'},
                             label: I18n.t('active_admin.has_many_delete'))
@@ -105,6 +101,17 @@ module ActiveAdmin
       end
 
       contents
+    end
+
+    def allow_destroy?(form_object, destroy_option)
+      !! case destroy_option
+         when Symbol, String
+           form_object.public_send destroy_option
+         when Proc
+           destroy_option.call form_object
+         else
+           destroy_option
+         end
     end
 
     def sorted_children(assoc, column)


lib/active_admin/view_helpers/method_or_proc_helper.rb
```diff
@@ -60,6 +65,7 @@ def call_method_or_proc_on(receiver, *args)
       if options[:exec]
         instance_exec(receiver, *args, &symbol_or_proc)
       else
+        # exec: false is deprecated.  TODO: raise if this branch called.
```
javierjulio: Should we throw a deprecation warning here? I'll make a reminder for us to remove anything deprecated once we start v2 in master.

varyonic: The deprecation is thrown at the top of the method, where exec is assumed always true.

varyonic:
Unfortunately this PR is missing some coverage.  I'm planning a different change before I circle back and try to complete this.

varyonic:
The remaining coverage issue looks a false positive to me.

lib/active_admin/base_controller/authorization.rb
```diff
@@ -100,7 +100,7 @@ def action_to_permission(action)
       end
 
       def dispatch_active_admin_access_denied(exception)
-        call_method_or_exec_proc active_admin_namespace.on_unauthorized_access, exception
+        instance_exec(self, exception, &active_admin_namespace.on_unauthorized_access.to_proc)
```
deivid-rodriguez: This seems backwards incompatible. Before we would allow passing a string, like '"to_s"', now we don't, I think ("to_s".to_proc NoMethodError: undefined method `to_proc' for "to_s":String).

Also, I'm not sure about the reduction in complexity. I feel it's nice to keep all this magic in a single place.

varyonic:
Big picture: I’m trying to make ActiveAdmin work with activeadmin/arbre#64, which is a significant change that tries to get Arbre and ActionView to play nice together.  The issue is that calls to view helpers via Arbre::Element#method_missing that return strings not meant to be rendered result in unwanted output.  Logging what happens in Element#method_missing has led me to much head scratching over view helper MethodOrProcHelper.  Personally I find it a not very cohesive module that tries to DRY code but actually obfuscates behavior behind case statements and options.

This PR contains two simplifications, which in retrospect I might have separated.  (1) Don’t include MethodOrProcHelper in FormBuilder.  The way it is used there (exec: false) is different from the way the helpers are used anywhere else.  (2) Don’t use call_method_or_proc in dispatch_active_admin_access_denied.  It’s just unnecessarily complex.

Is MethodOrProcHelper a public interface?  It’s not clear to me.  The module is included in multiple places and it is documented on rubydoc.info so I have been cautious to deprecate what could be removed.

PS. I found another redundant render_or_call_method_or_proc_on in Namespace#add_logout_button_to_menu given MenuItem#url also contains render_in_context...

lib/active_admin/form_builder.rb
```diff
@@ -107,6 +103,17 @@ def has_many_actions(has_many_form, builder_options, contents)
...
+    def allow_destroy?(form_object, destroy_option)
+      !! case destroy_option
+         when Symbol, String
+           form_object.public_send destroy_option
+         when Proc
+           destroy_option.call form_object
+         else
+           destroy_option
+         end
+    end
```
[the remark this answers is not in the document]
varyonic: The branch exec false exists solely for the benefit of FormBuilder#allow_destroy?  FormBuilder is a Formtastic::FormBuilder so why is that coupled to an Arbre view helper?

lib/active_admin/base_controller/authorization.rb
varyonic: Yes.  Why anyone would pass a method name as string here I really don't know, but backward compatibility would be assured by adding 
```
setting = active_admin_namespace.on_unauthorized_access
setting = setting.to_sym if setting.is_a?(String)
instance_exec(... &setting.to_proc)
```

lib/active_admin/form_builder.rb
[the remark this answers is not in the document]
varyonic: My 2¢: [it's not magic](http://ryanogles.by/2016/11/19/its-not-magic-you-just-dont-understand-it-yet.html). YAGNI.  Worst case we change to `helpers.call_method_or_proc_on...` and I have the change I need.

deivid-rodriguez:
Thanks for the context. I was considering this just an effort to make AA more maintainable, so I raised some concerns around the reduction of complexity since I'm not really sure about it making the code simpler. If you need to avoid including this module to fix specific issues, then go ahead.

Regarding considering `MethodOrProcHelper` public API, I honestly always considered it private. I mentioned the backwards compatibility concern because the patch cares about it. But I'm fine with considering it private and changing functionality as we please, even without deprecation.

lib/active_admin/base_controller/authorization.rb
deivid-rodriguez: I wouldn't really care about backwards compatibility here. I just mentioned it since the patch cares, so I assumed you had more context about external usage.

varyonic:
OK, I'm hearing keep the changes to `allow_destroy?` and `dispatch_active_admin_access_denied` but leave `MethodOrProcHelper` alone for now.  Does that sound right?  If so I'll rename the PR and update.

deivid-rodriguez:
@varyonic I would extract the `allow_destroy?` destroy method that calls `helpers.call_method_or_proc_on...`, yes.

Regarding `dispatch_active_admin_access_denied`, I don't really think the change buys us much. @Fivell or @javierjulio, what do you think?

deivid-rodriguez:
@varyonic I saw that you updated this, but you also need to remove the inclusion of `MethodOrProcHelper` from `authorization.rb`, right? I mean, that's the point of restoring the second commit?

PR #5089, opened 2017-07-19
Index page pagination

elbarto132:
The pagination on index pages was broken because the pagination code was not marked as html_safe and thus the html code was just displayed instead of being rendered. This fixes it

### DIFF
--- a/lib/active_admin/views/components/paginated_collection.rb
+++ b/lib/active_admin/views/components/paginated_collection.rb
@@ -107,7 +107,7 @@ module ActiveAdmin
           options[:right] = 0
         end
 
-        text_node paginate.html_safe collection, options
+        text_node paginate(collection, options).html_safe
       end
 
       include ::ActiveAdmin::Helpers::Collection
--- a/lib/active_admin/views/components/paginated_collection.rb
+++ b/lib/active_admin/views/components/paginated_collection.rb
@@ -107,7 +107,7 @@ module ActiveAdmin
           options[:right] = 0
         end
 
-        text_node paginate collection, options
+        text_node paginate.html_safe collection, options
       end
 
       include ::ActiveAdmin::Helpers::Collection


Fivell:
@elbarto132 , using AA a lot and never had such issue, can you add failing spec please?

javierjulio:
@Fivell I would think it would have to do with a specific version since I'm definitely not seeing this either. @elbarto132 what version of ActiveAdmin and Kaminari are you using?

elbarto132:
kaminari (1.0.1)
activeadmin (1.0.0 98365c9)
bildschirmfoto 2017-07-25 um 07 55 04

javierjulio:
@elbarto132 you link to a commit but list v1.0.0. They are not the same though. The version we cut as 1.0.0 has its own commit. What do you have specified in your Gemfile so that way we know for sure what version of AA is being used? Can you recreate this in a test app or in our bug report script template? It'll help us as we don't have any issues currently with this.

elbarto132:
If i put this in my gemfile
gem 'activeadmin', github: 'activeadmin'
As proposed here: https://activeadmin.info/documentation.html
It results in "bundle list" showing activeadmin (1.0.0 98365c9)

I just wanted to create an example project, but this stopped me from doing it: https://github.com/activeadmin/activeadmin/issues/5097

Am i supposed to have
gem 'activeadmin', github: 'activeadmin'
or
gem 'activeadmin'
in the gemfile?

javierjulio:
@elbarto132 hmm that's weird. Did you try instead recreating the issue with our [bug report template script](https://github.com/activeadmin/activeadmin/blob/master/CONTRIBUTING.md#4-did-you-find-a-bug)? I create integration tests to reproducible errors, I have recently with submitting the filter form. See if the pagination works by either outputting within the integration test and/or asserting that the element is there (this would work if its actual HTML rather than a string and thus confirming your result).

elbarto132:
My fault, i took some code from another project without Activeadmin so i didn't see the error there. The code was overwriting the paginate function as proposed here https://github.com/matenia/bootstrap-kaminari-views with some other stuff added by me and thus not being html safe anymore.

javierjulio:
@elbarto132 that makes sense. Glad you figured it out! You shouldn't need that for Kaminari anymore since I see you were using the latest version and it [has theme support which offers Bootstrap](https://github.com/kaminari/kaminari#themes).

Issue #5091, opened 2017-07-21
Error attempting to display active filters because wrong data is used


javierjulio:
@Fivell as we discussed on Slack here is an executable script that fails.

### Expected behavior

When filtering a model (e.g. User) that has a simple string field, and then that model has an association (e.g. Comment) using primary_key/foreign_key on the same string field, the page should render successfully as AA should know I'm filtering a string field on User, not on Comment but when displaying an active filter, it think its the string field on Comment which is wrong.

### Actual behavior

The page fails to render because we are pulling the wrong data for an active filter. We are using an association rather than just outputting the field on the model being filtered, for some reason we search through associations first or perhaps our logic is a little off in this case.

### How to reproduce

Run this executable script with a test case that fails and reproduces the same error I ran into. Again the filter is working because I can verify that the database query is correct, the issue is that the error comes from rendering what filters are active. Something is off in our logic. I've had other issues recently with active filters. I'll try to do more reproducible cases like this.

```ruby
begin
  require 'bundler/inline'
rescue LoadError => e
  STDERR.puts 'Bundler version 1.10 or later is required. Please update your Bundler'
  raise e
end

gemfile(true) do
  source 'https://rubygems.org'

  # Use local changes or ActiveAdmin master.
  if ENV['ACTIVE_ADMIN_PATH']
    gem 'activeadmin', path: ENV['ACTIVE_ADMIN_PATH'], require: false
  else
    gem 'activeadmin', github: 'activeadmin/activeadmin', require: false
  end

  # Change Rails version if necessary.
  gem 'rails', '~> 5.1.0'

  gem 'sass-rails'
  gem 'sqlite3', platform: :mri
  gem 'activerecord-jdbcsqlite3-adapter',
      git: 'https://github.com/jruby/activerecord-jdbc-adapter',
      branch: 'rails-5',
      platform: :jruby
end

require 'active_record'

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :active_admin_comments, force: true do |_t|
  end

  create_table :users, force: true do |t|
    t.string :full_name
    t.string :barcode
  end

  create_table :comments, force: true do |t|
    t.string :message
    t.string :barcode
  end
end

require 'action_controller/railtie'
require 'action_view/railtie'
require 'active_admin'

class TestApp < Rails::Application
  config.root = __dir__
[...]

PR #5092, opened 2017-07-21
resolve related class in active filters if possible

Fivell:
there are only few cases when it makes sense to detect related class in active filter by predicate
 - dropdown with single value (`eq` arel predicate)
 - dropdown/checkboxes with multiple values (`in` arel predicate)

so let's skip detecting unless in and eq arel predicates are used

[refs #5091]

### DIFF
--- a/lib/active_admin/filters/active_filter.rb
+++ b/lib/active_admin/filters/active_filter.rb
@@ -13,7 +13,7 @@ module ActiveAdmin
       def initialize(resource, condition)
         @resource = resource
         @condition = condition
-        @related_class = find_class
+        @related_class = find_class if find_class?
       end
 
       def values
@@ -65,6 +65,10 @@ module ActiveAdmin
 
       def ransack_predicate_name
         Ransack::Translate.predicate(condition.predicate.name)
+      end
+
+      def find_class?
+        ['eq', 'in'].include? condition.predicate.arel_predicate
       end
 
       # detect related class for Ransack::Nodes::Attribute


javierjulio:
@Fivell thanks so much for tackling this! 🙏🏻❤️ I've been testing locally today and its been working fine for part of it but we still have an issue where a completely unrelated association on a model will cause a failure if a filter is applied.

What I'm learning is that we can have associations on a model that are being filtered that has nothing to do with the filter at all, but the `find_class` lookup where it uses `foreign_key` fails on the unrelated association when this shouldn't happen.

I always wondered why we redid this code since what was there before worked well but since the update we've had numerous issues. Wouldn't it be beneficial to not be so fancy/dynamic here and keep the active filters simpler? Its caused us quite a few headaches. Here is a reproducible script of where it fails for us in production:

```ruby
begin
  require 'bundler/inline'
rescue LoadError => e
  raise e
end

gemfile(true) do
  source 'https://rubygems.org'

  if ENV['ACTIVE_ADMIN_PATH']
  else
  end

  gem 'rails', '~> 5.1.0'

  gem 'sass-rails'
  gem 'sqlite3', platform: :mri
  gem 'activerecord-jdbcsqlite3-adapter',
      git: 'https://github.com/jruby/activerecord-jdbc-adapter',
      branch: 'rails-5',
      platform: :jruby
end

require 'active_record'

  end

  create_table :posts, force: true do |t|
    t.integer :result_value
  end

  create_table :subscriptions, force: true do |t|
    t.integer :post_id
  end

  create_table :user_subscriptions, force: true do |t|
    t.integer :subscription_id
  end
end

require 'action_view/railtie'
require 'active_admin'

  config.root = __dir__

  config.eager_load = false
  Rails.logger  = config.logger
end

end

  has_one :subscription
  has_many :user_subscriptions, through: :subscriptions
end

class Subscription < ActiveRecord::Base
  belongs_to :post
end

class UserSubscription < ActiveRecord::Base
  belongs_to :subscription
end

ActiveAdmin.setup do |config|
end

Rails.application.initialize!

  content do
    'Test Me'
  end
end

ActiveAdmin.register Post do
  filter :result_value
end

  ActiveAdmin.routes(self)
end

require 'minitest/autorun'
require 'rack/test'
require 'rails/test_help'

  def test_active_filter_fails
    post = Post.create! result_value: 5000

    # This should be successful but it fails. Here I'm simulating
    # by submitting a filter for a field on the Post model, which
    # does work but when it comes to displaying active filters,
    # its doing the wrong thing and looking up an association and
    # fails on foreign_key look up on user_subscriptions association
    # even though its completely unrelated.
    #
    # In other words it doesn't matter what you filter, its just
    # having an association defined causes an error when you have
    # a filter applied.
    #
    # When running this test this is the error message you will get
    # which shows that active filters is failing because of an
    # unrelated association.
    #
    # ActionView::Template::Error (undefined method `klass' for nil:NilClass
    #   1: insert_tag renderer_for(:index)
    #
    get admin_posts_url, params: { "utf8" => "✓", "q[result_value_equals]" => "5000", "commit" => "Filter" }
    assert_response :success
  end

  private

  def app
    Rails.application
  end
end
```

Fivell:
@javierjulio we redid it  because there was some requests ,  take a look at original pr and related issue https://github.com/activeadmin/activeadmin/pull/4951

```
  has_one :subscription
  has_many :user_subscriptions, through: :subscriptions
end
```

@javierjulio , can you point me to docs, never saw such declaration before, since I always thought

```through``` should have value of existed association.


What I have is 
```
(byebug) Post.take.user_subscriptions
D, [2017-07-24T22:57:37.786245 #7675] DEBUG -- :   Post Load (0.2ms)  SELECT  "posts".* FROM "posts" LIMIT ?  [["LIMIT", 1]]
*** ActiveRecord::HasManyThroughAssociationNotFoundError Exception: Could not find the association :subscriptions in model Post

nil
```

javierjulio:
@Fivell yes you're right! I noticed something wrong too since `through` value was plural when it needed to be singular but the report still failed. Like you, I read the same, specifying the source is required for it to work. Thanks for following up!

javierjulio:
Since I know you're good with what's here I'll merge. Thanks again @Fivell! 👏🏻❤️

PR #5094, opened 2017-07-23
Handle table prefix & table suffix for ActiveAdminComment model

### LINKED ISSUE #5093: table name prefix & suffix not taken into account for active_admin_comments
Because I'm currently working with a legacy database, I have setup a table name prefix in `config/application.rb`.

```ruby
ActiveRecord::Base.table_name_prefix = 'prefix_'
```

It appears that ActiveAdmin does not pickup on that, which result in a Mysql error when trying to display the index page of the `Comment` resource or any individual resources:

### Expected behavior

ActiveAdmin should detect that a prefix has been set on ActiveRecord and use it when retrieving admin comments.

### Actual behavior

I think that since `table_name` is set `/lib/active_admin/orm/active_record/comments/comment.rb` the `table_name_prefix` is discarded.

```
ActiveRecord::StatementInvalid in Admin::CommentsController#index
Mysql2::Error: Table 'database.active_admin_comments' doesn't exist: SHOW FULL FIELDS FROM `active_admin_comments`
```

### Suggested fix

Something like:
```ruby
module ActiveAdmin

    self.table_name = "#{self.table_name_prefix}active_admin_comments"
[...]
```

I'll see if this work tomorrow and if yes I'll create a pull request.

PS: Note same thing is happening for the suffix.

Techbrunch:
Fixes #5093

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+* Fixed [#5093][] Handle table prefix & table suffix for `ActiveAdminComment` model
 * Fixed [#4173][] by including the default Kaminari templates [#5069][] by [@javierjulio][]
 * Fixed [#5043][]. Do not crash in sidebar rendering when a default scope is not
   specified. [#5044] by [@Fivell][].
--- a/lib/active_admin/orm/active_record/comments/comment.rb
+++ b/lib/active_admin/orm/active_record/comments/comment.rb
@@ -1,7 +1,7 @@
 module ActiveAdmin
   class Comment < ActiveRecord::Base
 
-    self.table_name = 'active_admin_comments'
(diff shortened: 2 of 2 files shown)


javierjulio:
This appears to be right as testing the following in a Rails console does show that setting `table_name` explicitly ignores the prefix and suffix. We have to provide it for it to work.

```sh
>> ActiveRecord::Base.table_name_prefix = 'prefix_'
=> "prefix_"
>> User.table_name
=> "prefix_users"

>> User.table_name= "usersss"
=> "usersss"

>> User.table_name
=> "usersss"
```

Fivell:
@Techbrunch please drop a line to changelog

Techbrunch:
@Fivell Done

javierjulio:
@Techbrunch thanks for the fix! 👏🏻❤️

PR #5096, opened 2017-07-25
Update active filters for dates to use "before" and "after" text

javierjulio:
This resolves the merge conflicts in #4622. I did this as I asked the original author to update but haven't heard back in a few weeks. I think the change is a very helpful one. I'm keeping the original commit but fixed the conflicts in a second one.

Fixes #4622


javierjulio:
Going to have to punt on this. 😞 I think its a great and valuable change but I missed some things and not sure why I can't get it to work. I don't know much about ransack and don't want to break filters. I just want to change what text appears for the predicates applied. Could be a good task to revisit in the future. Any input/pointers would be appreciated!

PR #5098, opened 2017-07-25
Add releasing instructions

javierjulio:
Just so we don’t forget how its done and be motivated to release more often. 😏

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,3 +61,20 @@ Tool                  | Description
 [Inherited Resources]: https://github.com/activeadmin/inherited_resources
 [Kaminari]: https://github.com/kaminari/kaminari
 [Ransack]: https://github.com/activerecord-hackery/ransack
+
+## Releasing
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -61,3 +61,20 @@ Tool                  | Description
...
+  ```sh
+  git checkout master
+  git fetch activeadmin
+  git rebase activeadmin/master
+  # If the release is 2.1 then this should be: 2-1-stable
+  git checkout -b N-N-stable
+  git push activeadmin N-N-stable:N-N-stable
+  ```
+
+* `bundle exec rake release`
```
deivid-rodriguez: I think `CONTRIBUTING.md` is a better fit for this.

javierjulio: Ah good catch! I thought that was just from the perspective of someone outside of AA contributing but I see have later list items that are for maintainers only. I'll move it.

PR #5099, opened 2017-07-26
Remove deprecated code

javierjulio:
This removes code that we deprecated in 1.x or earlier. I'm sure we have more we can remove. I focused mostly on what I thought I could handle well. Some other deprecations we could remove but depend on what range of Rails we'll go with. But that we are tackling in #5104.


javierjulio:
Sorry was a little too quick with the review requests. I'm going to push two more changes as I found other deprecations we can remove.

PR #5103, opened 2017-07-26
Change supported Ruby minimum to 2.3

javierjulio:
Ruby 2.2 is in in security maintenance phase and will be EOL very soon. By the time we release AA v2 it won’t be supported at all. Best to remove it and keep AA current all around.

I tried finding what I could that had to deal with Ruby version checks but I definitely might have missed something. Please let me know anything else I can change and/or remove.


varyonic:
> By the time we release AA v2 it won’t be supported at all.

Ruby 2.2 will be supported until March 2018.

javierjulio:
Exactly. What's our rate of releasing? 😂 I stand by this since Ruby 2.2 is only receiving security updates and is pointed out as reaching EOL very soon. We should drop this and focus AA v2 on what's going to be current and new.

varyonic:
Let's drop Ruby 2.2 when Rails drops 2.2, at least a few months from now.  If you are using Heroku then this kind of upgrade may appear trivial, but if you are stuck with manually administered production hosts with limited disk space then Ruby upgrades are not.  I'm not going to speculate on a release date for v2, that depends on everyone's contribution.

varyonic:
[Support of Ruby 2.2 has ended](https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/). Needs rebase.

javierjulio:
Great thanks. I'll take care of that later today.

javierjulio:
This is rebased. Force pushed changes.

PR #5104, opened 2017-07-26
WIP: Drop support for Rails 4.2

javierjulio:
We should [drop support for Rails 4.2 as it only receives severe security updates](http://guides.rubyonrails.org/maintenance_policy.html#severe-security-issues). By the time we do AA v2 it won't be supported and we should focus on what's current rather than what's no longer supported.

## Todo List

- [ ] search for `rails5?` checks and remove else conditions
- [ ] adjust Devise requirement to be 4.2+ instead?
- [ ] find and remove any old Rails support that is no longer needed


javierjulio:
@deivid-rodriguez instead of asking you to review, since there is other stuff to do here, it would be good to have your feedback particularly on the jruby build we had in Travis that was expected to pass but was using Rails 4.2. I know you mentioned that were more issues with jruby support with Rails 5.x.

deivid-rodriguez:
@javierjulio If we're going to drop support for Rails 4.2, something I do agree with, then I'm afraid you're right and we might drop official jruby support altogether. JRuby does not yet officially support Rails 5 as far as I know.

@varyonic @Fivell What do you think?

javierjulio:
It doesn't appear that it would be close to supporting it as the activerecord-jdbc-adapter is lacking in Rails 5 support and doesn't appear to have much activity. From what I've read only SQLite is tested. I say we cut our losses and move on. I can have the jruby removal stuff in one commit so if we want to revisit it in the future we can. I'll move forward with that since I can always drop the commit if we change our minds.

varyonic:
> By the time we do AA v2 it won't be supported... We should focus on what's current rather than what's no longer supported.

I think this is premature.  4.2.9 shipped only a month ago.  Security patches will continue to ship until next year.  I don't think dropping 4.2 will take much effort, so I'm not clear on the benefits of addressing this now.

javierjulio:
I don't think so. Its not worth the effort for security patches. The idea for bringing it up now is that we are doing a major release of AA and we should focus on what's current and supported instead of dropping after that date. Rails 4.2 is only for security updates but its not actively developed. Same with Ruby. Again the idea is to only focus on what's current and what's coming up.

deivid-rodriguez:
The easiest is probably to keep in sync with `Rails` and `Ruby` and drop support in a new minor version once _they_ drop support. That way will make most people happy, I think. Since we dropped really old versions support (1.9.3, Rails 4.0, ...), maintainance in this regard has been painless, so the only clear benefit for us I can actually see is faster build times.

So I say let's keep support a little longer.

I do think we can move _all_ jruby builds to `allowed_failures` though, since we haven't been able to get it consistently green.

varyonic:
> This is the right time. Its not worth the effort for just security patches. You're dragging out the inevitable.

@javierjulio We clearly have different perspectives.  From my view Rails is mature and the new features in new releases [rarely have compelling added value to my business](http://www.varyonic.com/2016/03/from-rails-3-2-to-rails-4-2-if-we-really-must/).  My upgrade cycle is driven by security fixes and my need to stay on supported versions for compliance purposes.  Upgrading earlier has to compete with other priorities, and if delaying a server Ruby upgrade six months lets resources address other issues so be it.  If you have a more specific technical reason than "This is the right time" then let’s discuss, otherwise I am 100% with @deivd-rodriguez, we can drop Rails 4.2 and 5.0 support the day 5.2 ships.  If Active Admin drops support earlier, then fine, I just won’t stay current on Active Admin.

javierjulio:
I do and I've stated it. These are for security releases only. Meaning these are not bug fixes. The maintenance policy makes this clear. Our viewpoints aren't entirely different. You are coming from the viewpoint of wanting the latest in AA in an older Rails release, that's your technical argument as far as I see. I get it you don't want to be left behind, but you will if you don't keep your dependencies current. Upgrading is important because the longer it goes unchecked the harder it becomes. Keep in mind we still have you covered since we have AA v1.x and we can continue to do fixes there for Rails 4.2.x as necessary, in fact we still will build travis with those configured builds from that 1-1-stable branch. I don't see any technical argument from you from keeping this from moving forward other than you wanting to be able to upgrade AA to v2 while still being on an older Rails. @deivid-rodriguez has mentioned he's for removing Rails 4.2. I am for doing the same. Again keep in mind we do have support for it but that is in AA v1.x where we still have Travis builds for. With that we have you and others covered while we can focus on where AA v2 will be.

deivid-rodriguez:
@javierjulio I did mention that, but after a second thought I don't see a clear benefit in doing it now. Normally you drop support to ease maintenance burden, but I haven't felt any burden from supporting Rails 4.2 recently, and it's something that can be very easily done at a later stage.

I'd say let's focus on the other things that need to be done rather than this. If supporting Rails 4.2 bites us somehow while doing that, we can reconsider.

javierjulio:
Then at a minimum we should remove the allowed failures for jruby since it doesn't support Rails 5. We should only keep what we know can work.

deivid-rodriguez:
JRuby does not officially support Rails 5 but the tests pass against the `rails-5` branch of `jruby/activerecord-jdbc-adapter`. Keeping that build running could be useful for people wanting to know how well (or bad) jruby would play with `activeadmin`.

So I would move _all_ jruby builds to `allow_failures`, so they are still run but they never get in the middle. Note that we have `fast_finish: true` in our config, so that means the `allowed_failures` entries won't delay CI feedback.
 
This is the same strategy [rails follows](https://github.com/rails/rails/blob/f55ecc6a7c7866a13fe9f5b84b295f2935c6ff1f/.travis.yml#L105-L109).

javierjulio:
Great! That's what we should be doing for jruby. It allows us to run it but without delaying the builds which is a good middle ground.

deivid-rodriguez:
@javierjulio As per #5512, I'm considering dropping Rails 4.2 support now, because it's starting to get in the middle (it's the last version including built-in XML serialization, and the modern alternative the opt-in `activemodel-serializers-xml` only supports Rails 5.0 or higher).

What do you think, should we rebase this PR?

javierjulio:
@deivid-rodriguez yes I agree, I think that's fine. 👍🏻 We should drop support for Rails 4.2. Regarding this PR though it might be easier to just close it and open a new one due to the number of conflicts. Is that something you might be able to do/submit? I'd like to do this but realistically I won't have time for it until late this week, I've been sick with a cold/flu since Friday. I'm trying to help out though where I can reviewing/approving here.

deivid-rodriguez:
No problem @javierjulio! I'll rebase this myself, lots of conflicts but should be easy to fix!

javierjulio:
Awesome thanks so much for updating!

deivid-rodriguez:
@javierjulio This is ready from my side. A couple of notes from me:

* I completely removed the `ActiveAdmin::Depedency::Adapter` module since it's no longer used. Any issues with that? I don't think anybody should be using this other than ourselves, internally.
* I didn't change the `devise` requirements because it was not necessary. We can handle that separately if/when it bites us.
* I can clean up git history before merging.

javierjulio:
@deivid-rodriguez it looks great. 👍🏻 Regarding each item:

1) I'm good with the removal. Less code is better. If we need this as you say it should be private (internal) only.
2) The commit message mentions a deprecation... took awhile but [I found what I likely came across](https://github.com/activeadmin/activeadmin/blob/b8a08e85576faeecc9688dd2da675f82d27da963/spec/rails_helper.rb#L26) but its just for the spec `rails_helper` so might not be worth it. I very likely updated the dependency version to just be a current major release to go with Rails 5.0+ support. If you'd rather leave as is, that's fine.
3) The commits are good! Feel free to change what you need to before merging. ❤️

Thanks again for doing this. 👏🏻 I'm hoping to be better soon. In the meantime, happy to help you where I can.

deivid-rodriguez:
@javierjulio I'm opened to dropping `devise 3` support (although I'd do it a bit later). In any case I'd prefer to discuss and do it separately. I squashed everything else on a single commit, it's ready!

javierjulio:
@deivid-rodriguez yes that's fine about Devise, thanks! I can't approve this though because I created the PR. If you can approve, I'll merge it in.

deivid-rodriguez:
Also removed 4.2 from the list of required branches! :+1:

javierjulio:
@deivid-rodriguez thanks!

senid231:
@javierjulio good job! now we can move from sprockets to webpacker

PR #5105, opened 2017-07-27
Fix app path setting during engine initialization

varyonic:
I offer this alternative to #5076 that does not rely on further changes to the Settings mixin, in particular read_default_setting().

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -21,6 +21,8 @@ module ActiveAdmin
       @namespaces = Namespace::Store.new
     end
 
+    setting :app_path, Rails.root
+
     # Load paths for admin configurations. Add folders to this load path
     # to load up other resources for administration. External gems can
     # include their paths in this load path to provide active_admin UIs
--- a/lib/active_admin/engine.rb
+++ b/lib/active_admin/engine.rb
@@ -1,5 +1,12 @@
 module ActiveAdmin
   class Engine < ::Rails::Engine
+    initializer "active_admin.load_app_path" do |app|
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -8,10 +8,6 @@ dist: trusty
 
 bundler_args: --without development
 
-branches:
-  only:
-    - master
-
```
javierjulio: Hmm that's weird, how come it was always running builds on PR's then?

deivid-rodriguez: @varyonic sometimes likes to run CI on a branch without opening a PR. With our current Travis CI settings, this is the only way to do it.

javierjulio: In this case running travis on their own fork correct?

varyonic: Yes.  Deleted this commit.

varyonic:
I looked at the CodeClimate issue and I'm not planning to address it.

javierjulio:
@varyonic @Fivell I don't want to hold this up but I was thinking about this over the weekend. I realized the question didn't come up about what is even for? I realized it seemed weird to have to specify the app path. Is it really necessary?

varyonic:
The idea is that app_path is valid even before Rails.root is set during initialization.  It's used to set load_path that in turn is used in #files to load the resource DSL.  Rails.root is only unset in certain corner cases, viz. application_spec.rb.  It might be appropriate, however, to update the [documented use of load_paths](https://github.com/activeadmin/activeadmin/blob/22da13f5ff330b1c89a19c026ee40432890e072f/docs/1-general-configuration.md)

PR #5106, opened 2017-07-27
Refactor inherited settings

varyonic:
Reimplement #2313 using composition and delegation instead of a mixin.

InheritableSetting was injecting itself into another class(!), which still needs a method override anyway (Namespace#read_default_setting).  SettingsNode is simpler to test and more reusable.

DeprecatedSetting and setting predicate (eg. foo?) not reimplemented as they appear no longer used.


varyonic:
According to CodeClimate this PR degrades Application from ‘D’ to ‘F’.  There are ways to fix this, but not that I see related to this PR.  I ‘approved’ this change in CodeClimate, whatever that might mean.  As to CodeCov I’m a little confused how testing for ‘raise_error’ is flagged as not covered.

lib/active_admin/application.rb
```diff
@@ -1,13 +1,50 @@
...
+    def settings
+      @settings ||= SettingsNode.new(self.class.default_settings)
+    end
+
+    def namespace_settings
+      @namespace_settings ||= SettingsNode.new(self.class.namespace_default_settings)
+    end
+
+    def respond_to_missing?(method, include_private = false)
+      [settings, namespace_settings].any? { |sets| sets.respond_to?(method) }
```
Fivell: @varyonic , I prefer always to add 
``` || super ```

in the end

File: lib/active_admin/namespace.rb [hunk 2]
```diff
@@ -41,6 +51,18 @@ def name
       @name.to_sym
     end
 
+    def settings
+      @settings ||= SettingsNode.new(application.namespace_settings)
+    end
+
+    def respond_to_missing?(method, include_private = false)
+      settings.respond_to?(method)
```
Fivell: || super

lib/active_admin/settings_node.rb
```diff
@@ -0,0 +1,47 @@
...
+    def fetch(name)
+      if settings.has_key?(name.to_sym)
+        settings.fetch(name.to_sym)
+      elsif @parent
+        parent.fetch(name)
+      end
+    end
+
+    def respond_to_missing?(method, include_private = false)
+      has_key?(method.to_s.chomp '=') || super
```
Fivell: this time respond_to_missing? calls super

varyonic:
¯\_(ツ)_/¯  Travis build fails.

Fivell:
@varyonic , not sure, but shouldn't be like
```ruby
 initializer "active_admin.load_app_path" do |app|
      ActiveAdmin.application.class.instance_exec(app) do |app|
        default_settings.app_path = app.root
        default_settings.load_paths = [File.expand_path('app/admin', app.root)]
      end
    end
```

varyonic:
This PR is on an older branch that hasn't been rebased yet... strange coincidence I think.

varyonic:
Rebased and fixed.

PR #5107, opened 2017-07-27
Remove unused flash JS component

javierjulio:
Its not referenced anywhere else except for a jasmine test but that doesn’t work (there’s no abstract method). No point in keeping this as its unused to begin with. Better to cut down on the JS we have and start fresh with #5081 and other work I have planned.


javierjulio:
One of the ruby builds failed but a weird one off error. I noticed some long delays yesterday. I re-ran the build now and everything checks out so its ready for review.

Issue #5108, opened 2017-07-27
[proposal] add opportunity to add a form to resource's sidebar


alex-bogomolov:
Recently a client asked if it is possible to add a form to the index page of resource. It would simplify the process of adding entries with a small amount of columns to the project without the need to click "Add new entry", fill the form, click "Save" and then return back to the index page.

I searched for this feature and realized that it's missing. So I suggest to add the following DSL to the resource:
```ruby
sidebar_form do |f|
   # ...
end
```
This DSL will generate a form in the sidebar section under the filters. When a user fills the form and clicks "Create", if the filled values are valid, the entry will be created and he will be redirected back to index page. If the values are not valid, the `new` action will be rendered.

I think this feature should be added to the project, because it will reduce amount of repetitive actions to do when adding a lot of entities and will improve the usability of ActiveAdmin.

varyonic:
FYI HE:labs wrote up a ['quick add' feature using fancybox](https://helabs.com/artigos/2014/05/28/quick-add-and-quick-edit-on-active-admin/) that may be of interest to others. I see your PR was closed after some discussion. I think this might make a good [plugin](https://github.com/activeadmin/activeadmin/wiki/Plugins).

PR #5109, opened 2017-07-27
[WIP] add opportunity to add a form to sidebar on resource's index page

alex-bogomolov:
This PR solves #5108. 

The newly added DSL:
```ruby
  sidebar_form do |f|
    f.input :keyword
  end
```

The form created by the new DSL:
active_admin_sidebar_form

This feature isn't finished yet and is in progress. I haven't modify the controller related code to redirect to index page when the entry was created with sidebar form yet. Also some stuff has to be done on the front-end part. All opinions about improving the code are welcome :)
@javierjulio @Fivell What do you think about this feature?

### DIFF
--- a/app/assets/stylesheets/active_admin/_base.scss
+++ b/app/assets/stylesheets/active_admin/_base.scss
@@ -27,6 +27,7 @@
 @import "active_admin/components/index_list";
 @import "active_admin/components/unsupported_browser";
 @import "active_admin/components/tabs";
+@import "active_admin/components/sidebar_form";
 @import "active_admin/pages/logged_out";
 @import "active_admin/structure/footer";
 @import "active_admin/structure/main_structure";
--- a/app/assets/stylesheets/active_admin/components/_sidebar_form.scss
+++ b/app/assets/stylesheets/active_admin/components/_sidebar_form.scss
@@ -1,0 +1,30 @@
+.sidebar_section.panel {
+  .cancel {
+    display: none;
+  }
+
+  fieldset {
+    box-shadow: none;
+
+    .input {
+      padding: 0;
+    }
+
+    legend {
+      span {
+        background-color: inherit;
+        background-image: none;
+        border: none;
+        box-shadow: none;
+        padding-left: 0;
+      }
+    }
+
+    margin-bottom: 0;
+  }
+
+  fieldset.actions {
+    padding-bottom: 0;
+    margin-top: 0;
+  }
+}
--- a/lib/active_admin.rb
+++ b/lib/active_admin.rb
@@ -122,6 +122,7 @@ require 'active_admin/error'
 # Require internal plugins
 require 'active_admin/batch_actions'
 require 'active_admin/filters'
+require 'active_admin/sidebar_form'
 
 # Require ORM-specific plugins
 require 'active_admin/orm/active_record' if defined? ActiveRecord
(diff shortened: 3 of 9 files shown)


javierjulio:
@alex-bogomolov thanks for continuing to contribute! I'm grateful you want to submit so don't want to discourage you. I don't think this is the right solution. The main issues I see here is that we shouldn't have a DSL specific to this use case, instead you should already be able to do this with the existing DSL. There are a few problems here. One is that now we have to support a specific sidebar. Second users now will expect custom sidebars for other things. But this doesn't mean you shouldn't be able to do this, I just think what we have should enable it already.

My first thought is that this is being done because the existing `form` DSL is for the page only which would make sense. We could have a new DSL for being able to insert forms on other pages. But I think we support a solution for you already that I've used and is better, is partials. Can a sidebar accept a partial at all? If not that might be a way to approach it. Second, what about using the `active_admin_form_for` directly? I think it might be better to have an option for that rather than tie this directly to sidebar.

alex-bogomolov:
@javierjulio I totally agree with you that new DSL for forms in the sidebar is excessive for a use case described in the related issue. You can display partials in a sidebar, but I think it's not the best solution for this use case. When writing a partial, the user has to be sure he uses the correct HTML markup so the form will look naturally in ActiveAdmin style. This is not very handy for those who aren't familiar with the project's styles. The partial solution also doesn't make the controller redirect to the index page after the resource is saved.

@javierjulio I have updated the code. Now the form which is used on the form page can be also rendered on the index page in a sidebar by calling `show_form_in_sidebar` method in resource's configuration file. I have added some styles for the form to look natural. Here's how it looks like now:
active_admin_sidebar_form

I also modified the `smart_resource_url` method and the form class to make the controller redirect to index page after resource being saved if the `redirect_to_index` parameter is present. @javierjulio what do you think about this approach?

javierjulio:
@alex-bogomolov I do think the partial solution would be great for this but doesn't mean it has to be the only solution. We should still be able to use the form within the partial or an existing form DSL within the sidebar. I currently use partials in both ERB or ARB successfully. I'd love to see what you had done to get partials working! We can only offer so much out of the box so that's why I suggest partials as a solution. 

I don't like having a DSL specific to sidebar forms. I'd rather we expose form functionality within sidebars and let users set it up there or where they like. The CSS is an interesting problem. I'd rather not introduce anymore CSS at the moment. Formtastic makes styling forms very painful. I'm trying to see if we can help move along a Formtastic v4 release (with ability to customize fieldset/list) or replace it with something else entirely like Simple Form (seems unlikely).

alex-bogomolov:
@javierjulio So, what do you think about the approach I used in the last commit? Should I continue developing this approach (adding a DSL method to display existing form DSL into the index page), or currently this approach is not relevant?

alex-bogomolov:
@javierjulio I didn't face problems with formtastic while working with CSS, I only had to override some of our own styles for forms (particularly, for the fieldsets.)

javierjulio:
No I don't think we'd want to introduce a DSL specific to sidebar for displaying a form. I think it would be better to either use partials for that for the sidebar, or look into having a DSL for displaying a form that can be used elsewhere too, not just sidebars. So for example, I would want to display a form within a partial. We do this currently by just using `semantic_form_for` directly which is all ActiveAdmin is wrapping. I don't think its a problem for users to use the tools available with ActiveAdmin rather than building this in. The solution technically is already there.

That's the exact problem with the CSS haha. Its bad having to override and turn off so many styles. I'll be improving this a lot once we tackle #3862. It'll take time but we'll get there!

PR #5112, opened 2017-07-28
Move all jruby travis builds to `allow_failures` section

javierjulio:
As we discussed that we'd move this to the `allow_failures` section. I just had another travis failure because the jruby rails42 build is failing in #5081. 😩 Sorry @deivid-rodriguez. 😞

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,8 @@ matrix:
 
   allow_failures:
     - rvm: jruby-9.1.12.0
+      gemfile: gemfiles/rails_42.gemfile
+    - rvm: jruby-9.1.12.0
       gemfile: gemfiles/rails_50.gemfile
     - rvm: jruby-9.1.12.0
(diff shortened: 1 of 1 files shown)


javierjulio:
@deivid-rodriguez cool, I just made that change but I don't see any builds for jruby so it means its not running them at all since it doesn't know what Gemfile to use. You can see that here: https://travis-ci.org/activeadmin/activeadmin/builds/258772337 I'll delete the second commit.

.travis.yml
```diff
@@ -25,7 +25,6 @@ script:
   - bundle exec rake docs:build
 
 rvm:
-  - jruby-9.1.12.0
```
deivid-rodriguez: You need to restore this entry I think. Otherwise it won't run any jruby build, indeed.

varyonic:
I pushed this through as it appears Travis moved us to Trusty and broke the build for JRuby.  There may be other fixes available but this was simplest.

PR #5119, opened 2017-08-03
[fix] do not display pagination info when there are no comments

alex-bogomolov:
This commit fixes the bug when redundant information was displayed if there were no comments in the resource:

bug_picture


javierjulio:
@Fivell no, he's showing an image of what is happening currently **before** his PR change. What he did is great and resolves the issue! I've noticed the same thing recently so this is ready. Thanks @alex-bogomolov! 🙏🏻❤️

Issue #2249, opened 2013-06-04
Inconsistent singular and plural naming with ActiveAdmin resources


kramerc:
When the `active_admin:install` generator is ran, it creates two files in `app/admin/` named `admin_user.rb` and `dashboard.rb`. These filenames are in singular form.

When the `active_admin:resource` generator is ran with the parameter `MyModel`, it creates a file in `app/admin/` with `my_models.rb`. This filename is [made plural in the resource generator](https://github.com/gregbell/active_admin/blob/02960bb5f81e43dbd5c02f870fd07d4c77bf5ab8/lib/generators/active_admin/resource/resource_generator.rb#L11).

As evidenced in issues #1960 and #1961, it seems the _de facto_ standard is to use the singular form for naming. Is this an oversight in the resource generator where the filenames should actually be in singular form?

PR #5120, opened 2017-08-03
Pluralize resource generator files (again)

### LINKED ISSUE #2334: ArgumentError: A copy of ApplicationController has been removed from the module tree but is still active

varyonic:
Reverts #2249.  Closes #2334.  See also #5076.

Having the same filename for models and resource DSL causes issues with the Rails autoloader.
Pluralizing is consistent with the Rails naming convention for controllers, which is much of what the DSL is generating.

### DIFF
--- a/docs/2-resource-customization.md
+++ b/docs/2-resource-customization.md
@@ -10,7 +10,7 @@ resource you must first create a Rails model for it.
 ## Create a Resource
 
 The basic command for creating a resource is `rails g active_admin:resource Post`.
-The generator will produce an empty `app/admin/post.rb` file like so:
+The generator will produce an empty `app/admin/posts.rb` file like so:
 
 ```ruby
 ActiveAdmin.register Post do
--- a/lib/generators/active_admin/resource/resource_generator.rb
+++ b/lib/generators/active_admin/resource/resource_generator.rb
@@ -12,7 +12,7 @@ module ActiveAdmin
 
       def generate_config_file
         @boilerplate = ActiveAdmin::Generators::Boilerplate.new(class_name)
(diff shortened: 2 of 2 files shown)


javierjulio:
@varyonic I've gone through the related issues before, there are many but I'm confused on how this is an issue. I'd like to know more to understand because I use the singular form with latest ActiveAdmin version on Rails 5.0.4 and my production config is set to eager load. Why wouldn't I be running into the issue? And I haven't ever run into this issue before.

varyonic:
Specifically https://github.com/activeadmin/activeadmin/issues/2334#issuecomment-249135360 and https://github.com/activeadmin/activeadmin/pull/5076#issuecomment-313926894

javierjulio:
@varyonic thanks! That first comment makes it clear and was helpful.

Fivell:
@varyonic  should we reopen #2249 again ?

varyonic:
> should we reopen #2249 again?

Not me.  I think I've followed up on https://github.com/activeadmin/activeadmin/pull/5076#issuecomment-318664361 to @mluts and I'm done.

Issue #5048, opened 2017-06-29
Look into dropping jQuery UI


javierjulio:
The jQuery UI library is massive and isn't seeing a lot of development. We should look into where its being used (e.g. calendar, tabs, modal) and see about replacing it with our own JS or smaller libraries dedicated for each.

varyonic:
See also #4034

rystraum:
I would like to help out on this. For now, we need a list on where jQuery is used, yes?

deivid-rodriguez:
I assume all of the jquery-ui components we currently vendorize use jquery under the hood. So we would need to replace each of these components with a standalone solution that does not depend on jquery. It's an ambitious task.

rystraum:
How open are you to replacing them with React-based components? That would require adding webpacker as a dependency. Just want to set the boundaries on what we can replace the vendorized libraries with.

deivid-rodriguez:
I'm pretty opened to replacing our assets with more modern solutions, not sure about others. Perhaps the task could be subdivided in two subtasks:

* Raplace `sprockets` based assets with `webpacker` based assets, while keeping the current implementation.
* Once we have `webpacker` in place, it should be way easier to replace the current components with more modern implementations, I think.

Word of caution though for the first task: the current state of the art around "webpacker + rails engines" is "pretty tricky". At least last time I checked was by no means straightforward how to set it up.

vfonic:
Ping @rystraum! In case you missed it, I just found this PR that you might be interested in having a look/try: #5855.

rystraum:
Thanks for the ping @vfonic. I was sort of stuck with this because I was following this thread https://github.com/rails/webpacker/issues/348. It didn't cross my mind that we can just write all the files to `app/javascript/activeadmin` upon running `rails g active_admin:install` instead. I'll try to move forward with this again.

vfonic:
I'd love to help. I'm not sure about the time available that I'll have, but let's see. Is there any repo where I can have a look? Maybe we could split some work?

Fire-Dragon-DoL:
I have been going through the various issues to figure out if I could remove jquery as a dependency and realized how painful must have been doing the work to migrate to webpacker, just to see it deprecated.

I wonder if activeadmin should do the following:
- offer an npm package with all the css, js and assets bundled, so that anybody can integrate that in their JS pipeline. Ideally provide these on a CDN too (importmap and all that) AND
[...]

PR #5121, opened 2017-08-03
Update our JS components

javierjulio:
This update is an overhaul of our JS files with various improvements. Not yet sure if this PR will completely close issue #5048 but it is related.

First, if applicable, it replaces the jQuery UI widget bridge with our own jQuery plugin interface since its easy to set up on our own. 

Second we'll be using event delegation on the document object, which now means we won't need to specify Turbolinks events like `page:load turbolinks:load` since this solves that too. 

Third we'll start to replace components like tabs and date picker with our own solution and thus removing the jQuery UI dependency entirely. 

From there we'll do general code improvements like consolidating code that deals with query strings (serializing and deserializing), on detecting Turbolinks and using a `js-` class name prefix for JS only hooks (or something similar).

### DIFF
--- a/app/assets/javascripts/active_admin/base.es6
+++ b/app/assets/javascripts/active_admin/base.es6
@@ -8,6 +8,7 @@
 //= require_self
 //= require ./ext/jquery
 //= require ./ext/jquery-ui
+//= require ./lib/active_admin
 //= require ./lib/batch_actions
 //= require ./lib/dropdown-menu
 //= require ./lib/has_many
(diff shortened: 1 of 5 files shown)


javierjulio:
@Fivell @deivid-rodriguez @varyonic I have a lot more I'd like to do but going through each file, I realize that I have to make CSS and Ruby changes. I can't do them in isolation as I've done with the files I have here. I'd like to get this in first so I can rebase and resolve conflicts on #3862 as that will be the next PR that will have to go in. I want to be making CSS/Ruby changes with that in and not having to deal with conflicts. Let me know if you have any concerns with what's here. Thanks!

PR #5126, opened 2017-08-05
add pt-BR missing translations 1-1-stable

renatocn:


varyonic:
Looks like PRs against 1-1-stable not being picked up by Travis.

javierjulio:
Our travis config specifies only the master branch. Either we remove it or specify stable-1-1. I checked the Rails repo to see what they specify but they don't even specify branches so they go with the travis default which my guess would be on all branches.

deivid-rodriguez:
The I'd say let's add the branch to `travis.yml`. It'll also give a quick hint of the currently supported branches, which seems nice.

varyonic:
While David has a valid point my vote is to follow Rails.

deivid-rodriguez:
Ok, so let's go back to our previous setting and test all branches by default. @renatocn Could you remove the whole `branches:` entry from our `.travis.yml` so that this gets tested?

renatocn:
@deivid-rodriguez sure. Done.

deivid-rodriguez:
Failing against `jruby` as expected, since the PR is not rebased on top of #5112. Merging, thanks @renatocn!

Issue #5090, opened 2017-07-19
Using paginated collection for embedded lists produces unusable download links


urkle:
### Expected behavior

What do you think should happen?

CSV/XML/JSON download links that go to the paginated lists not the main resource

### Actual behavior

What actually happens?

The CSV/XML/JSON links download the main resource being viewed.

### How to reproduce
```ruby
show do
    panel 'Sprocket buyers' do
      paginated_collection(sprocket.buyers.order('created_at desc').page(params[:sprocket_buyers_page]).per(15), param_name: 'sprocket_buyers_page') do
        table_for collection do
          column :buyer_name
          column :purchase_date
          column :quantity
          column :price
        end
      end
    end
end
```

The download links end up being  /admin/sprockets/1.json, etc..   instead of something that will give me the list of associated buyers to the sprocket..  e.g.  /admin/buyers.json?q[sprocket_id_eq]=1

Fivell:
@urkle , download links always use current request path to build links.

https://github.com/activeadmin/activeadmin/blob/7461a55e8c367cb6546c96ed7bfb90c952da2eb5/lib/active_admin/view_helpers/download_format_links_helper.rb#L21

Pr are welcome for this feature

leonardoprg:
Hey guys, please follow 

https://github.com/activeadmin/activeadmin/pull/5128

sposinff:
Would love for this to be resolved. 
However, until this is resolved, to avoid confusing our users, I am going to pass the option to hide the links.
```paginated_collection(resource.parents, download_links: false) do```
Another option is to create dynamic scopes
```
ActiveAdmin.register Product do
  Store.all.each do |store|
    scope("#{store.name}") { |scope| scope.where(store_id: store.id) }
  end
end
```

javierjulio:
This is a PR ready for this. Thank you.

PR #5128, opened 2017-08-06
5090: improve download links to deal with collection

leonardoprg:
Hey Guys, @urkle @Fivell 

I'm opening this PR, following #5090 

It's still in progress, (and missing tests) but I did open to discuss the code, and ask about two points.

1 - `[:admin, collection.name.constantize]`, I think hardcoded `:admin` here is not a good deal, is any helper that I can get default namespace/scope routes? 
I tried `ActiveAdmin.application.router...` but I not able to found any helper that returns it,
also I saw that `ActiveAdmin.application.default_namespace` returns `:admin` but not sure if it is related to routes or folder.

2 - I'm not sure about the use of ` params: collection.where_clause.to_h`, should we just get for `foreign_keys`? should we merge it with `request.query_parameters` ?, and, fore sure we need to apply Ransack protocol here.

let me know any suggestion/question/etc.. about code

Thanks

### DIFF
--- a/lib/active_admin/view_helpers/download_format_links_helper.rb
+++ b/lib/active_admin/view_helpers/download_format_links_helper.rb
@@ -14,12 +14,11 @@ module ActiveAdmin
         end
       end
 
-      def build_download_format_links(formats = self.class.formats)
-        params = request.query_parameters.except :format, :commit
+      def build_download_format_links(collection, formats = self.class.formats)
         div class: "download_links" do
           span I18n.t('active_admin.download')
           formats.each do |format|
-            a format.upcase, href: url_for(params: params, format: format)
+            a format.upcase, href: polymorphic_path([:admin, collection.name.constantize], params: collection.where_clause.to_h, format: format)
           end
         end
       end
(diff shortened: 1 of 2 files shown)


sposinff:
Without having looked at the specific code, it seems to me that the export works on an `index`, but not a `show`. There is probably some abstracted method that uses at `resources`, but on a show that object does not exist. So my thought... is it possible to pass resources to the csv builder?

leonardoprg:
well, works on an index is not necessarily true, the problem is `export links` are generated using `url_for` and it works well on index (from resource for example), but if you try to put a resource table in `dashboard/index` for example, it will generate export links to `dashboard#index`,
my fix is to when generate links take in consideration the `resource`, in this way, even you put a resource table in `dashboard#index` for `User`, it will generate links to export `user#index`.

sposinff:
`dashboard/index` is a `ActiveAdmin.register_page` which is different than `ActiveAdmin.register`. By index, I was referring to a `resource` index.

javierjulio:
This could use a rebase to see if CI passes.

PR #5131, opened 2017-08-11
Bump version to 2.0.0.alpha

thejspr:
As master is now tracking the 2.0 release, I thought it'd make it clearer by reflecting it with a version change. This will also make it clear that you are using 2.x when installing via Github.

Currently I see this even though I'm using the latest commit from master:

image


thejspr:
@deivid-rodriguez yes, fixed that typo :)

PR #5132, opened 2017-08-13
Cleanup Settings changes

varyonic:
Simplifies #5105 and #5106.  Deprecates a dead method, and uses a different approach for SettingsNode.  Each commit is related but stands on its own.

### DIFF
--- a/lib/active_admin/settings_node.rb
+++ b/lib/active_admin/settings_node.rb
@@ -5,6 +5,10 @@ module ActiveAdmin
       @parent = parent
     end
 
+    def register(name, value)
+      settings[name] = value
+    end
+
     def settings
       @settings ||= {}
     end
--- a/lib/active_admin/settings_node.rb
+++ b/lib/active_admin/settings_node.rb
@@ -1,50 +1,18 @@
 module ActiveAdmin
 
   class SettingsNode
-    def initialize(parent = nil)
-      @parent = parent
-    end
+    class << self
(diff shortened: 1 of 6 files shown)


javierjulio:
Replacing `method_missing` with another solution 👍🏻 Love it.

lib/active_admin/settings_node.rb
```diff
@@ -1,49 +1,30 @@
...
-      settings.has_key?(name.to_sym) || @parent && @parent.has_key?(name.to_sym)
-    end
+    class << self
+      def new(superclass = SettingsNode)
+        Class.new(superclass).tap do |klass|
+          superclass.instance_values.keys.each do |name|
+            klass.accessors(name)
+          end
+        end
+      end
```
deivid-rodriguez: This is the first time that I see the `.new` method of a class being overriden instead of using `initialize`. And if I read this properly, `SettingsNode.new` is not returning an instance of `SettingsNode` but a new class inheriting from it? I can't say I like it.

deivid-rodriguez: Would renaming `.new` to `.create` or something like that work? I would sleep better at night if it does :smile:

varyonic: Yes, I was considering that, let me revisit.

varyonic:
I removed a dead method and coverage dropped?

deivid-rodriguez:
By the way, I've enabled the new "based analysis" in codeclimate, we were using the "legacy analysis".

PR #5135, opened 2017-08-14
Test all branches by default

deivid-rodriguez:
On #5126 we decided to start testing all branches in Travis again. This PR cherry-picks that change into master.


deivid-rodriguez:
Sorry for this dance. Normally this change should've been introduced into master, and then backported to 1-1-stable, but I merged #5126 very quicky :sweat_smile:.

PR #5136, opened 2017-08-15
Fix undefined variable error in app generation

deivid-rodriguez:
This PR fixes a crash when generating our sample application because of a variable being evaluated in the wrong context.

### DIFF
--- a/spec/support/rails_template_with_data.rb
+++ b/spec/support/rails_template_with_data.rb
@@ -318,7 +318,7 @@ append_file "db/seeds.rb", "\n\n" + <<-RUBY.strip_heredoc
     ActiveAdmin::Comment.create!(
       namespace: :admin,
       author: AdminUser.first,
-      body: "Test comment #{i}",
+      body: "Test comment \#{i}",
       resource: categories.sample,
     )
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
This is a code snippet inside a heredoc, so we need to escape the variable, so that the correct snippet is generated. Otherwise the variable gets evaluated before the snippet generation, where it's undefined:

```pry
[1] pry(main)> code = <<~RUBY
[1] pry(main)*   5.times do |i|
[1] pry(main)*     puts "#{i}"
[1] pry(main)*   end
[1] pry(main)* RUBY
NameError: undefined local variable or method `i' for main:Object
from (pry):3:in `__pry__'


[2] pry(main)> code = <<~RUBY
[2] pry(main)*   5.times do |i|
[2] pry(main)*     puts "\#{i}"
[2] pry(main)*   end
[2] pry(main)* RUBY
=> "5.times do |i|\n" + "  puts \"\#{i}\"\n" + "end\n"


[3] pry(main)> eval(code)
0
1
2
3
4
=> 5
[4] pry(main)> 
```

PR #5139, opened 2017-08-15
Update tested railses

deivid-rodriguez:
This PR bumps the Rails versions we test on Travis.


deivid-rodriguez:
One build entry hit a network problem while installing phantomjs. Unrelated to this, so merging.

PR #5140, opened 2017-08-15
Remove debugging traces from tests

deivid-rodriguez:
Running tests currently open a browser window. This PR removes that, and other debugging traces.


javierjulio:
I was wondering what was causing that. Thanks!

PR #5150, opened 2017-08-20
StringSymbolOrProcSetting < DynamicSetting

varyonic:
This refactoring encapsulates dynamic behavior of settings, easing maintenance.
Declarative approach makes clear in NamespaceSettings which settings have dynamic behavior.
Reduces usage of view helper: replaces more than half of invocations of call_method_or_proc_on
originally introduced by #1033

### DIFF
--- a/lib/active_admin/dynamic_setting.rb
+++ b/lib/active_admin/dynamic_setting.rb
@@ -1,8 +1,14 @@
 module ActiveAdmin
 
   class DynamicSetting
-    def self.build(setting)
-      new(setting)
+    def self.build(setting, type)
+      (type ? klass(type) : self).new(setting)
+    end
+
+    def self.klass(type)
+      klass = "#{type.to_s.camelcase}Setting"
+      raise ArgumentError, "Unknown type: #{type}" unless ActiveAdmin.const_defined?(klass)
+      ActiveAdmin.const_get(klass)
     end
 
     def initialize(setting)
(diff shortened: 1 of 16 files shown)


lib/active_admin/views/footer.rb
```diff
@@ -6,21 +6,17 @@ def build(namespace)
...
 
-      def footer?
-        @namespace.footer.present?
-      end
-
       private
 
       def footer_text
-        helpers.render_or_call_method_or_proc_on(self, @namespace.footer)
+        @namespace.footer(self)
```
Fivell: @varyonic maybe we should cache such stuff  using instance variable  since you now use 
```footer_text.present?```

File: lib/active_admin/dynamic_setting.rb [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+
+    def value(*_args)
+      @setting
+    end
+  end
+
+  class StringSymbolOrProcSetting < DynamicSetting
+    def value(*args)
+      return @setting if @setting.is_a?(String) || @setting.nil?
+      (context = args.first) ? context.instance_eval(&@setting) : @setting.call
```
Fivell: @varyonic  would be nice to have yard docs or something like this for this method definition

Fivell: btw @varyonic  do we have StringSymbolOrProcSetting without a context ?

varyonic:
I copied the description of StringSymbolOrProcSetting from render_or_call_method_or_proc_on as they are almost the same, but `call_method_or_proc_on` is overkill as `instance_eval` is sufficient.

deivid-rodriguez:
I do like the declarativeness of `:string_symbol_or_proc` though, and keeping all settings in the same place. So your approach is probably fine... :+1:

PR #5152, opened 2017-08-22
Setup codeclimate

deivid-rodriguez:
I configured codeclimate for active admin.

Some time ago other contributors mentioned that it'd be nice to have style checks as a separate PR check instead of having to wait for the whole CI build to complete. This setup should allow that.

Note that I haven't yet moved `rubocop` to this `codeclimate` config (which would be the main benefit of this). I'm waiting until they add support for the latest `rubocop` version. According to https://github.com/codeclimate/codeclimate-rubocop/issues/99#issuecomment-322381645, it should be in the works.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -50,6 +50,10 @@ engines:
   markdownlint:
     enabled: true
 
+  rubocop:
+    enabled: true
+    channel: rubocop-0-49
+
 exclude_paths:
   - "gemfiles/vendor/**/*"
   - ".test-rails-apps/**/*"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,58 @@
+---
+
+engines:
+  duplication:
+    enabled: true
+
+    config:
+      languages:
+        javascript:
+
+        ruby:
+          mass_threshold: 100
+
+  fixme:
+    enabled: true
+
+    config:
+      strings:
+        - FIXME
+        - BUG
+
+  grep:
+    enabled: true
+
+    config:
+      patterns:
+        no-trailing-whitespace:
+          pattern: \s*$
+          annotation: "Don't leave trailing whitespace"
+          severity: minor
+          categories: Style
+          path_patterns:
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
`codeclimate-rubocop` 0.49 support is actually happening, so we could hold this a bit longer and I'll move  `rubocop` check to `codeclimate` in this PR.

deivid-rodriguez:
It's ready now!

varyonic:
Can you point at another project using this? I don't really understand the impact.

deivid-rodriguez:
Rails, for example: https://github.com/rails/rails/blob/master/.codeclimate.yml.

The idea is to have all code quality checks and linters separated from Travis. This was brought up by @javierjulio if I recall correctly in slack. It was not good to only find out about a style issue until the end of the Travis build.

Other than this, it does not have a lot of impact.

javierjulio:
Awesome! Thanks @deivid-rodriguez! ❤️

deivid-rodriguez:
Forgot to comment that I also added a [codeclimate-grep](https://github.com/codeclimate/codeclimate-grep) check to prevent trailing whitespace globally which I copied from https://github.com/decidim/decidim/blob/3c55b3ffe80919f1211227838d94a3dee59006be/.codeclimate.yml.

deivid-rodriguez:
@varyonic Cool with this?

deivid-rodriguez:
> It would be nice to move more links to bottom of README when there is a chance.

Done!

deivid-rodriguez:
I shouldn't have removed `rubocop` from our `Gemfile`, since even if we can still run rubocop via `codeclimate`, we can no longer run more advanced tasks like `rubocop --auto-correct`. Sorry about that. I'll re-add it if you agree.

javierjulio:
@deivid-rodriguez ah yes, you mean to run it locally? Yes I think that's fine and very helpful. Make a PR and I'll approve it for you.

deivid-rodriguez:
Yes, exactly!

PR #5154, opened 2017-08-23
Move JavaScript script tag in the body and add async and defer attributes

rhymes:
Hi,

I've made the following two modifications:

- moved <script> tags at the end of the <body>
- added "async" and "defer" attributes, only if the Rails assets concatenation is enabled

**Why at the end of the body**

None of the JavaScript in AA requires the browser to block the rendering to execute the script tag so we can put it right before the </body>. This will also allow older browser to render the pages faster.

**Why async and defer**

Basically async (introduced in HTML 5) tells the browser to execute the script tag asynchronously and defer tells the browser "execute this after you finished parsing the document and before the onload event".

Both attributes are ignored by browsers who don't support them.

They are disabled if the asset pipeline is not concatenating.

**Resources**

* [<script> tag](https://developer.mozilla.org/en/docs/Web/HTML/Element/script)
* [async and defer browser compatibility](https://developer.mozilla.org/en/docs/Web/HTML/Element/script#Browser_compatibility)
* [visual explanation of what async and defer do](http://www.growingwiththeweb.com/2014/02/async-vs-defer-attributes.html)


rhymes:
@varyonic @javierjulio is this PR still relevant? I can close it if it's not

Issue #5148, opened 2017-08-18
The Method member_action fails for a particular action name.


GeorgeLautenschlager:
### Expected behavior

The member_action method should generate a route.

### Actual behavior

The member_action method crashes with an error:
`wrong number of arguments (given 1, expected 0)`

### How to reproduce

Actual model names and attributes have been renamed, but here's the code as it stands:

```
ActiveAdmin.register SomeModel do
  menu parent: "parent"
  config.filters = false
  config.per_page = 50
  config.comments = false

  form do |f|
    f.inputs "Details" do
      f.input :attribute1
      f.input :attribute2
      f.input :file_attribute, as: :file, required: false, label: "File Attribute"
    end

    f.actions
  end

  action_item :process, only: :show  do
    link_to 'Process', process_admin_some_model_path(resource), method: :put
  end

  member_action :process, method: :put do
    resource.process!

    redirect_to resource_path
  end
end
```
What seems particularly odd about this is that the exact same code works fine if the action is renamed to 'generate'. Like so:

```
ActiveAdmin.register SomeModel do
  menu parent: "parent"
  config.filters = false
  config.per_page = 50
  config.comments = false

  form do |f|
    f.inputs "Details" do
      f.input :attribute1
      f.input :attribute2
      f.input :file_attribute, as: :file, required: false, label: "File Attribute"
    end

    f.actions
  end

  action_item :generate, only: :show  do
    link_to 'Process', generate_admin_some_model_path(resource), method: :put
  end

  member_action :generate, method: :put do
    resource.process!

    redirect_to resource_path
  end
end
```

aarek:
Hi. You are seeing this error because by adding `member_action` you redefined one of the `AbstractController::Base` core methods: [AbstractController::Base#process](http://api.rubyonrails.org/classes/AbstractController/Base.html#method-i-process).

The error at this moment is very cryptic so i added Pull Request to raise error informing that the method is already defined while registering resource in active admin

varyonic:
aarek's explanation is correct. Maybe it was a bad design decision to use the action name directly as a generated controller method name (eg. without appending "_action"). Leaving this open because too many users follow the GitHub default and don't search closed issues.

PR #5160, opened 2017-08-30
raise error when trying to redefine controller method

aarek:
This change addresses problem in #5148.
Raising error during application loading would be better than raising uncommunicative error during request

### DIFF
--- a/lib/active_admin/resource_dsl.rb
+++ b/lib/active_admin/resource_dsl.rb
@@ -131,6 +131,8 @@ module ActiveAdmin
     # action.
     #
     def action(set, name, options = {}, &block)
+      raise ArgumentError, "method #{name} already defined" if controller.method_defined?(name)
+
       set << ControllerAction.new(name, options)
       title = options.delete(:title)
 
--- a/spec/unit/action_builder_spec.rb
+++ b/spec/unit/action_builder_spec.rb
@@ -1,124 +1,156 @@
 require 'rails_helper'
 
 RSpec.describe 'defining actions from registration blocks', type: :controller do
-  let(:klass){ Admin::PostsController }
+  context 'when method with given name does not exist' do
+    let(:klass){ Admin::PostsController }
 
-  before do
-    load_resources { action! }
+    before do
+      load_resources { action! }
 
-    @controller = klass.new
+      @controller = klass.new
+    end
+
+    describe 'creates a member action' do
+      after(:each) do
+        klass.clear_member_actions!
+      end
+
+      context 'with a block' do
+        let(:action!) do
+          ActiveAdmin.register Post do
+            member_action :comment do
+              # Do nothing
+            end
+          end
+        end
+
+        it 'should create a new public instance method' do
+          expect(klass.public_instance_methods.collect(&:to_s)).to include('comment')
+        end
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Thanks!

varyonic:
Turns out I've been happily overwriting `clone` for years :-)  More seriously we have been defining the same member action with both get and patch, eg.
```ruby
member_action :change_payment_method do
  ...
end

member_action :change_payment_method, method: :patch do
  ...
end
```
This has stopped working now.  Perhaps this could be a warning instead of an error?

deivid-rodriguez:
What if we keep a blacklist of names, including only `process` for now?

varyonic:
Better than broken.

deivid-rodriguez:
@aarek Could you contribute another PR raising only when the `process` method is redefined?

aarek:
@deivid-rodriguez personally in this situation i think warning would be better - it will cover all cases, so in future releases error could be raised. I will prepare PR tomorrow if that's ok with you.

deivid-rodriguez:
But if I understood correctly, the warning is incorrect in the case @varyonic presented, no?

varyonic:
I could live with a warning and we (my employer) could change it if it bugged us enough.  I'm surprised what we are doing works, and I don't think it is documented behavior.  Normally one would do
```ruby
  member_action :add_check_payment, method: [:post, :get] do
    if request.get?
      ...
   else
      ...
    end
  end
```

aarek:
PR with warning here #5167

jpawlyn:
In order not to have this warning displayed (we get it whenever we run a test), would it be ok for me to raise a PR that added a boolean configuration variable to the initializer that controlled this logging with the default to display the warning?

varyonic:
@jpawlyn As far as I know you only get the warning if you are using undocumented behavior as I commented above along with the fix. I am unclear why a boolean configuration variable would be desirable.

jpawlyn:
@varyonic I guess it is undocumented behavior we're relying upon. Perhaps if I explain what our aim was, it'll be clear that there's a better to achieving it than overriding an existing action!

We have a User model for which we can update first and last name and also reset the password using the activeadmin edit page. However, if no password reset is required, we want to be able to modify the name fields and leave the password and password confirmation fields empty. To achieve this we override the `update` action in order to delete the password params if they are blank. If we don't remove the password params, devise validation fails.

We could create a custom form for password reset but it's handy being able to edit user attributes all in one place. Any suggestion for handling this would be much appreciated.

varyonic:
I see. We use a separate custom action for password reset, but we don't depend on Devise. Feedback from others welcome.

javierjulio:
@jpawlyn we have the same update functionality as you for our user records but instead we override the `update_resource` method. If it helps at all, this is in our `admin/user.rb` file:

```ruby
  controller do
    def update_resource(object, attributes)
      attrs, role = attributes

      if attrs[:password].blank?
        attrs.delete :password
        attrs.delete :password_confirmation
      end

      object.send :update, attrs
    end
  end
```

jpawlyn:
@javierjulio That's a great help, just what I needed, thanks a lot.

Issue #1833, opened 2012-12-30
active_admin conflicts gem will_paginate - need this in initialiser


rafaqat:
https://github.com/gregbell/active_admin/wiki/How-to-work-with-will_paginate

How to work with will_paginateNew Page Edit Page Page History
ActiveAdmin uses Kaminari for pagination and if you use will_paginate in your app, you need to configure an initializer for Kaminari to avoid conflicts. Put this in config/initializers/kaminari.rb

``` ruby
Kaminari.configure do |config|
  config.page_method_name = :per_page_kaminari
end
```

OR - put this in the same initializer if that doesn't work

``` ruby
if defined?(WillPaginate)
  module WillPaginate
    module ActiveRecord
      module RelationMethods
        def per(value = nil) per_page(value) end
        def total_count() count end
      end
    end
    module CollectionMethods
      alias_method :num_pages, :total_pages
    end
  end
end
```

PR #5161, opened 2017-08-30
Adding :entry_name so will_paginate works with current documentation

ErvalhouS:
I've been unsuccessfull on following current documentation about support of activeadmin along with will_paginate. The only way things work around here is following fully the instructions on [#1833](https://github.com/activeadmin/activeadmin/issues/1833) and adding explicit "entry_name" declaration when building comments view. Not sure if it suits all use cases, but sure does mine. Maybe better localization for entry_name would be needed.

### DIFF
--- a/lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb
+++ b/lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb
@@ -26,7 +26,7 @@ module ActiveAdmin
         def build_comments
           if @comments.any?
             @comments.each(&method(:build_comment))
-            div page_entries_info(@comments).html_safe, class: 'pagination_information'
+            div page_entries_info(@comments).html_safe, class: 'pagination_information', entry_name: I18n.t('active_admin.comments.resource')
           else
             build_empty_message
           end


lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb
```diff
@@ -26,7 +26,7 @@ def title
         def build_comments
           if @comments.any?
             @comments.each(&method(:build_comment))
-            div page_entries_info(@comments).html_safe, class: 'pagination_information'
+            div page_entries_info(@comments).html_safe, class: 'pagination_information', entry_name: I18n.t('active_admin.comments.resource')
```
deivid-rodriguez: Haven't tried this PR but after reading the specs we have for `:entry_name`:

https://github.com/activeadmin/activeadmin/blob/6038ea6f8fd048926d59ba2c8c29ca2c40d1d41a/spec/unit/views/components/paginated_collection_spec.rb#L94-L161

it seems that this would show "Display 1 resource" instead of "Display 1 comment". Is this what you want? Is it better? How is it related to `will_paginate`? Maybe there's a bug with `will_paginate` support when infering the class name automatically (no `:entry_name` specified)...

ErvalhouS: This is exactly what I meant about better internationalization. The only error I have ran into when trying to make it work on my project that has `will_paginate` was that one, all documentation instructions were OK, but that was never mentioned this problem and it was my case. I will update this later today with the I18n that makes it display as intended. I am not sure if it is a bug with `will_paginate`, I just read the error stack trace and fixed what it was broken in my case.

deivid-rodriguez: Hi @ErvalhouS, sorry for getting back so late. What's the stacktrace you're talking about? I still don't get what problem this is fixing...

imsinu9:
@ErvalhouS I am sure its late to ask this question but sure it will help saving some time for fellow visitors to this post.

Are you also using any bootstrap compatible rendering (popularly `will_paginate-bootstrap`) with `will_paginate`?. Actually there is no other configuration required with AA & `will_paginate` compatibility else than instructions on #1833. There has been similar issues with another gem check this out  thredded/thredded#654.

I recommend decommission of this PR.

deivid-rodriguez:
I'm going to go ahead and close this PR. @ErvalhouS feel free to answer our questions and reopen if you still feel this PR is needed!

PR #5165, opened 2017-09-04
Make ResourceController contain actual action methods only

chumakoff:
In any Rails controller public methods are considered as action methods and may be available through routes. So, we should avoid unnecessary public methods in controllers, which are not supposed to be action methods.

Now `ActiveAdmin::ResourceController.action_methods`  returns 

`#<Set: {"batch_action", "build_download_formats", "build_download_format_links", "index", "scope_chain", "only_render_implemented_actions", "render_or_call_method_or_proc_on", "render_in_context", "call_method_or_exec_proc", "call_method_or_proc_on", "new", "update", "create", "destroy", "edit", "show"}>`

But it should be

`#<Set: {"batch_action", "index",  "new", "update", "create", "destroy", "edit", "show"}>`

This PR fixes this problem.

### DIFF
--- a/lib/active_admin/base_controller.rb
+++ b/lib/active_admin/base_controller.rb
@@ -22,17 +22,17 @@ module ActiveAdmin
       attr_accessor :active_admin_config
     end
 
+    include Authorization
+    include Menu
+
+    private
+
     # By default Rails will render un-implemented actions when the view exists. Because Active
     # Admin allows you to not render any of the actions by using the #actions method, we need
     # to check if they are implemented.
     def only_render_implemented_actions
       raise AbstractController::ActionNotFound unless action_methods.include?(params[:action])
     end
-
-    include Authorization
-    include Menu
-
-    private
 
     # Calls the authentication method as defined in ActiveAdmin.authentication_method
     def authenticate_active_admin_user
--- a/lib/active_admin/helpers/scope_chain.rb
+++ b/lib/active_admin/helpers/scope_chain.rb
@@ -1,5 +1,6 @@
 module ActiveAdmin
   module ScopeChain
+    private
     # Scope an ActiveRecord::Relation chain
     #
     # Example:
--- a/lib/active_admin/resource_controller/action_builder.rb
+++ b/lib/active_admin/resource_controller/action_builder.rb
@@ -7,11 +7,21 @@ module ActiveAdmin
       module ClassMethods
 
         def clear_member_actions!
+          remove_action_methods(:member)
           active_admin_config.clear_member_actions!
         end
 
         def clear_collection_actions!
+          remove_action_methods(:collection)
           active_admin_config.clear_collection_actions!
+        end
+
+        private
+
+        def remove_action_methods(actions_type)
+          active_admin_config.public_send("#{actions_type}_actions").each do |action|
+            remove_method action.name
+          end
         end
       end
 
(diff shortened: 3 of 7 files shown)


lib/active_admin/base_controller/authorization.rb
```diff
@@ -1,8 +1,13 @@
...
   class BaseController < ::InheritedResources::Base
     module Authorization
-      include MethodOrProcHelper
       extend ActiveSupport::Concern
+      include MethodOrProcHelper
+
+      # prevent mixing of helper methods with controller action methods
+      MethodOrProcHelper.public_instance_methods(false).each do |method_name|
+        private method_name
+      end
```
varyonic: :scream: Why is MethodOrProcHelper included here @deivid-rodriguez ?  Oh yes, see dispatch_active_admin_access_denied in bb95d3d

deivid-rodriguez: So if we use the alternative you proposed in #5084:

```diff
- call_method_or_exec_proc active_admin_namespace.on_unauthorized_access, exception
+ instance_exec(self, exception, &active_admin_namespace.on_unauthorized_access.to_proc)
```

Then we will no longer need this? If that's the case, let's re-add it to #5084 then, you convinced me :)

chumakoff: Should I wait for #5084 to be merged, then rebase this one?

varyonic: @deivid-rodriguez  Unfortunately it is not so simple.  Since `MethodOrProcHelper` was included in `ResourceController` by #1817 `render_in_context()` has been used in two additional pull requests: #2089 (scope_to if/unless) and #3503 (batch_action options).  I have no problem with render_in_context(), but what I don’t like is that multiple generic instance_exec++ utility methods have gathered in a poorly cohesive module that is included in multiple places.  These two other references could be fixed with an explicit `MethodOrProcHelper.render_in_context` or by using `StringSymbolOrProcSetting.new(...).value`  Thoughts?

deivid-rodriguez: I see. An explicit `MethodOrProcHelper` receiver so we don't need to include the module sounds good, but you maybe need to `extend self` on the module in order to be able to make that call?

The other alternative sounds neat too! I pretty much trust your judgement and I apologize for being a pain in the ass sometimes... :)

varyonic: @chumakoff #5176 has been merged, you can go ahead and rebase now.

chumakoff: @varyonic Done. Added something new to `resource_controller/action_builder.rb`

lib/active_admin/base_controller.rb
```diff
@@ -22,18 +22,18 @@ class << self
       attr_accessor :active_admin_config
     end
 
+    include Menu
+    include Authorization
```
varyonic: See 7ec7d2f3f  My guess is this is an accidental (untested?) regression caused by a rebase.

chumakoff: @varyonic  I changed the order. But I am sure that the order of the includes can't affect the code that is in the `navigation_menu` block, since the block and related methods are never executed while `BaseController` is loaded. It is only definition at that moment.

chumakoff:
@varyonic Done.

PR #5166, opened 2017-09-04
Don't try to find an associated class if the attribute is a column on the model

### LINKED ISSUE #5155: Incorrect association foreign_key is used by ActiveFilter when the same column name exists on the model
### Expected behavior

I expected the filter to filter like it would with Ransacker.

### Actual behavior

The `find_class` override changes the meaning of the ransacker filter.

### How to reproduce

```ruby
begin
  require 'bundler/inline'
rescue LoadError => e
  raise e
end

gemfile(true) do
  source 'https://rubygems.org'

  if ENV['ACTIVE_ADMIN_PATH']
  else
  end

  gem 'rails', '~> 5.1.0'

  gem 'sass-rails'
  gem 'sqlite3', platform: :mri
  gem 'activerecord-jdbcsqlite3-adapter',
      git: 'https://github.com/jruby/activerecord-jdbc-adapter',
      branch: 'rails-5',
      platform: :jruby
end

require 'active_record'

  end

    t.string :full_name
  end

  create_table :user_devices, force: true do |t|
    t.string :user_id
    t.string :device_udid
    t.string :device_type
  end

  create_table :blocked_user_devices, force: true do |t|
    t.string :device_udid
    t.string :device_type
  end
end

require 'action_view/railtie'
require 'active_admin'

  config.root = __dir__

  config.eager_load = false
  Rails.logger  = config.logger
end

end

  has_many :user_devices
end

class UserDevice < ActiveRecord::Base
  belongs_to :user
  has_one :blocked_user_device, -> (object) { where(device_type: object.send(:device_type)) }, foreign_key: :device_udid, primary_key: :device_udid
end

class BlockedUserDevice < ActiveRecord::Base; end

ActiveAdmin.setup do |config|
end

Rails.application.initialize!

  content do
    'Test Me'
  end
end

ActiveAdmin.register User do
  filter :user_devices_device_udid_eq, label: 'UDID'
end

  ActiveAdmin.routes(self)
end

require 'minitest/autorun'
require 'rack/test'
require 'rails/test_help'

  def test_admin_root_success?
    get admin_root_url
    assert_response :success
  end

  def test_admin_users
    user1 = User.create! full_name: "Test McTester"
    UserDevice.create! user: user1, device_udid: '123', device_type: 'android'
    UserDevice.create! user: user1, device_udid: '456', device_type: 'ios'
    user2 = User.create! full_name: "Test McTester's Friend"
    UserDevice.create! user: user2, device_udid: '123', device_type: 'android'
    UserDevice.create! user: user2, device_udid: '456', device_type: 'ios'

    BlockedUserDevice.create device_udid: '456', device_type: 'android'

    # I expect this to filter and show all of the users who have devices
    # with the same device UDID. However, instead it searches for the
    # BlockerUserDevice due to its foreign_key on UserDevice.

    # It throws this exception:
    # F, [2017-08-24T14:02:33.944348 #64026] FATAL -- :
    # F, [2017-08-24T14:02:33.944370 #64026] FATAL -- : ActionView::Template::Error (Couldn't find BlockedUserDevice with 'id'=123):
    # F, [2017-08-24T14:02:33.944449 #64026] FATAL -- :     1: insert_tag renderer_for(:index)
    # F, [2017-08-24T14:02:33.944463 #64026] FATAL -- :
    # because instead of searching for UserDevice.where(device_udid: '123'), it's
    # actually looking for a BlockerUserDevice.find_by(id: '123'), which was
    # incorrectly selected by this line:
    # https://github.com/activeadmin/activeadmin/blob/0df2df0bf65a6e22dcf46d01dbaa12d6e995fb89/lib/active_admin/filters/active_filter.rb#L81

    get admin_users_url, params: { "utf8" => "✓", "q[user_devices_device_udid_eq]" => "123", "commit" => "Filter" }
    assert_response :success
  end

  private

  def app
    Rails.application
  end
end
```

## Workaround

I monkey-patched the `find_class?` method introduced with #5092 with this logic to work around the issue:

```ruby
module ActiveAdmin
  module Filters
    class ActiveFilter
      private

      def find_class?
        # This checks to see if there's a column name that matches the specified attribute
        # and skips looking for relationships if the attribute exists.
        return false if condition_attribute.klass.column_names.include?(name.to_s)

        super
      end
    end
  end
end
```

blimmer:
This PR fixes #5155. See that issue for more information.

cc @Fivell

### DIFF
--- a/lib/active_admin/filters/active_filter.rb
+++ b/lib/active_admin/filters/active_filter.rb
@@ -68,6 +68,8 @@ module ActiveAdmin
       end
 
       def find_class?
+        return false if condition_attribute.klass.column_names.include?(name.to_s)
+
         ['eq', 'in'].include? condition.predicate.arel_predicate
       end
 


blimmer:
@Fivell , I'm not quite sure what the best way to write a test for this behavior is.

Since the test suite relies on global models (User, Post, etc.), it feels kind of weird to create the exact situation that causes this problem.

The trick to exposing this problem is that you need models that look like this:

```ruby
  has_many :user_devices
end

class UserDevice < ActiveRecord::Base
  belongs_to :user
  has_one :blocked_user_device, -> (object) { where(device_type: object.send(:device_type)) }, foreign_key: :device_udid, primary_key: :device_udid
end

class BlockedUserDevice < ActiveRecord::Base; end
```

specifically, the `has_one` relationship with a different than normal foreign and primary key.

are you OK with me mutating the global models to add this kind of case for this one test?

Fivell:
@blimmer  can you take a look at failing specs, I will be able to rethink this case later, now  very busy :(

wspurgin:
@blimmer This monkey patch actually prevents `find_class?` from returning `true` in more simple situations. Like if you are searching a `belongs_to` filter. Using models from the specs:

```ruby
ActiveAdmin.register Post do
# ...
filter :author # this searches the author_id on the Post model
end
```
The Ransacker would still search the Post model (with our without the monkey patch), but with this patch the Current Filter section will display "Author ID equals some_value" versus "Author equals named_link_to_author_object".

I think this particular case might be better handled by manually letting active filter know that "for this filter, don't try to find a related class". Maybe along the lines of:

```ruby
filter :user_devices_device_udid_eq, label: "UUID", active_filter_search: false
```
Then `ActiveFilter` can use that as a switch to know that the developer has specifically requested for it not to do something fancy. I know its not the perfect solution, but it handles your specific case and is at least a fall back for other future cases if some behavior is undesired.

Thoughts?

blimmer:
@wspurgin - I think that makes sense - this monkey patch I've added definitely was just what I did to get it working. I'm open to that idea. I likely won't have time to work on this for a few weeks, but it'd be good to get an "OK" on that way forward from a maintainer before we try to tackle that.

javierjulio:
Not sure if this is still applicable with the changes made to filters since. Also several refactors. If so and can be updated, I'll reopen. Thank you.

PR #5174, opened 2017-09-13
Restore lint tools locally

deivid-rodriguez:
Follow up to #5152.

All yours @javierjulio!


deivid-rodriguez:
Thanks @javierjulio!

PR #5177, opened 2017-09-15
Refactor Router

varyonic:
This series of refactorings, heavily inspired by #3170, replaces all use of proc (x5) and instance_exec (x6) in Router with vanilla method calls, making it easier to understand and maintain.

### DIFF
--- a/lib/active_admin/router.rb
+++ b/lib/active_admin/router.rb
@@ -33,7 +33,7 @@ module ActiveAdmin
     def define_resource_routes(router)
       resources = @application.namespaces.flat_map{ |n| n.resources.values }
       resources.each do |config|
-        routes = resource_routes(router, config)
+        routes = proc { resource_routes(router, config) }
 
         # Add in the parent if it exists
         if config.belongs_to?
@@ -65,7 +65,6 @@ module ActiveAdmin
     end
 
     def resource_routes(router, config)
-      Proc.new do
         case config
         when ::ActiveAdmin::Resource
           router.resources config.resource_name.route_key, only: config.defined_actions do
@@ -90,7 +89,6 @@ module ActiveAdmin
           raise "Unsupported config class: #{config.class}"
         end
       end
-    end
 
     # Deals with +ControllerAction+ instances
     # Builds one route for each HTTP verb passed in
--- a/lib/active_admin/router.rb
+++ b/lib/active_admin/router.rb
@@ -40,12 +40,12 @@ module ActiveAdmin
           belongs_to = routes
           routes     = Proc.new do
             # If it's optional, make the normal resource routes
-            instance_exec &belongs_to if config.belongs_to_config.optional?
+            belongs_to.call if config.belongs_to_config.optional?
 
             # Make the nested belongs_to routes
             # :only is set to nothing so that we don't clobber any existing routes on the resource
             router.resources config.belongs_to_config.target.resource_name.plural, only: [] do
-              instance_exec &belongs_to
+              belongs_to.call
             end
           end
         end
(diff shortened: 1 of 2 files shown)


varyonic:
I don't understand why codecov/patch failed here.

deivid-rodriguez:
I think it's been a big coincidence that the coverage of your patch matches the global coverage of the project :astonished:. Probably your patch loses at the third decimal digit.

I normally use a more relaxed configuration for this in `codecov.yml`, such as

```
    patch:
      default:
        target: 50%
```

deivid-rodriguez:
Yeah, getting the patch 100% covered is another option :smiley:.

deivid-rodriguez:
By the way, codeclimate just rolled out their out code coverage status checks in PR, and they chose 80% as a default for this setting: https://codeclimate.com/changelog/59baa212ced24c026c0017b8.

lib/active_admin/router.rb
```diff
@@ -91,14 +83,23 @@ def resource_routes(router, config)
...
             end
           end
         else
           raise "Unsupported config class: #{config.class}"
         end
       end
+    end
+
+    # Deals with +ControllerAction+ instances
+    # Builds one route for each HTTP verb passed in
```
Fivell: @varyonic , I think this methods should be tested or defined as private

lib/active_admin/router.rb
```diff
@@ -55,40 +55,40 @@ def define_resource_routes(router)
...
+          end
 
-            router.collection do
-              config.collection_actions.each { |action| build_action(router, action) }
-              router.post :batch_action if config.batch_actions_enabled?
-            end
+          router.collection do
+            config.collection_actions.each { |action| build_action(router, action) }
+            router.post :batch_action if config.batch_actions_enabled?
           end
```
Fivell: @varyonic , is this big diff necessary?

lib/active_admin/router.rb
```diff
@@ -90,6 +83,18 @@ def resource_routes(router, config)
       end
     end
 
+    # Defines member and collection actions
+    def define_actions(router, config)
+      router.member do
+        config.member_actions.each { |action| build_action(router, action) }
+      end
+
```
Fivell: these might be not public too ?

lib/active_admin/router.rb
```diff
@@ -1,27 +1,21 @@
...
-    #
-    # Use this within the routes.rb file:
-    #
-    #   Application.routes.draw do |map|
-    #     ActiveAdmin.routes(self)
-    #   end
-    #
-    # @param router [ActionDispatch::Routing::Mapper]
-    def apply(router)
+    def apply(router = @router)
```
Fivell: @varyonic, not sure this commit  makes code more cleaner, I don't understand now if there   there are cases when different routers can be used for initialize and apply method ? in other words after router passed to constructor, why this method needs argument

Fivell: @varyonic  ignore this since it was fixed by next commit

lib/active_admin/router.rb
```diff
@@ -1,108 +1,113 @@
...
+        define_routes(config)
+      else
+        # Add on the namespace if required
+        define_namespace(config)
+      end
+    end
 
-          instance_exec &routes
-        end
+    def define_routes(config)
```
deivid-rodriguez: All of this methods are defining routes, could this method name be more clear?

lib/active_admin/router.rb
```diff
@@ -1,108 +1,113 @@
...
+      if config.belongs_to?
+        define_belongs_to_routes(config)
+      else
+        page_or_resource_routes(config)
```
deivid-rodriguez: For consistency, should these methods be named either `define_belongs_to_routes` & `define_page_or_resource_routes`, or `belongs_to_routes` & `page_or_resource_routes`?

varyonic: I agree the method naming is not great, I was focusing on unfolding the code and followed the examples of #3170.  Renaming methods retroactively with an interactive rebase can get tricky so I'll look at a final commit to do this.

varyonic: On second thoughts reviewing method names just left me wanting to refactor even further. I'ld rather close this technical excision of proc/exec and leave further improvements to a separate PR.

varyonic:
Thanks to @dmitry for showing this could be done.

PR #5180, opened 2017-09-16
Added MOVE button translation key; improved italian translations

blocknotes:


blocknotes:
I added "active_admin.filters.predicates.eq" also to en.yml because it's not translated in my language (italian).

In the first commit I added also the keys:
```
views.pagination.first
views.pagination.last
views.pagination.next
views.pagination.previous
views.pagination.truncate
```
Then I removed them because they are not defined in en.yml (for Travis) - shouldn't be added?

config/locales/en.yml
```diff
@@ -21,12 +21,14 @@ en:
...
     has_many_delete: "Delete"
     has_many_remove: "Remove"
+    move: "MOVE"
     filters:
       buttons:
         filter: "Filter"
         clear: "Clear Filters"
       predicates:
         contains: "Contains"
+        eq: "equals"
```
deivid-rodriguez: Not sure but I think this a "native" ransack predicate so that's why we don't provide translation. It should be translated in ransack? @Fivell?

config/locales/en.yml
```diff
@@ -21,12 +21,14 @@ en:
     has_many_new: "Add New %{model}"
     has_many_delete: "Delete"
     has_many_remove: "Remove"
+    move: "MOVE"
```
deivid-rodriguez: I guess we should keep the case consistent. I'm up for making this lowercase.

lib/active_admin/form_builder.rb
```diff
@@ -99,8 +99,8 @@ def has_many_actions(has_many_form, builder_options, contents)
       if builder_options[:sortable]
         has_many_form.input builder_options[:sortable], as: :hidden
 
-        contents << template.content_tag(:li, class: 'handle') do
-          "MOVE"
+        contents << template.content_tag(:li, class: 'handle sortable') do
```
deivid-rodriguez: Adding the "sortable" seems unrelated to this PR, but I think it makes sense to me.

config/locales/en.yml
blocknotes: I agree, I added it uppercase because in the source is uppercase

config/locales/en.yml
blocknotes: You are right, "eq" is in Ransack translations [ransack_en](https://github.com/activerecord-hackery/ransack/blob/master/lib/ransack/locale/en.yml). I could make a PR to Ransack for italian lang

lib/active_admin/form_builder.rb
javierjulio: Don't add this class so just remove the change please. It has been changed in #3862.

blocknotes: Yes :) I added it just to make easier to style it

config/locales/en.yml
javierjulio: This will change in #3862 but I agree we should have localized. That way I can use it in #3862 as alt text. Agree with @deivid-rodriguez on keeping case consistent. The text in the source will be removed anyway. Thanks!

lib/active_admin/form_builder.rb
javierjulio: Don't worry, we'll make sure it is in #3862 but we would prefer to avoid any changes here unless necessary and I don't think it is. I'll get that in #3862 for you. 😄

blocknotes: ook thanks :)
removed in the next commit

config/locales/it.yml
```diff
@@ -20,6 +20,7 @@ it:
     has_many_new: "Aggiungi nuovo/a %{model}"
     has_many_delete: "Rimuovi"
     has_many_remove: "Rimuovi"
+    move: "sposta"
```
deivid-rodriguez: You want this to start uppercase I guess?

blocknotes: Capitalized I suppose. Done

deivid-rodriguez:
It's ready! Ideally you'd squash all commits into two, one for each sentence in the PR title:

* Add MOVE button translation key.
* Improve italian translations.

I'd do it like this:

```bash
git reset HEAD~7 # undo all commits and get the whole diff in your working tree
git add -p # and add interactively the changes for the first commit
git commit # commit the first set of changes
git commit -a # add and commit the rest of the changes
git push --force # force-push to the PR once you're sure everything is fine
```

blocknotes:
Ok! One commit now

Issue #615, opened 2011-10-13
Nested namespaces


bradyhigginbotham:
I need to have a nested namespace:

/admin/namespace/resource.rb*

I'd like to know how I could create this through ruby ("rails g active_admin:resource namespace/Resource" didn't work)

and any additions to routes.rb? possibly like the following:

``` ruby
namespace "admin" do
  namespace "namespace" do
    resources :resources
  end
end
```

Thanks for any suggestions!
- just an example

snowmoonsoftware:
+1 struggling with this myself especially with the menus. See issue #3960

davidwparker:
What's the status on this?

I'd like to have dynamic active_admin, based on the parent's namespace in routes.

config/routes.rb
```
resources :marketplaces do
    ActiveAdmin.routes(self)
```

So I can go to `/marketplaces/:marketplace_id/admin`

config/initializers/active_admin.rb
```
config.namespace :admin do |admin|
    admin.site_title = "Admin"
    admin.authentication_method = :authenticate_admin!
  end
```

app/marketplaces/admin/dashboard.rb
```
  ActiveAdmin.register_page "Dashboard", namespace: :admin do
    menu priority: 1, label: proc{ I18n.t("active_admin.dashboard") }
...
```

Right now, I get the error:
```
uninitialized constant Marketplaces::Admin
```

When I try to go to http://localhost:3000/marketplaces/1/admin

Is this possible? Thanks!

PR #5185, opened 2017-09-24
Nested namespaces support

siutin:
This PR allows you to register resources/pages with nested namespaces. For example:

/app/admin/site1/foo/bar/posts.rb
```ruby
ActiveAdmin.register Post, namespace: [:site1, :foo, :bar] do
  ...
end
```

/app/admin/site2/demo/posts.rb
```ruby
ActiveAdmin.register Post, namespace: [:site2, :demo] do
  ...
end
```

It will generate routes like:

```
              admin_site1_foo_bar_root GET        /admin/site1/foo/bar(.:format)                    admin/site1/foo/bar/dashboard#index
                 admin_site2_demo_root GET        /admin/site2/demo(.:format)                       
batch_action_admin_site1_foo_bar_posts POST       /admin/site1/foo/bar/posts/batch_action(.:format) admin/site1/foo/bar/posts#batch_action
             admin_site1_foo_bar_posts GET        /admin/site1/foo/bar/posts(.:format)              admin/site1/foo/bar/posts#index
                                       POST       /admin/site1/foo/bar/posts(.:format)              admin/site1/foo/bar/posts#create
          new_admin_site1_foo_bar_post GET        /admin/site1/foo/bar/posts/new(.:format)          admin/site1/foo/bar/posts#new
         edit_admin_site1_foo_bar_post GET        /admin/site1/foo/bar/posts/:id/edit(.:format)     admin/site1/foo/bar/posts#edit
              admin_site1_foo_bar_post GET        /admin/site1/foo/bar/posts/:id(.:format)          admin/site1/foo/bar/posts#show
                                       PATCH      /admin/site1/foo/bar/posts/:id(.:format)          admin/site1/foo/bar/posts#update
                                       PUT        /admin/site1/foo/bar/posts/:id(.:format)          admin/site1/foo/bar/posts#update
                                       DELETE     /admin/site1/foo/bar/posts/:id(.:format)          admin/site1/foo/bar/posts#destroy
          admin_site1_foo_bar_comments GET        /admin/site1/foo/bar/comments(.:format)           admin/site1/foo/bar/comments#index
                                       POST       /admin/site1/foo/bar/comments(.:format)           admin/site1/foo/bar/comments#create
           admin_site1_foo_bar_comment GET        /admin/site1/foo/bar/comments/:id(.:format)       admin/site1/foo/bar/comments#show
                                       DELETE     /admin/site1/foo/bar/comments/:id(.:format)       admin/site1/foo/bar/comments#destroy
   batch_action_admin_site2_demo_posts POST       /admin/site2/demo/posts/batch_action(.:format)    admin/site2/demo/posts#batch_action
                admin_site2_demo_posts GET        /admin/site2/demo/posts(.:format)                 admin/site2/demo/posts#index
            edit_admin_site2_demo_post GET        /admin/site2/demo/posts/:id/edit(.:format)        admin/site2/demo/posts#edit
                 admin_site2_demo_post GET        /admin/site2/demo/posts/:id(.:format)             admin/site2/demo/posts#show
                                       PATCH      /admin/site2/demo/posts/:id(.:format)             admin/site2/demo/posts#update
                                       PUT        /admin/site2/demo/posts/:id(.:format)             admin/site2/demo/posts#update
             admin_site2_demo_comments GET        /admin/site2/demo/comments(.:format)              admin/site2/demo/comments#index
                                       POST       /admin/site2/demo/comments(.:format)              admin/site2/demo/comments#create
              admin_site2_demo_comment GET        /admin/site2/demo/comments/:id(.:format)          admin/site2/demo/comments#show
                                       DELETE     /admin/site2/demo/comments/:id(.:format)          admin/site2/demo/comments#destroy

```
---

In your initializers/active_admin.rb, you can define your own handler to the following methods.

* authentication_method
* current_user_method
* logout_link_path

For example:

``` ruby
  # config.authentication_method = :authenticate_admin_user!
  config.authentication_method = Proc.new do |namespace|
    if namespace && namespace.is_a?(Array)
      "authenticate_#{namespace.map(&:to_s).join('_').underscore}_admin_user!".to_sym
    else
      :authenticate_admin_user!
    end
  end
```
And In your routes.rb:

```ruby
  namespace :site1 do
    namespace :foo do
      namespace :bar do
        devise_for :admin_users, ActiveAdmin::Devise.config
      end
    end
  end
  namespace :site2 do
    namespace :demo do
      devise_for :admin_users, ActiveAdmin::Devise.config
    end
  end
```
---

You can checkout the demo here: [https://github.com/siutin/activeadmin-nested-namespaces-example](https://github.com/siutin/activeadmin-nested-namespaces-example)

### DIFF
--- a/lib/active_admin/namespace.rb
+++ b/lib/active_admin/namespace.rb
@@ -37,7 +37,7 @@ module ActiveAdmin
 
     def initialize(application, name)
       @application = application
-      @name = Array(name).first.to_s.underscore.to_sym
+      # @name_splits = name.split('/').map(&:to_sym)
       @name_path = application.build_name_path(name)
       @resources = ResourceCollection.new
       register_module unless root?
@@ -46,7 +46,7 @@ module ActiveAdmin
 
     def name
       Deprecation.warn "name replaced by name_path now that namespaces can be nested."
-      @name
+      name_path.first
     end
 
     def settings
@@ -91,7 +91,7 @@ module ActiveAdmin
     end
 
     def root?
-      @name == :root
+      name_path.first == :root
     end
 
     # Returns the name of the module if required. Will be nil if none
--- a/lib/active_admin/namespace.rb
+++ b/lib/active_admin/namespace.rb
@@ -33,18 +33,20 @@ module ActiveAdmin
 
     RegisterEvent = 'active_admin.namespace.register'.freeze
 
-    attr_reader :application, :resources, :menus
+    attr_reader :application, :resources, :menus, :name_path
 
     def initialize(application, name)
       @application = application
-      @name = name.to_s.underscore
+      @name = Array(name).first.to_s.underscore.to_sym
+      @name_path = application.build_name_path(name)
       @resources = ResourceCollection.new
       register_module unless root?
       build_menu_collection
     end
 
     def name
-      @name.to_sym
+      Deprecation.warn "name replaced by name_path now that namespaces can be nested."
+      @name
     end
 
     def settings
@@ -89,7 +91,7 @@ module ActiveAdmin
     end
 
     def root?
-      name == :root
+      @name == :root
     end
 
     # Returns the name of the module if required. Will be nil if none
@@ -100,11 +102,11 @@ module ActiveAdmin
     #   Namespace.new(:root).module_name # => nil
     #
     def module_name
-      root? ? nil : @name.camelize
+      root? ? nil : name_path.map(&:to_s).map(&:camelize).join('::')
     end
 
     def route_prefix
-      root? ? nil : @name
+      root? ? nil : name_path.map(&:to_s).join('_').underscore
     end
 
     # Unload all the registered resources for this namespace
@@ -147,12 +149,13 @@ module ActiveAdmin
     # @param [Hash] html_options An options hash to pass along to link_to
     #
     def add_logout_button_to_menu(menu, priority = 20, html_options = {})
-      if logout_link_path
+      computed_logout_link_path = logout_link_path.is_a?(Proc) ? logout_link_path.call(name_path) : logout_link_path
+      if computed_logout_link_path
         html_options = html_options.reverse_merge(method: logout_link_method || :get)
         menu.add id: 'logout', priority: priority, html_options: html_options,
-          label: ->{ I18n.t 'active_admin.logout' },
-          url:   ->{ render_or_call_method_or_proc_on self, active_admin_namespace.logout_link_path },
-          if:    :current_active_admin_user?
+                 label:   -> { I18n.t 'active_admin.logout' },
+                 url:   computed_logout_link_path,
+                 if:    :current_active_admin_user?
       end
     end
 
@@ -226,8 +229,10 @@ module ActiveAdmin
 
     # Creates a ruby module to namespace all the classes in if required
     def register_module
-      unless Object.const_defined? module_name
-        Object.const_set module_name, Module.new
+      # dynamically create nested modules
+      module_names = module_name.split("::").inject([]) { |n, c| n << (n.empty? ? [c] : [n.last] + [c]).flatten }
+      module_names.each do |module_name_array|
+        eval "module ::#{module_name_array.join("::")}; end"
       end
     end
 
--- a/lib/active_admin/namespace.rb
+++ b/lib/active_admin/namespace.rb
@@ -253,7 +253,7 @@ module ActiveAdmin
 
     class Store
       include Enumerable
-      delegate :[], :[]=, :empty?, to: :@namespaces
+      delegate :[]=, :empty?, to: :@namespaces
 
       def initialize
         @namespaces = {}
@@ -263,9 +263,19 @@ module ActiveAdmin
         @namespaces.values.each(&block)
       end
 
+      def [](key)
+        @namespaces[Array(key)]
+      end
+
       def names
+        Deprecation.warn "names replaced by name_paths now that namespaces can be nested."
+        @namespaces.keys.first
+      end
+
+      def name_paths
         @namespaces.keys
       end
+
     end
   end
 end
--- a/lib/active_admin/router.rb
+++ b/lib/active_admin/router.rb
@@ -19,7 +19,7 @@ module ActiveAdmin
         if namespace.root?
           router.root namespace.root_to_options.merge(to: namespace.root_to)
         else
-          router.namespace namespace.name, namespace.route_options.dup do
+          define_nested_namespace(namespace.name_path, namespace) do
             router.root namespace.root_to_options.merge(to: namespace.root_to, as: :root)
           end
         end
@@ -105,8 +105,18 @@ module ActiveAdmin
     end
 
     def define_namespace(config)
-      router.namespace config.namespace.name, config.namespace.route_options.dup do
+      define_nested_namespace(config.namespace.name_path, config.namespace) do
         define_routes(config)
+      end
+    end
+
+    def define_nested_namespace(names, namespace)
+      router.namespace names.first, namespace.route_options.dup do
+        if names.size > 1
+          define_nested_namespace(names.drop(1), namespace, &Proc.new)
+        else
+          yield
+        end
       end
     end
   end
--- a/lib/active_admin/scope.rb
+++ b/lib/active_admin/scope.rb
@@ -31,7 +31,7 @@ module ActiveAdmin
         raise "A string/symbol is required as the second argument if your label is a proc." unless method
         @id = ActiveAdmin::Dependency.rails.parameterize method.to_s
       else
-        @scope_method ||= name.to_sym
+        @scope_method ||= (name.is_a?(Array) ? name.join('_').underscore : name).to_sym
         @id = ActiveAdmin::Dependency.rails.parameterize name.to_s
       end
 
--- a/spec/requests/default_namespace_spec.rb
+++ b/spec/requests/default_namespace_spec.rb
@@ -51,7 +51,7 @@ RSpec.describe ActiveAdmin::Application, type: :request do
 
     it "should generate resource paths" do
       expect(resource1.route_collection_path).to eq "/test/categories"
-      expect(resource2.route_collection_path).to eq "/test/foo/bar/categories"
+      expect(resource2.route_collection_path).to eq "/foo/bar/categories"
     end
 
     it "should generate a log out path" do
@@ -72,7 +72,7 @@ RSpec.describe ActiveAdmin::Application, type: :request do
 
     it "should generate resource paths" do
       expect(resource1.route_collection_path).to eq "/abc_123/categories"
-      expect(resource2.route_collection_path).to eq "/abc_123/foo/bar/categories"
+      expect(resource2.route_collection_path).to eq "/foo/bar/categories"
     end
 
     it "should generate a log out path" do
--- a/spec/requests/default_namespace_spec.rb
+++ b/spec/requests/default_namespace_spec.rb
@@ -4,7 +4,19 @@ RSpec.describe ActiveAdmin::Application, type: :request do
 
   include Rails.application.routes.url_helpers
 
-  let(:resource) { ActiveAdmin.register Category }
+  before do
+    load_defaults!
+    ActiveAdmin.register(Category, namespace: [:foo, :bar])
+    reload_routes!
+  end
+
+  after :all do
+    load_defaults!
+    reload_routes!
+  end
+
+  let(:resource1) { ActiveAdmin.register Category }
+  let(:resource2) { ActiveAdmin.register Category, namespace: [:foo, :bar] }
 
   [false, nil].each do |value|
 
@@ -15,7 +27,8 @@ RSpec.describe ActiveAdmin::Application, type: :request do
       end
 
       it "should generate resource paths" do
-        expect(resource.route_collection_path).to eq "/categories"
+        expect(resource1.route_collection_path).to eq "/categories"
+        expect(resource2.route_collection_path).to eq "/foo/bar/categories"
       end
 
       it "should generate a log out path" do
@@ -37,7 +50,8 @@ RSpec.describe ActiveAdmin::Application, type: :request do
     end
 
     it "should generate resource paths" do
-      expect(resource.route_collection_path).to eq "/test/categories"
+      expect(resource1.route_collection_path).to eq "/test/categories"
+      expect(resource2.route_collection_path).to eq "/test/foo/bar/categories"
     end
 
     it "should generate a log out path" do
@@ -57,7 +71,8 @@ RSpec.describe ActiveAdmin::Application, type: :request do
     end
 
     it "should generate resource paths" do
-      expect(resource.route_collection_path).to eq "/abc_123/categories"
+      expect(resource1.route_collection_path).to eq "/abc_123/categories"
+      expect(resource2.route_collection_path).to eq "/abc_123/foo/bar/categories"
     end
 
     it "should generate a log out path" do
--- a/spec/unit/application_spec.rb
+++ b/spec/unit/application_spec.rb
@@ -145,7 +145,7 @@ RSpec.describe ActiveAdmin::Application do
         ActiveSupport::Deprecation.silence do
           expect(ns.name).to eq :new_namespace
         end
-        expect(ns.name_path).to eq [:admin, :new_namespace]
+        expect(ns.name_path).to eq [:new_namespace]
       end
     end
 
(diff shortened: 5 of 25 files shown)


lib/active_admin/namespace.rb
```diff
@@ -100,11 +109,25 @@ def root?
     #   Namespace.new(:root).module_name # => nil
     #
     def module_name
-      root? ? nil : @name.camelize
+      _name = @name
+      unless root?
+        if _name.is_a?(Array)
+          _name.map(&:to_s).map(&:camelize).join('::')
```
Fivell: @siutin  what about one liner like 
```
Array.wrap(@name).map(&:to_s).map(&:camelize).join('::')
```

lib/active_admin/namespace.rb
```diff
@@ -100,11 +109,25 @@ def root?
...
+        end
+      end
     end
 
     def route_prefix
-      root? ? nil : @name
+      _name = @name
+      unless root?
+        if _name.is_a?(Array)
+          _name.map(&:to_s).map(&:to_s).join('_').underscore
```
Fivell: @siutin  not sure double map needed here , same propose  with one liner

lib/active_admin/namespace.rb
siutin: Sure, I didn't know this before

lib/active_admin/namespace.rb
siutin: You are right, wrong copy-paste

siutin:
@Fivell Not sure how to simplify that for now, anyway I can still add some tests to the current approach.

lib/active_admin/router.rb
```diff
@@ -19,9 +19,17 @@ def define_root_routes
...
           router.root namespace.root_to_options.merge(to: namespace.root_to)
         else
-          router.namespace namespace.name, namespace.route_options.dup do
-            router.root namespace.root_to_options.merge(to: namespace.root_to, as: :root)
+          if namespace.name.is_a?(Array)
+            proc = Proc.new do
+              router.root namespace.root_to_options.merge(to: namespace.root_to, as: :root)
+            end
+            (namespace.name.reverse.inject(proc) { |n, c| Proc.new { router.namespace c, namespace.route_options.dup, &n } }).call
+          else
```
varyonic: Could this be simplified by extracting into a method that gets called recursively?

siutin: I pushed some changes based on what you said, and I do the same in define_namespace.

varyonic:
Is this expected behavior?
image

lib/active_admin/namespace.rb
```diff
@@ -226,8 +242,29 @@ def unload_resources!
 
     # Creates a ruby module to namespace all the classes in if required
     def register_module
-      unless Object.const_defined? module_name
-        Object.const_set module_name, Module.new
+      # dynamically create nested modules
+      module_names = module_name.split("::").inject([]) { |n, c| n << (n.present? ? [n.last] + [c] : [c]).flatten }
+      module_names.each_with_index do |module_name_array, i|
+        *prefix, parent, child = module_name_array
```
varyonic: My guess is this can be simplified by extracting a `parent_module` method that is called recursively.

siutin:
@varyonic Yes, that's intentional. Do you prefer like "/site3/demo"?

varyonic:
It's a minor issue, but yes.

spec/unit/routing_spec.rb
```diff
@@ -67,6 +67,32 @@
...
+      it "should route the new path" do
+        expect(new_admin_foo_bar_post_path).to eq "/admin/foo/bar/posts/new"
+      end
+
+      it "should route the edit path" do
+        expect(edit_admin_foo_bar_post_path(1)).to eq "/admin/foo/bar/posts/1/edit"
+      end
+    end
+
     context "when in root namespace" do
```
varyonic: What about routing of registered pages (not resources) for nested namespaces?

lib/active_admin/namespace.rb
siutin: Sure, I will think a bit on how to make it more general

spec/unit/routing_spec.rb
siutin: You're right, will add soon

lib/active_admin/namespace.rb
```diff
@@ -37,14 +37,23 @@ def setting(name, default)
...
     end
 
     def name
-      @name.to_sym
+      if @name.is_a?(Array)
+        @name
+      else
+        @name.to_sym
+      end
     end
```
varyonic: I propose treating this as a breaking change as follows:
```ruby
    def name
      Deprecation.warn "name replaced by namepath now that namespaces can be nested."
      namepath.first
    end

    # @param namepath [Array]
    def namepath=(namepath)
      namepath.shift if namepath.first == :root
      @namepath = namepath.map(&:to_s).map(&:underscore)
      default_namespace = application.default_namespace
      if default_namespace && default_namespace != :root
        @namepath.unshift default_namespace.to_s unless @namepath.first == default_namespace.to_s
      end
    end

    def namepath
      @namepath.map(&:to_sym)
    end
```
This approach leaves #name always returning the same symbol as before but warning of new functionality.  As #name is replaced by #namepath the latter always returns an array of symbols so no conditional logic is required to test the type of value returned.

siutin: Does that mean we will assign the namepath in initialize? And why do we need to convert the namepath into a underscored string?

varyonic: Yes, call namepath= from initialize.

siutin: Sure, I get the idea. And do you mind having a look of my recent changes? I can add the breaking changes on top.

lib/active_admin/router.rb
```diff
@@ -105,8 +111,25 @@ def define_belongs_to_routes(config)
...
+      if config.namespace.name.is_a?(Array)
+        define_nested_namespace(config.namespace.name, config.namespace) do
+          define_routes(config)
+        end
+      else
+        router.namespace config.namespace.name, config.namespace.route_options.dup do
+          define_routes(config)
+        end
+      end
+    end
```
varyonic: I think this simplifies to...
```ruby
    def define_namespace(config)
      names = Array(config.namespace.name)
      define_nested_namespace(names, config.namespace) do
        define_routes(config)
      end
    end
```

lib/active_admin/router.rb
```diff
@@ -105,8 +111,25 @@ def define_belongs_to_routes(config)
...
+
+    def define_nested_namespace(names, namespace)
+      curr, *rest = names
+      router.namespace curr, namespace.route_options.dup do
+        if rest.present?
+          define_nested_namespace(rest, namespace, &Proc.new)
+        else
+          yield
+        end
```
varyonic: This is fine, though I might tweak it like this...
```ruby
    def define_nested_namespace(names, namespace)
      router.namespace names.first, namespace.route_options.dup do
        if names.size > 1
          define_nested_namespace(names.drop(1), namespace, &Proc.new)
        else
          yield
        end
      end
    end
```

varyonic:
Resolves #615 (!)

siutin:
Added code for breaking changes recently. 
For now I prefer to use namespace's name (as original_name) instead of name_path for retrieving namespaces from the application.
Please take a look https://github.com/activeadmin/activeadmin/blob/42b73636a20d5af4d63d4e235c39e8e9a5a7bb52/lib/active_admin/application.rb#L77

spec/unit/application_spec.rb
```diff
@@ -140,17 +140,28 @@
 
   describe "#namespace" do
 
-    it "should yield a new namespace" do
-      application.namespace :new_namespace do |ns|
-        expect(ns.name).to eq :new_namespace
-      end
+    context "should yield a new namespace" do
+      before { ActiveSupport::Deprecation.behavior = :stderr }
```
siutin: I guess it should be fine to test breaking changes this way?

varyonic: Use `ActiveSupport::Deprecation.silence do ... end` in test.

siutin:
Using `namespace.name` as a key of `application.namespaces` is not enough to cover nested namespace support, so I come up 2 ideas below:
A) Use `name_path` as the key

``` ruby
namespaces[build_name_path(name)] ||= begin ...

namespaces[namespace.name]  # doesn't work
namespaces[build_name_path(namespace.name)] # works
namespaces[namespace.name_path] # works
```

B) Same as 1 but to run `build_name_path` inside `Namespace::Store` for every access.

```ruby
namespaces[name] ||= begin ...

namespaces[namespace.name]  # still works, so you can put deprecation warning here
namespaces[namespace.name_path] # works
```

I think A is good to go with but if the deprecation warning is needed I can go with B.

varyonic:
I agree (A) should be OK.

siutin:
@varyonic feel free to review overall changes.

siutin:
I find out I missed a commit https://github.com/activeadmin/activeadmin/pull/5185/commits/30454a68337f28d573158640cde1212e62c23794 when rebasing. Those squashed commits look good to me, I guess you could squash https://github.com/activeadmin/activeadmin/pull/5185/commits/30454a68337f28d573158640cde1212e62c23794 with https://github.com/varyonic/activeadmin/commit/1f3ae488191e4d024430e854664298c5d1192377.

varyonic:
If you have copied my branch into your repo and the squashed commits look good then rename your branch to `backup` and mine to `615-nested-namespaces`. Cherry pick your missed commit on top then use interactive rebase to merge it with `1f3ae48`. Also, the Travis build for you PR is now failing since @deivid-rodriguez re-enabled JRuby.

lib/active_admin/namespace.rb
```diff
@@ -33,18 +33,20 @@ def setting(name, default)
...
     RegisterEvent = 'active_admin.namespace.register'.freeze
 
-    attr_reader :application, :resources, :menus
+    attr_reader :application, :resources, :menus, :name_path
 
     def initialize(application, name)
       @application = application
-      @name = name.to_s.underscore
+      @name = Array(name).first.to_s.underscore.to_sym
+      @name_path = ActiveAdmin.application.build_name_path(name)
```
varyonic: @name_path = application.build_name_path(name) should work?

lib/active_admin/scope.rb
```diff
@@ -31,7 +31,7 @@ def initialize(name, method = nil, options = {}, &block)
         raise "A string/symbol is required as the second argument if your label is a proc." unless method
         @id = ActiveAdmin::Dependency.rails.parameterize method.to_s
       else
-        @scope_method ||= name.to_sym
+        @scope_method ||= (name.is_a?(Array) ? name.join('_').underscore : name).to_sym
```
varyonic: Should be using name_path here?

spec/requests/default_namespace_spec.rb
```diff
@@ -4,7 +4,19 @@
 
   include Rails.application.routes.url_helpers
 
-  let(:resource) { ActiveAdmin.register Category }
+  before do
+    load_defaults!
+    ActiveAdmin.register(Category, namespace: [:foo, :bar])
+    reload_routes!
+  end
```
varyonic: I'm not familiar with this, can you explain a little more?

spec/unit/application_spec.rb
```diff
@@ -142,19 +142,28 @@
 
     it "should yield a new namespace" do
       application.namespace :new_namespace do |ns|
-        expect(ns.name).to eq :new_namespace
+        ActiveSupport::Deprecation.silence do
+          expect(ns.name).to eq :new_namespace
+          expect(ns.name_path).to eq [:admin, :new_namespace]
```
varyonic: Minor nitpick, but `expect(ns.name_path).to eq [:admin, :new_namespace]` can be outside `Deprecation.silence` block.

spec/unit/application_spec.rb
```diff
@@ -142,19 +142,28 @@
...
+        end
       end
     end
 
     it "should return an instantiated namespace" do
       admin = application.namespace :admin
-      expect(admin).to eq application.namespaces[:admin]
+      expect(admin).to eq application.namespaces[[:admin]]
```
varyonic: It would be nice to have
```
class Namespace::Store
  def [](key)
    @namespaces[Array(key)]
  end
end
```
Then all three cases in https://github.com/activeadmin/activeadmin/pull/5185#issuecomment-343724223 (A) would work.

spec/unit/routing_spec.rb
```diff
@@ -15,7 +15,7 @@
   let(:namespaces) { ActiveAdmin.application.namespaces }
 
   it "should only have the namespaces necessary for route testing" do
-    expect(namespaces.names).to eq [:admin]
+    expect(namespaces.names).to eq [[:admin]]
```
varyonic: If Namespace::Store#names now returns a set of arrays instead of symbols I suppose it should be deprecated and replaced by Namespace::Store#name_paths

spec/unit/page_spec.rb
```diff
@@ -61,7 +61,9 @@ def config(options = {})
 
     describe "#namespace_name" do
       it "returns the name of the namespace" do
-        expect(config.namespace_name).to eq "admin"
+        ActiveSupport::Deprecation.silence do
+          expect(config.namespace_name).to eq "admin"
```
varyonic: This suggests something needs to be fixed in page.rb.  You still need to do pages for nested namespaces, right?

spec/unit/namespace/register_resource_spec.rb
```diff
@@ -148,14 +148,14 @@ module ::Mock; class Resource; def self.has_many(arg1, arg2); end; end; end
       it "should be namespaced" do
         namespace = ActiveAdmin::Namespace.new(application, :one)
         namespace.register Category
-        expect(defined?(One::CategoriesController)).to eq 'constant'
+        expect(defined?(Admin::One::CategoriesController)).to eq 'constant'
```
varyonic: I'm not completely clear why this changed?

varyonic:
Great progress!  Let me know when ready for another look.

wspurgin:
Great work so far. I have a general question about the DSL for registering a resource/page for a nested namespace. Right now the syntax uses an array:

```ruby
ActiveAdmin.register Post, namespace: [:admin, :blog] do
  # ...
end
```

and I know the implication is `[:admin, :blog] => 'admin/blog/'`, but the Array-style syntax feels wrong. Any time I see arrays that *aren't* explicitly joined, I treat it as being a set of independent "things". In this case it reads (to me), "Register 'Post' in the namespace*s* `:admin` and `blog`".

It's a pretty small, and it's most definitely a preference thing, but I think it might be worthy of at least a brief discussion. For example, I like that in the initializer, we used nested calls to `namespace` to register a nested namespace rather than an array-style syntax.

```ruby
namespace :admin do
  namespace :blog do
    # ...
  end
end
```

That to me is super clear. Whereas something like `namespace [:admin, :blog]` would not be. To me that reads "Configure the namespaces `:admin` *and* `:blog` like this:". Though, again, that's just my reading of it.

To make the registration process more clear, I have a few ideas, but they will ultimately add some typing (which IMHO isn't so bad if it helps readability). Writing something like this is what I'm thinking:

```ruby
ActiveAdmin.register Post, namespace: AcitveAdmin::Namespace.new([:admin, :blog]) do
end

# OR an explicit namespace

ActiveAdmin.register Post, namespace: 'admin/blog' do
end

# Maybe even a helper DSL function to reduce typing

ActiveAdmin.register Post, namespace: build_namespace(:admin, :blog, :foo) do
end

# That way passing an Array of namespaces will actually register the
# resource under multiple namespaces (if sharing functionality).

ActiveAdmin.register Post, namespace: [build_namespace(:admin, :blog), :admin, 'admin/demo'] do
end
```

These are just thoughts, not requirements, but I'd definitely like to hear other thoughts around it.

varyonic:
`namespace: 'admin/blog'` is a nice syntax, but regardless the internal representation should change from `name` (single symbol) to `name_path` (ordered set of symbols) and awareness of that change should propagate through the codebase by deprecating and replacing the attribute accessor.

siutin:
Thanks for reviewing and giving suggestions. 
I agree “namespace: ‘admin/blog’ looks clear. The reason I choose the Array way is that I feel more convenient to archieve the functionality, and it’s not a decision I made from the perspective of the DSL desig, so I think it’s make sense to further discuss this I understand the array-style syntax may not be a good solution and it can be confusing.

varyonic:
Might be as simple as
```ruby
@name_path = application.build_name_path(name.split('/'))
```
in Namespace#initialize.

spec/requests/default_namespace_spec.rb
siutin: This is registering the page in nested namespace before the test run, which is required for the test below.

siutin:
I changed the behavior of the name_path that make more sense. When the default namespace is :admin and the resource namespace is [:foo, :bar], we don't register it to /admin/foo/bar, but /foo/bar instead.

varyonic:
Sorry I did not get back to you on this, I see you have now published https://rubygems.org/gems/activeadmin-nested-namespaces

westonganger:
+1

PR #5186, opened 2017-09-26
Add jquery-ui dependency removal to changelog

joshmfrankel:
I recently ran into an issue upgrading ActiveAdmin where jquery-ui was required within active_admin.js but after 1.1.0 was packaged in the vendor folder. This PR adds documentation regarding the change in configuration. I placed it into a _Breaking Changes_ section but maybe it would be better to have this elsewhere. Open to feedback.

See: https://github.com/activeadmin/activeadmin/issues/5012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,10 @@
 
 * Ruby 2.1 support has been dropped [#5002][] by [@deivid-rodriguez][]
 * Replaced `sass-rails` with `sass` dependency [#5037][] by [@javierjulio][]
-* Removed `jquery-ui-rails` as a dependency. All jQuery UI assets are now
-packaged within the vendor directory. Please remove any direct inclusions of`//= require jquery-ui`. [#5052][] by [@javierjulio][]
+* Removed `jquery-ui-rails` as a dependency. The specific jQuery UI assets used
+  are now within the vendor directory. This will be replaced by alternatives and
+  dropped entirely in a major release. Please remove any direct inclusions
+  of `//= require jquery-ui`. [#5052][] by [@javierjulio][]
 
 ### Deprecations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 ## 1.1.0 [☰](https://github.com/activeadmin/activeadmin/compare/v1.0.0...v1.1.0)
 
-### Breaking Changes
-
-* jquery-ui-rails is no longer a gem dependency and all assets are now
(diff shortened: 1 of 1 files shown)


javierjulio:
@joshmfrankel hmm interesting, this shouldn't be a breaking change because you shouldn't be requiring those files in the active_admin.js file since they are part of base.js. Why do you require the internal files ActiveAdmin requires again? Is that documented somewhere or its how you installed AA?

joshmfrankel:
> Is that documented somewhere or its how you installed AA?

Hmm, yeah you are correct. I don't see it documented within the AA docs. We were running a much older version of AA "1.0.0.pre2". Looking back through our commit history, I had assumed this was part of the generator for the js file but it sounds like that might not be the case. 

Maybe instead of a breaking changes section we could add the above somewhere else (with some language modification). Thoughts?

javierjulio:
Yes I had come to the same conclusion as I checked the installation docs and also an app's history and see no addition for that in the file.

I'm open to changing the language to suggest that if you did require it directly then you'd have to remove it to avoid issues but its not a breaking change since its an internal dependency that shouldn't be referenced.

joshmfrankel:
@javierjulio Good call with the above suggestion. I moved the new note into the Removals section.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
 
 * Ruby 2.1 support has been dropped [#5002][] by [@deivid-rodriguez][]
 * Replaced `sass-rails` with `sass` dependency [#5037][] by [@javierjulio][]
+* Removed `jquery-ui-rails` as a dependency. All jQuery UI assets are now
+packaged within the vendor directory. Please remove any direct inclusions of`//= require jquery-ui`. [#5052][] by [@javierjulio][]
```
javierjulio: Just a small change to the second sentence with an additional sentence before your removal instruction:

> The specific jQuery UI assets used are now within the vendor directory. This will be replaced by alternatives and dropped entirely in a major release.

joshmfrankel: You got it :+1:

joshmfrankel:
@javierjulio I just noticed that the link to PR #5052 didn't generate properly in the changelog. I've made an update which directly links the text "#5052" to the proper PR. Should I go ahead and re-open this PR or start a new one?

javierjulio:
@joshmfrankel ah yes because it needs to be linked at the bottom of the file as you probably noticed. Yes please open a new PR and I'll merge it in. Thanks! ❤️

joshmfrankel:
> Yes please open a new PR and I'll merge it in.

Done and done. Thanks for your patience :-)

Issue #5182, opened 2017-09-19
Doubly nested input disappears when upgrading to activeadmin 1.0


alexaltair:
I have a form that, when maximally simplified, looks like this.
```ruby
ActiveAdmin.register Foo do
  form do |f1|
    f1.inputs 'DeepFoos' do
      f1.has_many :deep_foos do |f2|
        f2.input :name

        f2.inputs "DeeperFoos for #{f2.object.name}" do
          f2.has_many :deeper_foos do |f3|
            f3.input :color
          end
        end
      end
    end
  end
end
```
When I was using activeadmin 0.6.6, this rendered fine, with the double-nested resource. After updating to 1.0.0.pre5, the `DeeperFoos` input part of the form disappeared. There's no exception, and it's not in the html source code of the page. It's simply gone.

Weirdly, if I remove the line that says `f2.input :name`, the `DeeperFoos` input rendered again.

I can't find anything relevant in the [activeadmin upgrade docs](https://github.com/activeadmin/activeadmin/wiki/Upgrading-ActiveAdmin) or the [formtastic upgrade docs](https://github.com/justinfrench/formtastic/wiki/Upgrading-to-Formtastic-3.1) (which, as a dependency, got upgraded from 2.2.1 to 3.1.5).

javierjulio:
@alexaltair I'm not sure `has_many` can be nested like that. @varyonic would you have any idea if that is even or was supported? 😳

alexaltair:
@javierjulio We used it for several months on `activeadmin 0.6.6`, so it did used to work.

varyonic:
The simplest case is tested, see https://github.com/activeadmin/activeadmin/blob/0c342275cd90455ddf6b2787e7a3c3e00bfdaa69/spec/unit/form_builder_spec.rb#L898
However, other scenarios may have been broken by the rewrite.

alexaltair:
Agreed with @varyonic. The difference that I see between my example and the test one is the line `f2.input :name`, that is, the existence of any other input fields on the middle model. If I take that line out, it does render the deepest module.

xn:
Probably related. This only renders the `:value` input in the `has_many` At this point, I think the problem are the compounder refactors going on interfering with other AA plugins. It's a mess.
```form do |f|
  f.semantic_errors *f.object.errors.keys
  f.inputs "Foo" do
    f.input :a_boolean
    f.inputs do
      h3 "w00t"
[...]

PR #5190, opened 2017-09-29
Fix regression extracting HasManyBuilder

varyonic:
Extends a test in form_builder_spec to try to reproduce #5182  The test passes so more investigation required.

### DIFF
--- a/lib/active_admin/form_builder.rb
+++ b/lib/active_admin/form_builder.rb
@@ -58,7 +58,9 @@ module ActiveAdmin
       html = "".html_safe
       html << template.content_tag(:h3) { heading } if heading.present?
       html << template.capture { content_has_many(&block) }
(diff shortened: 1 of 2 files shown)


varyonic:
@alexaltair I updated the spec in bc4f4b5cab00523f7559aa91bd87fdea1223c536 and it still passes.  Is your form in a Rails partial template perhaps?

alexaltair:
@varyonic Huh. No, it's not in a partial. That test does look like the situation that fails for me.

varyonic:
Feel free to branch and tweak the test to see if you can make it fail.  I've temporarily included a commit that prints the generated HTML.

varyonic:
By great coincidence the test I added failed to reproduce the original reported issue but it DID detect a regression in #5196.

PR #5196, opened 2017-10-08
Refactor has many, extracting HasManyBuilder

varyonic:
Extracts has_many related attributes and methods into a form_builder decorator that can be nested to handle nested has_many blocks.  This is just a refactoring intended to make the implementation easier to understand and extend.

### DIFF
--- a/lib/active_admin/form_builder.rb
+++ b/lib/active_admin/form_builder.rb
@@ -38,19 +38,16 @@ module ActiveAdmin
   # nested decorators.
   class HasManyBuilder < SimpleDelegator
     attr_reader :has_many_form, :assoc
-    attr_reader :builder_options, :options
+    attr_reader :options
     attr_reader :heading, :sortable_column, :sortable_start
+    attr_reader :new_record, :destroy_option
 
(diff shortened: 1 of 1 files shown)


lib/active_admin/form_builder.rb
```diff
@@ -28,72 +28,93 @@ def cancel_link(url = {action: "index"}, html_options = {}, li_attrs = {})
...
+  class HasManyBuilder < SimpleDelegator
+    attr_reader :assoc
+    attr_reader :options
+    attr_reader :heading, :sortable_column, :sortable_start
+    attr_reader :new_record, :destroy_option
+
+    def initialize(has_many_form, assoc, options)
+      super has_many_form
+      @assoc = assoc
+      @options = extract_custom_settings!(options.dup)
```
zorab47: Would `extract_and_remove_custom_settings!` be clearer? Unless there is an established approach already.

varyonic: The name `extract_custom_settings!` is inspired by ActiveSupport `extract_options!` and `extract!`.  The ! is intended to signify that the argument is being mutated.

Issue #1123, opened 2012-03-11
Use group clause in scope cause error


janckerchen:
when I add group clause to scope like this

(I change the example code to make it clear after  latortuga commented)
# get count for each category.

class MyClass <  ActiveRecord::Base
  scope :category,  select("count(id) as count, category").group(:category)
end

ActiveAdmin.register MyClass do
   scope :category, :default => true
end

I get this error on rails page:

 Showing /usr/local/lib/ruby/gems/1.9.1/gems/activeadmin-0.4.3/app/views/active_admin/resource/index.html.arb where line #1 raised:
 ActiveSupport::OrderedHash can't be coerced into Fixnum

I notice that,  in rails console,  count attribute of scope with group is not Fixnum but Hash.  

I am not sure the issue belongs to rails or activeadmin,  somebody help me find the way~~

PR #5200, opened 2017-10-11
Fix bug when using a group by on a collection

jpbalarini:
I am using a `scoped_collection` on my controller that does a `GROUP BY`. I was getting this on my pagination control:
screen shot 2017-10-11 at 11 11 17 am
I found that the issue was on the `page_entries_info` method, since a `count` on a query that had a `group by` returns a Hash of elements with their respective count.

I also found that disabling pagination total on my index page with `pagination_total: false` wasn't working:
screen shot 2017-10-11 at 11 05 35 am
because of the same issue (count returns a Hash); the `+` method tries to sum an Integer with a Hash and that was fixed for the `build_pagination` method.

The 2 changes introduced in this PR fix those 2 problems.
Thanks!

### DIFF
--- a/lib/active_admin/views/components/paginated_collection.rb
+++ b/lib/active_admin/views/components/paginated_collection.rb
@@ -103,6 +103,8 @@ module ActiveAdmin
           # if there is another page or not, but the limit/offset make this
           # query fast.
           offset = collection.offset(collection.current_page * @per_page.to_i).limit(1).count
+          # if we have a collection with a group_by the count method returns a Hash
+          offset = offset.size if offset.is_a?(Hash) || offset.is_a?(ActiveSupport::OrderedHash)
           options[:total_pages] = collection.current_page + offset
           options[:right] = 0
         end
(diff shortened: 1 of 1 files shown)


jpbalarini:
Added some images explaining the issue I found.

varyonic:
Any new spec for that?

deivid-rodriguez:
@jpbalarini Could you add specs for this?

zorab47:
Can you share more on your use case? I've built similar reports but it requires the use of views. I'm curious if it should be solved differently.

Fivell:
@varyonic  seems duplicate of https://github.com/activeadmin/activeadmin/issues/1123

@jpbalarini see https://github.com/activeadmin/activeadmin/issues/1123#issuecomment-4473559

corsonknowles:
It still bugs me that this hasn't been fixed yet

irenehakes-handy:
Running into the `pagination_total: false` issue as well. Is there a fix in the works or another issue that could point me in the right direction?

javierjulio:
We've made some changes to pagination since then. I'm not sure if this is still an issue for most. In my experience, the issue here is with the query using a `GROUP BY` and not necessarily with ActiveAdmin. You would need to perform the grouping in a sub query or some other approach.

PR #5203, opened 2017-10-12
783: Update ActiveAdmin::DatabaseHitDuringLoad error message to offer…

taranda:
… more solutions.

### DIFF
--- a/lib/active_admin/error.rb
+++ b/lib/active_admin/error.rb
@@ -33,7 +33,14 @@ module ActiveAdmin
       super "Your file, #{file} (line #{line}), caused a database error while Active Admin was loading. This " +
             "is most common when your database is missing or doesn't have the latest migrations applied. To " +
             "prevent this error, move the code to a place where it will only be run when a page is rendered. " +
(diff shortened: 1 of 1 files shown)


taranda:
This PR will make the DatabaseHitDuringLoad error  message offer some more solutions which I think will prove useful for many users.

taranda:
When I first encountered this error and the recommendation was to add wrap the query in a proc, I did not immediately know what to do with that.  It took quite a bit of Googling and reviewing issues to finally figure out the solution.  I thought it would be nice to provide a more explicit solution to other developers who encounter this error.

varyonic:
Aha. Improvements to the documentation on how to use procs for filters, an example in demo.activeadmin.info and/or a wiki entry or link to a blog post would all be very welcome.  Maybe even a link to such docs in the error message.

deivid-rodriguez:
I'm not sure this is a good thing to do, so I'd rather not recommend this. Totally :+1: on improving the docs on how to defer query executions by wrapping them in a proc, though. Feel free to submit another PR for that @taranda! I'm going to close this PR for now, but feel free to continue the discussion!

PR #5206, opened 2017-10-16
Extract Sidebar View Component

varyonic:
Make it easier to override and customize all of the sidebar, not just the sections.


zorab47:
@varyonic what approaches do you take when overriding/extending components? It is primarily monkey-patching? Due to their construction it is difficult to re-associate the builder method (`sidebar` in this case) with a different component.

varyonic:
@zorab47 I will have another PR that shows my approach shortly.

PR #5215, opened 2017-10-20
Cache-able select filters

wspurgin:
I use ActiveAdmin often, and in some instances I'll have a select filter for an association that has an expensive collection to fetch for every request. Or I'll have many association filters such that grabbing all their collections in aggregate adds too much time to the request.

I created this caching feature for select filters for those situations in my own work. Since I love this framework, I thought it be good to offer it back and maybe get some improvements out of it! The premise is simple, if the select options tells us we should cache the filter, we'll store it in the `Rails.cache` (if caching is enabled). Like thus:

```ruby
  filter :author, cache: true, cache_key: -> { Author.cache_key }
# ...
```

I'm completely open to changes, particular around the options that should be used to enable caching for the filter. For example, should we only look for `:cache_key`? That way caching is only enabled when a `cache_key` is given (thereby encouraging not to have stale caches). Any thoughts or opinions would be most welcome.  Cheers :beers:

### DIFF
--- a/lib/active_admin/inputs/filters/base.rb
+++ b/lib/active_admin/inputs/filters/base.rb
@@ -38,6 +38,15 @@ module ActiveAdmin
           end
         end
 
+        def cache_filter(cache_key, &block)
+          full_key = "filter/#{cache_key}"
+          if Rails.configuration.action_controller.perform_caching
+            Rails.cache.fetch(full_key, &block)
+          else
+            yield
+          end
+        end
+
       end
     end
   end
--- a/lib/active_admin/inputs/filters/select_input.rb
+++ b/lib/active_admin/inputs/filters/select_input.rb
@@ -37,15 +37,27 @@ module ActiveAdmin
 
         # Provides an efficient default lookup query if the attribute is a DB column.
         def collection
-          if !options[:collection] && column
-            pluck_column
-          else
-            super
+          cache_on_demand do
+            if !options[:collection] && column
+              pluck_column
+            else
+              super
+            end
           end
         end
 
         def pluck_column
           klass.reorder("#{method} asc").distinct.pluck method
+        end
+
+        def cache_on_demand(&block)
+          if options[:cache]
+            cache_name = "#{klass.name}/#{method}"
+            cache_name = "#{cache_name}/#{options[:cache_key].call}" if options[:cache_key]
+            cache_filter(cache_name, &block)
+          else
+            yield
+          end
         end
 
       end
--- a/spec/unit/filters/resource_spec.rb
+++ b/spec/unit/filters/resource_spec.rb
@@ -125,4 +125,64 @@ RSpec.describe ActiveAdmin::Filters::ResourceExtension do
     expect(resource.sidebar_sections.first.name).to eq "filters"
   end
 
+  describe "cacheable filters" do
+    let(:view) { action_view }
+
+    context "with caching enabled" do
+      it "should cache the collection" do
+        action_controller_double = double(perform_caching: true)
+        configuration_double     = double(action_controller: action_controller_double)
+        cache_double             = double(fetch: "")
+
+        allow(Rails).to receive(:cache).and_return(cache_double)
+        allow(Rails).to receive(:configuration).and_return(configuration_double)
+        allow(view).to  receive(:collection_path).and_return("admin/posts")
+
+        resource.add_filter :author, as: :select, cache: true
+
+        # This is a hacky kind of way to force the select filters being created
+        view.active_admin_filters_form_for(Post.search, resource.filters)
+
+        expect(action_controller_double).to have_received :perform_caching
+        expect(cache_double).to have_received :fetch
+      end
+
+      it "should use a specified cache_key as part of full cache path" do
+        action_controller_double = double(perform_caching: true)
+        configuration_double     = double(action_controller: action_controller_double)
+        cache_double             = double(fetch: "")
+
+        allow(Rails).to receive(:cache).and_return(cache_double)
+        allow(Rails).to receive(:configuration).and_return(configuration_double)
+        allow(view).to  receive(:collection_path).and_return("admin/posts")
+
+        resource.add_filter :author, as: :select, cache: true, cache_key: -> { "foo-123" }
+
+        view.active_admin_filters_form_for(Post.search, resource.filters)
+
+        expect(action_controller_double).to have_received :perform_caching
+        expect(cache_double).to have_received(:fetch).with("filter/Post/author/foo-123")
+      end
+    end
+
+    context "with caching disabled" do
+      it "should not cache the collection" do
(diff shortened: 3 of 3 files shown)


varyonic:
I'm not sure how one tests this but it appears to need more coverage.  It's an interesting idea, awaiting more feedback from others.

wspurgin:
@varyonic The most I could think of was testing the structure and flow. Like "if the `:cache` is not set, then the filter is not cached" or "if the filter is specified to be cached, but caching is not enabled in the app, then the filter is not cached". Stuff like that. Unfortunately, setting up a spec to do that wouldn't be very "unit" testable. It requires setting up the namespace, registering a model, adding the filter, then actually executing a request (or pseudo-request) against that controller because the filter's build block doesn't get executed to render time. :thinking: Seems like a lot of overhead for such a small thing, and to be completely honest, I'd have no idea where such a spec should live in the current test suite..... The `resource_spec` was where I'd originally thought of putting it, but none of those specs really seem to test the creation side of the filters (just the DSL setup)... Honestly I'm a bit lost on how to go about setting up a reasonable spec... The most I've gotten is this

```ruby
  describe "cacheable filters" do
    let(:view) { action_view }

    context "a select filter with caching enabled" do
      it "should cache the collection" do
        action_controller_double = double(perform_caching: true)
        configuration_double     = double(action_controller: action_controller_double)

        allow(Rails).to receive(:cache).and_return(double(:fetch))
        allow(Rails).to receive(:configuration).and_return(configuration_double)

        resource.add_filter :author, as: :select, cache: true

        # This is a hacky kind of way to force the select filters being created
        view.active_admin_filters_form_for(Post.search, resource.filters)

        expect(action_controller_double).to have_received :perform_caching
        expect(Rails).to have_received(:cache)
      end
    end
  end
```
But it fails because for some reason [`MockActionView`](https://github.com/activeadmin/activeadmin/blob/master/spec/support/active_admin_integration_spec_helper.rb#L46-L59) doesn't have the URL helpers in this context? Don't see why it shouldn't.... Rails doesn't wait to define them.... right?

```
Failures:

  1) ActiveAdmin::Filters::ResourceExtension cacheable filters a select filter with caching enabled should cache the collection
     Failure/Error: url: collection_path,

     NameError:
       undefined local variable or method `collection_path' for #<ActiveAdminIntegrationSpecHelper::MockActionView:0x000000088120d8>
       Did you mean?  collection_select
     # ./lib/active_admin/filters/forms.rb:50:in `active_admin_filters_form_for'
     # ./spec/unit/filters/resource_spec.rb:142:in `block (4 levels) in <top (required)>'
```

wspurgin:
Fudged the `collection_path` for now (since it's not pertinent to the specs anyway). All feedback welcome!

wspurgin:
@Fivell I think that's a fair point. I would ask how "simple" AA's codebase truely is :wink:; though that sorta adds to the point doesn't it?

You *can* add the cache manually. Just as we can do many other things manually that are baked within AA now. This is a convenience feature certainly (if that is the argument against it). The reason why I have this patch to AA in all my work was to avoid having a bunch of clutter and duplication in my filters. I like this feature (but I'm biased of course), and I think it opens the door to encouraging more optimization work to the framework. However, I also recognize, that it is quite probable that most AA users don't require this level of control or convenience.  I think it really depends on how valuable other users of AA find this feature. If a lot of them find themselves plagued by slow page load times because of large collections on filters, I think this feature is valuable to that remedy that.

Perhaps a good solution for now is to let this MR sit for a while and get more feedback from maintainers, contributors, and users?

javierjulio:
Thanks for the contribution! I don't want to deter you but I'm not quite sure this is the right solution. The problem I see here is that the core issue at hand is still not addressed: large data queries. This PR still makes them happen but in turn bloats someone's cache which may be undesirable. Further since its loading so much data, that means its still creating all those objects and consuming memory. Are you using this feature in production? It would be great to see what the improvement was for you before and after.

As @fivell mentioned though its possible to do this without changes to AA which I think is ideal. We have to find the right balance to do something simple by default but allow a user to customize (typically done through a block) where need be.

I feel like the solution we should venture towards is adding some sort of autocomplete feature to handle the problem of not loading all associated data in a drop down. Like yourself, we've been frustrated with the long load times of an index page unless you specify explicit filters so that way we don't include the ones that are causing that issue. At work we've just started our first use of an autocomplete filter using `selectize-rails`. It might be better to put effort into something like this.

wspurgin:
Quite fair. I can spend some time gathering the before and after (we do use it in production), but I think you both have made good points. We don't cache any large objects. Usually arrays of ids and something searchable (like a name). We also use `selectize-rails` for our select filters. We use this caching to only create that arrays of `[ ... [searchable_text, id] ... ]` once (per a `cache_key`) over many requests. It's true that with `selectize.js` this can be done more asynchronously (with data urls pointing to the JSON that is searched), but other than writing custom json actions for every filter and some wonky JS, I can't see a cleaner way of accomplishing removing large slow quires from the initial request.

Like I said, great points. I'd be willing to put in some time discussing and implementing a workable method for creating autocomplete select filters. If caching the source request isn't part of that, I'm cool with it.

javierjulio:
@wspurgin ah I see. But then I can see others using this to cache full objects since its what the select filter loads by default. We can then say its an optional thing but at that point I think its worth not being in AA and just as long as we have the customization available to users to add that in to meet their specific use case if it works for them.

I really appreciate you taking the time to do this. While I mention the autocomplete suggestion, I know I'd love to see it, but my time is focused on a specific large part of AA (#3862) so while I'd love to see it I won't have the time. The other maintainers chip in where they can and what interests them and/or what they are best at. If this is something that interests you and you want to tackle it, I think we the AA team would love to collaborate on that and get it in. I don't know the amount of work involved but ideally I think we'd want someway to do autocomplete in AA, whether we do it by default, or make it easy to enable on case by case basis. 

I'd love to know more on how you using autocomplete with filters now. Any chance your AA source is online to see how you went about implementing it? Thanks again for taking the time! 🙏🏻❤️

wspurgin:
Thanks @varyonic @Fivell and @javierjulio for your discussion. I'll close this PR in favor of an eventual auto complete solution. Thanks for taking the time to look it over and share your thoughts.

@javierjulio Glad to have any part in such a cool and useful open source project! And great work on #3862, I've been looking forward to that for a long time. :+1:  Unfortunately, the full source of our AA implementation isn't available online for perusing. However, I'd be happy to discuss autocomplete implementation with AA (but perhaps via a different medium so I don't generate continual Github notifications)

javierjulio:
@wspurgin that's fine. If you have good commits just for the autocomplete changes, perhaps you could cherry pick them into your AA fork in a new branch. Would be a quick way to see what you did to get it in and have a discussion around it. If you have the time, give that a shot and open a new issue and we can talk about it. Thanks again!

Issue #4118, opened 2015-09-10
Exporting CSV runs out of memory.


MatUrbanski:
In my app when I export csv with around 100 thousand records heroku get out of memory. I noticed that when In my rails app I add to application.rb following line:

``` ruby
config.middleware.delete "ActiveRecord::QueryCache"
```

csv export takes much less ram than before and everything works fine. Is there any way to monkey patch csv export in active admin to work with disabled `ActiveRecord::QueryCache`?

adamdunkley:
@MatUrbanski We have the same problem here and I have verified your fix. What we find strange is that we don't use memory for our storage of query cache (we use `:mem_cache_store`). I also noted that when it is over the memory is not returned to the system so I think there is some sort of memory leak in `ActiveRecord::QueryCache` somewhere … perhaps it's something to do with how activeadmin paginates the orders or builds the CSV line? I ran a manual `GC.start` afterward and it cleared up the errant memory. 

I am going to try looking at the heap using https://github.com/Vasfed/heap_dump to see if I can figure out where the memory is being allocated from.

MatUrbanski:
I've fix that with overriding active_admin ```ActiveAdmin::CSVBuilder```
```
ActiveAdmin::CSVBuilder.class_eval do
  def build(controller, receiver)
    @collection = controller.send(:find_collection, except: :pagination)
    options = ActiveAdmin.application.csv_options.merge self.options
    columns = exec_columns controller.view_context

    if byte_order_mark = options.delete(:byte_order_mark)
      receiver << byte_order_mark
    end

    if options.delete(:column_names) { true }
      receiver << CSV.generate_line(columns.map{ |c| encode c.name, options }, options)
    end

    ActiveRecord::Base.uncached do
      @collection.find_each do |resource|
        resource = controller.send :apply_decorator, resource
        receiver << CSV.generate_line(build_row(resource, columns, options), options)
      end
    end
  end
end
```

seanlinsley:
@MatUrbanski interesting! I've never come across `uncached`, and it looks like exactly what we need.

yhourdel:
@MatUrbanski great fix, I've tested it inside our app in production and it works very well.
@seanlinsley I've forked the repo to be able to include this fix: should I make a pull request ?

joallard:
Whenever I try to download the CSV, I just get a 283-byte header with no records. I applied this patch in an initializer, no avail. CSV exporting in AA is my personal hell.

[...]

PR #5218, opened 2017-10-23
Do not cache queries that are used to construct CSV exports

fredngo:
Do not cache queries that are used to construct CSV exports, as it causes runaway memory consumption due to the ActiveRecord Query cache.

The majority of these one-off queries are issued in the `build_row` method.

This bug is documented here: https://github.com/activeadmin/activeadmin/issues/4118, where `ActiveRecord::Base.uncached` is used as a workaround.

### DIFF
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -71,8 +71,16 @@ module ActiveAdmin
     end
 
     def build_row(resource, columns, options)
-      columns.map do |column|
-        encode call_method_or_proc_on(resource, column.data), options
+      p = Proc.new do
+        columns.map do |column|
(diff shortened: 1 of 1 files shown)


varyonic:
Codecov coverage looks like maybe a line of dead code?

fredngo:
I can try to write a test that can target that particular [line of dead code](https://github.com/activeadmin/activeadmin/pull/5218/files#diff-da9014e0ac228dd5a8a7b0bd1dae16ecR83). I'll be back if I need some advice on how to do so, or when it's done...

varyonic:
@fredngo Does #5251 work for you?

fredngo:
I'll comment on #5251

Issue #5164, opened 2017-09-04
ActiveFilters ignores the primary_key setting of an association


Darhazer:
I have associations that are not based on the primary key of the table, but use some other key instead.
When filtering on such a field however, ActiveFilters correctly detects the associated class, but takes the value and uses it  as a primary key in a simple `find` call.

The association is set up like this:
`belongs_to :part, foreign_key: :sn, primary_key: :sn`

When filtered by sn, it correctly found that it should display a Part resource, but uses the `sn` value as `id` and either displays wrong Part or raises an exception, when there is no such id.
On a side note, there is no option to disable this box, in order at least not to raise errors

### Expected behavior

Find the part by `sn`

### Actual behavior

Tries to find part by `id`

wspurgin:
Related to #5223 because of the [`find`](https://github.com/activeadmin/activeadmin/blob/v1.1.0/lib/active_admin/filters/active_filter.rb#L22) in `ActiveFilter`. I think with some changes in that PR, this issue could also be resolved.

PR #5223, opened 2017-10-24
Fix Active Filter Labeling and Record Not Found 

wspurgin:
:wave: 

`ActiveFilter` was added to make the current filter sidebar a bit more humanized. It adds some cool things like smart, humanized filter names and, where possible, link to related resources with human readable names! Very cool. However, there's a few issues I saw when messing around a bit.

1. If a filter with a related class (as `ActiveFilter` defines it) is given conditions that have no matching records in the related class, `ActiveFilter` produces an `ActiveRecord::RecordNotFound` when trying to generate a set of values to display in the "Current Filters" sidebar.
2. If a filter is added to the resource that has a label specified, `ActiveFilter` ignores it in favor of pulling from `i18n`. This doesn't seem too terrible but can pose problems to the developer (see below)

The first issue I believe to be a regression. Even though, truthfully, there *is* no related record matching the filter criteria (and thus no resource records either), AA would display a blank slate in all other cases where a filter(s) produces no matching records.

The second issue, is more subtle. Most cases of custom labels are covered by the current implementation of `ActiveFilter`. After all, if you want the label of `Post#title` to be something other than "Title" just put it inside of `en.yml`, and `ActiveFilter` will do as you command. Here's the rub, imagine a more complex translation situation:

```yaml
# en.yml
en:
   activerecord:
      attributes:
         user:
            external_id: "%{client_name} External ID" 
#...
```
```ruby
ActiveAdmin.register User do
#...
filter :external_id, label: config.resource_class.human_attribute_name(:external_id, client_name: ENV["CLIENT"])
end
```

In this example, we have an interpolation within our translation because the translation changes based on who's bought our super shiny app. In this situation, `ActiveFilter` would produce a `I18n::MissingInterpolationArgument` when trying to render the Current Filter sidebar with this attribute.

While this example is a bit contrived, IMHO it's better to go with whatever the developer specifies as the attribute name. If the label on the filter is set, then that's takes precedence. If not, then the translations are certainly the next best thing.

Related issues:
- #5164

### DIFF
--- a/lib/active_admin/filters/active_filter.rb
+++ b/lib/active_admin/filters/active_filter.rb
@@ -19,7 +19,7 @@ module ActiveAdmin
       def values
         condition_values = condition.values.map(&:value)
         if related_class
-          related_class.where(related_class.primary_key => condition_values)
+          related_class.where(related_primary_key => condition_values)
         else
           condition_values
         end
@@ -89,16 +89,34 @@ module ActiveAdmin
       def find_class
         if condition_attribute.klass != resource_class && condition_attribute.klass.primary_key == name.to_s
           condition_attribute.klass
-        else
-          assoc = condition_attribute.klass.reflect_on_all_associations.
-            reject { |r| r.options[:polymorphic] }. #skip polymorphic
-            detect { |r| r.foreign_key.to_s == name.to_s }
-          assoc.class_name.constantize if assoc
+        elsif predicate_association
+          predicate_association.class_name.constantize
         end
       end
 
       def filter
         resource.filters[name.to_sym]
+      end
+
+      def related_primary_key
+        if predicate_association
+          predicate_association.active_record_primary_key
+        elsif related_class
+          related_class.primary_key
+        end
+      end
+
+      def predicate_association
+        @predicate_association = find_predicate_association unless defined?(@predicate_association)
+        @predicate_association
+      end
+
+      private
+
+      def find_predicate_association
+        condition_attribute.klass.reflect_on_all_associations.
+          reject { |r| r.options[:polymorphic] }. #skip polymorphic
+          detect { |r| r.foreign_key.to_s == name.to_s }
       end
     end
   end
--- a/lib/active_admin/filters/active_filter.rb
+++ b/lib/active_admin/filters/active_filter.rb
@@ -19,7 +19,7 @@ module ActiveAdmin
       def values
         condition_values = condition.values.map(&:value)
         if related_class
-          related_class.where(id: condition_values)
+          related_class.where(related_class.primary_key => condition_values)
         else
           condition_values
         end
@@ -28,11 +28,13 @@ module ActiveAdmin
       def label
         # TODO: to remind us to go back to the simpler str.downcase once we support ruby >= 2.4 only.
         translated_predicate = predicate_name.mb_chars.downcase.to_s
-        if related_class
-          "#{related_class.model_name.human} #{translated_predicate}".strip
+        if filter_label
+          "#{filter_label} #{translated_predicate}"
+        elsif related_class
+          "#{related_class_name} #{translated_predicate}"
         else
-          "#{attribute_name} #{translated_predicate}".strip
-        end
+          "#{attribute_name} #{translated_predicate}"
+        end.strip
       end
 
       def predicate_name
@@ -52,6 +54,18 @@ module ActiveAdmin
 
       def attribute_name
         resource_class.human_attribute_name(name)
+      end
+
+      def related_class_name
+        return unless related_class
+
+        related_class.model_name.human
+      end
+
+      def filter_label
+        return unless filter
+
+        filter[:label]
       end
 
       #@return Ransack::Nodes::Attribute
@@ -81,6 +95,10 @@ module ActiveAdmin
             detect { |r| r.foreign_key.to_s == name.to_s }
           assoc.class_name.constantize if assoc
         end
+      end
+
+      def filter
+        resource.filters[name.to_sym]
       end
     end
   end
--- a/lib/active_admin/filters/active_filter.rb
+++ b/lib/active_admin/filters/active_filter.rb
@@ -19,7 +19,7 @@ module ActiveAdmin
       def values
         condition_values = condition.values.map(&:value)
         if related_class
-          related_class.find(condition_values)
+          related_class.where(id: condition_values)
         else
           condition_values
         end
--- a/spec/unit/filters/active_filter_spec.rb
+++ b/spec/unit/filters/active_filter_spec.rb
@@ -132,4 +132,16 @@ RSpec.describe ActiveAdmin::Filters::ActiveFilter do
 
   end
 
+  context 'search has no matching records' do
+    let(:search) { Post.ransack(author_id_eq: "foo") }
+
+    it 'should not produce and error' do
+      expect { subject.values }.not_to raise_error
+    end
+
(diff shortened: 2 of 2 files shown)


wspurgin:
cc @Fivell : Since this is around the work you did to add `ActiveFilter`, I'd like any feedback you have around this change.

lib/active_admin/filters/active_filter.rb
```diff
@@ -19,7 +19,7 @@ def initialize(resource, condition)
       def values
         condition_values = condition.values.map(&:value)
         if related_class
-          related_class.find(condition_values)
+          related_class.where(id: condition_values)
```
zorab47: Using `id` here isn't ideal because the primary key column may be some other field on the record.

wspurgin: Yup. Great catch. :see_no_evil:

wspurgin:
@varyonic @zorab47 The [coverage failure for the last commit](https://codecov.io/gh/activeadmin/activeadmin/compare/95b8276d307aa86b7fbe27f870adfa24e5dbd378...c17be84e4e55866c9af7821e3db1257f510f3fea/diff#D2-183) looks like a false positive to me. But if the ✅ is desired, I can rejig that line so it is "covered" by codecov pov. Just let me know.

varyonic:
Right, CodeCov doesn't get it when you split a method chain across more than two lines.

wspurgin:
I amended the commit to put that method chain on one line then. 👍

wspurgin:
@zorab47 I think you'll need to re-approve these changes (since I added the association primary key support). Also let me know if a rebase is necessary.

lib/active_admin/filters/active_filter.rb
```diff
@@ -90,16 +90,27 @@ def find_class
         if condition_attribute.klass != resource_class && condition_attribute.klass.primary_key == name.to_s
           condition_attribute.klass
         else
-          assoc = condition_attribute.klass.reflect_on_all_associations.
-            reject { |r| r.options[:polymorphic] }. #skip polymorphic
-            detect { |r| r.foreign_key.to_s == name.to_s }
-          assoc.class_name.constantize if assoc
+          predicate_association.class_name.constantize if predicate_association
```
zorab47: Should this become an `elsif` for easier reading?

lib/active_admin/filters/active_filter.rb
```diff
@@ -90,16 +90,27 @@ def find_class
...
 
       def filter
         resource.filters[name.to_sym]
       end
+
+      def related_primary_key
+        if predicate_association
+          predicate_association.active_record_primary_key
+        else
+          related_class.primary_key if related_class
```
zorab47: Should this become an `elsif` for easier reading?

lib/active_admin/filters/active_filter.rb
wspurgin: It was like that previously, so I didn't know if it was convention to avoid if - elsif statements with no else.

lib/active_admin/filters/active_filter.rb
wspurgin: Like line 93, since a there were other places where it seemed that dangling if - elsif weren't preferred, I stuck with the if - else. But IMHO I agree that it should just be an if - elsif block

lib/active_admin/filters/active_filter.rb
```diff
@@ -76,12 +90,27 @@ def find_class
...
+        end
+      end
+
+      def predicate_association
```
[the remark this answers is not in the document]
wspurgin: Would you prefer `instance_variable_defined?` or just:
```ruby
def predicate_association
  @predicate_association ||= find_predicate_association
end

def find_predicate_association
  #...
end
```

wspurgin:
@zorab47 @Fivell I updated the PR with clearer if-elsif blocks, and caching of the predicate association. Appreciate the feedback :+1:

lib/active_admin/filters/active_filter.rb
```diff
@@ -75,13 +89,34 @@ def find_class?
...
+      def filter
+        resource.filters[name.to_sym]
+      end
+
+      def related_primary_key
+        if predicate_association
+          predicate_association.active_record_primary_key
+        elsif related_class
+          related_class.primary_key
         end
```
varyonic: Personally I like terse boolean operators over if/else:
```ruby
      def related_primary_key
        predicate_association && predicate_association.active_record_primary_key ||
          related_class && related_class.primary_key
      end
```

lib/active_admin/filters/active_filter.rb
```diff
@@ -54,6 +56,18 @@ def attribute_name
...
+
+        related_class.model_name.human
+      end
+
+      def filter_label
+        return unless filter
+
+        filter[:label]
+      end
+
```
varyonic:
```ruby
def filter_label
  filter && filter[:label]
end
```

lib/active_admin/filters/active_filter.rb
```diff
@@ -75,13 +89,34 @@ def find_class?
...
+      def filter
+        resource.filters[name.to_sym]
+      end
+
+      def related_primary_key
+        if predicate_association
+          predicate_association.active_record_primary_key
+        elsif related_class
+          related_class.primary_key
```
varyonic: Personally I like terse boolean operators over if/else:
```ruby
      def related_primary_key
        predicate_association && predicate_association.active_record_primary_key ||
          related_class && related_class.primary_key
      end
```

lib/active_admin/filters/active_filter.rb
```diff
@@ -75,13 +89,34 @@ def find_class?
...
         end
       end
+
+      def predicate_association
+        @predicate_association ||= find_predicate_association
```
Fivell: @wspurgin  since find_predicate_association can return nil, 

method can look like
```
return @predicate_association if defined? @predicate_association
@predicate_association ||= find_predicate_association
```

lib/active_admin/filters/active_filter.rb
wspurgin: Hurts readability though. You'll still need a block to return the proper primary key (as that only returns a `true` or `false`.

lib/active_admin/filters/active_filter.rb
wspurgin: Gotcha. That way we don't iterate every time if their is no predicate association. I'll make that update.

varyonic: `=` is sufficient if you know it's not defined.  Alternately:
```ruby
        @predicate_association = find_predicate_association unless defined?(@predicate_association)
        @predicate_association
```

lib/active_admin/filters/active_filter.rb
varyonic: It's a personal preference, maybe from writing too much C or perl. :-)

wspurgin: Definitely understand. I think I leave it as is for now though. It'll help me understand what's going on more quickly if I look at in a month than chained binary operators.  :see_no_evil:

wspurgin:
@varyonic Rebased.

wspurgin:
@varyonic That failure in the CI looks completely unrelated to this PR.  Looks broken on master too. I'm guessing that's what #5237 is for?

wspurgin:
@varyonic rebased and squashed. Thanks for the fixing the builds! :100:

PR #5233, opened 2017-11-04
View factory builder methods

varyonic:
Following on from #5206 pages/base feels quite dated compared to view components elsewhere in ActiveAdmin. I think that's because it relies on ViewFactory, which predates the inception of Arbre by several months (Feb vs Jun 2011). This PR adds a hook to ViewFactory to register builder_methods for the top level components, allowing pages/base and views/header to be simplified. ViewFactory could conceivably be deprecated and removed at a later date but I don't see that as a priority.


varyonic:
Updated and rebased.

varyonic:
Rebased.

varyonic:
:tada: Thank you!

PR #5248, opened 2017-11-20
Use activerecord-jdbc-adapter releases

deivid-rodriguez:
Finally jruby's activerecord adapters are getting traction and there's been releases for both Rails 5.0 and Rails 5.1. Let's see how they work.


deivid-rodriguez:
Jruby support is back now!

PR #5251, opened 2017-11-21
Mitigate memory bloat with CSV exports

f1sherman:
As reported in #4118 large CSV exports can cause memory bloat via `ActiveRecord::QueryCache`. This change disables the query cache while performing a CSV export to mitigate.

### DIFF
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -53,10 +53,12 @@ module ActiveAdmin
         csv << CSV.generate_line(columns.map{ |c| encode c.name, options }, csv_options)
       end
 
-      (1..paginated_collection.total_pages).each do |page|
-        paginated_collection(page).each do |resource|
-          resource = controller.send :apply_decorator, resource
-          csv << CSV.generate_line(build_row(resource, columns, options), csv_options)
+      ActiveRecord::Base.uncached do
+        (1..paginated_collection.total_pages).each do |page|
+          paginated_collection(page).each do |resource|
+            resource = controller.send :apply_decorator, resource
+            csv << CSV.generate_line(build_row(resource, columns, options), csv_options)
+          end
         end
       end
 
--- a/spec/unit/csv_builder_spec.rb
+++ b/spec/unit/csv_builder_spec.rb
@@ -234,6 +234,15 @@ RSpec.describe ActiveAdmin::CSVBuilder do
       builder.build dummy_controller, []
     end
 
+    it "should disable the ActiveRecord query cache" do
+      expect(builder).to receive(:build_row).twice do
+        expect(ActiveRecord::Base.connection.query_cache_enabled).to be_falsy
+        []
+      end
+      ActiveRecord::Base.cache do
+        builder.build dummy_controller, []
+      end
+    end
   end
 
   context "build csv using specified encoding and encoding_options" do


lib/active_admin/csv_builder.rb
```diff
@@ -53,11 +53,19 @@ def build(controller, csv)
...
-      (1..paginated_collection.total_pages).each do |page|
-        paginated_collection(page).each do |resource|
-          resource = controller.send :apply_decorator, resource
-          csv << CSV.generate_line(build_row(resource, columns, options), csv_options)
+      add_collection_to_csv = -> {
+        (1..paginated_collection.total_pages).each do |page|
+          paginated_collection(page).each do |resource|
+            resource = controller.send :apply_decorator, resource
+            csv << CSV.generate_line(build_row(resource, columns, options), csv_options)
+          end
```
varyonic: Could `add_collection_to_csv` be defined as a method instead of a lambda?

f1sherman: @varyonic sure thing, updated `add_collection_to_csv` to be a method!

lib/active_admin/csv_builder.rb
```diff
@@ -122,6 +115,15 @@ def humanize_name(name, resource, humanize_name_option)
 
     private
 
+    def add_collection_to_csv(csv, controller, csv_options)
+      (1..paginated_collection.total_pages).each do |page|
+        paginated_collection(page).each do |resource|
+          resource = controller.send :apply_decorator, resource
+          csv << CSV.generate_line(build_row(resource, columns, options), csv_options)
```
varyonic: Ugh. Looks like you have to pass in the local options from build instead of using self.options. I'm OK with leaving as a lambda, the options handling needs a deeper refactoring.

f1sherman: @varyonic no problem, I updated it to be a lambda again!

f1sherman:
@Fivell thanks for the review! 

I'm not seeing an obvious benefit to moving the change where you suggest. I like that the current change is scoped to where the ActiveRecord methods are called, vs. further up in the call stack. Can you help me understand why you feel your suggestion would be a better change?

I checked if `ActiveRecord` was defined as I noticed a precedent for that in [display_helper.rb](https://github.com/activeadmin/activeadmin/blob/a473cb68baee260021957ba3d6c3398459a4c551/lib/active_admin/view_helpers/display_helper.rb#L73). Great point about `ransack` being declared as a dependency, since `activerecord` is a dependency of `ransack` it doesn't seem possible that `ActiveRecord` is not defined, I've happily removed that unnecessary complexity!

Fivell:
@f1sherman 
>> 'm not seeing an obvious benefit to moving the change where you suggest.

I think disabling query cache is not  scope of csv builder , also csv builder can be replaced on resource level https://github.com/activeadmin/activeadmin/blob/5b711b6e22230e6e28be2a9d7367e9423d5dcc25/lib/active_admin/resource.rb#L50

f1sherman:
@Fivell thanks for the explanation. I tried the change that you suggested, however it does not appear to result in the query cache being disabled at the time `ActiveRecord` is called, likely because the `ActiveRecordBase::uncached` call is not in the scope of the block being set as the response body [here](https://github.com/activeadmin/activeadmin/blob/9af04e7516fc8b34bb126990a1b1a05b57f3b0c7/lib/active_admin/resource_controller/streaming.rb#L27). Let me know if I'm missing something or if there is an alternative implementation you would prefer.

Fivell:
@f1sherman  you're right I forgot about how streaming works

deivid-rodriguez:
So #5218 can be closed now?

fredngo:
Only a couple of comments here since I previously worked on https://github.com/activeadmin/activeadmin/pull/5218 to solve the same problem. Please take a look there.

1) `ActiveRecord::Base.uncached` will probably fail if the user is not using ActiveAdmin (e.g. using Mongoid)

2) The leak is actually in the `def build_row` method -- that's why I put my fix there instead of bubbling it up all the way into the `def build` method.

Anyway, this will solve the problem, either way I'm happy that the memory bloat will be gone.

PR #5253, opened 2017-11-28
Allow to apply custom decorations

faucct:
Restoring https://github.com/activeadmin/activeadmin/pull/4166 functionality while preserving the STI records instantiation (https://github.com/activeadmin/activeadmin/issues/4424).

### DIFF
--- a/lib/active_admin/resource_controller/data_access.rb
+++ b/lib/active_admin/resource_controller/data_access.rb
@@ -115,6 +115,7 @@ module ActiveAdmin
         get_resource_ivar || begin
           resource = build_new_resource
           resource = apply_decorations(resource)
+          resource = assign_attributes(resource, resource_params)
           run_build_callbacks resource
           authorize_resource! resource
 
@@ -127,7 +128,10 @@ module ActiveAdmin
       #
       # @return [ActiveRecord::Base] An un-saved active record base object
       def build_new_resource
-        scoped_collection.send method_for_build, *resource_params
+        scoped_collection.send(
+          method_for_build,
+          *resource_params.map { |params| params.slice(active_admin_config.resource_class.inheritance_column) }
+        )
       end
 
       # Calls all the appropriate callbacks and then creates the new resource.
--- a/spec/unit/resource_controller/data_access_spec.rb
+++ b/spec/unit/resource_controller/data_access_spec.rb
@@ -197,32 +197,36 @@ RSpec.describe ActiveAdmin::ResourceController::DataAccess do
   describe "build_resource" do
 
     let(:config) do
-      ActiveAdmin.register Post do
-        permit_params :body, taggings_attributes: [:id, :tag_id]
+      ActiveAdmin.register User do
+        permit_params :type, posts_attributes: :custom_category_id
       end
     end
 
-    let!(:tag) { Tag.create! }
+    let!(:category) { Category.create! }
 
     let(:params) do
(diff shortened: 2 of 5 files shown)


faucct:
ping

deivid-rodriguez:
This is addressing [this](https://github.com/activeadmin/activeadmin/pull/4883#issuecomment-347484534) @Fivell's comment, but also including the fix, right?

faucct:
Yes, it does.

deivid-rodriguez:
@Fivell what do you think?

spec/unit/resource_controller/data_access_spec.rb
```diff
@@ -197,32 +197,34 @@
...
+
     it "should return post with assigned attributes" do
-      expect(subject.body).to be_present
+      expect(subject).to be_a(User::VIP)
     end
 
     # see issue 4548
     it "should assign nested attributes once" do
-      expect(subject.taggings.size).to eq(1)
+      expect(subject.profile).not_to be_nil
```
Fivell: @faucct  I do not think this tests should be changed they addressed to issue https://github.com/activeadmin/activeadmin/issues/4548 about has_many , you replaced it with has_one example.

Fivell:
@deivid-rodriguez  not sure about this since @faucct  replaced part of specs instead of adding new specs

lib/active_admin/resource_controller/data_access.rb
```diff
@@ -127,7 +128,10 @@ def build_resource
       #
       # @return [ActiveRecord::Base] An un-saved active record base object
       def build_new_resource
-        scoped_collection.send method_for_build, *resource_params
+        scoped_collection.send(
+          method_for_build,
+          *resource_params.map { |params| params.slice(active_admin_config.resource_class.inheritance_column) }
```
Fivell: @faucct  I have many Form objects in AA so I would like to see it like 
```
scoped_collection.send( method_for_build, resource_params_for_build)
```

and 
```
def  resource_params_for_build
   if active_admin_config.resource_class.respond_to?(:inheritance_column)
      params.slice(active_admin_config.resource_class.inheritance_column)
  else 
     {}
   end
end
```

faucct: The solution you are suggesting would be valid in case you register a non-activerecord model. Is that what you do? Can you show me an example, so I could come up with test?

Fivell: @faucct  sort of this http://kenglish.co/using-form-pattern-with-an-activeadmin-controller/

faucct: Are you sure that a workaround to support a specific case of registering a non-activerecord resource belongs to activeadmin? You could fix this by defining an empty `inheritance_column` method in your form.

Fivell: @faucct  I already approved PR it is only one more suggestion to improve

faucct: So, merge?

faucct: ping

Fivell: @faucct  2 review needed for merge 
cc @varyonic

faucct:
Can this get a version bump, please?

faucct:
Thanks!

PR #5259, opened 2017-11-30
Use jQuery 3 as default minimum

wasifhossain:
Upgrading AA from 1.0.0 to 1.1.0 won't work as usual if jQuery 3 is not used in `active_admin.js.coffee`. jQuery UI functionalities does not work properly with jQuery 1 & 2 any more. Lots of type errors are thrown (`Uncaught TypeError: $(...).tableCheckboxToggler is not a function`) on any resource index page. Batch-Action button does not activate at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
   are now within the vendor directory. This will be replaced by alternatives and
   dropped entirely in a major release. Please remove any direct inclusions
   of `//= require jquery-ui`. [#5052][] by [@javierjulio][]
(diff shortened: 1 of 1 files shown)


varyonic:
See also #5256

varyonic:
These are related but there is no dependency. Ideally both the version check and the CHANGELOG comment should be backported to 1-1-stabe.

wasifhossain:
Now I see. Thanks for the direction. Going to close this PR and make another PR for `1-1-stable`.

Issue #1848, opened 2013-01-08
I18n problem with table columns on Dashboard only (works well on model page)


ByScripts:
Hi,

Supposing a **Page** model, with a **title** column.

In `admin/pages.rb` with something like this :

``` ruby
index do
  column :title

  // etc.
end
```

:white_check_mark:  The **title** column is correctly translated to french word **Titre** from `fr.activerecord.attributes.page.title`
:white_check_mark:  Each th and td for this column has a `class="title"` html attribute

Perfect.

Now in `admin/dashboard.rb` :

``` ruby
columns do
  column do
    panel "Whatever" do
      table_for Page.all do |page|
        column :title
      end
    end
  end
end
```

:x:  The "title" column is not translated

With :

``` ruby
        column t(:title), :title
```

:heavy_exclamation_mark:  The "title" column is correctly translated, **but** from `fr.title` directly (this is not DRY)
:heavy_exclamation_mark:  Each `th` and `td` for this column has a `class="titre"` html attribute (but the class attribute should not be translated)

So for the moment, I need to do this :

``` ruby
        column t(:title), :title, :class => :title
```

Not a big deal, but annoying, and I think the column behavior should be the same in both model page and dashboard.

Thank you

zorab47:
@ByScripts I think when you specify a `table_for` manually, you must provide the `i18n` key for translations to work. In your example (note the addition of `i18n`):

```ruby
columns do
  column do
    panel "Whatever" do
      table_for Page.all, i18n: Page do |page|
        column :title
      end
    end
  end
end
```

It isn't mentioned in the documentation, but see the [TableFor](https://github.com/activeadmin/activeadmin/blob/46526fc21917978f1d8253bc84082aeeb215ea45/lib/active_admin/views/components/table_for.rb#L14) component for more information.

deivid-rodriguez:
Mmmm, is that code to autodetect the `resource_class` (if `i18n:` is not specified) rigtht? Shouldn't it look into the class of the elements in the collection (instead of the class of the collection itself)?

Fivell:
@ByScripts  what version of AA do you use?

deivid-rodriguez:
This is a very old bug, maybe it's been fixed.

Fivell:
+1 for closing by https://github.com/activeadmin/activeadmin/commit/56f8aab55fda24d22a418e1d9e367eed40c180fc

PR #5263, opened 2017-12-01
Update table_for docs to show how to apply internationalization

zorab47:
Updates documentation for `table_for` to mention how to incorporate internationalization.

Should help alleviate issues like #1848.

### DIFF
--- a/docs/12-arbre-components.md
+++ b/docs/12-arbre-components.md
@@ -139,7 +139,7 @@ in `index_as_table`. It takes a collection and a hash of options and then
 uses `column` to build the fields to show with the table.
 
 ```ruby
-table_for order.payments do
+table_for order.payments, i18n: Payment do
   column(:payment_type) { |payment| payment.payment_type.titleize }
   column "Received On",     :created_at
   column "Details & Notes", :payment_details
@@ -147,7 +147,11 @@ table_for order.payments do
 end
 ```
 
-the `column` method can take a title as its first argument and data
+To maintain internationalization for the component, specify a resource to use for
+translations via the `i18n` named parameter. Unlike `index_as_table`, the
+translation resource cannot be automatically assumed for `table_for` components.
+
+The `column` method can take a title as its first argument and data
 (`:your_method`) as its second (or first if no title provided). Column also
 takes a block.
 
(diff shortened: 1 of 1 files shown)


zorab47:
Yes, it should be `Payment`. I'll update.

zorab47:
It depends on the collection/scope. In the case of an index action for a
resource, we can get the model name from the config. However, in a
‘table_for’, we’d have to get that information from the first value of the
collection. If the query is expensive it may be a big performance hit.

zorab47:
I suppose we might get similar info from the relation when it is an
ActiveRecord::Relation.

Are there other cases where it would be detrimental to access the
collection?

deivid-rodriguez:
But are you not supposed to iterate the collection anyways if you're going to show it in a table? :thinking:

zorab47:
Indeed. If we can get the necessary info without issuing 2 queries then it
would work.

deivid-rodriguez:
Wait, I was understanding the code wrong, it's calling `klass` on the collection, not `class`, which gives you the class of the elements in the relation. So I have no idea why the linked bug is happening in the first place, looks like it should work just fine?

Fivell:
@deivid-rodriguez @varyonic  I think also this is bug and can be fixed instead of adding docs

zorab47:
@deivid-rodriguez I'd guess it is calling `klass` to get the class of the model from an ActiveRecord::Relation. There are several scenarios to consider.

```ruby
Item.all.klass      # => Item
Item.all.class      # => Item::ActiveRecord_Relation
Item.all.to_a.class # => Array
```

Fivell:
@zorab47  ```klass```    is what we need it doesn't make additional query

deivid-rodriguez:
Yeah, so shouldn't it work? Isn't that exactly what we want?

Fivell:
@deivid-rodriguez  https://github.com/activeadmin/activeadmin/blame/8d58c9bdcbdfd634a366b423cf19f91d8ed7363a/lib/active_admin/views/components/table_for.rb#L15 

this line looks fine for me so I don't see why it doesn't work

deivid-rodriguez:
Yeah, same. :man_shrugging:

Fivell:
@zorab47  maybe related issue is about older version of AA 
https://github.com/activeadmin/activeadmin/commit/56f8aab55fda24d22a418e1d9e367eed40c180fc this should fix issue

deivid-rodriguez:
Here you go, that seems like it! I say let's close the related bug and update this PR to only document the `i18n` parameter but not mention any limitations.

zorab47:
Updated the docs to remove mention of limitations and rebased.

Issue #5198, opened 2017-10-09
filter behavior creates cross-site scripting (XSS) vulnerability


chrisbutler:
Found this old issue which calls out the problem, but only solves it for one specific case (and does so client-side using jQuery - which is probably not the most secure approach), I think: https://github.com/activeadmin/activeadmin/pull/3811

Here are the places in the code where user input is being rendered to the page without being sanitized:

https://github.com/activeadmin/activeadmin/blob/85d4690f824825fcd5f855146a56b7f98bcef37f/lib/active_admin/filters/active_sidebar.rb#L31

https://github.com/activeadmin/activeadmin/blob/8b15ff58a6d912899b46bf651354666d39f39ad0/lib/active_admin/batch_actions/views/batch_action_form.rb#L18

https://github.com/activeadmin/activeadmin/blob/9d0557ad05ece0da59a4df2c61468479093686a1/lib/active_admin/view_helpers/form_helper.rb#L14

### Expected behavior

What do you think should happen?

I should not be able to inject a `<script>` tag that will be executed when ActiveAdmin renders a page

### Actual behavior

What actually happens?

I can inject whatever I want because text input is not sanitized and then stored directly in `params[:q]`

### How to reproduce

- I enter `<script>alert('hax')</script>` in a text input in the filter sidebar
- I click filter
- I see an alert box with the text 'hax'
- I see a text input in the filter sidebar with value `<script>alert('hax')</script>`

(because this is currently the default behavior of ActiveAdmin, a reproduction repository should not be necessary)

chrisbutler:
p.s. happy to submit a PR once the issue is confirmed and a fix is decided on!

varyonic:
I confirmed the first of these.  No reason not to submit a PR.

justinsoong:
@chrisbutler pretty keen on this fix 👍

chrisbutler:
@justinsoong @deivid-rodriguez @markstory are all the issues fixed now?

deivid-rodriguez:
I think they are, can you check?

chrisbutler:
@deivid-rodriguez it does fix our issue! can we get an official release out with this?

PR #5265, opened 2017-12-03
Remove html safe to escape potential xss

justinsoong:


justinsoong:
fixes #5198

deivid-rodriguez:
Superseded by #5275 and #5284.

Thanks for getting this started!

PR #5267, opened 2017-12-04
Getting “unknown error: Chrome failed to start: crashed” in Travis CI.

varyonic:
Try https://github.com/travis-ci/travis-ci/issues/7150#issuecomment-340297038
@deivid-rodriguez Feel free to amend or replace.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 ---
 
 addons:
-  chrome: stable
+  apt:
+    packages:
+      - google-chrome-stable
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Looking at build logs, both approaches install the same version, so I don't think this is going to help. In my experience, what seems to fix these issues of chromedriver failing to start is to drop `chromedriver-helper` gem and install it manually.

That said, we can try this solution and see what happens.

PR #5275, opened 2017-12-05
No xss in filters sidebar

### LINKED ISSUE #5198: filter behavior creates cross-site scripting (XSS) vulnerability

deivid-rodriguez:
Fixes #5198.
Supersedes #5265.

### DIFF
--- a/spec/unit/pretty_format_spec.rb
+++ b/spec/unit/pretty_format_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe "#pretty_format" do
     mock_action_view.send *args, &block
   end
 
-  ['hello', 23, 5.67, 10**30, :foo, Arbre::Element.new.br(:foo)].each do |obj|
+  ['hello', 23, 5.67, 10**30, :foo, Arbre::Element.new.br].each do |obj|
     it "should call `to_s` on #{obj.class}s" do
       expect(pretty_format(obj)).to eq obj.to_s
     end
--- a/spec/unit/pretty_format_spec.rb
+++ b/spec/unit/pretty_format_spec.rb
@@ -8,10 +8,18 @@ RSpec.describe "#pretty_format" do
     mock_action_view.send *args, &block
   end
 
-  ['hello', 23, 5.67, 10**30, :foo, Arbre::Element.new.br].each do |obj|
+  ['hello', 23, 5.67, 10**30, :foo].each do |obj|
     it "should call `to_s` on #{obj.class}s" do
       expect(pretty_format(obj)).to eq obj.to_s
     end
+  end
+
+  it "normalizes Arbre elements" do
+    expect(pretty_format(Arbre::Element.new.br)).to eq("<br>\n")
+  end
+
+  it "sanitizes Arbre elements" do
+    expect(pretty_format(Arbre::Element.new.script('alert("foo");'))).to eq("alert(&amp;quot;foo&amp;quot;);\n")
   end
 
   shared_examples_for 'a time-ish object' do |t|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Bug fixes
+
+* Fixed [#5198][] Prevent XSS on sidebar's current filter rendering [#5272][] by [@deivid-rodriguez][]
+
 ## 1.1.0 [☰](https://github.com/activeadmin/activeadmin/compare/v1.0.0...v1.1.0)
 
 ### Bug Fixes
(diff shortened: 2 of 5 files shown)


spec/unit/pretty_format_spec.rb
```diff
@@ -8,12 +8,21 @@ def method_missing(*args, &block)
...
 
-  ['hello', 23, 5.67, 10**30, :foo, Arbre::Element.new.br(:foo)].each do |obj|
+  ['hello', 23, 5.67, 10**30, :foo].each do |obj|
     it "should call `to_s` on #{obj.class}s" do
       expect(pretty_format(obj)).to eq obj.to_s
     end
   end
 
+  it "normalizes Arbre elements" do
+    expect(pretty_format(Arbre::Element.new.br)).to eq("<br>\n")
```
deivid-rodriguez: This is an unexpected change in output from this patch. Before the ouput would be `<br />`, now it's `<br>`. Seems ok?

zorab47: https://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br#1946446

For more information.

zorab47:
Should it only be sanitize if the string isn't already consider `.html_safe?`? Perhaps users would optionally like to output HTML?

deivid-rodriguez:
Mmm, I think that's what `sanitize` does? It cleans up some backlisted tags but allows the ones considered harmless. That's why `<br>` passes through, but `<script>` disappears in the `pretty_format` unit tests.

spec/unit/pretty_format_spec.rb
javierjulio: Its fine. No need to be self closing in HTML5.

deivid-rodriguez: Only issue would be user searching for literal `<br />` on a text string but current filter will show `<br>`... Could be confusing somehow but seems reeeeally edge...

deivid-rodriguez: And an improvement anyways because previously nothing would be displayed, just some whitespace.

markstory:
There is still an issue in the `else` case when an `ActiveRecord::Base` instance is provided.
The else case uses ends up in `auto_link` or in `display_name`

https://github.com/activeadmin/activeadmin/blob/a473cb68baee260021957ba3d6c3398459a4c551/lib/active_admin/view_helpers/display_helper.rb#L72-L79

The `auto_link` method  uses `display_name` as a default value,  https://github.com/activeadmin/activeadmin/blob/8b15ff58a6d912899b46bf651354666d39f39ad0/lib/active_admin/view_helpers/auto_link_helper.rb#L14

The current code for `display_name` is missing escaping 

https://github.com/activeadmin/activeadmin/blob/a473cb68baee260021957ba3d6c3398459a4c551/lib/active_admin/view_helpers/display_helper.rb#L17-L36

This an acute problem when the ActiveRecord model implements one of the `name` methods defined in 

https://github.com/activeadmin/activeadmin/blob/26b7c841073cdddbdc8307ae821711654d4ac2c4/lib/active_admin/application_settings.rb#L27-L34

The end result is XSS from user supplied data stored in active record models.

deivid-rodriguez:
Thanks so much @markstory! Can you send us a patch fixing the other issues?

markstory:
@deivid-rodriguez Sure, would you like a pull request to this branch or to master?

deivid-rodriguez:
I think master is fine, it should be pretty independent from the code in this PR, right? Thanks!

rdlugosz:
Yes, this seems heavy handed. In our project, this strips the `target=` attribute out of links, so you cannot have certain links open in new tabs. This even happens when a string is marked as `html_safe`, (for example, if you wrap the link_to in a call to raw like: `raw(link_to "foo", "bar")`).

I'm all for sanitizing the strings, but perhaps the step should be ignored when strings are declared to be safe?

deivid-rodriguez:
Sorry about that, fixes welcome :pray:!

Or if the old situation is preferred, we can always revert my patch and release a new patch level version. I'm sorry I don't have time to contribute again now... :disappointed:

PR #5284, opened 2017-12-11
HTML encode display_name output.

markstory:
The display_name helper can end up invoking methods on ActiveRecord models or other arbitrary ruby objects that can contain unsafe user supplied data resulting in XSS holes.

Refs #5198
Refs #5265
Refs #5275


markstory:
Thanks!

PR #5286, opened 2017-12-11
Debug codecov

deivid-rodriguez:
We are having random coverage increases and decreases that cause some PR's to fail the codecov check (see https://codecov.io/gh/activeadmin/activeadmin/compare/6a0f2481761ee1387b30b9c89353faa4b4ffb359...870a474df28a6b331cc4205d9a484ee69c0f2f30/changes for an example).

The obvious fix for this is to add coverage for the "flaky" line, but out of curiosity I'd like to know why this happens.

These are some changes I made while investigating it. The useful one for this is to log the coverage percentage sent to codecov, that way I'll be able to see if the decrease happens consistenly for the same matrix entry or it's just fully random.

The other changes are just tweaks to the test run unrelated to this, maybe I should extract them to a separate PR?

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -35,11 +35,11 @@ require 'cucumber/rails'
 require 'rspec/mocks'
 World(RSpec::Mocks::ExampleMethods)
 
-Before do
+Around do |scenario, block|
   RSpec::Mocks.setup
-end
 
-After do
+  block.call
+
   begin
     RSpec::Mocks.verify
   ensure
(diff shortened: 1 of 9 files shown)


features/support/env.rb
```diff
@@ -88,18 +88,13 @@
...
-
-    # Reload Active Admin
-    ActiveAdmin.unload!
-    ActiveAdmin.load!
-  rescue
-    p $!
-    raise $!
-  end
+  # We are caching classes, but need to manually clear references to
+  # the controllers. If they aren't clear, the router stores references
```
varyonic: Such an awesome coincidence this fixes a problem I've been struggling with past two days.  Should be in ActiveAdminRequestHelpers also.

PR #5299, opened 2017-12-22
Display helper should not sanitize its input

faucct:
Following https://github.com/activeadmin/activeadmin/commit/0456e2d628f40ea6466a661c521bfd649b7388e4#r26411218

PR #5307, opened 2018-01-05
Fix #5306 HTML-safe strings getting sanitized again

### LINKED ISSUE #5306: HTML-safe strings get sanitized again
After updating from ActiveAdmin 1.1 to 1.2, I noticed that all links with a different method than GET stoped working corretly.

### How to reproduce

Given a configuation of:

```ruby
ActiveAdmin.register Post do
  actions :index, :destroy
  index do
    column do |post|
      link_to "Delete", "/admin/posts/1", method: :delete
    end
  end
end
```

### Expected behavior

When I go to Post#index I see a link: 

```ruby
<a rel="nofollow" data-method="delete" href="/admin/posts/1">Delete</a>
```

### Actual behavior

When I go to Post#index I see a link: 

```ruby
<a href="/admin/posts/1">Delete</a>
```

### Explanation

After looking through the changelog between 1.1 and 1.2 I found out that we started [preventing XSS](https://github.com/activeadmin/activeadmin/commit/3c708349774490a13ba2c4e747240d6794a22cbf), but in doing so we started calling [`sanitize` in all Strings](https://github.com/activeadmin/activeadmin/commit/3c708349774490a13ba2c4e747240d6794a22cbf#diff-b2d41db115eac24dba6f5e64bda8a690R69):

```ruby
def pretty_format(object)
  case object
  when String, Numeric, Symbol, Arbre::Element
    - object.to_s
    + sanitize(object.to_s)
  (...)
```

I think this had the unintended side effect of cleaning strings that are marked as HTML-safe, like the ones generated by `link_to`, so I think we should skip sanitization for `ActiveSupport::SafeBuffer`:

```ruby
def pretty_format(object)
  case object
  when ActiveSupport::SafeBuffer
    object.to_s
  when String, Numeric, Symbol, Arbre::Element
    sanitize(object.to_s)
  (...)
```

Am I correct in my assumption?

tiagotex:
When we started [preventing XSS](https://github.com/activeadmin/activeadmin/commit/3c708349774490a13ba2c4e747240d6794a22cbf) we started calling  [`sanitize` in all Strings](https://github.com/activeadmin/activeadmin/commit/3c708349774490a13ba2c4e747240d6794a22cbf#diff-b2d41db115eac24dba6f5e64bda8a690R69):

```ruby
def pretty_format(object)
  case object
  when String, Numeric, Symbol, Arbre::Element
    - object.to_s
    + sanitize(object.to_s)
  (...)
```

I think this had the unintended side effect of cleaning strings that are marked as HTML-safe, like the ones generated by `link_to`, so I think we should skip sanitization for `ActiveSupport::SafeBuffer`.

Fixes #5306

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Bug fixes
 
+* Fixed [#5306][] HTML-safe strings getting sanitized again [#5307][] by [@tiagotex][]
+
 ## 1.2.0 [☰](https://github.com/activeadmin/activeadmin/compare/v1.1.0...v1.2.0)
 
 ### Enhancements
@@ -216,6 +218,7 @@ Please check [0-6-stable][] for previous changes.
 [#4254]: https://github.com/activeadmin/activeadmin/issues/4254
 [#5043]: https://github.com/activeadmin/activeadmin/issues/5043
 [#5198]: https://github.com/activeadmin/activeadmin/issues/5198
+[#5306]: https://github.com/activeadmin/activeadmin/issues/5306
 
 [#4477]: https://github.com/activeadmin/activeadmin/pull/4477
 [#4759]: https://github.com/activeadmin/activeadmin/pull/4759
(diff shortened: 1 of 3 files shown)


tiagotex:
This was already fixed in #5299

PR #5308, opened 2018-01-05
Allow setting defaults for new records on form.has_many

jspanjers:
Use example:

```
  form do |f|
    f.inputs do
      f.has_many :comments, new_record: { text: 'Leave comment, value: Comment.new(body: 'Active Admin rules!') } do |b|
        b.input :body
    end
  end
```


faucct:
This would be great. I have also had this idea when trying to allow using records decorated with forms in activeadmin.

PR #5309, opened 2018-01-05
 Allow setting defaults for new records on form.has_many

jspanjers:
Use example:

```
  form do |f|
    f.inputs do
      f.has_many :comments, new_record: { text: 'Leave comment, value: Comment.new(body: 'Active Admin rules!') } do |b|
        b.input :body
    end
  end
```

### DIFF
--- a/docs/5-forms.md
+++ b/docs/5-forms.md
@@ -99,13 +99,18 @@ ActiveAdmin.register Post do
       end
     end
     f.inputs do
-      f.has_many :taggings, sortable: :position, sortable_start: 1 do |t|
+      f.has_many :taggings, sortable: :position,
+                            sortable_start: 1,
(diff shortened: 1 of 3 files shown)


Fivell:
@jspanjers interesting idea, don't you know if such behaviour can be emulated using :default option
```
f.has_many :comments, new_record: true  do |b|
        b.input :body, default: 'something'
    end

```

faucct:
This could be improved to make it possible to supply several new records to show multiple `Add ...` links for an STI class, e.g.:

```ruby
end
class UserComment < Comment
  belongs_to :user
end
class AnonymousComment < Comment
end

f.has_many :comments, new_record: [UserComment.new, AnonymousComment.new] do |fc|
  f.input :body
  f.input :user if f.object.is_a?(UserComment)
end
```

Issue #5117, opened 2017-08-02
Bug: comment routes appear even when comments are disabled


shaldengeki:
### Expected behavior

When disabling comments through `config.comments = false`, the comments-related routes should not be registered.

### Actual behavior

Comments-related routes (e.g. `/admin/comments`) appear in routes even if comments are disabled in configuration.

### How to reproduce

1. Add `config.comments = false` to `config/initializers/active_admin.rb`
2. Add `ActiveAdmin.routes(self)` to `config/routes.rb`
3. Run `rake routes | grep comments`

Comment-related routes appear.

(cc @alainbloch)

Fivell:
looks like this can be handled here 
https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/router.rb#L81

ghilead:
Perhaps registering the Comment resource should be avoided altogether for namespaces that disable it? 

https://github.com/activeadmin/activeadmin/blob/4b49319751b8c58363f46ae1d99e5bb1b7b65c39/lib/active_admin/orm/active_record/comments.rb#L23

vanegomez:
I have the same issue. I'm a bit confused on how you were able to solve it. Thank you

ghilead:
@vanegomez I ended up doing a workaround: created an empty custom resource page for ActiveAdmin::Comment. The routes are still there, but the page is blank. 

```
ActiveAdmin.register ActiveAdmin::Comment, as: 'Comment', namespace: :portal do
  index do
  end

  show do
  end
end
```

I traced back to where this was introduced and tried to understand the rationale, I don't think the fix as I proposed above is going to be effective. I tried to re-ignite the conversation here: 

https://github.com/activeadmin/activeadmin/issues/3695


Hope this helps!

mef79:
Found a workaround that prevents the apps from being generated that works for my purposes. It's not ideal, but I really wanted them gone.

```
# in config/intializers/active_admin.rb
# or anywhere else after ActiveAdmin is loaded but before ActiveAdmin.routes gets called

module NamespaceWithoutComments
  def register(resource_class, options = {}, &block)
    super unless resource_class == ActiveAdmin::Comment
  end
end

module ActiveAdmin
  class Namespace
    prepend NamespaceWithoutComments
  end
end
```

Consider before using:
- future compatibility concerns - if any of the active admin code that this references changes and you update, this snippet might cause problems in your app
[...]

Issue #3695, opened 2014-12-12
Filters, comments and global AA config


pranas:
I'd like to propose a change to config behaviour. Currently if you set `config.filters = false` or `config.allow_comments = false` in your global AA config, you can no longer use those features:

```
activeadmin/lib/active_admin/filters/resource_extension.rb:62:in `add_filter': Can't remove a filter when filters are disabled. Enable filters with 'config.filters = true' (ActiveAdmin::Filters::Disabled)
```

I would like to be able to disable them but still have an option to add filters and comments where I want them. 

Default filters became a common problem in our projects: when the data set grows association filters start to consume a lot of memory. I'd like to make `config.filters = false` a standard practice in my organisation. By default our admin pages wouldn't have any filters, we would add just the ones we want.

The similar situation is with AA comments, I don't want them on every model. I'd better have comments disabled by default but add them where I need to.

I can make a PR if I get a thumbs up.

swilgosz:
I know it is already done, but I think it would be nice to include new syntax in default generated 'config/active_admin.rb' file or information about this change.

timoschilling:
Wait a day, I want to rethink about it

ghilead:
It seems that this PR introduces the issue described here

https://github.com/activeadmin/activeadmin/issues/5117

I was about to add a PR to only register the Comment page when the namespace has comments enabled, but dug deeper and found this PR.

Do you guys have any suggestion on how to achieve both ?

TiSer:
Do you have some way to disable all filters in all models globally for now?

When I try to `config.filters = false` at activeadmin's initialisers I have the same error mentioned at the start of the topic: `Can't remove a filter when filters are disabled`.

PR #5311, opened 2018-01-08
do not register the comments controller if config.comments == false

ryanoboril:
Setting `config.comments = false` is supposed to **completely** disable comments functionality, but it doesn't exclude comments from the routes after calling ActiveAdmin.routes(self) and there is no way to remove a registered resource.  This PR addresses this issue by preventing the Comments controller from being registered at all if config.comments is set to false.

### DIFF
--- a/lib/active_admin/orm/active_record/comments.rb
+++ b/lib/active_admin/orm/active_record/comments.rb
@@ -20,70 +20,72 @@ ActiveAdmin.autoload :Comment, 'active_admin/orm/active_record/comments/comment'
 # Walk through all the loaded namespaces after they're loaded
 ActiveAdmin.after_load do |app|
   app.namespaces.each do |namespace|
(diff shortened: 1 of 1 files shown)


varyonic:
See also #3695 and #5117. The intended behavior is that a default for config.comments may be set at the namespace level and overridden at the resource level.

zorab47:
Ideally it would check to verify there are no ActiveAdmin resources that have comments enabled, right?

Bialogs:
Comment tables are also left in the database when setting comments to false...
Removing that table in conjunction with setting comments to false will obviously cause an application error when browsing to the route. Documentation should be updated to show that a migration to drop the comments table will finish completely disabling comments (to include removing?).

javierjulio:
I'm open to a configuration for removing the comments routes and not registering the controller if it's disabled but perhaps best revisited since the existing configuration also controls on what resources it does or doesn't show up. New suggestions can be brought up in #5117.

PR #5319, opened 2018-01-15
workaround chromedriver crashing on Travis

varyonic:
Try —no-sandbox option to workaround chromedriver crashing on Travis,…
… eg. https://travis-ci.org/activeadmin/activeadmin/builds/325657550

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -52,7 +52,15 @@ require 'capybara/cucumber'
 require 'capybara/session'
 require 'selenium-webdriver'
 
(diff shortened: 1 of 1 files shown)


Fivell:
disable-gpu option is necessary for the time being until some bugs are resolved.

no-sandbox option also appears to be necessary to get Chrome running inside a Docker container for GitLab's CI/CD environment.

In our final implementation we changed this to conditionally add the headless option unless you have CHROME_HEADLESS=false in your environment.

PR #5324, opened 2018-01-22
Views::ActiveAdminForm#create_another_checkbox styles are fixed

faucct:
Before:

After:

Is there any way to spec this?


faucct:
@Fivell, merge?

PR #5334, opened 2018-01-28
Add priority option to action item 

andreslemik:
Able to add action item to the beginning of action items via adding `priority: 0`.

```ruby
action_item :view, only: :show, priority: 0 do
  link_to 'View on site', post_path(post) if post.published?
end
```
image

### DIFF
--- a/lib/active_admin/resource/action_items.rb
+++ b/lib/active_admin/resource/action_items.rb
@@ -24,10 +24,8 @@ module ActiveAdmin
       #                         this action item on.
       #                 :except: A single or array of controller actions not to
       #                          display this action item on.
-      #                 :first: Place action item to the beginning of action_items array.
       def add_action_item(name, options = {}, &block)
-        method_name = options.delete(:first) ? :unshift : :<<
-        action_items.public_send method_name, ActiveAdmin::ActionItem.new(name, options, &block)
+        self.action_items << ActiveAdmin::ActionItem.new(name, options, &block)
       end
 
       def remove_action_item(name)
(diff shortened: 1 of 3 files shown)


varyonic:
Interesting idea, I would like to see more feedback from others. My first thought is to add a `priority` attribute, like for menu items.

Fivell:
@varyonic  I think priority is better one, since we already have such option for sidebar

andreslemik:
@varyonic, @Fivell  I was change this like for sidebar.

spec/unit/resource/action_items_spec.rb
```diff
@@ -32,6 +32,16 @@
       expect(resource.action_items.first.html_class).to eq("action_item test")
     end
 
+    it 'should be ordered by priority' do
+      resource.add_action_item :first, priority: 0 do
+        # Empty ...
```
Fivell: please also add one action item without priority to this example for testing default priority

lib/active_admin/resource/action_items.rb
```diff
@@ -110,6 +110,10 @@ def initialize(name, options = {}, &block)
     def html_class
       "action_item #{@options[:class]}".rstrip
     end
+
+    def priority
+      @options[:priority] || 10
```
Fivell: @andreslemik  please fix docs for add_action_item method with including description for new option

andreslemik: @Fivell, all done

Fivell:
@varyonic  what do you think ?

PR #5345, opened 2018-02-05
Manually add ID for polymorphic associations

jstumbaugh:
This is a recent bug I encounter when upgrading from 1.1 to 1.2.1.

Filtering on polymorphic associations throwing an error when rendering the page. After investigation, I found that #5238 , which was a great addition, removed the `_id` field from being added to the filter search for polymorphic associations.

The whole point of #5238 was to search on associations with a primary key other than ID. However, for polymorphic associations, I think it is OK to assume the field searching on is the ID column.

### DIFF
--- a/lib/active_admin/inputs/filters/select_input.rb
+++ b/lib/active_admin/inputs/filters/select_input.rb
@@ -15,7 +15,14 @@ module ActiveAdmin
             "#{reflection.through_reflection.name}_#{reflection.foreign_key}"
           else
             name = method.to_s
-            name.concat "_#{reflection.association_primary_key}" if reflection_searchable?
+
+            if reflection_searchable?
+              name.concat "_#{reflection.association_primary_key}"
+
+            elsif reflection_polymorphic?
+              name.concat "_id"
+            end
+
             name
           end
         end
@@ -50,6 +57,10 @@ module ActiveAdmin
 
         def reflection_searchable?
           reflection && !reflection.polymorphic?
+        end
+
+        def reflection_polymorphic?
+          reflection && reflection.polymorphic?
         end
 
       end


wspurgin:
Probably not safe to assume `_id` even for polymorphic associations. It'll be the probable common case, but not true in all scenarios. More over, if you have a polymorphic select filter, how would including just *one* of the two fields in the select be of value? Imagine this scenario:

If you selected in the Vehicle select dropdown "Car #1" which is the Car sub model with id equal to 1, the filtered results might also include "Truck #1" or "Boat #1" unless you also have a filter for the Vehicle type.

That's why polymorphic relationships aren't supported in the select filter by default. There's no *single* identifier the select filter can pass to ransack to indicate both fields for the polymorphic lookup. If you want to tackle that beast, I'm all for it 👍  I think it'll require some custom ransacker searchers though

jstumbaugh:
@wspurgin good point. @zorab47 and I ended up fixing it on our end with custom ransack filters. I think filtering on polymorphic associations would be a fantastic addition, but not part of this PR. It can be closed I think.

jstumbaugh:
We were using the [rails/globalid](https://github.com/rails/globalid) gem to filter on our polymorphic models. I think that now that it is included with rails, we could potentially use that to filter on polymorphics.

wspurgin:
Yup that's what I was thinking as well actually. @zorab47 and I used that in the past for to support this exact same scenario

javierjulio:
What do you all do for filtering polymorphic associations? I'd be interested to know. I've always just specified separate filters for both fields as it didn't seem that it would be possible with a single one, for example, with a billable polymorphic association we'd have:

```ruby
  filter :billable_type
  filter :billable_id, label: 'Billable Id'
```

jstumbaugh:
Using the global ID, we have used something similar to the following: 

```ruby
filter :vehicle_id, label: "Vehicle", as: :select,
  collection: proc { (Car.all + Truck.all).map { |v| [v.name, v.try(:global_id)] } }
```

with a custom ransacker as
```ruby
ransacker :vehicle_id, formatter: proc { |v|
  v.gsub("GlobalID-", "")
} do |parent|
  Arel.sql("(vehicle_type || '-' || vehicle_id)")
end
```

wspurgin:
Yup ☝️ that's exactly it

wspurgin:
That's also the reason I made https://github.com/activeadmin/activeadmin/pull/5215 :joy: To cache all those expensive polymorphic filter collections

PR #5347, opened 2018-02-07
Add option to ignore warning when redefining method

### LINKED ISSUE #5316: Warning: method `status` already defined
Just upgraded to version 1.2.1 and now constantly get this warning message:

`Warning: method `status` already defined`

No idea what is causing it.

Running rails 5.1.4, Ruby 2.4.1

jstumbaugh:
Fixes #5316.

[This commit](https://github.com/activeadmin/activeadmin/commit/c9e7f9ac9aee5b4ad4a58ba5216488cb5e7f4972#diff-c854d1a85b2cb3e8e48e1a559229039a) was added to warn about redefining a controller method. In my instance, we added the method to override the default controller's method. This merge adds an option when defining an action to ignore the warning message. It can be done similar to:

```ruby
member_action :method_name, ignore_warning: true
```

Otherwise, it will still continue to warn about the overwritten method.

### DIFF
--- a/lib/active_admin/resource_dsl.rb
+++ b/lib/active_admin/resource_dsl.rb
@@ -133,7 +133,11 @@ module ActiveAdmin
     # action.
     #
     def action(set, name, options = {}, &block)
-      warn "Warning: method `#{name}` already defined" if controller.method_defined?(name)
+      display_warning = !options.delete(:ignore_warning)
+
+      if controller.method_defined?(name) && display_warning
+        warn "Warning: method `#{name}` already defined"
+      end
 
       set << ControllerAction.new(name, options)
       title = options.delete(:title)
--- a/spec/unit/action_builder_spec.rb
+++ b/spec/unit/action_builder_spec.rb
@@ -143,6 +143,18 @@ RSpec.describe 'defining actions from registration blocks', type: :controller do
       end
     end
 
+    describe 'ignoring the warning about redefining a member action' do
+      let :action! do
+        ActiveAdmin.register Post do
+          member_action :process, ignore_warning: true
+        end
+      end
+
+      it 'writes warning to $stderr' do
+        expect($stderr.string).not_to include('Warning: method `process` already defined')
+      end
+    end
+
     describe 'defining collection action' do
       let :action! do
         ActiveAdmin.register Post do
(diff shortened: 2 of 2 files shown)


varyonic:
Perhaps you can `undef` the original method, that would make explicit your intent to replace it.

jstumbaugh:
That is a great idea, I didn't know about `undef`. Thanks for the suggestion!

zorab47:
Can you rebase your branch to fix CI?

varyonic:
Given the workaround I am :+1: for closing.

zorab47:
@varyonic that works for me. It would probably be best to add it to the docs too.

varyonic:
Doc PR welcome, though I'm unclear how many others will encounter this.

gglin:
@jstumbaugh @zorab47 in case there are no plans to add docs, can you post a code snippet of the `undef` workaround?  adding `undef method_name` right above the overriding `member_action` definition doesn't seem to work, as the member action method isn't defined on the ActiveAdmin::ResourceDSL class

jstumbaugh:
You have to undef the method within the controller before defining the member/collection action.

```ruby
controller do
  undef :my_action
end

collection_action :my_action do
   ...
end
```

gglin:
ah of course. I had tried `undef` in the controller but the member action was defined before the controller instead of after.  Thanks for your help!

machty:
I get this message for overriding `Kernel#clone` via `member_action :clone`, which seems perfectly reasonable/common. Having to `undef` feels like a crufty workaround. Maybe the rule should be: if Rails doesn't complain/break when you override a method on a controller, neither should AA?

```
[6] pry(#<ActiveAdmin::ResourceDSL>)> controller.method(:clone)
=> #<Method: Class(Kernel)#clone>
[7] pry(#<ActiveAdmin::ResourceDSL>)> controller.method(:clone).source_location
=> nil
```

zorab47:
@machty you'd prefer it only warn when overriding controller methods and not those inherited elsewhere?

machty:
@zorab47 that seems OK to me if that were easy to determine but I'm not sure there's a good way of fetching/determining a list of controller-only methods that shouldn't be overridden. And generally speaking, this kind of issue is going to be common in duck-typed languages like Ruby and if it were a bad enough footgun in Rails, then Rails would have already solved it, and I'm not sure it makes sense for AA to try and prevent this footgun if the giant Rails community hasn't already found a way.

zorab47:
@machty alternatively, there could be a config to indicate whether or not to produce the warnings. That is if the warnings are the issue. It would be like a "I know what I'm doing" mode.

varyonic:
I don't think being able to turn off the warning is desirable, you can always undef if you know what you are doing. If anything it needs a more explicit message that details where the original method was defined.

PR #5348, opened 2018-02-07
Allow passing a namespace to the Pundit adapter

iangreenleaf:
Pundit now allows you to pass an array to the policy methods to look up policies under a namespace (varvet/pundit#331). I know opinions on this vary, but for a lot of projects it would be helpful to have a different set of policies for the admin interface. This adds an optional config setting to use a namespace when ActiveAdmin looks up policies. So I could, for example, set:

```ruby
```

and then keep my admin policies in `Admin::ProductPolicy` etc while the rest of my site uses the non-namespaced `ProductPolicy`.


iangreenleaf:
Note: this is only available starting in Pundit 1.1.0 and will probably blow up on older Pundit versions. I'm not sure how you'd like to deal with that, so let me know.

zorab47:
Would it be feasible for this to be an alternative adapter (pundit with namespaces)? Or might that just confuse things?

iangreenleaf:
Hi @zorab47! It would certainly be possible to do it as a separate adapter, but I think it might add more confusion than it's worth. The code would probably end up pretty duplicated. And I think most people using the adapter wouldn't consider this addition something that fundamentally alters the adapter, it's more like a small adjustment. Is there a reason you're thinking a separate adapter might be better?

iangreenleaf:
I've added some specs that aren't super related to the PR, in the hopes that making the code coverage green will make this more appealing to the reviewers 😀

zorab47:
I do like the idea, but worry that the change will require existing Pundit setups to migrate their code in some way. What are your thoughts on the upgrade path? Does this require existing setups to consider a namespace? Or is this primarily a dependency upgrade?

iangreenleaf:
@zorab47 there should be no migration required for existing setups. Anyone who doesn't wish to use a namespace can keep using it the same way they always have and it should continue to work exactly the same. The only upgrade issue is the dependency upgrade—anyone who wants to use namespaces will need to have Pundit `>= 1.1.0`. Now that I think about it, anyone not using namespaces will actually be fine with older versions of Pundit, so you wouldn't have to force an upgrade on them if you didn't want to.

iangreenleaf:
Hey there, anything else I can do to get this merged?

PR #5351, opened 2018-02-08
Fix documentation on CanCan::Ability for Dashboard

eikes:
The ActiveAdmin::Page namespace_name is a [string](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/page.rb#L60), not a symbol. Otherwise the authorization will fail and result in an infinite loop, if the dashboard is your admin root.


eikes:
I am not sure why a change in documentation breaks the build...

PR #5358, opened 2018-02-21
[CI] Test against Ruby 2.5

nicolasleger:


varyonic:
Yes, but that pushes the CI run time over 75 minutes. Do we really need to test *every* version of MRI? Would min (2.2) and max (2.5) be sufficient?

deivid-rodriguez:
I'm fine with testing only min and max. And you're the most involved in maintainace now, so I'll support your decision.

As a personal opinion, I would probably move on with #5103 and #5104 now, at least unofficially. My previous point regarding those not being a maintainance burden is starting not to hold that much. For example, last time I was contributing we had to manually cherry-pick a view helper method from `rails` core only available on Rails 5. And in any case, by the `allowed_failures` entry being added here, Rails 4.2 seems not to be usable with latest ruby right now...

javierjulio:
I thought at first when this was submitted that Travis would test against the latest patch release for the specified minor version. Interesting that it does all patch releases. I don't think its necessary to test all of them. We should stay current to the recent one but I don't see a need or desire to go further than that.

deivid-rodriguez:
@javierjulio I I think you understood properly. Travis does not test all patch level releases by default. We only test the latest patch level release for every minor release, but that's still a long time in our case.

deivid-rodriguez:
@nicolasleger Almost there but you need to remove the allowed_failures entry that you added (since it no longer fails),  and also update the jruby version in the previous `allowed_failures` entry so that the exclusion is correctly picked up. Thanks!

.travis.yml
```diff
@@ -47,5 +48,8 @@ matrix:
   fast_finish: true
 
   allow_failures:
-    - rvm: jruby-9.1.15.0
+    - rvm: jruby-9.1.16.0
       gemfile: gemfiles/rails_52.gemfile
+    - rvm: 2.5.0
+      # devise 3.5.10 syntax incompatible error
+      gemfile: gemfiles/rails_42.gemfile
```
deivid-rodriguez: Please remove these 3 lines, Rails 4.2 + ruby 2.5.0 should be passing now. Thanks!

nicolasleger:
Sorry I mixed up the things. Looks good now, thanks.

varyonic:
The codecov/project failure is noise and can be ignored. Can you squash please?

.travis.yml
deivid-rodriguez: It seems that a wrong squash happened? This allow_failures entry is back there.

nicolasleger: Pfiou indeed ! Now updated, thanks.

PR #5359, opened 2018-02-22
Scopes groups

leio10:
This is a feature proposal. The idea is to add a group attribute to scopes to allow to show them grouped in the index page.

This can be useful when working with different kind of scopes for a model. For example, a user can have a state and a role. If we add the following DSL:
```ruby
scope :active, group: :state
scope :inactive, group: :state

scope :admin, group: :role
scope :user, group: :role
scope :guest, group: :role
```
Then, users index page will show something like this in the scopes section:
```
( active | inactive ) ( admin | user | guest )
```

With this PR adding groups to scopes only provides a visual help for admin users. In the future it also could be used to allow the user to select a scope for each group, but I guess that could be more complex to implement.

### DIFF
--- a/app/assets/stylesheets/active_admin/components/_scopes.scss
+++ b/app/assets/stylesheets/active_admin/components/_scopes.scss
@@ -6,5 +6,8 @@
       font-size: 0.9em;
       line-height: 10px;
     }
+    &:first-child a {
+      margin-left: 10px;
+    }
   }
 }
--- a/docs/3-index-pages.md
+++ b/docs/3-index-pages.md
@@ -214,6 +214,23 @@ end
 Scopes can be labelled with a translation, e.g.
 `activerecord.scopes.invoice.expired`.
 
+### Scopes groups
+
+You can assign group names to scopes to keep related scopes together and separate them from the rest.
+
+```ruby
+# a scope in the default group
+scope :all
+
+# two scopes used to filter by status
+scope :active, group: :status
+scope :inactive, group: :status
+
+# two scopes used to filter by date
+scope :today, group: :date
+scope :tomorrow, group: :date
+```
+
 ## Index default sort order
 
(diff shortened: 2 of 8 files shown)


leio10:
I've fixed a small bug in CSS and modified sample application to use groups for scopes. This is how it looks now with three groups:

imagen

zorab47:
Nice feature addition! Adding a bit of [documentation](https://github.com/activeadmin/activeadmin/blob/master/docs/3-index-pages.md#index-scopes) would be very helpful.

Do you think the new CSS class `scopes-group` is necessary? Is it possible to instead have multiple `ul.scopes` elements for grouping? (just me thinking out loud)

leio10:
Thanks for the feedback @zorab47!!

Regarding documentation, I had added some code sample in my first commits. Anyway, I reviewed them again and [added an explanation](https://github.com/leio10/activeadmin/blob/scopes_groups/docs/3-index-pages.md#scopes-groups) to improve it.

Regarding styles, you were right, I was adding some unnecessary classes. I modified the widget code and the stylesheet to keep changes minimum and I like it much more now. :blush:

zorab47:
Should the new `group` option accept a proc like `default` or `if` does for dynamic groupings? I can't think of a reason at the moment, but I bet someone might. Thoughts?

Looks great as is. 👍

leio10:
@zorab47 I can't think a situation where that could be needed. I have needed this feature several times, and it has always been with a fixed list of scopes and groups or looping through a collection, in which case the group could be extracted from the item. The second case would be something like this:
```ruby
scope :all
Tag.all.each do |tag|
  scope tag.name, group: tag.category.name, -> { where(tag: tag) }
end
```

varyonic:
Looks good. I think the CSS changes can be squashed.

Issue #5258, opened 2017-11-30
Comment wasn't saved, text was empty - when adding comment


hendricius:
### Expected behavior

After a blank installation I want to be able to add comments to a user.

### Actual behavior

I am adding a comment after a fresh install and then only get the error `Comment wasn't saved, text was empty.`

Did a small video with the issue here: https://app.hyfy.io/v/ab4tNqZiULU/

### How to reproduce

* Latest active admin version from github or `1.1.0`
* User model
* Enable comments

hendricius:
Has anyone had a similar issue before?

8vius:
Having the same issue

OthmanAmoudi:
did you try migrating the database ?
in your terminal try:
`rails db:migrate
`

zorab47:
Might it be related to permitted parameters?

clay-f:
same issue here

<img width="1014" alt="screen shot 2018-01-30 at 11 55 51" src="https://user-images.githubusercontent.com/16079417/35547541-9550ce16-05b4-11e8-8970-81806c4db54d.png">


this is terminal info

<img width="790" alt="screen shot 2018-01-30 at 11 57 21" src="https://user-images.githubusercontent.com/16079417/35547580-d02b4a84-05b4-11e8-949f-cf01ae693ec6.png">

user table:

```
  create_table "users", force: :cascade do |t|
    t.string "name"
    t.string "password_digest"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
  end
```



activeadmin version: 1.2.1
rails version: 5.1.4
ruby version: 2.5.0
mac os: 10.13

rainchen:
OK, I got it.
This is what happing when fails:

```
resource.errors
=> #<ActiveModel::Errors:0x00007fbf30436bc8
 @base=
  #<ActiveAdmin::Comment:0x00007fbf317f32b0
   id: nil,
   namespace: "admin",
   body: "11",
   resource_type: "Site",
   resource_id: 1,
   author_type: nil,
   author_id: nil,
   created_at: nil,
   updated_at: nil>,
 @details={:author=>[{:error=>:blank}]},
 @messages={:author=>["must exist"]}>

 ```

comment's author is nil and validation fails.

This issue is caused by Rails 5 makes belongs_to association required by default, here is [detail](https://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html)

[...]

PR #5360, opened 2018-02-23
allow to create an anonymous comment without authentication

rainchen:
related to #5258 

BTW, this issue is hard to debug in test env since there is a setting in config/environments/test.rb as:

```
if Rails::VERSION::MAJOR >= 5
    config.active_record.belongs_to_required_by_default = false
end
```

### DIFF
--- a/features/comments/commenting.feature
+++ b/features/comments/commenting.feature
@@ -27,6 +27,26 @@ Feature: Commenting
     And I should see "Hello from Comment"
     And I should see a comment by "admin@example.com"
 
+  Scenario: Create an anonymous comment without authentication
+    Given a show configuration of:
+      """
+        # disable authentication, but Comment's author is optional, it should be able to create a comment as author is anonymous
+        ActiveAdmin.application.authentication_method = false
+
+        ActiveAdmin.register Post
+      """
+    And I am logged out
+    When I go to the last post's show page
+    And I add a comment "Hello from Comment"
+    Then I should see a successful create flash
+    Given a show configuration of:
+      """
+        # reset authentication settings
+        ActiveAdmin.application.authentication_method = :authenticate_admin_user!
+        # re-register Post to make other tests pass
+        ActiveAdmin.register Post
+      """
+
   Scenario: View resource with comments turned off
     Given a show configuration of:
     """
--- a/lib/active_admin/orm/active_record/comments/comment.rb
+++ b/lib/active_admin/orm/active_record/comments/comment.rb
@@ -5,9 +5,10 @@ module ActiveAdmin
     self.table_name = "#{table_name_prefix}active_admin_comments#{table_name_suffix}"
 
     belongs_to :resource, polymorphic: true, optional: true
-    belongs_to :author, polymorphic: true
+    belongs_to :author, polymorphic: true, optional: true
 
     validates_presence_of :body, :namespace, :resource
+    validates_presence_of :author, if: -> { Rails.application.config.active_record.belongs_to_required_by_default && ActiveAdmin.application.authentication_method != false }
 
     before_create :set_resource_type
 
--- a/spec/unit/comments_spec.rb
+++ b/spec/unit/comments_spec.rb
@@ -23,6 +23,45 @@ RSpec.describe "Comments" do
       expect(comment.author).to eq(user)
     end
 
+    context "when authentication disabled" do
+      before do
+        ActiveAdmin.application.authentication_method = false
+      end
+
+      it "doesn't need an author" do
+        comment.valid?
+        expect(comment.errors[:author]).to be_empty
+      end
+    end
+
+    context "when authentication enabled" do
+      before do
+        ActiveAdmin.application.authentication_method = :authenticate_admin_user!
+      end
+
+      context "and application not using `belongs_to_required_by_default`" do
+        before do
+          allow(Rails.application.config.active_record).to receive(:belongs_to_required_by_default).and_return(false)
+        end
+
+        it "doesn't need an author" do
+          comment.valid?
+          expect(comment.errors[:author]).to be_empty
+        end
+      end
+
+      context "and application using `belongs_to_required_by_default`" do
+        before do
+          allow(Rails.application.config.active_record).to receive(:belongs_to_required_by_default).and_return(true)
+        end
+
+        it "needs an author" do
+          expect(comment).to_not be_valid
+          expect(comment.errors[:author]).to eq(["can't be blank"])
+        end
+      end
+    end
+
     it "needs a body" do
       expect(comment).to_not be_valid
       expect(comment.errors[:body]).to eq(["can't be blank"])


features/comments/commenting.feature
```diff
@@ -27,6 +27,34 @@ Feature: Commenting
...
+    Given a show configuration of:
+      """
+        # without :required options for rails 5
+        if Rails::VERSION::MAJOR >= 5
+          # assume Rails.application.config.active_record.belongs_to_required_by_default = true
+          ActiveAdmin::Comment.belongs_to_required_by_default = true
+          ActiveAdmin::Comment.belongs_to :author, polymorphic: true
+        else # use required: true for rails 4
+          ActiveAdmin::Comment.belongs_to :author, polymorphic: true, required: true
+        end
```
deivid-rodriguez: Just a question, no need to change anything in the PR. If we enable `belongs_to_required_by_default = true` in our test app (which we definitely should), would this whole block be needed for the test?

features/comments/commenting.feature
```diff
@@ -27,6 +27,34 @@ Feature: Commenting
...
+
+        # diable authentication
```
deivid-rodriguez: Small typo, s/diable/disable

lib/active_admin/orm/active_record/comments/comment.rb
```diff
@@ -4,7 +4,7 @@ class Comment < ActiveRecord::Base
     self.table_name = "#{table_name_prefix}active_admin_comments#{table_name_suffix}"
 
     belongs_to :resource, polymorphic: true
-    belongs_to :author,   polymorphic: true
+    belongs_to :author,   polymorphic: true, required: false
```
deivid-rodriguez: Would it make sense to add a presence validation to the model conditional on whether authentication is enabled or not? That way we could keep the same integrity validation for the 90% case (authentication enabled).

features/comments/commenting.feature
rainchen: no need, but I think there are many tests need to be fixed with `belongs_to_required_by_default = true`

features/comments/commenting.feature
rainchen: you got sharp eyes, bro. I suggest updating these in future updates.

lib/active_admin/orm/active_record/comments/comment.rb
rainchen: I don't quite remember the details, it looks like Rails has changed the default value for this option, so I have to set it to false.

rainchen: “ensures that no comments are created without an author” that's not true since I saw that there was a comment created by an "anonymous" author but not raising an exception.

deivid-rodriguez: Yeah, but that would be the case when authentication is disabled, right? I'm proposing to add a conditional presence validation only when authentication is enabled.

features/comments/commenting.feature
deivid-rodriguez: I merged #5495. If you'd like to wrap this up, please rebase and remove the whole block above!

rainchen: nice work, I will update this PR.

deivid-rodriguez: Thanks @rainchen!

deivid-rodriguez:
ping @rainchen

rainchen:
@deivid-rodriguez I'm in a heavy development sprint, I will try to fix this in this weekend

deivid-rodriguez:
Thanks for the heads up @rainchen, please take your time! It took me 8 months to review this PR after all... :sweat_smile:

rainchen:
@deivid-rodriguez I have updated the codes, plz review.

lib/active_admin/orm/active_record/comments/comment.rb
```diff
@@ -4,7 +4,14 @@ class Comment < ActiveRecord::Base
...
     belongs_to :resource, polymorphic: true
-    belongs_to :author,   polymorphic: true
+
+    # NOTE: after rails 5, :required is set to true by default and is deprecated.
+    belongs_to_optional_flag = if Rails::VERSION::MAJOR < 5
+                                 { required: false }
+                               else
+                                 { optional: true }
+                               end
+    belongs_to :author,   { polymorphic: true }.merge(belongs_to_optional_flag)
```
deivid-rodriguez: Up until now, `author` was compulsory for Rails 5 aplications with `belongs_to_required_by_default` and authentication enabled. Sould we add something like

```ruby
validates :author, presence: true, if: -> { Rails.application.config.active_record.belongs_to_required_by_default && ActiveAdmin.application.authentication_method != false }
```

rainchen: I think the major problem is the `belongs_to` default value for `: required ` is different in rails 4 and 5. If we want to keep the same behavior both in rails 4 and 5, we still have to change the `:required` value, and it will auto setup a validation, adding a `validates :author` is needless. 

refs:
https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-belongs_to

```
:required
When set to true, the association will also have its presence validated. This will validate the association itself, not the id. You can use :inverse_of to avoid an extra query during validation. NOTE: required is set to true by default and is deprecated. If you don't want to have association presence validated, use optional: true.
```

deivid-rodriguez: Yeah, sorry, I didn't explain it right. I mean turning off the associations validation (via `required: false` or `optional: true`, _and_ adding the presence validation that wouldn't be needless if we turn the association's validation off.

deivid-rodriguez: By the way, we'll be dropping support for Rails 4 soon, so the different values here won't be a problem (although I don't fully understand why they are a problem now).

rainchen: maybe we can have a meta tag like "TODO: refactor this after dropping support for Rails 4", then we can filter these codes quickly using `rake notes`

rainchen: > Still, what's wrong with
> 
> ```ruby
> belongs_to_optional_flag = if Rails::VERSION::MAJOR < 5
>                                  { required: false }
>                                else
>                                  { optional: true }
>                                end
> belongs_to :author,   { polymorphic: true }.merge(belongs_to_optional_flag)
> 
> validates :author, presence: true, if: -> { Rails.application.config.active_record.belongs_to_required_by_default && ActiveAdmin.application.authentication_method != false }
> ```
> 
> I think it would keep the same behavior in the general case, but still fix this bug?

we should not use`validate :author, presence: true` in any case, since `:author` is optional

deivid-rodriguez: Author is not currently optional for Rails 5 aplications with `belongs_to_required_by_default`. I'd like to make sure those applications still keep that behavior, but I'd like to be able to fix this bug too. Does my proposal not achieve that?

rainchen:
```
belongs_to_optional_flag = if Rails::VERSION::MAJOR < 5
                                 { required: false }
                               else
                                 { optional: true }
                               end
belongs_to :author,   { polymorphic: true }.merge(belongs_to_optional_flag)
```

these codes will set `optional: true` for  Rails 5 applications, even `belongs_to_required_by_default`.

deivid-rodriguez: Yes, exactly, that's why I'm suggesting to also add a presence validation that's only applied when `belongs_to_required_by_default` and authentication are enabled, in order to "restore" the validation in every case except for the one this bug applies to.

rainchen: I don't think so.
for Rails 4, `belongs_to :author` is optional, since ':require' is `false` by default.
for Rails 5, we should still keep `belongs_to :author` is optional, no matter Rails framework changed the `belongs_to_required_by_default` to whatever value.

deivid-rodriguez: Ok, let's try to agree on the basics.

Say an application is currently using Rails default (`belongs_to_required_by_default = true`) and not changing that. Let's say they're also using authentication. We can assume that all the comments in their comments table have an author. Then say they run a script to bulk load some comments with a bug that tries to create some comments without an author. With the current code, the `Comment` model will raise validation errors and prevent that. If we introduce this new code and set `belongs_to :author, polymorphic: true, optional: true`, we'll change that behavior and validation errors will no longer be raised for comments without author.

Do we agree on that?

rainchen: IMO, I think this is a bugfix but not a new feature, because when using Rails 4, comment's author is optional.
But Rails 5 has changed it to required by default. "the framework has changed the application's behavior", so I'm thinking this is a bug. This PR is going to fix this issue.

Otherwise, we have to write 2 test cases, one is for Rails 4 and the other one is for Rails 5 since the behavior is changed in your written cases.

deivid-rodriguez: I see your point now. According to you the expected behavior is that of Rails 4, and this fix simply restores that behavior. Specially since up until very recently we were using `belongs_to_required_by_default = false` in our specs, I understand how you may assume that.

However, I see this differently, ActiveAdmin has supported Rails 5 for a long time now, and when Rails 5 was released we didn't change this `belongs_to` relationship to be optional. That means we "accepted" the new Rails 5 feature of required `belongs_to` associations and decided to use it (but unintentionally introduced this bug). I think changing this behavior now means removing a feature that some applications could miss. This is why I suggest to keep the validation when it makes sense.

I'd like other @activeadmin/maintainers to chime in, if possible. I know it's a lenghty discussion, but I would find more opinions valuable.

rainchen: no problem, I can wait for the final conclusion. You're very professional, I'm glad to discuss and make this clear.

rainchen: ok, I will update it.

rainchen:
@deivid-rodriguez hi, how is this PR going? I saw all CI jobs had passed

deivid-rodriguez:
Yeah. I rebased it but still debating with myself whether the best path forward is the current solution (just make the author optional), or the one I had proposed (add more complex logic that respects whether authentication is enabled and whether the application has `belongs_to_required_by_default` set).

PR #5362, opened 2018-02-25
New form_helper_method setting

blocknotes:
To use SimpleForm / Formtastic / any other form system.

Usage example (with `config.form_helper_method = :simple_form_for` in the initializer):

```rb
  form builder: ::SimpleForm::FormBuilder do |f|
    f.input :published
    f.input :name
    f.input :category
    f.submit
  end
```

In my opinion another setting could be added to specify a default form builder (ActiveAdmin::FormBuilder or something else).

### DIFF
--- a/lib/active_admin/namespace_settings.rb
+++ b/lib/active_admin/namespace_settings.rb
@@ -40,6 +40,9 @@ module ActiveAdmin
     # is a currently authenticated admin user
     register :authentication_method, false
 
+    # The method to call to create forms
+    register :form_helper_method, :semantic_form_for
+
     # The path to log user's out with. If set to a symbol, we assume
     # that it's a method to call which returns the path
     register :logout_link_path, :destroy_admin_user_session_path
--- a/lib/active_admin/views/components/active_admin_form.rb
+++ b/lib/active_admin/views/components/active_admin_form.rb
@@ -27,7 +27,7 @@ module ActiveAdmin
         @resource = resource
         options = options.deep_dup
         options[:builder] ||= ActiveAdmin::FormBuilder
-        form_string = helpers.semantic_form_for(resource, options) do |f|
+        form_string = helpers.send( ActiveAdmin.application.form_helper_method, resource, options ) do |f|
           @form_builder = f
         end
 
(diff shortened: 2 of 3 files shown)


varyonic:
Interesting idea. My concern is that this will raise expectations that ActiveAdmin will support SimpleForm with has_many. SimpleForm with has_many support would be an great plugin.

blocknotes:
@varyonic sure - I'm working on it :)
Currenty I'm monkey patching the build method to make it work, that's the reason of my PR.

But it's not easy to maintain the exact compatibility. I'm also building other methods like inputs, actions, etc.

Any suggestion is appreciated. Preview:
```
module ActiveAdmin
  class SimpleFormBuilder < ::SimpleForm::FormBuilder
    include ::Formtastic::Helpers::InputsHelper

    attr_accessor :already_in_an_inputs_block

    def action( method, options = {} )  # -> Formtastic::Helpers::ActionHelper
      case method
      when :button, :reset
        button :button, options[:label], type: method
      when :submit
        button :submit, options[:label]
      else
        template.link_to options[:label], options[:url]
      end
    end

    def cancel_link(url = {action: "index"}, html_options = {}, li_attrs = {})  # -> ActiveAdmin::FormBuilder
      li_attrs[:class] ||= "cancel"
      li_content = template.link_to I18n.t('active_admin.cancel'), url, html_options
      template.content_tag(:li, li_content, li_attrs)
    end

    def has_many( assoc, options = {}, &block )  # -> ActiveAdmin::FormBuilder
      # TODO: improve me
      html = template.simple_fields_for assoc, &block

      tag = @already_in_an_inputs_block ? :li : :div
      html = template.content_tag( tag, html, class: "has_many_container #{assoc}" )
      # html = template.content_tag(tag, html, class: "has_many_container #{assoc}", 'data-sortable' => sortable_column, 'data-sortable-start' => sortable_start)
      # template.concat(html) if template.output_buffer

      html
    end
  end
end
```

varyonic:
I think monkey patching is a reasonable approach until there is a more complete solution.

blocknotes:
Beta version of the plugin ready: [activeadmin_simple_form](https://github.com/blocknotes/activeadmin_simple_form)

blocknotes:
Stale PR :)
Closing it.

PR #5364, opened 2018-02-26
Also look into private method

maxime-carbonneau:
Add :include_private to respond_to? so the helper can be included into a controller

### DIFF
--- a/lib/active_admin/view_helpers/auto_link_helper.rb
+++ b/lib/active_admin/view_helpers/auto_link_helper.rb
@@ -35,7 +35,7 @@ module ActiveAdmin
 
       # Returns the ActiveAdmin::Resource instance for a class
       def active_admin_resource_for(klass)
-        if respond_to? :active_admin_namespace
+        if respond_to? :active_admin_namespace, :include_private
           active_admin_namespace.resource_for klass
         end
       end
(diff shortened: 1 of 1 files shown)


zorab47:
The second argument to [`respond_to?`][1] is a boolean. I'm not sure if `link_to` is available within the controller. You'd likely have to call `link_to` from `view_context`, which may not achieve what you're looking for...

What's the goal of this change? 

[1]: http://ruby-doc.org/core-2.5.0/Object.html#method-i-respond_to-3F

maxime-carbonneau:
I would like to access auto_url_for from the controller.

maxime-carbonneau:
But I can achieve the same result by overloading the method in the controller.

zorab47:
Looks OK to me. Thoughts @varyonic or @Fivell ?

Fivell:
@zorab47  I'm not sure I understand the reason of this change.  Can you summarize this ?

maxime-carbonneau:
The idea is to let the controller use the method auto_url_for :
```ruby
def destroy
  destroy! do
    redirect_to auto_url_for(resource.imageable)
  end
end
```

javierjulio:
Not sure there's any interest or need for this. If so, we can reopen.

Issue #432, opened 2011-08-30
has_many problem; undefined method 'new_record?'


nickpellant:
I'm trying to customise a ActiveAdmin form for a Recipe model that has a has_many relationship with Step.

``` ruby
class Recipe < ActiveRecord::Base
  has_many :steps
end

class Step < ActiveRecord::Base
  acts_as_list :scope => :recipe

  belongs_to :recipe
end
```

I have the following in my ActiveAdmin file with relation to this:

``` ruby
form do |f|
  f.has_many :steps do |ing_f|
    ing_f.inputs
  end
end
```

The following error is thrown when I try to load the form:

```
undefined method `new_record?' for nil:NilClass
```

I've isolated it so far to the has_many method but I'm lost past this. Any advice and help would be appreciated!

tiagodll:
I wish I could upvote this :+1:

bobzhen:
thank you!

dkniffin:
This should be documented under the "Nested Resources" section of this page: https://activeadmin.info/5-forms.html

thisismydesign:
There should be a complete working example in the docs imo.

PR #5366, opened 2018-03-01
Add note about accepts_nested_attributes

dkniffin:
I just ran into an error `undefined method 'new_record?' for nil:NilClass`, and luckily I found my answer in #432, but I feel like this should be noted in the docs

### DIFF
--- a/docs/5-forms.md
+++ b/docs/5-forms.md
@@ -84,6 +84,7 @@ your model uses `has_one`:
 
 ```ruby
 ActiveAdmin.register Post do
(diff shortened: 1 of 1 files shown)


dkniffin:
If anyone reviews this in the next few minutes, please hold off on merging it. I might have some more edits to make.

varyonic:
Thanks. Please squash into one commit.

dkniffin:
@varyonic Alright that's done. I'm not sure why travis failed in that build...

PR #5367, opened 2018-03-05
It should be possible to specify multiple new has_many record links

faucct:


faucct:
ping

faucct:
ping

zorab47:
Can you describe the use case and provide more background on this change?

faucct:
We use this feature to make it possible to create `STI` records: instead of exposing `#type` select input each subclass has its own "Add record" link with a customized form which could have more or less inputs. Persisted records also get customized forms.
Here is another usecase – https://github.com/activeadmin/activeadmin/pull/5309

faucct:
ping

faucct:
ping

faucct:
ping

PR #5368, opened 2018-03-05
Add missing dutch translations

dennisvdvliet:
The yaml file for Dutch was missing some translation. Here they are.


dennisvdvliet:
Thanks for the swift reply 
🏎🏎🏎🏎

PR #5370, opened 2018-03-06
Add better support for rendering lists

dkniffin:
Before, lists of items (arrays and ActiveRecord relations) would render poorly. For example, if there's a `post` with tags of `[:a, :b, :c]`, then you render that with AA like so:

```rb
show do
  attributes_table do
    row :tags
  end
end
```

you'd get some thing like "Tags: Tag"

Now, it loops over those lists and joins them with a comma, so you'd get something like "Tags: a, b, c"

### DIFF
--- a/lib/active_admin/view_helpers/display_helper.rb
+++ b/lib/active_admin/view_helpers/display_helper.rb
@@ -69,14 +69,22 @@ module ActiveAdmin
           object.to_s
         when Date, Time
           I18n.localize object, format: active_admin_application.localize_format
+        when Array
+          format_collection(object)
         else
           if defined?(::ActiveRecord) && object.is_a?(ActiveRecord::Base) ||
              defined?(::Mongoid)      && object.class.include?(Mongoid::Document)
             auto_link object
+          elsif defined?(::ActiveRecord) && object.is_a?(ActiveRecord::Relation)
+            format_collection(object)
           else
             display_name object
           end
         end
+      end
+
+      def format_collection(collection)
+        safe_join(collection.map { |item| pretty_format(item) }, ', ')
       end
 
       def boolean_attr?(resource, attr, value)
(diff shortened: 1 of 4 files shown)


dkniffin:
It looks like the missing coverage is around arrays. I'll take care of that at some point soon.

dkniffin:
I think that's a valid concern. However, I think in those cases, people will likely override this functionality with a block or a decorator anyways. In my case, I have a lot of associations like this, where I have very few records, and it'd be nice if AA had a reasonable default behavior that simply rendered that list. It would help me avoid doing `.map(&:name).join(", ")` in so many different places.

Also, FWIW, I just tested it with 6000 records, and it does take a few seconds to load, but it does render it just fine.

dkniffin:
@zorab47 or anyone else: any chance of getting this merged?

Fivell:
@dkniffin  can you share how it looks like ?

dkniffin:
Sure thing. I've got a model `Plant`, with `has_many :security_users`. `SecurityUser` has an `email` attribute, and I've seeded 1000 security users for this plant. Then in AA, I've got this:

```rb
ActiveAdmin.register Plant do
  show do
    attributes_table do
      row :security_users
    end
  end
end
```

Here's the before:

screen shot 2018-03-15 at 9 09 56 am

And the after:

screen shot 2018-03-15 at 9 11 07 am

Fivell:
cc @zorab47 @varyonic

zorab47:
Does joining with a comma work across languages? Should it be internationalized or use the `#to_sentence` helper? 

http://devdocs.io/rails~5.1/array#method-i-to_sentence

dkniffin:
Great question. I haven't really done I18n before. If someone can guide me to what needs to be done for that, I'd be fine doing the work.

PR #5376, opened 2018-03-15
Keeps filter input_html as a proc

kjeldahl:
Adds dup of filter options to keep input_html as a proc so it will be evaluated on each form rendering

### DIFF
--- a/lib/active_admin/filters/forms.rb
+++ b/lib/active_admin/filters/forms.rb
@@ -57,8 +57,11 @@ module ActiveAdmin
           filters.each do |attribute, opts|
             next if opts.key?(:if)     && !call_method_or_proc_on(self, opts[:if])
             next if opts.key?(:unless) &&  call_method_or_proc_on(self, opts[:unless])
-            opts[:input_html] = instance_exec(&opts[:input_html]) if opts[:input_html].is_a?(Proc)
-            f.filter attribute, opts.except(:if, :unless)
+
+            filter_opts = opts.except(:if, :unless)
+            filter_opts[:input_html] = instance_exec(&filter_opts[:input_html]) if filter_opts[:input_html].is_a?(Proc)
+
+            f.filter attribute, filter_opts
           end
 
           buttons = content_tag :div, class: "buttons" do


zorab47:
Is the only difference adding a call to `dup` the opts? Is this change fixing it by maintaining the Proc instead of replacing it with a string?

kjeldahl:
Yes and yes.
Just like the :if and :unless procs are maintained and not overwritten by their first invocation.

lib/active_admin/filters/forms.rb
```diff
@@ -57,7 +57,10 @@ def active_admin_filters_form_for(search, filters, options = {})
           filters.each do |attribute, opts|
             next if opts.key?(:if)     && !call_method_or_proc_on(self, opts[:if])
             next if opts.key?(:unless) &&  call_method_or_proc_on(self, opts[:unless])
-            opts[:input_html] = instance_exec(&opts[:input_html]) if opts[:input_html].is_a?(Proc)
+            if opts[:input_html].is_a?(Proc)
+              opts = opts.dup
```
zorab47: It would be better to avoid assigning the `opts` variable because it can be confused with the block variable with same name. Something like this would be a better approach:

```ruby
filters.each do |attribute, opts|
  next if opts.key?(:if)     && !call_method_or_proc_on(self, opts[:if])
  next if opts.key?(:unless) &&  call_method_or_proc_on(self, opts[:unless])

  filter_opts = opts.except(:if, :unless)
  filter_opts[:input_html] = instance_exec(&fitler_opts[:input_html]) if filter_opts[:input_html].is_a?(Proc)

  f.filter attribute, filter_opts
end
```

kjeldahl: Indeed. What you propose is much better. I have updated the code

kjeldahl:
@varyonic it is squashed now.

varyonic:
Fixes #5377

mauro-oto:
@deivid-rodriguez any chance we could get this in a future 1.4.2 or 1.5.0 release? I'm currently using 1.4.1 plus a monkeypatch to address the problem.

deivid-rodriguez:
I released 1.4.2 with this fix!

Issue #5369, opened 2018-03-05
Destroy errors on nested resources don't create an error flash message


joker-777:
### Expected behavior

If a resource has a nested resource which depends on the parent object on destroying

`has_many :nested_resources, :dependent => :destroy`

then active admin should show an error message too.

### Actual behavior

Active admin shows a positive destroy flash message but the object is actually not destroyed.

### How to reproduce

Have two models with the `has_many` relationship where the parent destroys the children automatically.
Have some validations in the children model which fail when you try to destroy the parent object.
This error case will not be shown to the user as a flash message but instead, a positive message is shown.

### Nice to have

The concrete error messages should be shown in the flash message. Right now only a generic error message is shown as a flash message when the parent object has error messages.

domangi:
This is due to the fact that when an association prevents the destroy, normally by throw(:abort), the error message is written only on the association object itself, while the errors array of the object you are trying to delete remains empty.

``` ruby

class Post < ActiveRecord::Base
   ...
   has_many :taggings, dependent: :destroy
end

class Tagging < ActiveRecord::Base
  ...
  belongs_to :post
  before_destroy do
    self.errors.add(:base, "Tagging cannot be deleted")
    throw(:abort)
  end
end
``` 

Assume you have a post object with an associated tagging, when you try to destroy the post object it will be prevented by the before_destroy callback in the tagging object, but only the tagging object errors will have an entry, while the post object errors messages will be empty, and active admin will show a success message even if the object was non deleted.

To fix this problem you have to ensure that there is an error message in the object your try to delete. You can do it manually in your associated object before throwing abort, in the example above it would be something like this:

``` ruby
class Tagging < ActiveRecord::Base
  ...
  belongs_to :post
  before_destroy do
    self.post.errors.add(:base, "Tagging cannot be deleted")
    throw(:abort)
  end
end
``` 

Or you can override the destroy method in active_admin in order to check if the destroy method was successful, and add an error message if it failed:

``` ruby
# lib/destroy_action_monkey_patch.rb
module ActiveAdmin
  class ResourceController < BaseController
    module DataAccess
      def destroy_resource(object)
        run_destroy_callbacks object do
          if object.destroy
            true
          else
            object.errors.add(:base, "Not destroyed") if object.errors.empty?
            false
          end
        end
      end
    end
  end
end
```

I prefer this method because IMHO if the destroy fails then it should never say "successfully destroyed..."

[...]

PR #5385, opened 2018-03-25
5369 destroy errors on nested resources

domangi:
This pull-request solves a bug described in https://github.com/activeadmin/activeadmin/issues/5369
When a user tries to destroy an object, but the operation fails, for example because a throw(:abort) occurs in a before_destroy callback, without writing an error message on the object itself, then active admin will not destroy the object but it will show a positive flash message saying "Successfully deleted ....". Same happens when making a bulk delete.

In this solution the destroy outcome is checked and if it fails and the error messages are empty, then a generic error message is added, so that an error message is shown to the user instead of the wrong success message. In the case of bulk delete, successful and failed destroy operations are counted and presented to the user together, for example "Successfully destroyed 3 posts." and "Failed to destroy 2 posts.".

### DIFF
--- a/lib/active_admin/batch_actions/resource_extension.rb
+++ b/lib/active_admin/batch_actions/resource_extension.rb
@@ -62,19 +62,32 @@ module ActiveAdmin
         }
 
         add_batch_action :destroy, proc { I18n.t('active_admin.delete') }, destroy_options do |selected_ids|
+          destroyed_count = 0
+          not_destroyed_count = 0
+
           batch_action_collection.find(selected_ids).each do |record|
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@domangi Sorry for getting back so late. In https://github.com/activeadmin/activeadmin/issues/5369#issuecomment-375938722 you said that you wanted to write some tests before sending a PR. Are you still interested in doing that? This PR is looking good but needs some tests indeed.

deivid-rodriguez:
ping @domangi

domangi:
Yes, of course. But I need some hint on how to setup the test.

PR #5391, opened 2018-04-10
Relax Rails dependency for Rails 5.2.0

junk0612:
Rails 5.2.0 has been released!

- http://weblog.rubyonrails.org/2018/4/9/Rails-5-2-0-final/

PR #5394, opened 2018-04-11
Capybara is causing CI errors due to 3.0 release

zorab47:
Require Capybara < 3.0 until the issue is resolved. See #5391


varyonic:
FWIW 3.0 was released last week https://github.com/teamcapybara/capybara/blob/master/UPGRADING.md

PR #5396, opened 2018-04-12
Add Rails 5.2 support.

gobijan:
This pull request enables Rails 5.2 support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,11 +27,12 @@ script:
   - bundle exec rake docs:build
 
 rvm:
+  - 2.2.10
+  - 2.3.7
+  - 2.4.4
+  - 2.5.1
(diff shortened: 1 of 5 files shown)


gobijan:
From what I see there might be an issue with the cucumber-rails gem which has the railites dependency set to ['>= 4', '< 5.2']. Could be that the cucumber-rails gem needs update first to support rails 5.2 and then the 5.2 tests can pass.

zorab47:
Thanks, but I'd prefer to split this into separate PRs. One PR would be to handle the ruby and gemfile updates.

And the error with Cucumber specs is resolved on master by disallowing Capybara 3.0. See #5394.

zorab47:
Closing in favor of #5399.

PR #5399, opened 2018-04-13
Allow Rails 5.2

zorab47:
Allows Rails 5.2 by adjusting the restriction on railties to be `<= 5.3`.

Why 5.3? Incase there are releases between 5.2 and 6.0 that require changes to support. See discussion in #5391 and #4916.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ least version `1.14.3`.
 Select the Gemfile for your preferred Rails version, preferably the latest:
 
 ```sh
-export BUNDLE_GEMFILE=gemfiles/rails_51.gemfile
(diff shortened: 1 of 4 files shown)


zorab47:
@varyonic do you know how easily these changes can be backported?

varyonic:
Fixed!

rafaelfranca:
> In case there are releases between 5.2 and 6.0 that require changes to support.

It will not exist any release between 5.2 and 6.0

zorab47:
Agreed @rafaelfranca. But we'd like to avoid the beta and release candidates until the project is ready:

```ruby
Gem::Version.new("6.0.0.rc1") < Gem::Version.new("6.0") # => true
```

rafaelfranca:
Cool, that is a good reason. 6.0.x should also reject it, but I think 5.3 is fine for that.

Issue #1665, opened 2012-09-12
Localize date picker


mspanc:
In 0.5.0, even if locale of whole activeadmin is changed, date picker remains to load with "en" locale.

PR #5408, opened 2018-05-11
Allow proc label in datepicker input

tiagotex:
Currently datepicker input doesn't allow to pass label as a proc, this is needed when supporting multiple languages.

### DIFF
--- a/lib/active_admin/inputs/datepicker_input.rb
+++ b/lib/active_admin/inputs/datepicker_input.rb
@@ -9,6 +9,13 @@ module ActiveAdmin
         end
       end
 
+      # Can pass proc to filter label option
+      def label_from_options
+        res = super
+        res = res.call if res.is_a? Proc
+        res
+      end
+
       private
       def datepicker_options
         options = self.options.fetch(:datepicker_options, {})
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Master (unreleased)
 
+### Enhancements
+
+#### Minor
+
+* Allow proc label in datepicker input [#5408][] by [@tiagotex][]
+
 ### Bug fixes
 
 * Fixed the string representation of the resolved `sort_key` when no explicit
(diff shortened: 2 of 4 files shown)


tiagotex:
Is there something else I can do here?

lib/active_admin/inputs/datepicker_input.rb
```diff
@@ -1,6 +1,8 @@
 module ActiveAdmin
   module Inputs
     class DatepickerInput < ::Formtastic::Inputs::StringInput
+      include ::Formtastic::Inputs::Base
```
deivid-rodriguez: I don't think this inclusion is needed?

varyonic:
Related to #1665?

deivid-rodriguez:
Somewhat, I think this one is about i18n (translating the input label), and that other issue is about l10n (localized month names, week day names, date formats, and so on).

lib/active_admin/inputs/datepicker_input.rb
tiagotex: You are right, `Formtastic::Inputs::StringInput` already includes `Base`

tiagotex:
Somehow this got lost in my todo list 🙈

Rebased the PR and added the change log entry.

Thank you for the awesome work maintaining ActiveAdmin 💚

javierjulio:
@deivid-rodriguez @tiagotex I don't want to delay this if you'd like to get this merged in now but with the change I was wondering if we had this documented? I took a quick earlier and I didn't see anything. I bring it up as I've seen more of these type of changes come in which I think is great but wondered perhaps if its valuable to also [include an example in the docs](https://github.com/activeadmin/activeadmin/blob/master/docs/5-forms.md#datepicker)? Sorry I did mean to comment with this earlier today but forgot. 😞

deivid-rodriguez:
I agree we should document this :+1:. Could you add an example to the docs, @tiagotex?

tiagotex:
@javierjulio Thank you for reminding me about the docs.

I added the follow information to the datepicker section:

```
Datepicker also accepts the `:label` option as a string or proc to display.
If it's a proc, it will be called each time the datepicker is rendered.
```

PR #5422, opened 2018-06-05
Added option to tweak csv_builder batch sizes

DarkKnighth:
My first pull request ever. If I missed something please tell me! I haven't created an issue for this but since I already made a hack for it in a project I thought I should pull request it.

This is an update to the csv_builder. By default the csv_builder have a static batch_size of 1000 and always writes all data to file. I think this is a good default and have not changed the default behavior. In some cases though it might be a good idea to ensure that the feature does not impact performance of the server for the regular users while the admin user wants a csv file made. To enforce this changing the batch_size and adding a max_batch_page_count option so that a concise number of records cannot be exceeded and therefore only impacts the server in a predictive manner for a predictive amount of time.

Specifications:
max_batch_page_count:
  if negative, 0 or nil the default behavior of writing all batches
  otherwise the amount of batches made cannot exceed the number given
batch_size:
  if negative, 0 or nil the default of 1000 records
  otherwise the number given is the assigned as the size of the batches 

Edit: Updated specification

### DIFF
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -39,13 +39,22 @@ module ActiveAdmin
       @columns << Column.new(name, @resource, column_transitive_options.merge(options), block)
     end
 
+    def page_count
+      max_page_count = @options[:max_batch_page_count] || 0
+      batch_page_count = paginated_collection.total_pages
+      batch_page_count = max_page_count if max_page_count > 0 && batch_page_count > max_page_count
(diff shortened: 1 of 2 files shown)


DarkKnighth:
I don't know how to fix the "Cognitive Complexity" and still retain functionality.

DarkKnighth:
@varyonic  I left the tests in a separate commit to make them easily removable/editable.

PR #5428, opened 2018-06-15
Indicate how to actually show the comments area

rossmeissl:
I couldn't for the life of me get comments to show up until I looked through the code and found this method.


javierjulio:
We do have this documented but only in examples for the Show Pages doc. Adding it here is helpful. Thanks! I'll approve but let other maintainers chime in if they want the wording changed.

rossmeissl:
Thanks @javierjulio!

PR #5434, opened 2018-07-01
Update Policy template

NickWarm:
Pundit gem's policy template has replace ` Scope < Struct.new(:user, :scope)`  with `Scope class`, so I update ActiveAdmin's policy template.

ref:
- https://github.com/varvet/pundit#scopes
- [pundit's application policy template](https://github.com/varvet/pundit/blob/master/lib/generators/pundit/install/templates/application_policy.rb#L37-L48)

### DIFF
--- a/spec/support/templates/policies/active_admin/comment_policy.rb
+++ b/spec/support/templates/policies/active_admin/comment_policy.rb
@@ -1,6 +1,6 @@
 module ActiveAdmin
   class CommentPolicy < ApplicationPolicy
-    class Scope < Struct.new(:user, :scope)
+    class Scope < Scope
       def resolve
         scope
       end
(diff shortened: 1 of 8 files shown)


zorab47:
Looks good. And this seems backward compatible, right?

On a side note: These types of changes make me wonder if the authorization libraries should be extracted into separate gems.

varyonic:
Timo [suggested extraction before](#3571) and I would support such a move.

varyonic:
Scope was added [4 years ago](https://github.com/varvet/pundit/blame/66e49fa55b7eda82e850cfb7b2db2fce923b7bd3/lib/generators/pundit/install/templates/application_policy.rb#L37-L48), that would make it [pre-1.0](https://github.com/varvet/pundit/releases)

PR #5438, opened 2018-07-17
Update 5-forms.md

dlederle:
On the live docs site this line renders into the sidebar and is unreadable.


javierjulio:
That it does: https://activeadmin.info/5-forms.html#nested-resources. Thanks!

PR #5446, opened 2018-08-08
Fix resource filtering by association when the resource has custom primary key

wasifhossain:
## Scenario

Given a resource (`Product`) has a custom primary key (`:sku`) and it has a `belongs_to` association (`Brand`) which is available in the filter, if we filter products by a brand, https://github.com/activeadmin/activeadmin/pull/5223 (introduced in `1.2.0`) throws an error:

```
ActionView::Template::Error: PG::UndefinedColumn: ERROR: column brands.sku does not exist 
LINE 1: SELECT "brands".* FROM "brands" WHERE "brands"."sku" = '36'
```

This PR uses `predicate_association.association_primary_key` instead of `predicate_association.active_record_primary_key`, which returns the primary key of the association (Brand) instead of the resource (Product), that will generate the desired query:

```
SELECT "brands".* FROM "brands" WHERE "brands"."id" = '36'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
   [@chumakoff][]
 * Fixed `if:` scope option when a lambda is passed [#5501][] by [@deivid-rodriguez][]
 * Comment validation adding redundant errors when resource is missing [#5516][] by [@deivid-rodriguez][]
+* Fixed resource filtering by association when the resource has custom primary key [#5446][] by [@wasifhossain][]
 
 ## 1.3.0 [☰](https://github.com/activeadmin/activeadmin/compare/v1.2.1...v1.3.0)
 
@@ -297,6 +298,7 @@ Please check [0-6-stable][] for previous changes.
 [#5343]: https://github.com/activeadmin/activeadmin/pull/5343
 [#5399]: https://github.com/activeadmin/activeadmin/pull/5399
 [#5401]: https://github.com/activeadmin/activeadmin/pull/5401
+[#5446]: https://github.com/activeadmin/activeadmin/pull/5446
 [#5464]: https://github.com/activeadmin/activeadmin/pull/5464
 [#5501]: https://github.com/activeadmin/activeadmin/pull/5501
 [#5408]: https://github.com/activeadmin/activeadmin/pull/5408
@@ -337,5 +339,6 @@ Please check [0-6-stable][] for previous changes.
 [@timoschilling]: https://github.com/timoschilling
 [@TimPetricola]: https://github.com/TimPetricola
 [@varyonic]: https://github.com/varyonic
+[@wasifhossain]: https://github.com/wasifhossain
 [@wspurgin]: https://github.com/wspurgin
 [@zorab47]: https://github.com/zorab47
--- a/lib/active_admin/filters/active_filter.rb
+++ b/lib/active_admin/filters/active_filter.rb
@@ -102,7 +102,7 @@ module ActiveAdmin
 
       def related_primary_key
         if predicate_association
-          predicate_association.active_record_primary_key
+          predicate_association.association_primary_key
         elsif related_class
           related_class.primary_key
         end
--- a/spec/support/rails_template.rb
+++ b/spec/support/rails_template.rb
@@ -92,7 +92,7 @@ create_file 'app/models/category.rb', <<-RUBY.strip_heredoc, force: true
   end
 RUBY
 
-generate :model, 'store name:string'
+generate :model, 'store name:string user_id:integer'
 
 generate :model, 'tag name:string'
 create_file 'app/models/tag.rb', <<-RUBY.strip_heredoc, force: true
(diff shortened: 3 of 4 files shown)


lib/active_admin/views/components/paginated_collection.rb
```diff
@@ -102,8 +102,8 @@ def build_pagination
           # you pass in the :total_pages option. We issue a query to determine
           # if there is another page or not, but the limit/offset make this
           # query fast.
-          offset = collection.offset(collection.current_page * @per_page.to_i).limit(1).count
-          options[:total_pages] = collection.current_page + offset
+          another = collection.offset(collection.current_page * @per_page.to_i).any?
+          options[:total_pages] = collection.current_page + (another ? 1 : 0)
```
deivid-rodriguez: This changes are new. Are they intentional?

wasifhossain: @deivid-rodriguez I see. Actually I made that change some days ago in my local branch to fix another issue that was trivial. 

Let's assume we have a `collection` that contains `ActiveRecord::QueryMethods#group` method by dint of which, it returns a relation of `hash` instead of array-convertible relation. The old code (`limit(1).count`) could only count the size if its a collection of array, but return a hash when the collection query contains `group_by` instead of count. For that, we would need `limit(1).count.count`, which is not pretty either.

With the old code, it would return an error on line#106 (`options[:total_pages] = collection.current_page + offset`) when the collection query contains `group_by`:
```
TypeError (Hash can't be coerced into Integer)
```

So I rethought the process and found the whole point is to determine if there is another page or not, as commented on line#103. So I used a safe solution `.any?` and added `1/0` based on `true/false` that **does the same thing as the old code** but will also be **able to handle collections that contains grouped result**.

---

As it seems to be a trivial change, I thought to push it along with this PR. But if you suggest, I will remove this change from this PR.

I am sorry if it makes issue in any way.

Thanks!

deivid-rodriguez: I think this is a fix for https://github.com/activeadmin/activeadmin/issues/1123, which also has a PR for it already [opened](https://github.com/activeadmin/activeadmin/pull/5200), although your solution seems better to me at first sight?.

Let's remove it from this PR and handle it separately in another PR, ok?

deivid-rodriguez: Feel free to submit your solution as a separate PR. I think the biggest issue preventing that PR (or yours) from being merged is that we'd like a spec for it!

wasifhossain: Ah I see! will keep in mind to include the spec for that in a new PR :+1:

wasifhossain:
@deivid-rodriguez pardon me if I sound naive. I am eager to know when do the PRs usually get merged once **approved**. I think I need to rebase my branch again on top of the latest changes before the merge.

Thanks!

deivid-rodriguez:
@wasifhossain Once a second maintainer have a look at this PR and approves it, it will get merged. We decided a while ago that it's good to have a second pair of eyes on the changes we make :)

Feel free to rebase again this PR. Sometimes I ask contributors to do this, specially on very old PRs, but in this case I don't think it's strictly necessary.

wasifhossain:
thanks for insights :+1: Also I will keep an eye on any rebase request.

javierjulio:
@wasifhossain feel free to rebase if you can! I think it helps to keep a branch current because I've had problems (in other projects, not AA) where you merge something in but then the build fails. The issue though would have been caught in advance if the PR that was merged were rebased first and then force pushed to have CI confirm with a successful build here. That way it prevents any scrambling to fix whatever the mainline branch is (in our case its master).

wasifhossain:
just did it anyways! thanks @javierjulio  :)

javierjulio:
@wasifhossain yes I noticed because the checks refreshed just before I hit the "Comment" button.  Nice work! 👌🏻😏 Thanks for doing that. I think it's valuable and the right thing to do. Appreciate it!

deivid-rodriguez:
Maybe we could start using https://github.com/bors-ng/bors-ng. We use it for https://github.com/bundler/bundler and it's great. Not only because it ensures that PRs at tested against latest master _before_ they are merged, but because we no longer need to bother contributors with rebases!

javierjulio:
@deivid-rodriguez thanks for sharing! I didn't know about that tool. I will be looking into it this week and see if something we can use at work actually. If we can use that here I think that would be great and less hassle to our contributors! I've been bitten by that problem I described a few times in the past so its why I think its best to rebase first and confirm before merging.

javierjulio:
@deivid-rodriguez I'm not familiar with this area but I can approve if we can't get another review here. I'm going to request review from @Fivell and @zorab47 since they reviewed the related linked PR. Also they will likely have a better understanding of the changes here.

wasifhossain:
Hi @deivid-rodriguez, is there anything that I can do for this PR now :)

deivid-rodriguez:
No, let's get it in since @zorab47's approved and your answer to the backwards compatibility question is :+1:.

wasifhossain:
I really appreciate all of your valuable suggestions :+1: Thanks!

deivid-rodriguez:
Thanks to you for your contribution!

javierjulio:
@wasifhossain thank you for following up and seeing this through! We really appreciate your contribution. ❤️

PR #5451, opened 2018-08-14
Update method name to specify intended meaning

kazu9su:
I guess this method intended to specify adding `destroy` action item instead of `show` 🤔 
I'd appreciate if you would confirm this PR.

### DIFF
--- a/lib/active_admin/resource/action_items.rb
+++ b/lib/active_admin/resource/action_items.rb
@@ -58,7 +58,7 @@ module ActiveAdmin
       def add_default_action_items
         add_default_new_action_item
         add_default_edit_action_item
-        add_default_show_action_item
+        add_default_destroy_action_item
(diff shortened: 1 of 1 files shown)


varyonic:
I agree that @timoschilling may have used inconsistent naming when he extracted these in #3053. The change was supposed to be related to #3463 and #2625 but on review I don't understand if it was followed through. Given the method is private what is your interest in this change?

kazu9su:
Thanks to response me, @varyonic 
I just suppose whether it would be confusing for someone who will investigate or update this features because I felt a little bit weird when I found it.
But of course there is no necessity to update the private method name right now, so I close this PR if you think it is no need to do.

kazu9su:
OK, thanks again for considering this issue @varyonic ! ;)

PR #5452, opened 2018-08-15
Disable destroy buttons in has_many_form if false is passed as an option

omitter:
Don't create "remove buttons" on has_many forms when build_options gets passed `allow_destroy: false`, even on `.new_record?`s.
Rationale: You might have new records with children already populated and don't want the user to change their count.

### DIFF
--- a/lib/active_admin/form_builder.rb
+++ b/lib/active_admin/form_builder.rb
@@ -107,7 +107,7 @@ module ActiveAdmin
     end
 
     def has_many_actions(form_builder, contents)
-      if form_builder.object.new_record?
+      if form_builder.object.new_record? && builder_options[:allow_destroy] != false
         contents << template.content_tag(:li) do
           remove_text = remove_record.is_a?(String) ? remove_record : I18n.t("active_admin.has_many_remove")
           template.link_to remove_text, "#", class: "button has_many_remove"


varyonic:
This is failing in Travis, are you working on it?

omitter:
Sorry, I committed and thought all would be fine, I couldn't have been more wrong. I'm sorry, but my patch is currently incorrect. The problem is, the file in question differs greatly between master and 1.3.1. I tested the change on 1.3.x and it works there for me, but I don't know how to apply it to master, which's `has_many_actions` method changed drastically (different arguments). I don't understand why this is so. Can I somehow send you a patch for 1.3.1? I'm sorry to ask, but I don't know how to do this (and didn't find it in the contributing guide).

varyonic:
The changes are probably due to #5196. No, I am not accepting patches to 1.3.1.

varyonic:
master (2.alpha) has been branched since 1.1 in July 2017.

omitter:
Okay, I see. Then, when is 2.0 going to be released? If not any time soon, why would you not accept a patch for 1.x? If created the correct commit here [branch: disableable_destroy_button](https://github.com/omitter/activeadmin/tree/disableable_destroy_button). I could make a pull request for 1-3-stable or 1-4-stable if you're fine this it.

PR #5458, opened 2018-08-23
I18n for Comments

mauriciopasquier:
I wanted to have a default installation (i.e. with comments enabled) translated by default, so I added what was missing.

This repetition:

        comment: "Comment"
        active_admin/comment: "Comment"

is needed because `comments_registration_name` is `Comment` so it looks it up without namespace for title and menu translation. The rest is used by filters and scopes generated in the index.

I added [hints](https://github.com/glebm/i18n-tasks#fine-tuning) for **i18n-tasks** about those ActiveRecord-related strings in the comments file so it can report them as missing when using `i18n-tasks missing`.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,6 +1,20 @@
 en:
+  activerecord:
+    models:
+      comment: "Comment"
+      active_admin/comment:
+        one: "Comment"
+        other: "Comments"
+    attributes:
+      active_admin/comment:
+        author_type: "Author type"
+        body: "Body"
+        created_at: "Created"
+        namespace: "Namespace"
+        resource_type: "Resource type"
+        updated_at: "Updated"
   active_admin:
-    dashboard: Dashboard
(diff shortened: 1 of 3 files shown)


mauriciopasquier:
Sorry, I fixed-up and rebased, I had missed some errors before.

mauriciopasquier:
> Always keep quoted values in our YAML files. Not very important, but would be nice.

I noticed that style, that's why I changed `active_admin.dashboard` although it wasn't related to the commit per se. But I'm not sure how to enforce this at code level.

> Make sure we update the i18n hints list when/if new columns are added to the comments table. This happens very rarely, so it does not seem like a big deal.

I was hesitant to introduce hints for that reason but I think the pros are enough :)

deivid-rodriguez:
Thanks so much @mauriciopasquier! :heart:

mauriciopasquier:
Thank you all for this gem :)

PR #5462, opened 2018-08-30
Remove codeclimate

deivid-rodriguez:
Hello!

I was reviewing #5460 and we're having trouble getting codeclimate enabled, it just won't trigger. Since a while I noticed that I usually only want to use codeclimate to run linters for me, but that ends up being more complicated than running the linters directly in CI. I find this solution better because:

* It's more flexible. The rubocop version that will be used will be the version in your `Gemfile`. With the previous solution you need to manually make sure that the rubocop version in your `Gemfile` matches the `Rubocop` version used by Codeclimate. Letting those get unsyncronized is confusing because you can get failures in CI which you can't see locally.

* We don't depend on a separate external service that needs to be configured, can have issues, and so on.

The downside is that moving the checks to the CI requires a bit a ruby code. I don't think it's a big deal but I understand if someone does not want to maintain this code.

Let me know what you think!


deivid-rodriguez:
If we agree with this, `codeclimate` will just be disconnected, so the current codeclimate failures can be ignored.

deivid-rodriguez:
I think this is a good step in your preferred direction then, since we completely get rid of code metrics, and regarding linters, we keep what we currently had, but with less surprises.

deivid-rodriguez:
Great, let's do it then!

deivid-rodriguez:
I disconnected codeclimate and will follow up with a backport PR to the latest stable branch removing it too.

PR #5464, opened 2018-09-03
Stop passing Proc as sort_key

chumakoff:
**First**

https://github.com/activeadmin/activeadmin/blob/ac5481c9bc2db03bb5bfb6516c1b2fb1fe1b710c/lib/active_admin/views/components/table_for.rb#L179
```
        # If you pass a block to be rendered for this column, the column
        # will not be sortable unless you pass a string to sortable to
        # sort the column on:
        #
        #   column('Username', sortable: 'login'){ @user.pretty_name }
        #   # => Sort key will be 'login'
        #
        def sort_key
```

This hasn't been true since https://github.com/activeadmin/activeadmin/commit/b9a84532b04880319670e8bd97f65b57806cda17

Confirmed by this test: https://github.com/activeadmin/activeadmin/blob/ac5481c9bc2db03bb5bfb6516c1b2fb1fe1b710c/spec/unit/views/components/table_for_spec.rb#L376

I removed the comment, as `:sort_key` resolution now doesn't depend on whether the block is passed to the `:column` method or not.

**Second**

There is a related problem. If a proc is passed to the `:column` method and `:sortable` option is boolean, then `:sort_key` is set to the string representation of the proc (something like `"Proc:0x00007fb3c40c8080@/home/user_name/app_folder/app/admin/model_name.rb:99_method_name"`). 

So, the same value is present in the url as the `order` parameter value:

`http://host/admin/model_name?order=Proc:0x00007fb3c40c8080@/home/user_name/app_folder/app/admin/model_name.rb:99_method_name`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Bug fixes
 
+* Fixed the string representation of the resolved `sort_key` when no explicit `sortable` attribute is passed [#5464][] by [@chumakoff][]
+* Fixed docs on the column `sortable` attribute (which actually doesn't have to be explicitly specified when a block is passed to column) [#5464][] by [@chumakoff][]
+
 ## 1.3.0 [☰](https://github.com/activeadmin/activeadmin/compare/v1.2.1...v1.3.0)
 
 ### Enhancements
--- a/docs/3-index-pages/index-as-table.md
+++ b/docs/3-index-pages/index-as-table.md
@@ -131,9 +131,7 @@ When a column is generated from an Active Record attribute, the table is
 sortable by default. If you are creating a custom column, you may need to give
 Active Admin a hint for how to sort the table.
 
-If a column is defined using a block, you must pass the key to turn on sorting. The key
-is the attribute which gets used to sort objects using Active Record.
-
+You can pass the key specifying the attribute which gets used to sort objects using Active Record.
 By default, this is the column on the resource's table that the attribute corresponds to.
 Otherwise, any attribute that the resource collection responds to can be used.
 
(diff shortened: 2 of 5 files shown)


deivid-rodriguez:
@chumakoff It looks great to me, I would just add a couple of changelog entries to the things being fixed here.

chumakoff:
@deivid-rodriguez Done. Also updated the documentation.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Bug fixes
 
+* Prevent passing Proc as order (sort_key) parameter [#5464][] by [@chumakoff][]
```
deivid-rodriguez: I find this description a bit confusing. This PR is not preventing a block being passed anywhere, right? What's doing is:
* Fixing the string representation of the resolved `sort_key` when no explicit `sortable` attribute is passed.
* Fixing docs on how the `sort_key` is chosen (which actually ignores the block passed to `column`).

If this is correct, can we clarify it?

deivid-rodriguez: Thanks for updating the docs as well, by the way! :heart:

chumakoff: I am fine with your description. Should I add it to the changelog as it is (two items)? Do you think fixing docs is worth mentioning in the changelog?

deivid-rodriguez: I have the same doubt, but I guess it doesn't hurt, so I'd add both, yeah.

chumakoff: Let's then formulate the second change like this:
* Fixed docs on the column `sortable` attribute (which is actually doesn't have to be explicitly specified when a block is passed to column)

Issue #5457, opened 2018-08-23
Are Ransack's features like `*_cont_all` or `*_cont_any` supported?


kewubenduben:
Are Ransack's features like `*_cont_all` or `*_cont_any` supported?

I'm trying to look for a filter that uses `contains all` for multiple keyword.

If this is possible in any way, can someone point out on how to do this?

Thanks

jzisser9:
I was just looking for support for `i_cont` as well. AA doesn't seem to support it yet. I might take a crack at adding support for it.

kewubenduben:
> I was just looking for support for `i_cont` as well. AA doesn't seem to support it yet. I might take a crack at adding support for it.

Check out my PR#5465

It wasn't merge though due to not having a spec.

I'm currently using it at the moment

PR #5465, opened 2018-09-04
Support Ransacks _all and _any requests

kewubenduben:


kewubenduben:
I'm kinda new into writing test code. I'm not sure on where to start.

Anyway, I'll try to write a test.  And any help would be appreciated.

varyonic:
Related #5457

deivid-rodriguez:
@kewubenduben Can you explain a bit how you plan to use this feature? From looking at the implementation, I'm guessing you want to do something like

```
filter :name, filters: [:cont_any]
```

and then the user specifies in the text box each of the keywords she wants to match separated by spaces?

kewubenduben:
@deivid-rodriguez 

Yes correct, I wanted the keyword search separated by spaces.

I've had this situation where I want to search from `first_name` and `last_name` from a single field.

```
filter :full_name_query_cont_any, as: :string, multiple: true, label: "Full Name"
```
```
# User Model
  ransacker :full_name_query do |parent|
    Arel::Nodes::NamedFunction.new('LOWER',
      [Arel::Nodes::NamedFunction.new('concat_ws',
        [Arel::Nodes::SqlLiteral.new("' '"), parent.table[:first_name], parent.table[:last_name]])])
  end
```

This is for example.

first_name = 'Black Hawk'
last_name = 'Down Under'

Searching for `black under` would return any of the names(first and last) that contains `black` or `under`. Instead of only matching for `black under`.

I understand that this works with `tags` too, but `tags` look a bit ugly on doing this.

PR #5474, opened 2018-09-12
Changelog fixes

deivid-rodriguez:
I noticed a bunch of broken bad links in our changelog, which makes it harder to review the changes being introduced. Also, I noticed some inconsistencies in punctuation (see https://github.com/activeadmin/activeadmin/pull/5472#discussion_r217077696). This PR fixes both things and introduces some specs so that these problems are never reintroduced again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 #### Major
 
-* Rails 5.2 support [#5343][] by [@varyonic][], [#5399][], [#5401][] by [@zorab47][].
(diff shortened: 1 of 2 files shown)


javierjulio:
@deivid-rodriguez nice! I dig it. 👍🏻 Thanks.

deivid-rodriguez:
Kudos to the rubocop project, since that's where I got these ideas and the base code from this checks from :)

javierjulio:
@deivid-rodriguez great! This is quite handy. I'll remember this for other projects. Thanks!

PR #5475, opened 2018-09-12
Remove chromedriver-helper dependency

deivid-rodriguez:
This gem currently has a hard to debug [issue](https://github.com/flavorjones/chromedriver-helper/issues/57), which makes installing it break _other_ projects's specs which use `chromedriver` but don't use the `chromedriver-helper` gem.

Also, we're pinning it already because of other issues.

And finally, we sometimes get some weird spec failures when `chromedriver` fails to start, so who knows if this will also fix that.

This gem does very little (download a binary and put it in your path), I don't think it's worth using it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ bundler_args: --without development
 cache: bundler
 
 before_script:
+  - bin/install_chromedriver.sh
   - unset _JAVA_OPTIONS
 
 before_install:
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
> And finally, we sometimes get some weird spec failures when chromedriver fails to start, so who knows if this will also fix that.

Confirmed, this PR does not fix that. :disappointed: https://travis-ci.org/activeadmin/activeadmin/jobs/427946918

deivid-rodriguez:
@varyonic I also disabled email notifications, but didn't mention it in the PR. Just to confirm that you saw the change and are ok with it.

PR #5483, opened 2018-09-20
Fix polymorphic path generation for a renamed resource with a belongs_to association

### LINKED ISSUE #3647: optional belongs_to for renamed active_admin resources creates incorrect url helpers

``` ruby
  navigation_menu :default
```

```
```

ndbroadbent:
Fixes #3647

I updated `features/renamed_resource.feature` to use an optional `belongs_to`, which caused an error:

```
  ActiveAdmin.register User
  ActiveAdmin.register Blog::Post, as: 'Post' do
    belongs_to :user, optional: true
    ...
```

```
$ cucumber features/renamed_resource.feature
Using the default profile...
....F----------

(::) failed steps (::)

undefined method `new_admin_blog_post_path' for #<Admin::PostsController:0x00007fed11c57960>
Did you mean?  new_admin_post_path
               new_admin_user_post_path (ActionView::Template::Error)
./lib/active_admin/resource/action_items.rb:69:in `block in add_default_new_action_item'
./lib/active_admin/views/action_items.rb:9:in `instance_exec'
./lib/active_admin/views/action_items.rb:9:in `block (2 levels) in build'
./lib/active_admin/views/action_items.rb:8:in `block in build'
./lib/active_admin/views/action_items.rb:7:in `each'
./lib/active_admin/views/action_items.rb:7:in `build'
./lib/active_admin/views/title_bar.rb:50:in `build_action_items'
```

This PR fixes that error.

I tried a few different approaches, but I think this is the cleanest way to solve it, even if it involves monkey-patching some model instances. I couldn't see any other way to override `#model_name` while only affecting the ActiveAdmin context (and not breaking the rest of the Rails application.)

Another solution might be to completely fork the `#polymorphic_path` and `#polymorphic_url` methods, and just have some custom code for ActiveAdmin, instead of calling out to the Rails helpers. But I think a small monkey-patch is probably better than copying lots of code and potentially getting out of sync with Rails.  There might also be some changes in `inherited_resources` that could make this easier. But anyway, I think this PR should work for now.


ndbroadbent:
I came back to this a bit later, and wanted to see if there might be a better way to solve it. I read through the source of [actionpack-5.2.0/lib/action_dispatch/routing/polymorphic_routes.rb](https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/actionpack/lib/action_dispatch/routing/polymorphic_routes.rb), and I'm finally understanding how it all fits together. `@key_strategy` is a Proc defined [here](https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/actionpack/lib/action_dispatch/routing/polymorphic_routes.rb#L198-L204), and it just calls `singular_route_key` and `route_key` on the `model_name`.

I tried rewriting the `polymorphic_*` calls to use plain strings (e.g. `["admin", "posts", 1]`), but I wasn't able to get that to work. You can't pass plain integer ids to the `polymorphic_*` helpers, only model instances or classes.

I also tried to see if I could monkey-patch a small part of the `ActionDispatch::Routing::PolymorphicRoutes::HelperMethodBuilder` class, but I don't think it can be done. There's just a lot of very long methods, and it's not written in a very extensible way. I couldn't see any way to pass in a new `:model_name` option without copy/pasting large chunks of the code. So that's what I ended up doing.

I decided that it would probably be better if ActiveAdmin just used it's own custom copy of the `PolymorphicRoutes` code. I've duplicated it and created a `ActiveAdminHelperMethodBuilder` class. It turned out to be a lot of work to actually rewrite the method to support `resource_name` and `resource_class`, because we only want to use the `resource_name` for the actual model class, and need to defer to `#model_name` for other models to support `#belongs_to` associations.

But I think this is the right way to do it, and it doesn't involve any monkey patching. I've pushed this as a separate commit, so you can take a look at both approaches and let me know which one I should squash.

ndbroadbent:
Sorry, didn't see the rubocop errors. Fixed now

varyonic:
For reference `polymorphic_mapping` was implemented in rails/rails#23138 for Rails 5.1.

ndbroadbent:
@varyonic Whoops just saw the broken tests for Rails 5.0 and 4.2. Sorry about that, not sure why I'm not getting email notifications from Travis CI.

I'll look into fixing it for 5.0 and 4.2 as well.

varyonic:
Related activeadmin/inherited_resources#455

deivid-rodriguez:
@ndbroadbent So sorry I didn't get to this earlier. A PR with a very similar code to the second approach used here was opened in #5823, but it didn't mention this PR and I had forgotten about it, so I reviewed without looking at both of the commits in here and your helpful insights.

I ended up creating #5826, which turns out to be similar to the _first_ approach you used. I feel doing something like that is better than duplicating a lot of Rails code and having to make sure it stays in sync. It would also introduce a lot of code that our specs are not covering, so it would be hard to know how much of it was actually needed or not.

If you have time, it would be great if you could have a look at #5826 and tell me what you think. I understand though that it's been a while and you might no longer be interested or have no time.

Before merging #5826 I will squash the commits, and add you as an author to the squashed commit to recognize your attribution.

Thanks so much for fixing this longstanding issue. I'll close this in favor of #5826, let me know if that's ok too.

PR #5485, opened 2018-09-21
Use 'pry-byebug' instead of just 'pry', so that you can step into/over code with 's' and 'n'

ndbroadbent:
Pry is great but I was really missing the step-by-step debugging commands from [pry-byebug](https://github.com/deivid-rodriguez/pry-byebug)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem 'rake'
 gem 'parallel_tests'
 
 # Debugging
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
We should've specified `platform: :mri` here, since this gem is only mri-compatible, I'm adding that in #5487 as a follow up!

PR #5486, opened 2018-09-21
Fixes #5235: removes passwords from default csv export

### LINKED ISSUE #5235: CSV includes sensitive attributes by default
When defining a resource for the Devise `User` class, Active Admin by default includes sensitive data like encrypted passwords in the CSV output accessible via the download links.

While Devise [itself takes precautions](https://github.com/plataformatec/devise/blob/657507b705fa4cad269c3c06460cd836743e1297/lib/devise/models/authenticatable.rb#L104) to exclude sensitive data from JSON and XML serializations, the default CSV builder includes all attributes.

To avoid exposing the data, one needs to add a `csv` block to the resource definition.

While I understand, that this more or less also applies to the HTML view of the default index table, I have found the download endpoints to be far easier to miss.

chrp:
AA comes with devise which in turn by default saves a bcrypt-hashed password in a column encrypted_password. So it's actually fairly easy to leak all pw hashes via the default CSV export. This is a pragmatic way to avoid this pitfall.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Enhancements
 
+### Security Fixes
+
+* Prevent leaking hashed passwords via user CSV export and adds a config option for sensitive attributes [#5486][] by [@chrp][]
+
 #### Minor
 
 * Allow proc label in datepicker input [#5408][] by [@tiagotex][]
@@ -310,6 +314,7 @@ Please check [0-6-stable][] for previous changes.
 [#5501]: https://github.com/activeadmin/activeadmin/pull/5501
 [#5408]: https://github.com/activeadmin/activeadmin/pull/5408
 [#5516]: https://github.com/activeadmin/activeadmin/pull/5516
+[#5486]: https://github.com/activeadmin/activeadmin/pull/5486
 
 [@5t111111]: https://github.com/5t111111
 [@aarek]: https://github.com/aarek
@@ -350,3 +355,4 @@ Please check [0-6-stable][] for previous changes.
 [@wasifhossain]: https://github.com/wasifhossain
 [@wspurgin]: https://github.com/wspurgin
 [@zorab47]: https://github.com/zorab47
+[@chrp]: https://github.com/chrp
--- a/features/filter_attributes.feature
+++ b/features/filter_attributes.feature
@@ -1,0 +1,46 @@
+Feature: Filter Attributes
+
+  Filtering sensitive attributes
+
+  Background:
+    Given a configuration of:
+    """
+      ActiveAdmin.register User
+    """
+    Given I am logged in
+    And a user named "John Doe" exists
+    And I am on the index page for users
+
+  Scenario: Default index page
+    Then I should not see "Encrypted"
+    But I should see "Age"
+
+  Scenario: Default new form
+    Given I follow "New User"
+    Then I should not see "Encrypted"
+    But I should see "Age"
+
+  Scenario: Default edit form
+    Given I follow "Edit"
+    Then I should not see "Encrypted"
+    But I should see "Age"
+
+  Scenario: Default show page
+    Given I follow "View"
+    Then I should not see "Encrypted"
+    But I should see "Age"
+
+  Scenario: Default CSV export
+    Given I follow "CSV"
+    Then I should not see "Encrypted"
+    But I should see "Age"
+
+  # TODO: JSON
+  # Scenario: Default JSON
+  #   Given I follow "JSON"
+  #   Then I should not see "encrypted"
+  #   But I should see "age"
+
+  Scenario: Default XML
+    Given I follow "XML"
+    Then I should not see "encrypted"
--- a/features/support/paths.rb
+++ b/features/support/paths.rb
@@ -14,6 +14,8 @@ module NavigationHelpers
       "/admin"
     when /the new post page/
       "/admin/posts/new"
+    when /the new user page/
+      "/admin/users/new"
     when /the login page/
       "/admin/login"
     when /the first post custom status page/
--- a/lib/active_admin/application_settings.rb
+++ b/lib/active_admin/application_settings.rb
@@ -35,5 +35,8 @@ module ActiveAdmin
 
     # To make debugging easier, by default don't stream in development
     register :disable_streaming_in, ['development']
+
+    # Remove sensitive attributes from being displayed, made editable, or exported by default
+    register :filter_attributes, [:encrypted_password, :password, :password_confirmation]
   end
 end
--- a/lib/active_admin/resource/attributes.rb
+++ b/lib/active_admin/resource/attributes.rb
@@ -24,7 +24,7 @@ module ActiveAdmin
       end
 
       def reject_col?(c)
-        primary_col?(c) || sti_col?(c) || counter_cache_col?(c)
+        primary_col?(c) || sti_col?(c) || counter_cache_col?(c) || filtered_col?(c)
       end
 
       def primary_col?(c)
@@ -39,6 +39,9 @@ module ActiveAdmin
         c.name.end_with?('_count')
       end
 
+      def filtered_col?(c)
+        ActiveAdmin.application.filter_attributes.include?(c.name.to_sym)
+      end
     end
   end
 end
--- a/lib/active_admin/views/components/active_admin_form.rb
+++ b/lib/active_admin/views/components/active_admin_form.rb
@@ -57,6 +57,12 @@ module ActiveAdmin
           end
           insert_tag(SemanticInputsProxy, form_builder, *args, &wrapped_block)
         else
+          # Set except option to prevent sensitive fields from being shown in forms by default.
+          opts = args.extract_options!
+          opts[:except] ||= []
+          ActiveAdmin.application.filter_attributes.each { |e| opts[:except] << e }
+          args << opts
+
           proxy_call_to_form(:inputs, *args, &block)
         end
       end
--- a/lib/generators/active_admin/install/templates/active_admin.rb.erb
+++ b/lib/generators/active_admin/install/templates/active_admin.rb.erb
@@ -152,6 +152,13 @@ ActiveAdmin.setup do |config|
   #
   # config.before_action :do_something_awesome
 
+  # == Attribute Filters
+  #
+  # You can exclude possibly sensitive model attributes from being displayed,
+  # added to forms, or exported by default by ActiveAdmin
+  #
+  config.filter_attributes = [:encrypted_password, :password, :password_confirmation]
+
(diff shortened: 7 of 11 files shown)


javierjulio:
Nice, you're right. This is a good idea. I wonder if it would be better though to define the field names that should be excluded in config instead? Perhaps model it after Rails' filter_parameter_logging initializer? That way ActiveAdmin can provide a sensible default for its common setup but easily allow users to specify additional fields.

chrp:
Alright, let me give it a try. And since I am at it, would you think it makes sense to also change ``ActiveAdmin::Resource::Attributes.default_attributes`` in order to prevent leaking hashes on show pages?

javierjulio:
@chrp I'm not familiar with that. Do you have an existing issue or seen a related one mention that at all? If not I'd suggest creating one as one of the other maintainers may be familiar and able to offer help/advice on what to do.

deivid-rodriguez:
If we make this a config, I think it makes sense to filter out whatever is in there from `ActiveAdmin::Resource::Attributes.default_attributes` too :+1:

varyonic:
Referencing #5235

chrp:
I'll be working on this next week & will at least make this PR green but hopefully also add the suggested config option.

chrp:
I've added a config option ``filter_attributes`` which influences the ``default_attributes`` of a resource. Removed those attributes from default forms, too (which also makes cucumber tests green).

javierjulio:
@chrp thanks for following up! It looks good so far. 👍🏻 Let's see how the tests do.

lib/active_admin/form_builder.rb
```diff
@@ -11,6 +11,16 @@ def input_wrapping(&block)
...
 
+module Formtastic
+  class FormBuilder
+    # Prevent sensitive fields from being shown in forms by default
+    # NOTE: Formastic::Helpers::InputsHelper.inputs calls
+    #   ``Formtastic::FormBuilder.skipped_columns`` explicitly so we need to
+    #   monkeypatch instead of use inheritance
+    self.skipped_columns += ActiveAdmin.application.filter_attributes
+  end
+end
```
deivid-rodriguez: Do you think this is something that we should fix (or at least report to) formtastic? Also, could you add a spec for the monkeypatch?

chrp: I've opened a [PR for formtastic](https://github.com/justinfrench/formtastic/pull/1273). Not sure wether there is much support for subclassing from FormBuilder from their side because they explicitly call Formtastic::FormBuilder in various other places, too.

chrp:
I've found a way to avoid a monkeypatch completely. Let me wait for test coverage report to add some more specs if necessary.

deivid-rodriguez:
Sorry @chrp, I added some checks so that we don't introduce broken links in the changelog, you need to add the full link to this PR at the bottom of the changelog. This is quite counterintuitive right now, same problem happened to myself yesterday on #5501... :sweat_smile:. I think we could consider switching to writing the links "inline", but for now this is how it's working, and it's great that we don't get broken links in the changelog :)

chrp:
Is there any trick regarding the line length in changelog?

chrp:
While building the tests I figured that the default JSON formatter would export ``encrypted_password`` as well. What is the supposed way to customize it? More specific where would I inject defaults which still can be overridden by a user?

deivid-rodriguez:
@chrp Not sure... I would've expected that `resource_attributes` would be used in there too? :thinking:

@Fivell maybe knows since I recall he did some work on the different formatters?

chrp:
> @chrp Not sure... I would've expected that `resource_attributes` would be used in there too? 🤔

Not really. It's using inherited_resources' default controller behavior, see [here](https://github.com/activeadmin/activeadmin/blob/be2f19547b1fcfb6d1be2b9503cda94ed9b5f795/lib/active_admin/resource_controller.rb#L15). You can inject there (as for example the authorization layer does) but I am not sure were to inject a new default behavior.

deivid-rodriguez:
Let me have a look...

deivid-rodriguez:
So, I think regarding json the problem is that nil attributes seem to be ommited by default. That seems.. fine? Not sure.

I did this and the spec passed

```diff
diff --git a/features/filter_attributes.feature b/features/filter_attributes.feature
index 904830bb..c375df98 100644
--- a/features/filter_attributes.feature
+++ b/features/filter_attributes.feature
@@ -8,7 +8,7 @@ Feature: Filter Attributes
       ActiveAdmin.register User
     """
     Given I am logged in
-    And a user named "John Doe" exists
+    And a user named "John Doe" of age 47 exists
     And I am on the index page for users
 
   Scenario: Default index page
@@ -35,11 +35,10 @@ Feature: Filter Attributes
     Then I should not see "Encrypted"
     But I should see "Age"
 
-  # TODO: JSON
-  # Scenario: Default JSON
-  #   Given I follow "JSON"
-  #   Then I should not see "encrypted"
-  #   But I should see "age:"
+  Scenario: Default JSON
+    Given I follow "JSON"
+    Then I should not see "encrypted"
+    But /I should see ""Age":/
 
   # TODO: The default XML export is useless
   # Scenario: Default XML
diff --git a/features/step_definitions/factory_steps.rb b/features/step_definitions/factory_steps.rb
index c07d2c1d..95bee66a 100644
--- a/features/step_definitions/factory_steps.rb
+++ b/features/step_definitions/factory_steps.rb
@@ -23,6 +23,10 @@ Given /^a (user|publisher) named "([^"]*)" exists$/ do |type, name|
   create_user name, type
 end
 
+Given /^a (user|publisher) named "([^"]*)" of age (\d+) exists$/ do |type, name, age|
+  create_user(name, type).update!(age: age)
+end
+
 Given /^a store named "([^"]*)" exists$/ do |name|
   Store.create! name: name
 end
```

Regarding XML, I think the problem is that the test app does not define any xml serializer for the User model, so it just doesn't work. I guess we can omit that test?

deivid-rodriguez:
By the way I tried creating a test application with this branch (`bundle exec rake local s`), and I got

```
(...)
        rake    db:seed
rake aborted!
NoMethodError: undefined method `random' for String:Class
/home/deivid/Code/activeadmin/.test-rails-apps/rails-5.2.1/db/seeds.rb:16:in `block in <main>'
/home/deivid/Code/activeadmin/.test-rails-apps/rails-5.2.1/db/seeds.rb:10:in `collect'
/home/deivid/Code/activeadmin/.test-rails-apps/rails-5.2.1/db/seeds.rb:10:in `<main>'
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/bootsnap-1.3.2/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:50:in `load'
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/bootsnap-1.3.2/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:50:in `load'
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:281:in `block in load'
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:253:in `load_dependency'
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:281:in `load'
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/railties-5.2.1/lib/rails/engine.rb:551:in `load_seed'
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/tasks/database_tasks.rb:281:in `load_seed'
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/railties/
```

chrp:
@deivid-rodriguez I am having a bit of a hard time to do something about the codecov target. I guess we could exclude /features/support/*.rb from codecov?

chrp:
> By the way I tried creating a test application with this branch (`bundle exec rake local s`), and I got
> ...

I just rebased my work on current master and now get the same issue so this must have been introduced with the change of how the test app is built.

deivid-rodriguez:
@chrp Don't worry about codecov, it's noisy but we use it just as an orientation not as a hard requeriment (I got curious thought how those lines are uncovered, usually uncovered lines on test files mean that those lines are dead code... :thinking:, maybe is just codecov acting weirdly).

chrp:
Alright, I found a simpler way to exclude filtered attributes from forms which avoids the buggy skipped_columns of Formtastic. I've added a spec for XML nonetheless, as it will fail as soon as someone adds a proper XML exporter that would not check for filtered attributes. 

For JSON I am a bit at a loss. The way to do it would probably be to build a JsonBuilder similar to CsvBuilder which makes the output a bit more controllable and then use it in ResourceController::Streaming where there is already a comment hinting at this option. 

As this is quite a lot of work and out of scope for this PR my solution would be to disable JSON export by default and leave a comment in the initializer template to make sure that people who enable JSON downloads know about this potential risk. 

What do you think @deivid-rodriguez @Fivell ?

deivid-rodriguez:
@chrp Yeah, that'd be great!

@Fivell If you get some time, it'd be great if you were able to double check this PR, since you've worked on this export links before :)

javierjulio:
@chrp thanks a lot for following through on this, it's great work. 👍🏻 Thanks @deivid-rodriguez too for helping out here. I'll let @Fivell chime in if he can.

chrp:
Rebased, squashed and I've added a suggestion to disable JSON exports by default as they cannot be protected by filter_attributes by default but are as unsafe as CSV exports.

deivid-rodriguez:
I say let's get the bug fix in first, and discuss disabling JSON exports separately.

I think we can backport and release this bug fix right now, but disabling the exports would probably be better done for the next major version.

chrp:
I would consider the JSON exporting a encrypted_password a part of the bug. If we only close the CSV loophole we just go half the way. Happy to revert the JSON-related commit, though if you feel it is a bit too radical. I just did not find any more subtle solution that is quick to implement.

deivid-rodriguez:
I see your point, and I admit we're leaving an open issue here. I just feel, as you say, it's too radical and doesn't really solve the problem. People currently using JSON exports will just reenable them, and will be still suffering from the issue after doing that.

chrp:
Alright, I've removed the commit disabling JSON-exports by default and now this should be ready to merge.

chrp:
@deivid-rodriguez Thank you! Please to contribute after having been using AA for 5 years.

javierjulio:
Great job! Thanks again for contributing! 👍🏻❤️

PR #5488, opened 2018-09-21
Build only protected branches

deivid-rodriguez:
To save CI resources and not have to wait so much for PRs to get green.

We briefly enabled this kind of setting but decided to remove it again in https://github.com/activeadmin/activeadmin/pull/5126. The reason for reverting it was that it is what Rails does. However, this is not actually true, as can be seen by checking a random Rails PR and observing it only has a single TravisCI check (continuous-integration/travis-ci/pr). It _is_ true that they don't have this setting in the TravisCI config, but they probably have it configured in TravisCI UI.

I propose to restore the old way, because it saves quite a bit of CI time.

The only gotcha of this approach that I know of is that in some case you want a feature branch tested on TravisCI without opening a PR for it, you need to temporarily add the branch to the CI config. In my opinion, the regular time savings are worth this little inconvenience!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ addons:
     packages:
       - google-chrome-stable
 
+branches:
+  only:
+    - master
+    - 1-4-stable
+
 language: ruby
 
 sudo: false


deivid-rodriguez:
@javierjulio Yes, normally you just open a PR to kick CI in but sometimes it happens to me that I just want to run some experiments that are not necessarily meant to become a PR. In that case, I temporarily add a commit to add the branch I'm working on to `.travis.yml` to kick CI in.

javierjulio:
@deivid-rodriguez that sounds good, I just think it would be better that you (and others in general) open a PR anyway! I do that often even with "experiments". Show your work and be proud of it! Its helpful to you and others because likely someone will come across it that can either help you out or has an interest in it. No harm with an additional PR. It can always be labeled as such (experimental, on hold, in progress, etc.) and if anything, closed.

javierjulio:
Ohhh @deivid-rodriguez... 🙄😂 You don't have to say much at all. You can simply state that is an experiment and at minimum what you are trying to accomplish. You can also add notes and resources. I do this as it shares not just the code/experiment but the material too openly so others can pitch in and comment if there is any interest. It really doesn't hurt. Consider it next time an experiment or idea comes up!

deivid-rodriguez:
I will consider it, specially now that branches don't build by themselves :stuck_out_tongue_winking_eye:. Thanks!

PR #5492, opened 2018-09-26
Remove unnecessary requires

deivid-rodriguez:
The helper is already required through the `.rspec` file. Their presence is unlikely to be a problem but can be annoying in some situations. For example, if you introduce a syntax error in the helper, you will get it printed hunders of times, one per file, since the require never succeeds and it's retried everytime a new spec file is loaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,6 @@ spec/rails
 *.sqlite3-journal
 .test-rails-apps
 public
-.rspec
 .rspec_failures
 .rails-version
 .rbenv-version
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,1 @@
+--require rails_helper
--- a/.simplecov
+++ b/.simplecov
@@ -1,6 +1,3 @@
-# TODO: Always run simplecov again once
-# https://github.com/colszowka/simplecov/issues/404,
-# https://github.com/glebm/i18n-tasks/issues/221 are fixed
 SimpleCov.start do
   add_filter 'spec/rails/'
 end
--- a/spec/bug_report_templates_spec.rb
+++ b/spec/bug_report_templates_spec.rb
@@ -1,5 +1,3 @@
-require 'spec_helper'
-
 RSpec.describe 'bug_report_templates' do
   subject do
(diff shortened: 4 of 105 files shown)


deivid-rodriguez:
Turns out that I had a local `.rspec` file in my working copy, but this file was not in the repo. I think it's good to add it for the reasons above!

deivid-rodriguez:
There was a codecov failure. It was due to the changes in this PR, but it was within our "allowed decrease threshold" as specified in `codecov.yml`, so I don't understand why the PR was marked as failed. So,

* I contacted `codecov` to ask about it.
* I fixed the issue, that was due to the fact that after these changes, `rails_helper` was being required before `simplecov`, not after it as it should.

features/support/env.rb
```diff
@@ -6,7 +6,7 @@
 
 ENV['RAILS_ENV'] = 'test'
 
-require 'simplecov'
+require 'simplecov' if ENV["COVERAGE"] == "true"
```
varyonic: I think it is OK to delete or squash branch commits instead of reverting them.

deivid-rodriguez: Yes, absolutely! But I prefer to add separate commits initially to keep track of the changes I make. Once I have a final working version, I cleanup history.

In this case I realized that the removal of optional coverage is not good, because when you run single specs, so get an annoying delay for coverage report generation. So I restored that and wanted to check that I'm not breaking anything. After that I'll cleanup the branch.

deivid-rodriguez:
I had no news from codecov, by the way, just automatic emails, and there slack is [down](https://slack.codecov.io/). So... :man_shrugging:.

deivid-rodriguez:
I rebased this PR and added a somewhat related change: factor out the non Rails related rspec configuration to `spec_helper`, so spec files that don't need rails, don't need to require `rails_helper`.

javierjulio:
@deivid-rodriguez great job with the `spec_helper` change! I've made use of just requiring that file in a test file or two before in a Rails project so I think its worthwhile. Thanks for doing that. 👍🏻

PR #5496, opened 2018-09-29
Unnecessary spec stuff removal

deivid-rodriguez:
Test applications were generating, tweaking and installing a default Gemfile, but that `Gemfile` was not being used. So it's better not to generate it for faster development application generation and to reduce confusion.


deivid-rodriguez:
I rebased this PR and also removed a bunch of early requires and configuration in `rails_helper` that are not really necessary.

ping @activeadmin/maintainers

PR #5498, opened 2018-09-30
Bump capybara to latest 3.8 version

deivid-rodriguez:
As you can imagine by the diff in the specs, it was a lot of fun to figure it out :rofl:. Those were `&nbsp;` characters (I guess copied direcly from the HTML?) and not just whitespaces. And capybara doesn't like those because it just checks for the actual output the user sees.


deivid-rodriguez:
The coverage decrease is expected here: the "@debug" tag is not meant to be covered by specs. I tried it manually though, and it works :)

Issue #5484, opened 2018-09-21
Tabs are not correctly created when using non-latin characters as title


omitter:
This is probably a low priority bug, but it kept me debugging for hours.

Using this code, tab links are not created correctly:
```ruby
tabs do
  tab "テスト" do
    raw "Whatever"
  end
end
```

The id generated for the tab link is blank. This caused my app to reload itself continuously. Afaik jquery-ui interpreted it as link that should get loaded via ajax.

I think the problem is that `title.parameterize` returns `""` for non latin text [here](https://github.com/activeadmin/activeadmin/blob/4b862ed13f9f2a60d4fad48ffe2629e594dadde4/lib/active_admin/views/components/tabs.rb#L18).

As the source shows this can easily be circumvented by passing an `id` value to the method, but I would appreciate it when a clear error is raised (`raise "…" if fragment.blank?`) or if valid ids were generated (i.e. `title.parameterize.presence ||"id_"+SecureRandom.alphanumeric`). Thanks for considering.

deivid-rodriguez:
Hi @omitter! Thanks for the report. I added this spec locally but it passes for me. Can you help me reproducing this with a spec?

```diff
diff --git a/spec/unit/views/components/tabs_spec.rb b/spec/unit/views/components/tabs_spec.rb
index 8d6b9ef2..3ffc027b 100644
--- a/spec/unit/views/components/tabs_spec.rb
+++ b/spec/unit/views/components/tabs_spec.rb
@@ -67,5 +67,21 @@ RSpec.describe ActiveAdmin::Views::Tabs do
         expect(tabs.find_by_tag('span').first.content).to eq('tab 1')
       end
     end
+
+    context "when tab has non-latin characters" do
+      let(:tabs) do
+        render_arbre_component do
+          tabs do
+            tab "テスト" do
+              raw "whatever"
+            end
+          end
+        end
+      end
+
+      it "should create a tab navigation bar based on the symbol" do
+        expect(tabs.find_by_tag('li').first.content).to include("テスト")
+      end
+    end
   end
 end
```

omitter:
@deivid-rodriguez Thanks for looking into this. If I understand the spec correctly, you're looking for the wrong thing. The `content` is fine, it's the `id` that gets set to `id="#"` (for any tag without latin letters). Sorry, I can't check it further before next week.

deivid-rodriguez:
You're absolutely right, what was I thinking? :flushed: I'll have another look!

[...]

PR #5508, opened 2018-10-09
Tab ids will use non-ascii characters if title is non-ascii only

omitter:
Also an error is raised if the id is blank

See discussion [here](https://github.com/activeadmin/activeadmin/issues/5484).

### DIFF
--- a/lib/active_admin/views/components/tabs.rb
+++ b/lib/active_admin/views/components/tabs.rb
@@ -16,6 +16,8 @@ module ActiveAdmin
 
       def build_menu_item(title, options, &block)
         fragment = options.fetch(:id, title.parameterize)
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@omitter You're totally right (again!). In my mind I had asked you to submit a PR... (sorry :flushed:). At this point, may I ask you to add some specs for this?

omitter:
I'm sorry, but I don't have the time to set up the testing and writing the specs currently. If it's a quick job, I think it's best when someone of the project does it.

deivid-rodriguez:
@omitter I totally understand, no problem. We'll try to wrap this up ourselves! :+1:

panasyuk:
Don't worry guys i'll take the specs.

PR #5509, opened 2018-10-10
CircleCI

deivid-rodriguez:
I'd like to compare the difference in build time between Circle and Travis, so I'm adding a CircleCI configuration to try it out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,175 @@
+---
+
+version: 2.1
+
+steps: &steps
+  - checkout
+
+  - run:
+      name: Cat current Gemfile to a specific file so caching works
+      command: cat "$BUNDLE_GEMFILE.lock" > current_gemfile.lock
+
+  - restore_cache:
+      keys:
+        - bundler-dependencies-{{ checksum "current_gemfile.lock" }}
+
+  - run:
+      name: Install dependencies
+      command: bundle install --without development --path vendor/bundle --retry 3 --jobs 3
+
+  - run:
+      name: Install chromedriver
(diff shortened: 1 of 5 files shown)


javierjulio:
So much nicer! Great to see in the Checks section a status for each ruby/rails combination build. 👌🏻

deivid-rodriguez:
So the comparison was not very appropriate since when I opened the PR, TravisCI was not idle but was running another build already. But the improvement seems clear.

CircleCI build completed in 35 minutes and the TravisCI has not even yet started because the previous build is still running (for 52 minutes now).

deivid-rodriguez:
I suppose we can do this for now, and keep the TravisCI build temporarily just for `jruby` until we get an answer on https://github.com/circleci/circleci-images/issues/266?

javierjulio:
I think that's fine. Should we update the Travis config then to only be jRuby as part of this PR? At least to test and see the results in the Checks section?

Since the Travis check has a "Required" badge, the CircleCI ones don't. With the suggested change wouldn't we flip that? Perhaps that would speed up the check process to a completed state since it wouldn't have to wait on Travis to complete as it wouldn't be required now. What do you think?

deivid-rodriguez:
Wow, I pushed an extra commit and CircleCI automatically cancelled the running build and started the latest one. Sweet.

deivid-rodriguez:
@javierjulio This is ready from my side. I also added `jruby-9.1.17.0` to the TravisCI matrix for completeness.

deivid-rodriguez:
Wowowooooo, thanks, let's do it! Fast and non-redundant builds for the win!

Issue #5440, opened 2018-07-25
XML serialization broken in Rails 5.2 if activemodel-serializers-xml not present


tf:
### Expected behavior

Clicking the XML-Link below a table should result in a XML serialization of the table including the model attributes.

### Actual behavior

Instead, the output looks like this:

```
<orders type="array">
  <order type="Order">#<Order:0x0055acb3e56388></order>
</orders>
```

### How to reproduce

http://demo.activeadmin.info/admin/orders.xml 

### Solution/Workaround

Adding `activemodel-serializers-xml` to the applications `Gemfile` fixes the issue. Maybe this should be a dependency or added by the install generator by default?

varyonic:
`active-model-serializers` was extracted in Rails 5. It was subsequently added to Draper v3 which is often used with, but not required by ActiveAdmin. Perhaps adding it as a direct dependency would be the simplest fix.

PR #5512, opened 2018-10-12
XML serialization broken in Rails 5.2 if activemodel-serializers-xml not present

mapeveri:
Related to issue #5440


deivid-rodriguez:
Funny, I think this explains why we haven't managed to get the XML spec passing in https://github.com/activeadmin/activeadmin/pull/5486#issuecomment-427903098.

Note that here you're only adding it to our test environment, to reintroduce this as a runtime dependency you would need to add it to the gemspec file.

I'm not sure how to best fix this, but I tend towards reintroducing the dependency? My question: are XML links included by default on index pages? If that's the case, I think they should work by default? If not, we could not introduce the dependency but suggest the user to install it when she introduces an XML download link, and the dependency is not there?

Thoughts @activeadmin/maintainers?

timoschilling:
I would prefer to not add it as a strict dependency, we should do something like with devise which is optional.

deivid-rodriguez:
Yeah, that makes sense. I'd like to investigate whether this work out of the box in the past, and when it got broken to decide how to move forward, but in principle it seems like the best option would be that XML links are not there by default but can be optionally enabled through configuration and adding that gem.

mapeveri:
Hi everyone, I do not finish understanding the changes that I would have to do. 
Thanks!

deivid-rodriguez:
I understand, we haven't decided what to do yet, so it's not clear what you should do...

I investigated a bit more. As @varyonic points out in https://github.com/activeadmin/activeadmin/issues/5440#issuecomment-407828830, the `activemodel-serializers-xml` gem was extracted out of `rails` in `rails 5.0`. And it's included as a runtime of `draper` since drapper `3.0`, because that version only supports `rails >= 5.0`, so they fixed the functionality being extracted out of Rails by depending on the new gem.

As a result, the XML download links that activeadmin creates by default on all index pages don't currenly work unless you're using Rails 4.2 _or_ you're using a more recent Rails _and_ draper 3.0 (or any other dependency that happens to require `activemodel-serializers-xml`).

The current situation is bad, and although I think the mid-long term goal should be making XML download links and thus the `activemodel-serializers-xml` dependency optional, I think the quick fix for the time being is to drop Rails 4.2 support and exchange it with the `activemdel-serializers-xml` dependency.

mapeveri:
@deivid-rodriguez Anything that can help, let me know.

deivid-rodriguez:
Ok so now that we've dropped Rails 4.2 support, my vote here is to depend on `activemodel-serializers-xml` for now, so that the XML links generated by activeadmin by default actually work.

The future direction would be to not generate those links by default, but instead make them optional, and prompt the user to install the gem while they are enabled. Since this is more work, I prefer to readd the dependency for now and open a ticket to make the dependency optional in the future.

What do you think @timoschilling?

javierjulio:
@deivid-rodriguez yes, that would be the easier thing to do now. Ideally though it should be optional. I'm fine with what you propose. 👍🏻

deivid-rodriguez:
I say let's do this.

@mapeveri There are several things that are needed in order to get this ready:

* Rebase on top of latest master.
* Add the dependency as a runtime dependency in the `activeadmin.gemspec` file.
* Run `bundle install`, and then `bundle exec rake bundle` and commit the result to update all lock files.
* I'd like a high level spec to ensure that the XML download works.

Gemfile
```diff
@@ -5,6 +5,7 @@ eval_gemfile(File.expand_path("Gemfile.common", __dir__))
 gem "rails", "~> 5.2.x"
 gem "devise", "~> 4.4"
 gem "draper", "~> 3.0"
+gem "activemodel-serializers-xml"
```
deivid-rodriguez: Since we are adding a runtime dependency in the gemspec, we no longer need this addition to the Gemfile.

activeadmin.gemspec
```diff
@@ -21,6 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'arbre', '>= 1.1.1'
   s.add_dependency 'formtastic', '~> 3.1'
   s.add_dependency 'formtastic_i18n'
+  s.add_dependency 'activemodel-serializers-xml'
```
deivid-rodriguez: Can you add `~> 1.0` here, so that we never unintentionally allow a potentially breaking release?

mapeveri:
@deivid-rodriguez Ready. Can you explain a little more about this "As per how to test this, can you add a But I should see "Age" line at the end of this file?". Thanks!

deivid-rodriguez:
@mapeveri Thanks for working on this. The last commit seems to have introduced a bad merge, since it includes 1763 additions and 2026 deletions... :O. Can you review that, and rebase this PR?

Regarding your question about the test, if you look at 

https://github.com/activeadmin/activeadmin/blob/84de643cee0f732c12232debac53d408e88c0302/features/filter_attributes.feature#L1-L46, you'll see that all scenarios include the expectation

```gherkin
But I should see "Age"
```

except for the one about XML downloads. Once this PR lands, XML downloads will work in all situations, so we can add that expectation to the XML scenario as well.

PR #5515, opened 2018-10-14
Bump rubocop and other rubocop tweaks

deivid-rodriguez:
Rubocop version was quite old already so I bumped it to the latest version.

Also I added a couple of cops:

* `Style/FrozenStringLiteralComment`. We were fully consistent in this regard until I accidentally added a couple of those. Let's keep the code free of those comments until we add full support for frozen string literals.

* `Lint/ParenthesesAsGroupedExpression`. This is because I noticed some code that was a bit hard to read on a review, and it was because of a style issue of this kind. Since I prefer not making style review comments and let those be handled by a bot, I propose to add this cop.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,10 +10,7 @@ AllCops:
     - vendor/bundle/**/*
 
   Include:
-    - "**/*.rake"
-    - "**/Gemfile"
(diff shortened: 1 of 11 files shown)


deivid-rodriguez:
Thanks to you for all the reviews @javierjulio!

PR #5520, opened 2018-10-17
Workaround jruby bug inside Capybara

deivid-rodriguez:
This should get the jruby build green. It took me a while but I pinpointed the issue to a change introduced in Capybara 3.0 that made it hit a jruby bug.

See https://github.com/teamcapybara/capybara/issues/2115 for more information.

Hopefully this is addressed in capybara so we can stop pointing to my fork, but for now I'd like to get the build green :)

### DIFF
--- a/Gemfile.common
+++ b/Gemfile.common
@@ -37,7 +37,7 @@ group :development do
 end
 
 group :test do
-  gem 'capybara', '~> 3.0'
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
Yay!

javierjulio:
Awesome to see that the full build is now green! Really great work, thanks @deivid-rodriguez! 👌🏻

PR #5521, opened 2018-10-17
Remove more Rails 4.2 lefovers

deivid-rodriguez:
This PR is a follow up to #5104. It removes a bunch of other stuff we were including in order to support Rails 4.


javierjulio:
@deivid-rodriguez if Timo can't get back to this since you pushed a new commit, let me know and I'll approve this. Thanks!

deivid-rodriguez:
Thanks so much @javierjulio and @timoschilling for your reviews!

PR #5523, opened 2018-10-17
Refactor weird specs

deivid-rodriguez:
The jruby bug we were hitting before #5520 were hitting specs that were a bit "weird". They were including activeadmin modules directly into RSpec examples, that are dynamic classes, and stubbing method calls to `self` on them. Rewriting the specs to be a little bit more "standard" made the bug no longer hit them.

I ended up finding the real culprit, but the rewritten specs look better to me anyways, so I propose to include these changes.

### DIFF
--- a/spec/unit/filters/filter_form_builder_spec.rb
+++ b/spec/unit/filters/filter_form_builder_spec.rb
@@ -3,7 +3,7 @@ RSpec.describe ActiveAdmin::Filters::ViewHelper do
   # Setup an ActionView::Base object which can be used for
   # generating the form for.
   let(:helpers) do
-    view = action_view
+    view = mock_action_view
     def view.collection_path
       "/posts"
     end
@@ -366,7 +366,7 @@ RSpec.describe ActiveAdmin::Filters::ViewHelper do
     # Setup an ActionView::Base object which can be used for
     # generating the form for.
     let(:helpers) do
-      view = action_view
+      view = mock_action_view
       def view.collection_path
         "/categories"
       end
--- a/spec/unit/pretty_format_spec.rb
+++ b/spec/unit/pretty_format_spec.rb
@@ -1,26 +1,66 @@
 require 'active_admin/view_helpers/display_helper'
 
 RSpec.describe "#pretty_format" do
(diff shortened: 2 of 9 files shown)


spec/unit/pretty_format_spec.rb
```diff
@@ -49,33 +89,44 @@ def method_missing(*args, &block)
...
-      post = Post.new
-      expect(self).to receive(:auto_link).with(post) { "model name" }
-      expect(pretty_format(post)).to eq "model name"
+      expect(view).to receive(:auto_link).with(obj).and_return("model name")
+      expect(formatted_obj).to eq "model name"
     end
   end
 
   context "given an arbitrary object" do
+    let(:obj)  { Class.new.new }
```
javierjulio: This seemed odd at first glance. Just to confirm is the double `.new` on purpose? I see it was that way originally though.

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -3,7 +3,7 @@
   # Setup an ActionView::Base object which can be used for
   # generating the form for.
   let(:helpers) do
-    view = action_view
+    view = mock_action_view
```
javierjulio: I'm confused a little with these changes because, the original method was also called `mock_action_view` or at least that was what was moved and edited. Is `action_view` a method/variable we can remove or its elsewhere? I mention this since I've seen this changed in several other files but nothing changing something called `action_view` so was I was curious.

javierjulio: You're right! I missed that! I see it now in the diff. I knew I was missing something. Just wanted to be sure. Thanks! Yes I think what you did is a great improvement. It helps make things a little clearer. 👍🏻

spec/unit/pretty_format_spec.rb
javierjulio: Ok cool I'll just leave this as is then and resolve conversation. Thanks!

javierjulio:
@deivid-rodriguez now that there is a conflict, please ping me when you resolve and push. I'll make sure to approve again for you.

PR #5526, opened 2018-10-18
Ability remote: true for batch actions

atanych:
Sometime we don't need to reload page (e.g. BG jobs). And ability to submit batch action through ajax might be useful.

Example:

```
  batch_action :some_action, remote: true do |ids|
     ...
  end
```


deivid-rodriguez:
Hello @atanych!

Does this provide any visual feedback about the completion of the request? Does this disable the link/button once clicked until the request finishes, so it's not submitted multiple times? By the way, if you rebase on top of #5527, build should go green.

atanych:
Hey @deivid-rodriguez!

about visual feedback.

This solution immediately:
1. Closed dropdown
2. Unselected checkboxes
3. Disabled batch action button

javierjulio:
Not sure it's feasible for us to take on another dependency and an outdated one too. I'd be open to a solution here if we can revisit with a new approach since it's been so long. I'm not sure how simple it would be overall with batch actions.

PR #5528, opened 2018-10-18
Add stuff for managing gemfiles

deivid-rodriguez:
Having multiple gemfiles and lock files in the repo is useful for environment determinism, but it takes a bit of work to keep up to date.

This PR adds some tools to facilitate that:

* A spec to check that lockfiles are kept in sync.
* A task to bundle them all in one command: `bundle exec rake bundle`.

This way PRs adding or removing dependencies (like #5512), should be easier to handle.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,10 @@ steps: &steps
   - checkout
 
   - run:
+      name: Install a specific bundler version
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
Thanks @javierjulio!

PR #5530, opened 2018-10-19
Fix parallel test setup

### LINKED ISSUE #4257: Run tests in Travis in parallel
Travis provides 1.5 "virtual" CPU cores for each VM. Their documentation hints that you may see gains if you use the parallel_tests gem (that we already have set up for local test runs), and the [parallel_tests wiki](https://github.com/grosser/parallel_tests/wiki/Distributed-Parallel-Tests-and-Travis-Support#travis-ci-support) describes how to get it working on Travis.

So with parallel_tests to run two workers on Travis, how long does it take for our test suite to run?

Note this requires **order independent cucumber tests**.

deivid-rodriguez:
This PR is a work in progress to fix #4257, and hopefully get a 4x speed up in CI.


deivid-rodriguez:
OMG, this actually worked!!! :tada: :tada:

javierjulio:
@deivid-rodriguez haha awesome, I love it. 😍 I took a quick look at CircleCI. It seems this has cut down each build in a little more than half! That's great work, thanks! 👏🏻

deivid-rodriguez:
@javierjulio Yeah, I'm happy! :)

I think actual test time should be 4 times faster now, but given that we do other things that we're not parallelizing (linting and doc generation), the final improvement ends up being smaller. That actually suggests another improvement we can do: run linting and doc generation as a separate job, a single time instead of for every ruby version.

javierjulio:
Ah yes of course, that makes sense! They are running in every build. For both linters and docs, we'd use the default Gemfile/setup right?

deivid-rodriguez:
Yeah, default Gemfile, and latest MRI I guess.

deivid-rodriguez:
I splitted lint and doc generation to a separate job on #5531.

Also, I discovered that the reason for the speed up not being 4x is actually that tests are not splitted evenly per processor by running time, so the slowest group acts as a bottleneck. I'll add some more changes to fix that.

.circleci/config.yml
```diff
@@ -48,10 +48,23 @@ version: 2.1
   attach_workspace:
     at: spec/rails
 
+.restore_test_times: &restore_test_times
+  restore_cache:
+    keys:
+      - testtime-{{ .Branch }}-{{ .Environment.CIRCLE_JOB }}
+      - testtime-master-{{ .Environment.CIRCLE_JOB }}
```
javierjulio: What's the reason for restoring both dynamically from a feature branch and master? Not that this is wrong, but just by looking at how we save times, I didn't understand.

deivid-rodriguez: What parallel_tests does is split all the tests in groups, and run each group in parallel. If you don't specify any criteria, it will split in groups of the same number of tests randomly. That means, for example, that all integration specs could fall in one group, and all model specs in another. And that means the group with the integration tests will be a bottleneck.

To solve this issue, and ideally have no bottlenecks, parallel_tests can optionally record the time it took for each test to run, and then in subsequent runs reuse that information to find the optimal split between tests so that all groups take approximately the same time. So this is what we're doing here, record that information so that our builds always split tests into groups optimally.

The edge case is that on the first push to a branch we will have no recorded information for that branch. If that is the case, we look at what was last recorded for the master branch, which in most cases will still be a good approximation. This `keys:` section does not mean we're restoring both, it means we're first looking at the first key, and it that's not present in the cache, we look for the second one.

Hope that all makes sense!

deivid-rodriguez:
@javierjulio Absolutely, I recommend "commit per commit" reviewing, as usual!

.circleci/config.yml
javierjulio: Yes that does! It makes a lot of sense. I realize now I've done basically this but on what was CircleCI 1.0 as they had a tool for splitting tests which essentially seemed to just pick sets of spec files and specified that list to individual Rspec commands that would run. I remember about saving the timings of that too. The timings here come from the format option: `--format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log` which is what I had to use as well but with RspecJunitFormatter and the `circleci tests` command on their older platform. Thanks for clarifying! It helped as I now know it's all doing essentially the same thing. 👍🏻

tasks/application_generator.rb
```diff
@@ -0,0 +1,85 @@
...
+    def base_env
+      { 'BUNDLE_GEMFILE' => ENV['BUNDLE_GEMFILE'] }
+    end
+
+    def running_mode
+      parallel ? "in parallel" : "sequentially"
+    end
+
+    def correctly_configured_app?
+      app_exists? && !(parallel ^ parallel_tests_setup?)
```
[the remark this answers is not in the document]
deivid-rodriguez: Yeah, absolutely, it's not very straighforward, I'll clarify it!

Something I'm considering though is that we might not need these checks at all? We could _always_ configure the app in parallel (with multiple DBs), and I think sequential tasks would just work too, they would run on the first db.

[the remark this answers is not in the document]
deivid-rodriguez: I'll have a look tomorrow at this!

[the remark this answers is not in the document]
javierjulio: Sounds good to me, thanks!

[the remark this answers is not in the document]
deivid-rodriguez: Ok, I simplified this in https://github.com/activeadmin/activeadmin/pull/5530/commits/22fdce1ff4e76b381006ca28f1141244e1f0dee1, but now I'm thinking that maybe we don't need the sequential test tasks at all? There's just wrappers on top of rspec / cucumber. If someone wants to run specific specs (or all) sequentially, she can use rspec or cucumber directly?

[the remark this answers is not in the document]
deivid-rodriguez: So removed all the sequential tasks and went "parallel by default". One can still run sequential tasks by using `rspec` or `cucumber` directly, but the test tasks use parallelization by default.

deivid-rodriguez:
@javierjulio I added a few more tweaks, I thing everything is simpler now since we don't keep duplicate tasks for the parallel / sequential case. I also did a bit of squashing to keep the number of commits reasonable.

Let me know what you think!

deivid-rodriguez:
CI takes less than 20 minutes now!

tasks/application_generator.rb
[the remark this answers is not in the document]
javierjulio: Ok that's good to know because after reviewing the changes I was going to ask. I think that's a good change. It a big test suite so to get it running even faster is a huge help! Also nice that it's consistent with what and how we run it in CI. 👍🏻

tasks/local.rake
```diff
@@ -1,18 +1,19 @@
...
 task :local do
-  Rake::Task['setup'].invoke(false,
-                             'development',
-                             'rails_template_with_data')
+  ActiveAdmin::ApplicationGenerator.new(
+    rails_env: 'development',
+    template: 'rails_template_with_data'
+  ).generate
 
   app_folder = ".test-rails-apps/rails-#{Rails::VERSION::STRING}"
```
javierjulio: Since we also have this in the ApplicationGenerator class, I wonder if we can avoid that duplication. If you don't want to store access to an instance of the generator, perhaps an idea could be that the `generate` method returns the `app_dir` value? It doesn't seem far fetched since that's really what the method deals a lot with and would signify what's there. Or perhaps something else?

tasks/application_generator.rb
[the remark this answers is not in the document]
javierjulio: So to confirm that I understand right, the rspec and cucumber tests are run in parallel locally against one database?

tasks/local.rake
deivid-rodriguez: Sounds like a good idea to me! I'll add it!

deivid-rodriguez:
> So to confirm that I understand right, the rspec and cucumber tests are run in parallel locally against one database?

The technique used by parallel tests to run tests in parallel is to create multiple databases (one per process/thread), and run each process/thread against a separate DB, so that they don't interfere with each other. What I'm doing in this PR is creating the test app with all those DBs by default. If you run tests in parallel locally, each process/thread runs against those DBs (a different one for each). If you run tests sequentially instead (through `bundle exec rspec` or `bundle exec cucumber`), then the first of the DBs will be used and things will work as if the app had a single DB configured. Also, for the _development_ app, I'm not doing any "parallel setup", only a single DB is created.

tasks/local.rake
deivid-rodriguez: I pushed https://github.com/activeadmin/activeadmin/pull/5530/commits/78e2aca1e5e2d4feb944414cb84fca68fabf0454!

.gitignore
```diff
@@ -26,6 +26,7 @@ tmtags
 
 ## Project (general)
 tags
+tmp
```
deivid-rodriguez: I was in doubt whether this entry belongs in the project's or in the global gitignore... :man_shrugging:

javierjulio: Sorry, a little confused. You mean in another `gitignore` file completely? Oh wait you mean in just the same file but a different section? I think it's fine as is. 👍🏻

javierjulio:
Ok that makes sense. Thanks! So that means when you run tests locally (the default which is the full suite) it runs in parallel and thus makes use of multiple databases. 👍🏻 Just when you run `rspec` or `cucumber` individually and directly, it's sequential. Love it. 👏🏻

.gitignore
deivid-rodriguez: Normally I put on my global gitignore (`~/.git/ignore`) stuff specific to my workflow but unrelated to the project (for example, that temp file my editor "x" generates) , but in the project's `.gitignore` stuff specific to the project. In this case I was not sure in which category "tmp" falls.

Anyways, seems fine as is indeed :+1:!

javierjulio: Ah ok that’s what I thought you meant but had me confused haha. Yes I think it’s fine here. Doesn’t hurt. Thanks! 👍🏻

PR #5531, opened 2018-10-20
Lint and docs as single job

deivid-rodriguez:
In this PR I'm trying to further speed up and save resources in CI by running linting and doc generation on a single job, not on all of them.


tasks/lint.rake
```diff
@@ -94,4 +94,16 @@ namespace :lint do
...
+  desc "RSpec specs for linting project files"
+  task :rspec do
+    puts "Linting project files..."
+
+    abort unless system(
+      "rspec",
+      "spec/gemfiles_spec.lint.rb",
+      "spec/changelog_spec.lint.rb",
+      "spec/i18n_spec.lint.rb"
+    )
```
deivid-rodriguez: These specs are actually linting files, so I moved them to the lint task as well. The rename is so that they don't get run by default with the regular `rspec` run.

.circleci/config.yml
```diff
@@ -2,42 +2,70 @@
...
+    command: bundle install --without development --path vendor/bundle --retry 3 --jobs 3
+
+.install_chromedriver: &install_chromedriver
+  run:
+    name: Install chromedriver
+    command: bin/install_chromedriver.sh
+
+.save_cache: &save_cache
+  save_cache:
+    key: bundler-dependencies-{{ checksum "current_gemfile.lock" }}
```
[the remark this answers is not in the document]
deivid-rodriguez: Indeed, done!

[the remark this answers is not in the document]
javierjulio: Awesome thanks! I was reviewing on my phone and couldn't jump to the other PR, so let me just double check everything.

.circleci/config.yml
```diff
@@ -2,42 +2,80 @@
 
 version: 2.1
 
-steps: &steps
+.install_bundler: &install_bundler
+  run:
+    name: Install a specific bundler version
+    command: gem install bundler -v 1.16.6
```
javierjulio: This is fine but I couldn't remember the reason for this. Are we using a specific version rather than the latest so its predictable and ensure the same one is used for all builds?

.circleci/config.yml
```diff
@@ -2,42 +2,80 @@
...
-      name: Install dependencies
-      command: bundle install --without development --path /home/circleci/project/vendor/bundle --retry 3 --jobs 3
-
-  - run:
-      name: Install chromedriver
-      command: bin/install_chromedriver.sh
+jobs:
+  lint_and_docs:
+    docker:
+      - image: circleci/ruby:2.5.1
```
javierjulio: I noticed here and in other places too we are using Ruby 2.5.1 but 2.5.3 has already been released. Should we open a new PR to update that? Or do you prefer we keep at 2.5.1 for the time being?

.circleci/config.yml
deivid-rodriguez: Yeah. The thing is, our base docker images are updated almost inmediately after bundler releases new versions. So, for example, we had previously `1.16.5` in the lock file at the `BUNDLED_WITH` section, but the images had already 1.16.6 installed. In that situation, the bundle binstub provided by the docker image was refusing to run with an error about the versions not matching. So yeah, in order to run a predictable version for all builds, and be able to update it at our own pace, I added this step installing the specific version that we want.

.circleci/config.yml
deivid-rodriguez: Yeah, I almost used 2.5.3 here, but decided to stay with 2.5.1 for now, and separately update everything on a separate PR.

javierjulio: It occurred to me that we weren't on Ruby 2.5.2 but saw now that it was released only days ago and that 2.5.3 was to correct an issue in that patch release. I was just wondering why we hadn't updated sooner but now it makes sense. Sounds good!

.circleci/config.yml
javierjulio: Great thanks! I figured it was rather important/detailed. Do you think there would be value in having a comment in this file as to why we need a specific version? I ask because I'm so used to seeing that the latest is installed I figured its good to have a reason here. If not its fine.

deivid-rodriguez: Sure! I'll improve the description of the step, the job is going to fail if we remove it I think, but a better description is never a bad idea :smiley:

deivid-rodriguez: I'm actually not super happy about this because updating the bundler version we're using means updating it in all lock files, and here. Maybe a better idea is to lock the base docker images that we use, so they don't get updated unless we do it explicitly? :thinking:.

PR #5533, opened 2018-10-22
Bump rubies used in CI to latest versions

deivid-rodriguez:
As suggested in https://github.com/activeadmin/activeadmin/pull/5531#discussion_r226872505.


javierjulio:
Good to know, thanks! Yeah its affecting us at work too since we also use CircleCI. It should be back up soon though. 🙏🏻

PR #5534, opened 2018-10-22
Split test app creation to a separate job

deivid-rodriguez:
This is some initial implementation of https://github.com/activeadmin/activeadmin/pull/5530#issuecomment-431696455.

It turns out that the main reason why jruby is so slow is the creation of the test app, because at lot of subprocesses are spawned, and jruby has a very slow startup. This should alleviate that by creating a single test app per supported Rails version.

I haven't actually tried any of this in CI, will see how it goes once Github is back up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -291,6 +291,15 @@ jobs:
 workflows:
   version: 2
 
+  #
+  # The workflow has three types of jobs:
+  #
+  # * A lint job to run style checks.
+  # * Jobs that create test apps for each supported rails version.
+  # * Jobs that run tests, one for each supported [ruby_version, rails_version]
+  #   combination. Every test job for a specific rails version reuses the test
+  #   app created for that specific rails version.
+  #
   build:
     jobs:
       - lint_and_docs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,10 +33,25 @@ version: 2.1
     paths:
       - vendor/bundle
 
+.create_test_app: &create_test_app
+  run:
+    name: Create test app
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
This is now ready even if the WIP check won't go away for some reason!

.circleci/config.yml
```diff
@@ -33,10 +33,25 @@ version: 2.1
     paths:
       - vendor/bundle
 
+.create_test_app: &create_test_app
```
javierjulio: I had to walk through this for a bit and in CI to double check but all makes sense. At first I had it wrong. Perhaps add a small note: "A test app is shared across Ruby versions but only for one Rails version."?

deivid-rodriguez: I added a general comment about the workflow in the workflows section in https://github.com/activeadmin/activeadmin/pull/5534/commits/61a17ce92eebdbb438ad8860d92cda6886609ec4, let me know what you think! Not very good at doc writing, but I did include some bullet points to try make it easier to read :smiley:.

javierjulio: That’s great thanks! 😄

javierjulio:
@deivid-rodriguez I take it the new commit/push fixed the WIP issue too?

deivid-rodriguez:
Apparently, yeah! Note that the WIP app has changed and now using the `work in progress` label is a [paid feature](https://github.com/wip/app/issues/89#issuecomment-432447327). So only changing the PR title will work to set a WIP status on the PR from now on.

javierjulio:
Ok cool thanks!

PR #5535, opened 2018-10-22
Fix jruby errors and warnings

deivid-rodriguez:
This PR fixes some warnings and errors that are printed to the screen when running specs under jruby:

* One is a `parser` ruby version mismatch warning that comes from requiring rubocop. I fixed it by not requiring rubocop (or other lint tools, since I was at it) unless necessary.
* The other one is some errors printed by `puma`. They don't seem to affect the outcome of the build, but they're annoying. I tried using the `webrick` server that comes with ruby and it works just fine with similar run times, so I propose to use that. `webrick` is not something people use in production, but I don't think using `puma` here brings a lot of benefits.

### DIFF
--- a/Gemfile.common
+++ b/Gemfile.common
@@ -26,9 +26,6 @@ gem 'yard'                        # Documentation generator
 gem 'redcarpet', platforms: :mri  # Markdown implementation (for yard)
 gem 'kramdown', platforms: :jruby # Markdown implementation (for yard)
 
-# Application server
-gem 'puma', '~> 3.12'
-
 group :development do
   # Debugging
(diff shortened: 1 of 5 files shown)


javierjulio:
@deivid-rodriguez let me know if you got a chance to spot check this now that CI ran. I'll review if its ready.

deivid-rodriguez:
It should be ready although the current green build is noticiably slower than the last completed build in master (from yesterday). I think it's because of the caching up after the outage, so once the current master build finishes, I'll rebase this one so we can compare build times under similar conditions.

deivid-rodriguez:
It seems like there's not much difference in performance so :+1:.

PR #5548, opened 2018-11-02
Avoid rendering huge associations

DanielHeath:
This is a relatively small change that substantially reduces the pain caused by #2692 

When an association has a large number of records (default 20), no dropdown is built. 

If it also has a column that looks likely to be a name, a 'starts_with' text field is added to the default filters instead.


lib/active_admin/namespace_settings.rb
```diff
@@ -103,5 +103,17 @@ class NamespaceSettings < DynamicSettingsNode
 
     # Include association filters by default
     register :include_default_association_filters, true
+    register :maximum_association_filter_arity, 20
```
varyonic: I'm in favor of a higher value, say 256, that will preserve country selects.

DanielHeath: Would country selects be done with an association table? I'd assumed you would use the builtin rails select.

deivid-rodriguez:
This seems like a good idea for 2.0, I'll review as soon as I can! :+1:

DanielHeath:
Is there anything still blocking this?

features/index/filters.feature
```diff
@@ -249,3 +249,36 @@ Feature: Index Filtering
...
+    And an index configuration of:
+    """
+      ActiveAdmin.register Category
+      ActiveAdmin.register Post do
+        config.namespace.maximum_association_filter_arity = 3
+      end
+    """
+    And I am on the index page for posts
+    Then I should see "CATEGORY" within "#filters_sidebar_section"
+     And I should not see "CATEGORY NAME STARTS WITH" within "#filters_sidebar_section"
```
deivid-rodriguez: Can you left align this `And` keyword with the rest?

features/index/filters.feature
```diff
@@ -249,3 +249,36 @@ Feature: Index Filtering
...
+    Given a post with the title "Compass and Sextant" in category "Navigation" exists
+    And an index configuration of:
+    """
+      ActiveAdmin.register Category
+      ActiveAdmin.register Post do
+        config.namespace.maximum_association_filter_arity = 3
+      end
+    """
+    And I am on the index page for posts
+    Then I should see "CATEGORY" within "#filters_sidebar_section"
```
deivid-rodriguez: I'm very confused, this expectation is failing for me when running this locally :thinking:.

features/index/filters.feature
```diff
@@ -249,3 +249,36 @@ Feature: Index Filtering
...
+        config.namespace.filter_method_for_large_association = '_contains'
+      end
+    """
+    And I am on the index page for posts
+    Then I should see "CATEGORY NAME CONTAINS" within "#filters_sidebar_section"
+    When I fill in "Category name contains" with "Astro"
+    And I press "Filter"
+    Then I should see "Star Signs"
+     And I should see "Constellations"
+     And I should not see "Compass and Sextant"
```
deivid-rodriguez: Left align these two tags as well.

features/index/filters.feature
```diff
@@ -249,3 +249,36 @@ Feature: Index Filtering
...
+      end
+    """
+    And I am on the index page for posts
+    Then I should see "CATEGORY" within "#filters_sidebar_section"
+     And I should not see "CATEGORY NAME STARTS WITH" within "#filters_sidebar_section"
+    Given an index configuration of:
+    """
+      ActiveAdmin.register Category
+      ActiveAdmin.register Post do
+        config.namespace.maximum_association_filter_arity = 2
```
deivid-rodriguez: Should we also test the case where  a `starts_with` box is rendered (maximum arity reached, and no explicity filter method configured)?

lib/active_admin/namespace_settings.rb
deivid-rodriguez: If the countries live on a separate table, the yes, right? I tend to agree with @varyonic's opinion of setting a higher value here.

DanielHeath:
I've updated this in response to all comments.

DanielHeath:
Fixed the last linter error and rebased on latest master. Is there anything else this needs to be merged?

DanielHeath:
Seems that https://github.com/deivid-rodriguez/chandler has yanked a revision we were depending on, causing all the tests to start failing :/

deivid-rodriguez:
Yes, my fault, the branch had conflicts with upstream's master so I force-pushed it! I'll fix it tomorrow, sorry for the trouble.

deivid-rodriguez:
I'll also do a final review pass, thanks for this!

lib/active_admin/filters/resource_extension.rb
```diff
@@ -128,7 +128,18 @@ def default_association_filters
...
-          filters = poly.map(&:foreign_type) + not_poly.map(&:name)
+          # Check high-arity associations for filterable columns
+          high_arity, low_arity = not_poly.partition{ |r| r.klass.count > namespace.maximum_association_filter_arity }
+
+          search_columns = namespace.filter_columns_for_large_association.map(&:to_s)
+
+          # Remove high-arity associations with no searchable column
+          high_arity.reject! {|r| ! r.klass.column_names.find {|name| search_columns.include?(name)}}
+
+          high_arity = high_arity.map {|r| r.name.to_s + "_" + r.klass.column_names.find {|name| search_columns.include?(name)}.to_s + namespace.filter_method_for_large_association}
```
deivid-rodriguez: Can we extract some duplicated logic to methods? Maybe

```ruby
def search_columns
  @search_columns ||= namespace.filter_columns_for_large_association.map(&:to_s)
end

def searchable_column_for(relation)
  relation.klass.column_names.find {|name| search_columns.include?(name)}}
end
```

Also, are the `to_s` in this line necessary?

DanielHeath: Good idea splitting it out. 
The `to_s` allows comparison between strings and symbols.

DanielHeath: Updated.

DanielHeath:
Implemented your suggestion

DanielHeath:
@deivid-rodriguez anything else required on this?

DanielHeath:
Anything else still to do on this before merging?

DanielHeath:
Added changelog and rebased. 99% sure I'd already fixed all the outstanding style issues - were there any others left?

DanielHeath:
Lint should pass now too.

deivid-rodriguez:
@DanielHeath One more rebase and style fixes are needed here. I can do it myself, but I'm not sure if I can force-push to your fork's master (I guess I can if you didn't uncheck the "Allow edits from maintainers" checkbox when creating the PR). Are you ok with me doing that? :)

DanielHeath:
Did you forget to push chandler commit c46770b45 ? I can't `bundle install` anymore.

DanielHeath:
Happy for you to do the style fixes - I can't get `bundle` to work after rebasing.

deivid-rodriguez:
@DanielHeath I broke the build in the exact same way as last time, sorry again. Should be fixed now, and not happen again.

deivid-rodriguez:
bors try

DanielHeath:
Any luck with Bors?

deivid-rodriguez:
Nah, sorry for the spam, by the way. You're going to have to do the last rebase and style fixing... :pray:. Sorry for how hard it's being to get this in, it's a great improvement!

DanielHeath:
@deivid-rodriguez should be sorted now :)

DanielHeath:
@varyonic all good to merge?

DanielHeath:
Who has access to merge?

lib/active_admin/filters/resource_extension.rb
```diff
@@ -128,13 +128,30 @@ def default_association_filters
           # remove deeply nested associations
           not_poly.reject! { |r| r.chain.length > 2 }
 
-          filters = poly.map(&:foreign_type) + not_poly.map(&:name)
+          # Check high-arity associations for filterable columns
+          high_arity, low_arity = not_poly.partition { |r| r.klass.count > namespace.maximum_association_filter_arity }
```
zorab47: Counting records has potential to impact performance in unexpected ways. There should also be an option to skip this check altogether. Another aspect of the performance impact is a additional query per association.

I'd also consider this as a major change, and require a major version bump. Is there about upgrading doc to add this new configuration to?

Fivell: @zorab47  agree this should not be a default behaviour

seanlinsley: Another option would be to ask the database for an approximate count from its internal statistics. That should be good enough for this use case (it's not 100% accurate, but querying for the approximate count is effectively free).

Postgres has [`reltuples`](https://wiki.postgresql.org/wiki/Count_estimate) and MySQL has [`table_rows`](https://dba.stackexchange.com/a/151790). Does ActiveAdmin support any other databases?

DanielHeath: If the check is being skipped / it's not the default behavior, does it still require a major version bump?

DanielHeath:
If we issued a `limit N` and got back `N` records, we could assume there are probably more and render the alternative UI.

DanielHeath:
Alright, I've implemented the `limit` approach, which should mean the size query is only hitting one disk page.

DanielHeath:
@zorab47 do these changes address your concerns?

DanielHeath:
@zorab47 @deivid-rodriguez was there anything still outstanding on this?

deivid-rodriguez:
I'm sorry @DanielHeath for not getting back to this. I'll have a look during the next days :+1:

deivid-rodriguez:
I want to setup a sample application with an association with lots of records to see this in action, but the technique you used looks like it should work and a good idea :+1:. I plan to have a closer look tomorrow.

deivid-rodriguez:
This works great. The count + limit trick seems to work as expected

""""
 (0.2ms)  SELECT COUNT(*) FROM (SELECT  1 AS one FROM "posts" LIMIT ?) subquery_for_count  [["LIMIT", 257]]
""""

Does it make sense to keep the current behaviour under `maximum_association_filter_arity = :unlimited` or something like that?

I figure some people will still always want select boxes, because their association are not expected to grow much, and if they do, the grow will be noticed more easily through a select box, and thus be easier to keep under control?

DanielHeath:
@deivid-rodriguez Took me awhile to get back to this (hectic month) but I've added the `:unlimited` option.

DanielHeath:
Anything left before I can get this merged? I'm starting to lose steam / excitement, tbh.

deivid-rodriguez:
I'm so sorry for the delay @DanielHeath, thanks for the remainder and your patience.

From my side, I would also document the ":unlimited" value for the new configuration. Also, I think it probably makes sense to make this an opt-in feature and keep ":unlimited" as a default for now. It's a non trivial change introducing three new configurations, so maybe we can use a few 2.x releases to test them out and gather feedback before toggling the default behavior?

Not sure what others think but this seems inline with @zorab47's comment in https://github.com/activeadmin/activeadmin/pull/5548#discussion_r257686754. What do you think?

DanielHeath:
IMO, the practical result of the current default behavior is frequently to crash both the server and the clients browser with an OOM. 

I'm less-than-clear on what makes that a good default to keep, but I'd be happy to wait a few releases before switching it on by default anyways.

Agree the new switch needs documentation - do you think adding it as a comment to the default config file is enough? Where else should it be covered?

deivid-rodriguez:
I essentially agree with you but I think it's better to change this in a more gradual manner, by providing a way to opt in to change it in the next minor version (so that affected users have a way out), but not changing the behavior by default (so that users that are happy with the current behavior don't get the breaking change on a minor release).

> Agree the new switch needs documentation - do you think adding it as a comment to the default config file is enough? Where else should it be covered?

Yes, I think a comment in the config file is a good place to document this. We can also comment the other new options, explaining what each one does and how it affects the generated filters.

We could also add some general documentation about how auto-generated filters work and how they can be controlled (maybe in `docs/3-index-pages-md`), but we can leave that for a separate PR.

Also, there seems to be a merge conflict now (sorry again for leaving this unattended for so long...).

DanielHeath:
I've defaulted it to `:unlimited` for 2.3 and `256` for 2.4 and up. 

Willing to change it to just be `:unlimited` if you would prefer to choose the release to enable it manually, though.

DanielHeath:
@deivid-rodriguez anything else required on this? The changelog churn is getting a bit much.

deivid-rodriguez:
> Willing to change it to just be :unlimited if you would prefer to choose the release to enable it manually, though.

Yes, don't include any logic that checks future version. Set it to disabled by default, and we'll enable it later.

Thanks again for your patience.

DanielHeath:
Done :)

DanielHeath:
Fixed the lint errors, should be good to go :)

deivid-rodriguez:
Specifically @zorab47, have all your concerns been addressed?

lib/active_admin/filters/resource_extension.rb
```diff
@@ -129,12 +129,36 @@ def default_association_filters
           not_poly.reject! { |r| r.chain.length > 2 }
 
           filters = poly.map(&:foreign_type) + not_poly.map(&:name)
+
+          # Check high-arity associations for filterable columns
+          max = namespace.maximum_association_filter_arity
+          if max != :unlimited
+            high_arity, low_arity = not_poly.partition do |r|
+              r.klass.limit(max + 1).count > max
```
zorab47: Could this be improved by removing the `ORDER BY` clause if present? I think that can be done via `reorder("")`. My concern is that a query with a `LIMIT` can still be slow if it is ordered.

DanielHeath: Is `r.klass` a scope? I thought it was an AR class (which, I guess, could have a `default_scope`)

deivid-rodriguez:
@zorab47 Does this look ok now? Do you have any other insights? Thanks! :heart:

zorab47:
Yes, that covers my concerns. 👍

DanielHeath:
@deivid-rodriguez anything else remaining?

deivid-rodriguez:
There's just one merge conflict in the changelog that needs to be addressed, but other than that everybody seems happy with landing this :)

DanielHeath:
Alright, resolved that

CHANGELOG.md
```diff
@@ -18,6 +19,7 @@
 
 #### Minor
 
+* Render a text filter instead of a select for large associations (opt-in) [#5548] by [@DanielHeath]
```
deivid-rodriguez: You left this line here unintentionally, it seems.

DanielHeath: Oops - fixed.

Extazystas:
Hi @DanielHeath ✋. Could you please provide any example of how to use it? 
`Render a text filter instead of a select for large associations (opt-in)` sounds promising but there are no wiki or documentation how to use if in form e.g. I have a Booking edit form with User select input. Booking belongs_to User. How should I organize my form to enable text filter by user name? 
Thanks in advance :shipit:

Issue #5553, opened 2018-11-08
Batch action with turbolinks attaching click event twice


sgara:
Hi,
We'be been using activeadmin with turbolinks quite successfully for some time. While migrating to Turbolinks 5, I encountered an issue on batch actions : seems like both `ready` and `turbolinks:load` are trigger, which causes click handler to be attached twice. 

I've recreated a simple project from scratch to check if that wasn't coming from our side and I do reproduce it. Basically just a brand new default rails app with activeadmin. [Source code](https://github.com/Augment/activeadmin-issue-demo) and [live demo](https://activeadmin-issue-demo.herokuapp.com/admin/items) are available. Just go to `admin/items` directly, select items then `Batch Actions` > `Delete Selected Items`. The modal is then opened twice as you can see in this screenshot :
<img width="1280" alt="screenshot 2018-11-08 at 22 45 28" src="https://user-images.githubusercontent.com/4999575/48229356-21731080-e3a8-11e8-8a34-b2f22c081ab1.png">

Looks like every version between 1.0.0 and 1.3.1 are affected (master might be as well). A possible solution would be to detach previously attached click handler [here](https://github.com/activeadmin/activeadmin/blob/1-3-stable/app/assets/javascripts/active_admin/lib/batch_actions.js.coffee#L5) like this :
```coffeescript
$('.batch_actions_selector li a').off('click').on 'click', (e)->
```

I'd be happy to do a PR if that solution seems acceptable. I'm open to other solutions/workarounds/.. as well. 
Best!

PR #5562, opened 2018-11-13
Detach previous batch_action click handler before attaching new one (coffeescript)

sgara:
Following #5553, here is a propositional change to solve batch action modal being opened twice when using turbolinks. This ensures any previously attached click handler is being removed (including own `batch_actions` handler) before attaching own handler.

I'm targeting `1-3-stable` branch as master has moved from coffeescript to es6, so that's the current stable branch in production. Patch could easily back ported to all `1-x-stable` branch, at your discretion. Incoming PR to solve that issue on master as well.


sgara:
Seeing that 1.4 is close to release (cf. #5566), it'd be great to have some feedback on this 😇 
Should I switch target branch to `1-4-stable` maybe?

deivid-rodriguez:
We're no longer maintaining 1.x branches, so I'll close this!

PR #5563, opened 2018-11-13
Detach previous batch_action click handler before attaching new one (es6)

sgara:
Following #5553 and #5562, here is the es6 version of the same patch to ensures any previously attached click handler is being removed (including own batch_actions handler) before attaching own handler.

### DIFF
--- a/app/assets/javascripts/active_admin/lib/batch_actions.es6
+++ b/app/assets/javascripts/active_admin/lib/batch_actions.es6
@@ -3,7 +3,7 @@ const onDOMReady = function() {
   // Use ActiveAdmin.modal_dialog to prompt user if
   // confirmation is required for current Batch Action
   //
-  $('.batch_actions_selector li a').click(function(event){
+  $('.batch_actions_selector li a').off('click').on('click', function(event){
     let message;
     event.stopPropagation(); // prevent Rails UJS click event
     event.preventDefault();


deivid-rodriguez:
Hi @sgara! Thanks for working on this! Is it possible to add a test? Maybe clicking on the batch action and check that a single modal opens?

deivid-rodriguez:
Also, can you add a changelog entry for this bug fix following the current changelog format? Remember to add links to your github profile and to this PR at the bottom of the file.

sgara:
Hi @deivid-rodriguez, thanks for the reply! I'd be happy to add the test, just wondering where to add the "turbolinks" gem, is `:test` section of `Gemfile.common` a good place to do it?

deivid-rodriguez:
Yes, that's the place! One you add it, run `bundle exec rake bundle` so that it gets added to the lock file for every Rails version we support :+1:.

roufique7:
I miss you

deivid-rodriguez:
I did some work to try reproducing this and write a regression spec on a separate test application that uses `turbolinks`, but I couldn't get it to reproduce.

I did try the fix manually and it works, so I propose to merge this as is (with a changelog entry).

@activeadmin/maintainers What do you think?

sgara:
Sorry for the lack of news! I did try to add some tests as well but failed to do so as well. Adding turbolinks and all necessary conditions was quite heavy, and then the spec couldn’t reproduce the issue. We had a cucumber scenario that was reproducing it in our own app though so it is doable somehow.

deivid-rodriguez:
Yes, I had the exact same problem. After enabling turbolinks in the test application, the cucumber scenario that I wrote would not reproduce the issue for some reason :grimacing:...

So, do you plan to give it another try? I'm very curious about why it wouldn't reproduce in a test environment.

sgara:
I can give it one more attempt over the next few days. I’d say that if I’m still stuck on Tuesday you can just merge it as is. I don’t want to spend months on such a small change 😅

deivid-rodriguez:
Yeah, I totally understand. I'm just curious about what's going on here, so thanks for giving it one more try :)

sgara:
Finally got a test reproducing the issue and fix 👍 
Closing this one in favour of #5815 (can push to this branch anymore)

PR #5501, opened 2018-10-06
Fix `if:` scope option with lambdas

deivid-rodriguez:
Fixes #5494.

PR #5564, opened 2018-11-14
Last backports

### LINKED ISSUE #5499: Documentation does not match released features
### Expected behavior

If I go to the documentation site, and I see a feature documented, I should be able to use it or at least be informed about the series it was (or will be released).

### Actual behavior

Features are documented in the website as soon as the land into master, but the release cycle is much slower.

### How to reproduce

See #5497 for an example.

deivid-rodriguez:
I completed the review of changes present in master, and the backporting. This release is a bit... weird, because you normally don't backport that many changes, but let's see how it works out. This is a sum up of the changes being backported here (and in general), and the result is my final proposal for the 1.4 release.

* Changelog, chores, and build changes.
* Documentation, so we can fix #5499 without issues.
* All translation additions and fixes.
* A couple more features:
  * Allow proc label in datepicker input. #5408
  * Add `priority` option to `action_item`. #5334
* A bunch of bug fixes:
  * Fixed the string representation of the resolved `sort_key` when no explicit `sortable` attribute is passed. #5464
  * Fixed docs on the column `sortable` attribute (which actually doesn't have to be explicitly specified when a block is passed to column). #5464
  * Fixed `if:` scope option when a lambda is passed. #5501
  * Comment validation adding redundant errors when resource is missing. #5517
  * Fixed resource filtering by association when the resource has custom primary key. #5446
  * Fixed "create anoter" checkbox styling. #5324


deivid-rodriguez:
Thanks!

deivid-rodriguez:
No problem, I'll finally do the release after we merge #5566!

PR #5577, opened 2018-11-19
Bump dependencies

deivid-rodriguez:
I haven't yet found a good way to automatically update dependencies. So I'll be running `bundle exec rake bundle[update]` periodically (which runs `bundle update` against every gemfile).


deivid-rodriguez:
Thanks for having a look @timoschilling! :smiley:

javierjulio:
Thanks @timoschilling! 😄👋🏻

PR #5580, opened 2018-11-21
Backport fix for sample app generation

deivid-rodriguez:
I don't think we'll release a 1.5 version and we'll go directly to 2.0. However, we might release 1.4.x bug fix releases, and being able to generate a sample application against 1-4-stable might help fixing any bugs that we want to fix on the branch.


javierjulio:
That's a good idea! Let me know when this is ready and I'll approve for you. Thanks!

deivid-rodriguez:
It's ready @javierjulio. Thanks!

PR #5583, opened 2018-11-21
Fix bug in menu item link with method delete

tiagotex:
Menu item links with method delete send a post request instead of a delete request.

In active admin 1.3.1
```html
<a rel="nofollow" data-method="delete" href="/users/sign_out">Logout</a>
```
In active admin 1.4.0
```html
<a method="delete" href="/users/sign_out">Logout</a>
```

This regression was introduced in https://github.com/activeadmin/activeadmin/commit/8e5c00d11ff67ffef7f761807acc6953c6c69ea7#diff-4a22354c40f395659ed3fd7cad761bf5R47

Browsers don't support the DELETE Verb,but rails uses it, so `link_to` puts the verb in `data-method` instead of the traditional `method` allowing rails to do it's magic. Source: https://stackoverflow.com/a/7465962

This restores the behavior present in 1.3.1

I created a feature test but I'm not sure this is the correct way to test this, would you suggest another way?

### DIFF
--- a/features/users/logging_out.feature
+++ b/features/users/logging_out.feature
@@ -3,7 +3,27 @@ Feature: User Logging out
   Logging out of the system as an admin user
 
   Scenario: Logging out successfully
-    Given I am logged in
+    Given a configuration of:
+    """
+      ActiveAdmin.setup do |config|
+        config.logout_link_method = :get
+      end
+    """
+    And I am logged in
     When I go to the dashboard
+    Then I should see the element "a[data-method='get']:contains('Logout')"
+    And I follow "Logout"
+    Then I should see "Login"
+
+  Scenario: Logging out sucessfully with delete method
+    Given a configuration of:
+    """
+      ActiveAdmin.setup do |config|
+        config.logout_link_method = :delete
+      end
+    """
+    And I am logged in
+    When I am on the dashboard
+    Then I should see the element "a[data-method='delete']:contains('Logout')"
     And I follow "Logout"
     Then I should see "Login"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@
   convert that file to plain JS (`//= require active_admin/base` if you
   didn't add any stuff to it).
 
+### Bug Fixes
+
+* Fix menu item link with method delete. [#5583][] by [@tiagotex][]
+
 ## 1.4.0 [☰](https://github.com/activeadmin/activeadmin/compare/v1.3.1...v1.4.0)
 
 ### Enhancements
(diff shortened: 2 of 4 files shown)


tiagotex:
Yeah, `add_logout_button_to_menu` is affected.

Changed the tests to test the logout button instead

deivid-rodriguez:
Nice! So this highlights an important missing cucumber scenario: logging out.

I would create a new file `features/users/logging_out.feature` and do the inverse tests to `features/users/logging_in.feature`. Something like "given that I am logged in, when I click logout, I can see a successful flash and a login link", for both `logout_link_method = :delete` and `logout_link_method = :get`.

In addition to that, you can maybe restore the [previous test](https://github.com/activeadmin/activeadmin/pull/5583/commits/4219ebc043b6023baaabbab0c26d5bb066f739f5#diff-7400c5c3977a37172831a973123066b7R41) that checks for the specific link html content.

What do you think?

features/users/logging_out.feature
```diff
@@ -5,9 +5,25 @@ Feature: User Logging out
   Scenario: Logging out successfully
     Given a configuration of:
     """
-      ActiveAdmin.register Post
+      ActiveAdmin.setup do |config|
+        config.logout_link_method = :get
+      end
     """
-    And I am logged in
+     And I am logged in
```
deivid-rodriguez: At least by looking at the existing logging in and logging out features, we seem to be using left alignment for cucumber keywords (`When`, `And`, `Given`, `Then`). Yesterday I starred a [gherkin-lint](https://github.com/vsiakka/gherkin-lint) project that might (or might not) help enforcing this, but for now let's keep the existing style so we can keep the diff free of style changes.

deivid-rodriguez:
Indeed! I missed that! I added a nitpick, but other than that I'm ready to merge this! :smile:

Again, thanks so much for fixing this! <3

features/users/logging_out.feature
tiagotex: I totally understand, I also prefer to keep things consistent 👍

features/users/logging_out.feature
```diff
@@ -5,9 +5,25 @@ Feature: User Logging out
   Scenario: Logging out successfully
     Given a configuration of:
     """
-      ActiveAdmin.register Post
```
deivid-rodriguez: Last thing I noticed from looking at the final diff is that this setup is unnecessary. I opened #5584 to clean it up in all cucumber scenarios related to user sessions. Once we merged that PR we can rebase this one, and get it in! :+1:

deivid-rodriguez: It's merged now! Can you rebase this PR and squash it to a single commit?

tiagotex: done 🙂

javierjulio: Great thanks!

deivid-rodriguez:
Thank you @tiagotex! Would you like to open a backport PR to 1-4-stable so we can release this bugfix? :smiley:

tiagotex:
Will do 👌

PR #5586, opened 2018-11-23
Backport fix bug in menu item links with method delete

tiagotex:
As suggested in https://github.com/activeadmin/activeadmin/pull/5583#issuecomment-441078851

To back port my change I also had to include your commit `
Remove unnecessary setup from session features`.

I'm not sure if this is the correct way to back port things since I've never done it before. Let me know if there's something I should fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Master (unreleased)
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
Going in!

PR #5588, opened 2018-11-23
Remove the "allow-rescue" cucumber tag

deivid-rodriguez:
None of the current tests seem to require this.


deivid-rodriguez:
Codecov check should have passed, according to our configuration, I opened a [ticket](https://github.com/codecov/codecov-ruby/issues/41) about it.

PR #5593, opened 2018-11-26
Use chandler to automate GitHub releases

deivid-rodriguez:
It's very common to forget to create a Github release every time you release a new version to rubygems. It's also duplicate work since it basically consists of duplicating the relevant entries in the changelog. So, I'm adding [chandler](https://github.com/mattbrictson/chandler) to our release task so everything is done automatically upon release.

As part of this change, I'm simplifying the style for changelog links. This is because the [releases page](https://github.com/activeadmin/activeadmin/releases) autolinks users and PR references (just like in github comments), so the brackets end up being redundant. By simplifying the links, it reads slightly better. Chandler could probably reconcile this, but I think this works for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -225,7 +225,7 @@ index download_links: ->{ can?(:view_all_download_links) || [:pdf] }
 ```
 
 * Comments menu can be customized via configuration passed to `config.comments_menu`. [#4187][] by [@drn][]
-* Added `config.route_options` to namespace to customize routes. [#4731][] by [@stereoscott[]]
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
Another solution that would make this read perfectly on the releases page would be to inline all links in the changelog instead of using link reference definitions. But that would hurt a bit the readability of the raw changelog. I don't have a strong opinion on either option.

deivid-rodriguez:
I added some release docs about chandler!

deivid-rodriguez:
I discovered that some of the new changelog specs were not being run (thanks to tracking coverage :)), and fixing that led to discovering a broken link on it!

This should be finally ready.

PR #5595, opened 2018-11-27
Remove codecov and use codeclimate instead

deivid-rodriguez:
Codecov keeps behaving weirdly and they don't have very good support at the moment, so I'd like to use codeclimate for coverage reporting instead.


javierjulio:
Sounds good! I know someone at CodeClimate if we ever need help, perhaps I can ask. Let me know when this is ready and I'll approve for you. Thanks!

PR #5596, opened 2018-11-27
Fix a link for comparing the versions v0.6.3 and v1.0.0 [skip ci]

akinomaeni:


deivid-rodriguez:
Thanks @akinomaeni, can you update the commit message to remove the "[ci skip]" part.

In this case this is a very simple fix where we could probably indeed skip CI, but we have several static checks that affect almost every file type (including the changelog), so I don't want to give the false impression that we support this tag. We want CI to run everytime.

akinomaeni:
I see. I will update the message.

deivid-rodriguez:
@javierjulio Since this PR is likely to conflict with #5593, merge this one first and I'll then fix conflicts myself on the other PR.

Issue #5518, opened 2018-10-14
Consider setting up bors-ng for evergreen master and never needed rebases


deivid-rodriguez:
### Expected behavior

* Master is always green.
* Contributors to never need to rebase PRs if they don't have any conflicts with master.

### Actual behavior

* Merging a bad combination of PRs could result on a red master. We don't have continuous deployment, we're not even a final application, so it doesn't sound like a huge deal. But situations like this usually require "follow-up PRs" that can be avoided.

* Contributors are asked to rebase their PRs, sometimes multiple times, which results in extra work and extra CI resources.

### How to reproduce

1. Open a PR enforcing a new rubocop rule. It's green :heavy_check_mark:
1. Open a PR with a new cool feature but not following the new rule. It's green :heavy_check_mark:
1. Merge the rubocop PR.
1. Merge the feature PR.
1. See the new master status: :heavy_multiplication_x:

### Proposed fix

Use https://bors.tech/. Instead of clicking on "Merge", you tell bors to do it through `@bors r+`. The bot will take care of fast-forwarding the changes on top of the latest master on a temporary branch, get that tested, and only if all status checks succeed, merge it.

This is a good visual explanation of the problem bors-ng solves: https://bors.tech/essay/2017/02/02/pitch/.

notriddle:
I bet @indirect would help set up bundlerbot for you. 😏

deivid-rodriguez:
@notriddle Since I got you here, do you think the hosted version will work for us, or do you recommed that we set it up on heroku or somewhere else ourselves (I think this is what bundler does?)?

notriddle:
There are basically three reasons that the hosted version wouldn't work:

* You are using private, or Enterprise, GitHub repositories
* You need better uptime and security guarantees than "a stranger on the internet runs it"
* You want custom patches

Most people who've asked me for help self-hosting work for companies that do proprietary software, so I'm guessing the first reason is the most popular reason to self-host.

But, even if you're OSS, most big projects should self-host. Remember what happened to homu.io? I intentionally made bors-ng easy to self-host just to avoid repeating that mess.

deivid-rodriguez:
I see, thanks! So I think I would try out first the hosted version (since it involves zero setup), and if we're happy with it after a while move to a self-hosted version.

PR #5600, opened 2018-11-29
Build pushes to all supported branches in CI

deivid-rodriguez:
Previously we had "build only PRs" turned on on CircleCI config, but CircleCI makes an exception with the master branch, and still runs pushes to it.

In my mind, 1-4-stable (or whatever the last stable branch is) and master should have the same setup, meaning when we push to 1-4-stable, CI should be run. This is particularly interesting because right after release, it feels good that the exact state of the branch you are releasing from has been tested on CI.

Also (although this is only a side effect of these changes), this makes it possible again to run CI on a branch you have not opened a PR from (by adding the branch to the list of branches in the CircleCI config). Although @javierjulio has encouraged us to not do this, I think it's still interesting to support it, so we can fit the workflow of more developers.

Finally, this makes it explicitly in the CircleCI config which branches we support at a given time, which is also interesting.


deivid-rodriguez:
It sounds like this is not currently possible: once I add the configuration, PRs no longer build. The feature was requested [here](https://discuss.circleci.com/t/is-there-a-way-to-build-all-prs-to-specific-branches-and-those-branches-themselves/24851), but no response yet.

Closing.

javierjulio:
@deivid-rodriguez perhaps it's worth commenting that we want to do this on ActiveAdmin and to give a reason why, basically share what we are trying to do here, that way it should hopefully clear up why its helpful/needed.

If we revisit this I won't have an issue with it by the way, I just feel there are some nice advantages to going PR only that you helped me realize and detail. If this helps keep CI consistent for master and stable branch releases though I'm fine with that. Thanks! 👍🏻

deivid-rodriguez:
> @deivid-rodriguez perhaps it's worth commenting that we want to do this on ActiveAdmin and to give a reason why, basically share what we are trying to do here, that way it should hopefully clear up why its helpful/needed.

I tried to do this in the PR description, paragraphs 2 and 4 would be the reasons to do it. I'll try to explain it again. When we release from a non master branch, CI has not been run for the exact code that it's going to be released. The closest thing to it is the last PR to that branch, but it's not 100% the same thing. In any case, this is not supported by CircleCI, and when we implement #5518 it will be fixed anyways, so no big deal! :smiley:.

deivid-rodriguez:
Exactly!

CircleCI by default does not build pushes to any branches, but makes an exception for master (see attached screenshot). I would like it to make exceptions for each supported branch, not only master. I tried to do that by toggling the UI option off, and making the changes proposed in this PR, but the result of it is that PR's are no longer built, only pushes to branches, that's why I closed this.

project settings activeadmin activeadmin circleci

deivid-rodriguez:
Aaaaaah, you were indeed misunderstood, now I see what you meant! :+1:

javierjulio:
@deivid-rodriguez haha no problem! I figured that was the case. I was wondering why was I getting a lengthy description and not CircleCI. 😂 Thanks! Sorry for the confusion.

PR #5601, opened 2018-11-30
Simplify gitignore

deivid-rodriguez:
This PR simplifies the gitignore file by removing stuff that I consider each developer should configure on her global gitignore instead.

But in reality this is just an excuse to open a PR and confirm that the configuration I changed for #5600 is not really working.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 tmp
 coverage
 .yardoc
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@javierjulio I added back the `.ruby-version` entry!

PR #5605, opened 2018-11-30
Enable RSpec/HookArgument cop

deivid-rodriguez:
The `:each` argument to RSpec hooks is redundant. This cop enforces that we don't pass it.


deivid-rodriguez:
Usually `rubocop-rspec` is harder to enforce than plain `rubocop`. But with our approach of only enabling selected cops that make sense for us, it's all good :)

PR #5609, opened 2018-11-30
Simplify cucumber setup

deivid-rodriguez:
And make it consistent with RSpec setup.

The "chore day" should be over now :)


Origa7jeffrey:
22424811909

PR #5610, opened 2018-12-01
More test setup improvements

deivid-rodriguez:
This is the last PR with test setup improvements. Now I think I'm fully happy with it, and tests have 0 warnings :)


javierjulio:
Haha don’t feel bad about it. I say keep the improvements coming! Always good to do. You mention 0 warnings, the test suite had many previously? I wasn’t aware. Thanks! 👍🏻

deivid-rodriguez:
No, not many, just one :). This PR fixes that one warning.

deivid-rodriguez:
Thanks @javierjulio!

PR #5611, opened 2018-12-01
Batch action regression when passing form as proc/lambda

buren:
Given

```ruby
# form = {type: ["a", "b"]}      # Works as intended
form = -> { {type: ["a", "b"]} } # ArgumentError

ActiveAdmin.register Post do
  batch_action(:flag, form: form) do
    redirect_to collection_path, notice: "Successfully flagged 10 posts"
  end
end
```

it will fail with `ArgumentError, wrong number of arguments (given 1, expected 0)`.
However if we instead pass a `Hash` instead of `proc`/`lambda` to `:form` then it works as intended.

I've added a passing feature test for the `Hash` case and a failing one for the `proc`/`lambda` case.

---

_Update_: This bug is _not_ present in `v1.3.x`, but _is_ present in `v1.4.x` - it was introduced in `v.1.4.0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 * Prevent leaking hashed passwords via user CSV export and adds a config option for sensitive attributes. [#5486] by [@chrp]
 
+### Bug Fixes
+
+* Fix `form` parameter to `batch_action` no longer accepting procs. [#5611] by [@buren] and [@deivid-rodriguez]
+* Fix passing a proc to `scope_to`. [#5611] by [@deivid-rodriguez]
+
 ### Removals
 
 * Rails 4.2 support has been dropped. [#5104] by [@javierjulio] and [@deivid-rodriguez]
@@ -373,6 +378,7 @@ Please check [0-6-stable] for previous changes.
 [#5537]: https://github.com/activeadmin/activeadmin/pull/5537
 [#5583]: https://github.com/activeadmin/activeadmin/pull/5583
 [#5608]: https://github.com/activeadmin/activeadmin/pull/5608
+[#5611]: https://github.com/activeadmin/activeadmin/pull/5611
 
 [@5t111111]: https://github.com/5t111111
 [@aarek]: https://github.com/aarek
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
I'm guessing the fact that the build passed means that this regression is not present on master? :thinking:

deivid-rodriguez:
Ha! It is a problem with the CI setup. Tests [failed](https://circleci.com/gh/activeadmin/activeadmin/5921?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), but they still report a success status.

I'll fix that issue and then I'll give it a try at fixing this bug! Thanks for the regression scenarios!

deivid-rodriguez:
Rebased on top of the CI fix applied on master, there should be now a proper regression scenario on this branch :+1:. I'll have a look into fixing the issue.

deivid-rodriguez:
Ok, it should be fixed now!

I initially fixed the issue via 71657a4, but I noticed that removing the whole input validation like in c394f1a did not make any tests fail, so I added an extra spec in 36bb5d21632adfe018957c72f5352f091ee0e149 to catch that.

@buren Can you check that this branch fixes the regression in your application?

buren:
@deivid-rodriguez yep that fixes the regression 👍

Thank you! ⭐️

_sidenote/fyi_: (unrelated from the above) I got `NoMethodError, undefined method 'delete' for :warning:Symbol` when calling `status_tag` – is there expected breaking chances in v2 for `status_tag`?

deivid-rodriguez:
Great! I'll release 1.4.3 with this fix soon!

Regarding `status_tag`, I do recall some breaking changes about it, but I also remember that we were emitting deprecation warnings about it... I'll try to have a closer look at it tomorrow.

deivid-rodriguez:
By the way, this regression seems to have been caused by https://github.com/activeadmin/activeadmin/pull/5176, so it sounds that a similar regression would've been introduced on dynamic `scope_to`'s. I'll have a look at this tomorrow as well.

deivid-rodriguez:
There was indeed the same regression in `scope_to` which should be fixed now. But now the fix for it makes me think that https://github.com/activeadmin/activeadmin/pull/5611/commits/71657a413ce733eaf9d121c54203f34da5f28a54 might not work in all cases and `render_in_context` might be the better fix. I'll investigate that and add another spec to catch that if necessary.

deivid-rodriguez:
This should be ready to go now!

deivid-rodriguez:
@buren I changed the fix a bit, mind giving it another try in your app?

buren:
@javierjulio, @deivid-rodriguez is works 👍

javierjulio:
@buren ok great! @deivid-rodriguez I'll get this in now. Thanks guys!

PR #5612, opened 2018-12-02
Fix CI status codes

deivid-rodriguez:
This PR fixes the problem detected by #5611.


deivid-rodriguez:
It does not actually close that issue, it only makes the regression spec actually fail CI.

deivid-rodriguez:
Thanks for having a look!

javierjulio:
Ah ok sorry! Thanks then for removing and also resolving the issue. 👍🏻

PR #5613, opened 2018-12-02
Less jruby shelling out

deivid-rodriguez:
This is a bit of a blind guess to try to fix jruby intermittent failures.


deivid-rodriguez:
I rerun CI in #5611 at about the same time as this PR, and while that one consistently fails on jruby, this one passed. So I think this might do the trick.

I think it is enough to make sure the bug report template works against the latest ruby & rails version, that's the main thing this PR changes.

deivid-rodriguez:
Also, this makes jruby jobs faster, of course, because of the high cost of spawing processes on jruby.

PR #5615, opened 2018-12-03
Better selenium capybara driver

deivid-rodriguez:
This PR catches up with the official way upstream capybara uses to register a chrome headless driver, and also increases selenium's default timeout to try to fix random CI failures like https://circleci.com/gh/activeadmin/activeadmin/6357.


deivid-rodriguez:
Also added a couple of extra tweaks to the cucumber setup.

PR #5617, opened 2018-12-03
Remove `load_defaults!`

deivid-rodriguez:
I wanted to do this improvement in the test suite for a while but I never got to it. Since we're having more flakies than usual at the moment, I decided to tackle it.

Previous specs were full with calls to `load_defaults!` and `reload_routes!` at random places, probably to reset state left around by other specs. There's two things I didn't like about that:

* The `load_defaults!` helper loaded several resources that most of the specs didn't need.
* It was pretty arbitrary that these lines were there, it's better that each spec takes care of cleaning after itself.

So in this PR I completely remove `load_defaults!`. If a resource is considered to be needed by all specs, it should live inside the generated application as a standalone file in `app/admin`. Instead, I'm using a new helper, `with_resources_during`, that receives the current example and a resource definition block, a defines those resource only for the duration of the example.

Doing this cleanup revealed a few state leakages that I tried to fix. It's likely that more will appear, I'll try to fix those as they come up.


javierjulio:
@deivid-rodriguez thanks for tackling this! After going through this in Inherited Resources, I know we'll be much better off here. Let me know when its ready and/or if you need help. I'll do what I can. Thanks!

deivid-rodriguez:
@javierjulio It's ready. Maybe some flakies will appear, but I'll deal with those as they come up!

PR #5622, opened 2018-12-04
Slim down packaged gem

deivid-rodriguez:
I was making sure that I was not releasing any unintented stuff with 1.4.3 and I noticed that the packaged gem of activeadmin includes many more files than necessary.

This PR slims down the activeadmin gem to include only the files needed at runtime, and also adds some other minor tweaks to the gemspec.


deivid-rodriguez:
I manually built the gem with the changes in this PR and installed it on one of my applications. Everything seems to be ok.

deivid-rodriguez:
Thanks Javier!

PR #5626, opened 2018-12-04
Remove one shelling out level in CI

deivid-rodriguez:
Trying to make jruby less flaky. We're getting a lot of "killed" errors, so I'm trying this based on https://github.com/jruby/jruby/issues/3739#issuecomment-199156622. According to that issue, this should've already been fixed on `jruby-launcher`, but I'm unclear how that library is distributed, so I'll try this directly anyways and see if it does the trick.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,9 @@ version: 2.1
 .run_tests: &run_tests
   run:
     name: Run tests
-    command: COVERAGE=true PARALLEL_TEST_PROCESSORS=4 bundle exec rake test
+    command: |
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Didn't work. I'll try another approach: stop shelling out when running `parallel_tests`.

deivid-rodriguez:
Also, I tried to include a fix for [this failure](https://circleci.com/gh/activeadmin/activeadmin/7012) in https://github.com/activeadmin/activeadmin/pull/5626/commits/1f8697f41cf182ae5198aced120dece6e3762463.

PR #5631, opened 2018-12-06
I18n tasks new checks

deivid-rodriguez:
Latest `i18n-tasks` version has a new check to find translations that don't use a correct number of interpolation variables with respect to the main language. I'm not sure if this check makes sense in all languages, so I wanted to ask.

Rails does not seem crash, whether you pass more or less variables than expected. If you call `I18n.t("translation", unknown_variable: "something")`, it ignores the variable. If conversely, you pass `I18n.t("translation")` to a translation defined as `translation: "Something %{missing_variable}"`, it interprets the "%" as a literal and translates literally to "Something %{missing_variable}".

Pinging the latest translators for each affected language, in case they can give feedback on whether this is an improvement or not on their language:

* Czech: @14113.
* Latvian: @patizivs.
* Romanian: @chalmagean.
* Slovakian: @mirelon.

Not pinging anyone for Lithuanian, since the new check just detected a plain translation bug in there where interpolation variables had been mistakenly translated.


config/locales/sk.yml
```diff
@@ -40,18 +40,18 @@ sk:
     sidebars:
       filters: "Filtre"
     pagination:
-      empty: "Nenájdený."
-      one: "Zobrazená  <b>1</b> položka"
-      one_page: "Počet zobrazených položiek %{n}"
-      multiple: "<b>%{from}&nbsp;-&nbsp;%{to}</b> z <b>%{total}</b>"
-      multiple_without_total: "<b>%{from}&nbsp;-&nbsp;%{to}</b>"
+      empty: "Nenájdený %{model}."
+      one: "Zobrazená  <b>1</b> %{model}"
```
mirelon: If the model is of female gender (položka = item), it is correct.
But, if e.g. the model is "auto", then pagination.one should be "Zobrazené 1 auto"

config/locales/sk.yml
```diff
@@ -40,18 +40,18 @@ sk:
...
+      one_page: "Počet zobrazených %{model} %{n}"
```
mirelon: An inflection should be used (plural genitiv)

config/locales/sk.yml
```diff
@@ -40,18 +40,18 @@ sk:
...
+      multiple: "%{model} <b>%{from}&nbsp;-&nbsp;%{to}</b> z <b>%{total}</b>"
```
mirelon: An inflection should be used (plural nominativ)

deivid-rodriguez:
@mirelon Thanks so much for the feedback, so in your opinion in these cases it's better to completely omit the model name and leave them as they are now, right?

Gender inflections seem pretty complicated to implement, although it would seem like a nice project :)

config/locales/sk.yml
deivid-rodriguez: You mean there are different pluralization for nouns in Slovakian, according to the rule they play in the sentence? I think we only support `model` or `plural_model`... :disappointed:

deivid-rodriguez:
> Yes, omitting the model name is the best solution for now...

Ok! I'll add exceptions for these hard keys, then. :+1:

I will look for some gem wrapping up the declination rules (https://en.wikipedia.org/wiki/Slovak_declension). But still, guessing gender and animate/inanimate of the model would be pretty hard...

Yeah, I think the solution would involve each model explicitly informing of its own gender. For the standard pluralization rules, you can add support [here](https://github.com/davidcelis/inflections).

deivid-rodriguez:
@mirelon I restored the previous versions of the locale file! Thanks for your feedback :+1:

PR #5632, opened 2018-12-07
Tackle Rails 5.2.2 support

deivid-rodriguez:
I'm guessing we'll probably need to bump the minimum ransack dependency.


deivid-rodriguez:
I propose to bump the minimum ransack dependency to ensure activeadmin always works under rails 5.2.2. This can be solved manually by end users by updating `ransack`, but I think this might be better? Thoughts @activeadmin/maintainers?

javierjulio:
@deivid-rodriguez I think that's best. It's been painless to update too. We haven't had issues. This would also mitigate issues reported here when all that's needed is to update ransack. I think its good we keep updating our minimum requirements as we do more releases. 👍🏻

deivid-rodriguez:
Ok!

nicolas-brousse:
Any idea of when this will be released?

deivid-rodriguez:
No, sorry. But you can run `bundle update ransack` and you should be fine.

nicolas-brousse:
My bad I though version was locked with the current version of ActiveAdmin. Thanks for your answer.

PR #5634, opened 2018-12-07
Nicer runtime dependencies

deivid-rodriguez:
Currently running `gem build activeadmin.gemspec` display a bunch of open-ended dependency warnings. The warnings are true. Most of our dependencies are open-ended, so if a dependency released a backwards incompatible version, even if it followed semantic versioning, activeadmin would pick it up and break.

So, I'm proposing to bound our dependencies so that it's less likely that 3rd party releases break activeadmin.

I also added a check to ensure the `gem build activeadmin.gemspec` builds without warnings going forward.

### DIFF
--- a/spec/gemspec_spec.lint.rb
+++ b/spec/gemspec_spec.lint.rb
@@ -14,7 +14,7 @@ RSpec.describe "gemspec sanity" do
     expect(build[1]).not_to include("WARNING")
   end
 
-  it "suceeds" do
+  it "succeeds" do
     expect(build[2]).to be_success
   end
 end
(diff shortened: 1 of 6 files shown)


spec/gemspec_spec.lint.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  let(:build) do
+    Open3.capture3("gem build activeadmin.gemspec")
+  end
+
+  it "has no warnings" do
+    expect(build[1]).not_to include("WARNING")
+  end
+
+  it "suceeds" do
```
javierjulio:
```suggestion
  it "succeeds" do
```

PR #5636, opened 2018-12-09
Gherkin lint

deivid-rodriguez:
I sometimes make comments on PRs that add cucumber features about little inconsistencies in the style (for example, https://github.com/activeadmin/activeadmin/pull/5548#discussion_r239782596). I'd like a basic gherkin style to be enforced automatically, so this is what this PR does.

### DIFF
--- a/.gherkin-lintignore
+++ b/.gherkin-lintignore
@@ -1,0 +1,1 @@
+vendor/bundle/**/*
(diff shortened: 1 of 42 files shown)


deivid-rodriguez:
> My only concern was using node and updating CI but I remember you had updated the base image to include node recently so we are good. Thanks!

Yeah, I had the same concern, but I noticed that I needed to add it to the lint job in 3da2364c9029feed4359f9b43cbbbd470e1aafb3, so I guess it's fine.

PR #5640, opened 2018-12-10
Flakiness fix attempt

deivid-rodriguez:
I found a potential explanation to the flakies we've been seeing recently:

Some specs create and remove files on the fly. Since we run tests in parallel, the creating and removal of these files by one process can affect autoloading in other processes.

So my attempt to fix this is to run this kind of specs that modify files sequentially.

Let's see if it works...


javierjulio:
Awesome thanks David! I merged in that PR. Let me know when this is ready.

deivid-rodriguez:
Thanks for having a look! Hopefully now we have a CI run that's both fast and stable! :crossed_fingers:

PR #5645, opened 2018-12-13
Fix gemfile consistency checks

deivid-rodriguez:
The release of bundler 1.17.2 broke our build because whereas our gemfiles are currently locked to 1.17.1, the new docker images already include (and use) bundler 1.17.2. That causes source control changes in the locked version when running `bundle install` and makes the checks fail.

This PR introduces binstubs which are meant precisely to avoid this kind of mismatches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ version: 2.1
 .install_dependencies: &install_dependencies
   run:
     name: Install dependencies
-    command: bundle install --path /home/circleci/project/vendor/bundle --retry 3 --jobs 3
+    command: bin/bundle install --path /home/circleci/project/vendor/bundle --retry 3 --jobs 3
 
 .install_chromedriver: &install_chromedriver
   run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2.1
 .install_bundler: &install_bundler
   run:
     name: Install a specific bundler version
-    command: gem install bundler -v 1.17.1
+    command: gem install bundler -v 1.17.3
 
 .copy_current_gemfile: &copy_current_gemfile
   run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,11 @@
 
 version: 2.1
 
+.install_rubygems: &install_rubygems
+  run:
+    name: Install a specific rubygems version
+    command: gem update --system 3.0.2
+
 .install_bundler: &install_bundler
   run:
     name: Install a specific bundler version
@@ -115,6 +120,7 @@ version: 2.1
 
 .test_steps: &test_steps
   - checkout
+  - *install_rubygems
   - *install_bundler
   - *copy_current_gemfile
   - *restore_cache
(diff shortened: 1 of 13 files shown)


deivid-rodriguez:
Yeah, I can see how this is a bit confusing without more explanations... Sorry!

The `bundle` binstub is a special binstub generated by bundler (through `bundle binstubs bundler`), although I slightly modified it to remove some stuff I don't think we need (for example, old rubies support), and to fit our code style. In particular, I could have removed the `cli_arg_version` too, because it implements a feature of rubygems binstubs that none of us probably uses (if you specify `_<version>_`
 after a binstub, it uses that version. For example, `bundle _1.17.0_ install` will use the 1.17.0 version of `bundler` if present).

The idea of this binstub is that it reads the `BUNDLED_WITH` section of the lockfile (the bundler version the lock file was generated with), and makes sure that exact version is used when `bundler` is run. The point of this is the same point as having a lock file, only for `bundler` itself: make sure a deterministic fixed version is always used by developers and CI, in order to avoid "it works on my environment" kind of problems.

As per why this discrepancy appeared: The base docker images have been updated to use the latest version of bundler (1.17.2) whereas our lockfiles are locked to the previous one (1.17.1). We have a check that runs `bundle install` on every gemfile, and makes sure no version control differences are generated, so that we keep our lock files in sync with their respective gemfiles. Unless we use a `bundler` binstub for this check, the latest version of bundler will be used and that will generate a lock file with `BUNDLED_WITH 1.17.2`, thus generating these source control differences.

A simpler way to fix this was to actually upgrade the `BUNDLED_WITH` section of every lockfile to 1.17.2, but I wanted to fix it "properly" since it is my opinion, no external release of a gem should ever break our build. But I will do a follow up PR to upgrade this section too.

deivid-rodriguez:
@javierjulio I removed the `cli_arg_version` method, but I also tried upgrading to `bundler 1.17.2` and it crashed on `jruby`, I'm not sure why.

Just upgrading the locked bundler version like I did in #5646 worked without any further changes, though.

I think this is not the right solution since any release of `bundler` will break our build, but since the repo is very actively maintained at the moment it shouldn't be a big deal.

So, I downgraded this PR to "Work in Progress" until I can find the root cause for these jruby crashes, and I propose #5646 instead to get the build green for now.

javierjulio:
@deivid-rodriguez sorry I wasn't able to get back to you yesterday. I didn't have time to respond as I had company events on Wednesday and Thursday. 

What you mentioned all makes sense. I didn't know the binstub was generated by Bundler. I figured you had just written it. The info is good to know so thanks for clarifying! It was very helpful. I agree that it shouldn't break our builds.

I merged in #5646 so we should be good for now. Before I also wanted to mention, thanks for also updating the docs to make sure to use the binstub! Let me know where you get with this and we'll resolve it. Thanks!

deivid-rodriguez:
The reason for the jruby failure is that I was installing `1.17.1` but the locked version was `1.17.2`. Since the jruby images didn't come with bundler 1.17.2 by default, hence the failure.

I made sure the version of bundler locked in the `Gemfile.lock` file and the version we install on CircleCI match, so that we don't depend on the specific version of bundler the docker images include.

deivid-rodriguez:
Also made sure rubygems 3.0.2 is installed since a resolver bug was fixed in this version.

PR #5650, opened 2018-12-18
Added workaround for non-transliteratable tab titles

### LINKED ISSUE #5484: Tabs are not correctly created when using non-latin characters as title

panasyuk:
Fixed code that generates fragment link and id for tab content for non-transliteratable tab titles.

Fixes #5484.
Closes #5508.

### DIFF
--- a/lib/active_admin/views/components/tabs.rb
+++ b/lib/active_admin/views/components/tabs.rb
@@ -32,7 +32,7 @@ module ActiveAdmin
 
       def fragmentize(string)
         result = string.parameterize
-        result = Base64.urlsafe_encode64(string) if result.blank?
+        result = URI.encode(string) if result.blank?
         result
       end
     end
--- a/lib/active_admin/views/components/tabs.rb
+++ b/lib/active_admin/views/components/tabs.rb
@@ -32,7 +32,7 @@ module ActiveAdmin
 
       def fragmentize(string)
         result = string.parameterize
-        result = string.hash if result.blank?
+        result = Base64.urlsafe_encode64(string, padding: false) if result.blank?
         result
       end
     end
--- a/lib/active_admin/views/components/tabs.rb
+++ b/lib/active_admin/views/components/tabs.rb
@@ -15,7 +15,8 @@ module ActiveAdmin
       end
 
       def build_menu_item(title, options, &block)
-        fragment = options.fetch(:id, title.parameterize)
+        fragment = options.fetch(:id, fragmentize(title))
+
         html_options = options.fetch(:html_options, {})
         li html_options do
           link_to title, "##{fragment}"
@@ -23,8 +24,16 @@ module ActiveAdmin
       end
 
       def build_content_item(title, options, &block)
-        options = options.reverse_merge(id: title.parameterize)
+        options = options.reverse_merge(id: fragmentize(title))
         div(options, &block)
+      end
+
+      private
+
+      def fragmentize(string)
+        result = string.parameterize
+        result = string.hash if result.blank?
+        result
       end
     end
   end
--- a/lib/active_admin/views/components/tabs.rb
+++ b/lib/active_admin/views/components/tabs.rb
@@ -32,7 +32,7 @@ module ActiveAdmin
 
       def fragmentize(string)
         result = string.parameterize
-        result = Base64.urlsafe_encode64(string, padding: false) if result.blank?
+        result = Base64.urlsafe_encode64(string) if result.blank?
         result
       end
     end
--- a/spec/unit/views/components/tabs_spec.rb
+++ b/spec/unit/views/components/tabs_spec.rb
@@ -59,7 +59,7 @@ RSpec.describe ActiveAdmin::Views::Tabs do
       end
 
       it "should create a tab navigation bar based on the symbol" do
-        expect(tabs.find_by_tag('li').first.content).to include "Overview"
+        expect(tabs.find_by_tag('li').first.content).to include("Overview")
       end
 
       it "should create a tab with a span inside of it" do
@@ -71,7 +71,9 @@ RSpec.describe ActiveAdmin::Views::Tabs do
       let(:tabs) do
         render_arbre_component do
           tabs do
-            tab '🤗'
+            tab '🤗' do
+              'Tab content'
+            end
           end
         end
       end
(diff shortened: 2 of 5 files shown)


lib/active_admin/views/components/tabs.rb
```diff
@@ -15,17 +15,26 @@ def build(&block)
...
-        options = options.reverse_merge(id: title.parameterize)
+        options = options.reverse_merge(id: fragmentize(title))
         div(options, &block)
       end
+
+      private
+
+      def fragmentize(string)
+        result = string.parameterize
+        result = string.hash if result.blank?
```
timoschilling: This is not safe! `string.hash` returns different values in each process.
```
➜ ruby -e 'puts "string".hash'
-1157288931546009962
➜ ruby -e 'puts "string".hash'
3581747432009468955
```
This means that the tabs didn't work across different server instances!

timoschilling:
maybe `'👍'.ord == 128077` can help

lib/active_admin/views/components/tabs.rb
deivid-rodriguez: Is this `id` used in the server side? I thought we just needed it to be unique per page and play nice with our js.

deivid-rodriguez:
> maybe '👍'.ord == 128077 can help

That's a per-character thing, so wouldn't work for us...

```
> "Title 1".ord # => 84
> "Title 2".ord # => 84

File: lib/active_admin/views/components/tabs.rb (hunk 1 above)
Comment (timoschilling) [reply lib/active_admin/views/components/tabs.rb]:
```suggestion
        result = Digest::MD5.hexdigest(string) if result.blank?
```

spec/unit/views/components/tabs_spec.rb
```diff
@@ -66,5 +66,29 @@
...
+      end
+
+      let(:subject) { Capybara.string(tabs.to_s) }
+
+      it "should create a tab navigation bar based on the string" do
+        expect(subject).to have_content('🤗')
+      end
+
+      it "should have tab with id based on hash of the string" do
+        expect(subject).to have_selector("div##{'🤗'.hash}")
```
timoschilling:
```suggestion
        expect(subject).to have_selector("div##{Digest::MD5.hexdigest('🤗')}")
```

spec/unit/views/components/tabs_spec.rb
```diff
@@ -66,5 +66,29 @@
...
+      end
+
+      it "should have link with fragment based on hash of the string" do
+        expect(subject).to have_selector(%{a[href="##{'🤗'.hash}"]})
```
timoschilling:
```suggestion
        expect(subject).to have_selector(%{a[href="##{Digest::MD5.hexdigest('🤗')}"]})
```

timoschilling:
`Digest::MD5.hexdigest` should do the trick!

lib/active_admin/views/components/tabs.rb
timoschilling:
```suggestion
        result = Digest::MD5.hexdigest(string) if result.blank?
```

timoschilling: yes for the js it will work, but if you reload the page it will maybe break, depending on which server or which process your hit with the request.

panasyuk: Inconsistent string hash values would brake nothing in terms of navigation. Though using MD5 is definitely better.
P.S.
Thanks for the tip @timoschilling

timoschilling: It will be break if you are on an tab and reload the page an get the request from a different server / instance.

You can reproduce it it:
- one an tab with hash value as id
- restart your server (if you use pow or puma local, you need to restart that too)
- reload the page and you will be on the first tap instatt of the hash value tab

deivid-rodriguez: I don't see how it could break, but I'm fine with using digests too :+1:. It might help with caching, maybe, and I guess it gives more uniqueness guarantees, which in the unlikely case of collisions, would indeed break navigation.

deivid-rodriguez: I saw https://github.com/activeadmin/activeadmin/pull/5650#discussion_r now, nice catch!

panasyuk: What about Base64 encoded string? It's shorter and more recognizable.

deivid-rodriguez: No problem, we have Timo! :)

deivid-rodriguez: By the way, I checked this on Chromium and we have this problem for all kind of tabs, and without restarting the server. It sounds like we need to update the current URL with the tab id.

But when we fix that, the issue would reproduce as Timo said, and his suggestion would fix it.

deivid-rodriguez: > What about Base64 encoded string? It's shorter and more recognizable.

Yeah, that sounds even better! [`urlsafe_encode64`](https://ruby-doc.org/stdlib-2.5.3/libdoc/base64/rdoc/Base64.html#method-i-urlsafe_encode64) maybe?

panasyuk: @deivid-rodriguez Sure 🙂

panasyuk:
@timoschilling @deivid-rodriguez I'm sorry for being noizy. Eventually i came to `URI.encode(title)` which is much better for URLs.

lib/active_admin/views/components/tabs.rb
```diff
@@ -15,17 +15,26 @@ def build(&block)
...
+
         html_options = options.fetch(:html_options, {})
         li html_options do
           link_to title, "##{fragment}"
         end
       end
 
       def build_content_item(title, options, &block)
-        options = options.reverse_merge(id: title.parameterize)
+        options = options.reverse_merge(id: fragmentize(title))
```
timoschilling: `URI.encode` is a good catch, but I think with it we no longer need `fragmentize` and can do:
```suggestion
        options = options.reverse_merge(id: URI.encode(title))
```

panasyuk: But it will brake existing fragments.
```ruby
'a b'.parameterize #=> "a-b"
URI.encode('a b') #=> "a%20b"
```

panasyuk: Some apps can still rely on parameterized titles.

deivid-rodriguez:
Thank you so much @panasyuk for fixing this, and @timoschilling for your helpful review!

panasyuk:
Thank you. You are welcome guys. I'm happy to contribute to the gem. But where would the problem be without your attention to details? Great job @timoschilling @deivid-rodriguez

PR #5649, opened 2018-12-17
Reorganize README

deivid-rodriguez:
This PR is motivated by #5648. I'm totally happy with the idea of having an opencollective page, but we need to stay fair towards Tidelift, which is our currently preferred means of funding.

This PR improves the contributing section by:

* Merging the "Want to contribute?" and "Want to support us?" section into a single one, since it's all about contributing after all.
* Adding some thanks in there, since we're grateful for all contributions :)
* Expanding a bit on what a Tidelift subscription means, and how it is a good thing for organizations.
* Reduce a bit the visibility of Liberapay, since it was previously more visible than Tidelift.

Supersedes #5648.
Closes #5559.

PR #5652, opened 2018-12-19
Link directly to activeadmin donation's page

deivid-rodriguez:
After the README cleanup in #5649, we lost the link to directly donate through Liberapay. Without it, it is much harder to donate through this platform, so I'm restoring it.


javierjulio:
No problem, I got you! 💪🏻

PR #5656, opened 2019-01-01
Bundler 1.17.3

deivid-rodriguez:
Make sure we use the latest bundler 1.17.3 in CI.


deivid-rodriguez:
#5645 is a better way to fix this I think, closing.

PR #5657, opened 2019-01-02
Test against ruby 2.6

deivid-rodriguez:


.circleci/config.yml
```diff
@@ -208,6 +208,7 @@ jobs:
   lint_and_docs:
     docker:
       - image: circleci/ruby:2.6.0-node
+        user: root
```
javierjulio: @deivid-rodriguez do you have a build where you saw this happen whether it was as log output or a build failure?

deivid-rodriguez: Yes, see for example the `gem update` step in https://circleci.com/gh/activeadmin/activeadmin/11012. It's not failing the step, for some reason, but the rubygems update is failing due to permissions.

In any case, I finally found the culprit here and it doesn't seem related to the rubygem versions we're using, so I see 2 options:

* Always use the `root` user, so we can upgrade and use a consistent version of rubygems.
* Don't use a consistent version of `rubygems`, but instead use the one vendored with ruby.

deivid-rodriguez: Let's go with the second!

deivid-rodriguez: Turns out rubygems 3.0.2 is actually needed for the lint_and_docs jobs. Not exactly sure why but it looks like a bug fix in the latest rubygems is needed. Given that, I feel option 1 is better.

deivid-rodriguez:
@javierjulio This is ready.

Funny enough, this PR was running into bundler/bundler#6937, which for other reasons, I started looking at yesterday. Only then I found out that we were hitting it. The issue can be workarounded by running `gem update --system 3.0.2` (ruby 2.6 ships with rubygems 3.0.1), so when fixing the rubygems installation by running the command as root, everything started working.

PR #5659, opened 2019-01-06
Bundler 2

deivid-rodriguez:
This PR upgrades the bundler locked version to 2.0.1. Bundler 2 provides version autoswitching  according to the version present in the lockfile, so a binstub for bundler itself is no longer needed.

In order for bundler 2 to work in CI, I also needed to override the `BUNDLER_VERSION` environment variable because of https://github.com/docker-library/ruby/issues/246. The fix for this has already been added to the official docker images for ruby, but have not yet been deployed. Once the fixed images are available, the overriding of the `BUNDLER_VERSION` environment variable can be removed.


deivid-rodriguez:
Thanks @zorab47! :heart:

Issue #2283, opened 2013-06-18
pagination uses COUNT(*) which doesn't work for large databases


amonks:
When loading a multiple-page index, the pagination thing renders "Displaying [Things] 1 - 30 of [SELECT COUNT(*) FROM "Things"] in total"

On a large enough database, this COUNT(*) query can require a prohibitive amount of time and memory (see #2100 )

There should be a simple way to turn off the "of [total] in total" part

PR #5660, opened 2019-01-11
Use collection length instead of running COUNTs for limited collections

MmKolodziej:
There have been many issues with this method's speed. While investigating some performance issues today, I've stumbled upon it as well and saw a lot of queries like:
```sql
SELECT COUNT(*) FROM (SELECT  1 AS one FROM `table` LIMIT 30 OFFSET 0) subquery_for_count
```
generated.

The purpose of this PR is not to issue such SELECTs for the main collection being rendered, where standard pagination methods are being applied. The collection will be loaded anyway, so we may simply check how many objects we have.
If the collection is unlimited (e.g. using `scope` to display a filtered collection), original behavior is preserved.

### DIFF
--- a/lib/active_admin/helpers/collection.rb
+++ b/lib/active_admin/helpers/collection.rb
@@ -5,6 +5,7 @@ module ActiveAdmin
       # 2. correctly handles the Hash returned when `group by` is used
       def collection_size(c = collection)
         return c.count if c.is_a?(Array)
+        return c.length if c.limit_value
 
         c = c.except :select, :order
 
--- a/spec/unit/views/components/paginated_collection_spec.rb
+++ b/spec/unit/views/components/paginated_collection_spec.rb
@@ -241,6 +241,15 @@ RSpec.describe ActiveAdmin::Views::PaginatedCollection do
         .not_to make_database_queries(matching: "SELECT COUNT(*) FROM \"posts\"")
     end
 
+    it "makes no COUNT queries to figure out the last element of each page" do
+      require "db-query-matchers"
+
+      undecorated_collection = Post.all.page(1).per(30)
+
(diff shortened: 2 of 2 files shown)


waymondo:
Investigating superfluous COUNT queries lead me to this PR which solves our performance issues.

munen:
I found this issue, because I also see redundant 'count' queries. The original issue from 2013 (https://github.com/activeadmin/activeadmin/issues/2283) claims a fix. But as documented in this closed issue from 2013-2015 it doesn't seem to work for everybody: https://github.com/activeadmin/activeadmin/issues/2638

We have integrated this line into our codebase as a patch and are happy to save 50% of the query time! No downsides have been seen so far.

munen:
@pawelma You've reviewed and approved these changes about a year ago. I'm pinging in case this PR just was forgotten to be merged. Apologies for the notification if there is another reason not to merge it.

MmKolodziej:
We've also used this in our codebase, for over 2 years now, without any issues.

pawelma:
@munen, unfortunately, I don't have write access which means that I'm unable to merge it ;(

deivid-rodriguez:
I forced-push to this branch to add a spec and will be merging it shortly.

@MmKolodziej I hope you were not using this branch directly and I didn't mess up anything :pray:.

javierjulio (approved):

javierjulio:
Thank you!

deivid-rodriguez:
Thanks!!

allisonphillips:
I also came across this because `pagination_total` is not working for me. I appreciate that your spec here is testing whether the query is made -- the original PR where `pagination_total` was added did not test for this (though to be fair, the developer flagged that they were concerned about their tests and wanted feedback/guidance on them in the PR description)

PR #5661, opened 2019-01-15
Unify specifying label in show -> attributes_table

### LINKED ISSUE #167: Specify label for attributes_table row
I want ability to specify label for row in attrubtes_table like for index column headers. E.g.:

``` ruby
show do
  attributes_table do
    row :name
    row "Student", :user
  end
end
```

As I understand from code it's impossible right now

### LINKED ISSUE #3359: Ability to change labels in show -> attributes_table
Is it possible?

mirelon:
Up to now, the only way to specify a label in a row in attributes_table was by using the first argument. The data was then taken from executing a block:
```
attributes_table do
  row 'Renamed' do |r|
    r.username
  end
end
```

This pull request allows for syntax:
```
attributes_table do
  row 'Renamed', :username
end
```

The same syntax is used by index:
```
index do
    column 'Renamed', :username
end
```

Closes #3359, closes #167

### DIFF
--- a/lib/active_admin/views/components/attributes_table.rb
+++ b/lib/active_admin/views/components/attributes_table.rb
@@ -22,6 +22,7 @@ module ActiveAdmin
       def row(*args, &block)
         title   = args[0]
         options = args.extract_options!
+        data  = args[1] || args[0]
         classes = [:row]
         if options[:class]
           classes << options[:class]
@@ -36,7 +37,7 @@ module ActiveAdmin
           end
           @collection.each do |record|
             td do
-              content_for(record, block || title)
+              content_for(record, block || data)
             end
           end
         end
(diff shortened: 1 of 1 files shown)


mingchang:
So this has been sitting around for a year due to an extra space in the pull request?

```
lib/active_admin/views/components/attributes_table.rb:25:13: C: Layout/ExtraSpacing: Unnecessary spacing detected.
        data  = args[1] || args[0]
            ^

355 files inspected, 1 offense detected
```

deivid-rodriguez:
The reason is not the space, but that nobody has been able to review it yet. I'll try to find some time.

mingchang:
Cool, got it. I think this is a great addition to ActiveAdmin due to much cleaner code in customizing "show" pages. Would clear up some confusion I've seen around the net as well from devs who found this new syntax much more intuitive. Thanks for the quick response!

javierjulio:
@mirelon I will accept this if a test is added. Thank you.

PR #5662, opened 2019-01-17
Undeprecate external asset registration

### LINKED ISSUE #5170: Including external JavaScript that you can't just copy
#5060 deprecated including javascripts in favor of using the asset pipeline. While good in spirit, this makes the world for external javascripts that you can't just copy locally a lot harder [as pointed out](https://github.com/activeadmin/activeadmin/pull/5060#issuecomment-325805074).

Would be great if we had an easy (tm) way to also include these sorts of javascript resources (like google maps etc.).

Thanks for all the work!
Tobi

deivid-rodriguez:
We might reevaluate this in the future, but for now I'm reverting while we figure out good enough alternatives so we can unblock releasing other improvements.

Closes #5170.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 #### Minor
 
 * Add better support for rendering lists. [#5370] by [@dkniffin]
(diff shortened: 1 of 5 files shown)


benjaminwood:
Is it expected that this change will make it into a 1.x release? Or is it held for 2.0?

Issue #5209, opened 2017-10-17
wrong dependency to rails


timoschilling:
### Expected behavior

Active Admin's gemspec didn't specify a dependency to the `rails` gem, which is ok, while for example `inherited_resources` specifies a dependency to `actionpack`.

For this reason I expect that I can use `activeadmin` without using `rails`, I only want to use `actionpack` (the reason is that I don't need actioncable).

### Actual behavior

If I run any command which boots the rails app, I get this error:
```
ActiveAdmin::DependencyError:
  To use rails you need to specify it in your Gemfile.
```

### How to reproduce

Just change a standard Gemfile from
```ruby
gem 'rails'
gem 'activeadmin'
```
to
```ruby
gem 'railties'
gem 'activerecord'
gem 'actionpack'
gem 'actionview'
gem 'actionmailer'
gem 'activejob'
gem 'activeadmin'
```

### How to fix it

This is my current hotfix
```ruby
require 'active_admin/dependency'
module ActiveAdmin
  module Dependency
    class Matcher
      def initialize(name)
        name = 'railties' if name.to_s == 'rails' # should work with 'actionpack' too
        @name = name
      end
    end
    module Adapter
      class Railties < Rails
      end
    end
  end
end
```

javierjulio:
@timoschilling I don't understand why you're doing that. From what I know, to do what you want, you still specify the rails gem as a dependency but instead in `application.rb` replace `require "rails/all"` with explicit requires for the gems you want to use, for example:

```ruby
require "active_record/railtie"
require "action_controller/railtie"
require "action_view/railtie"
require "action_mailer/railtie"
require "active_job/railtie"
```

I bring this up because we too don't use actioncable, at least as of now we don't want to include it until we plan to use it. If I can find the resource as to why I did it this way, I'll share.

timoschilling:
@javierjulio with your way you still need to install `actioncable` and it's dependencies. If we fix the `activeadmin` dependency checking, that's not needed.

javierjulio:
@timoschilling correct but its not loaded. From what I understand that's the way to get around to do what you want. The dependency doesn't change since its still Rails. I'm not sure if its intended to install these gems directly, although of course nothing stopping anyone from doing that. I wanted to share what we were doing since that's how we get around not loading other gems we aren't using which can be very useful.

zorab47:
Any updates here?

timoschilling:
[...]

PR #5670, opened 2019-01-23
fix dependency detection

timoschilling:
fixes: #5209


deivid-rodriguez:
Should we actually add `actionpack` as a dependency? :thinking:

timoschilling:
we have a implicit dependency through `railties` which should be enough

PR #5677, opened 2019-02-07
Fix chandler reference again

deivid-rodriguez:
Hopefully the PR will get attended soon. For now I'm rebasing it periodically fixing conflicts, and thus breaking our build :sweat_smile:.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 eval_gemfile(File.expand_path("Gemfile.common", __dir__))
 
 group :release do
-  gem 'chandler', git: 'https://github.com/deivid-rodriguez/chandler', branch: 'submit_link_references' # Github releases from changelog
+  gem 'chandler', git: 'https://github.com/deivid-rodriguez/chandler', tag: '0.8.0_plus_submit_link_references' # Github releases from changelog
 end
 
 group :lint do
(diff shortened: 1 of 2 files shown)


javierjulio:
@deivid-rodriguez no problem at all! I've dealt with this actually because its how I use my ActiveAdmin fork. A suggestion to avoid this would be when you make changes/rebase/etc do that on a new branch. You could name it with a version or date. This way it won't break other builds pointing to an old revision.

deivid-rodriguez:
Yeah! I always realized a killed the revision we're depending on right after I force push :sweat_smile:. Should I create a copy named differently from the branch that holds the chandler PR then?

javierjulio:
I've been there. 😂 It's up to you really. As long as you create a new branch from the latest changes you had, and then make any updates on that new branch, it'll be fine. You can name it how you like. You could use a date or semantic version in the branch name if that helps.

By the way I was fine to merge this in but I didn't because it didn't request review or have the `ready` label so I wasn't sure if you preferred to hold on this for now. Let me know as I can approve and merge in!

deivid-rodriguez:
I think using a different branch to which I don't force push is a good idea! Let me get this ready.

javierjulio:
@deivid-rodriguez great thanks! Set me as reviewer once its ready and I'll get this in for you. 🚀

deivid-rodriguez:
I pointed to a tag instead, which seems like a better fit here! :+1:

PR #5682, opened 2019-02-21
Replace cancan with cancancan

coorasse:
cancan project is deprecated and not updated since some years. Switch to `cancancan` (maintained version) to avoid confusion.

### DIFF
--- a/docs/0-installation.md
+++ b/docs/0-installation.md
@@ -11,7 +11,7 @@ gem 'activeadmin'
 
 # Plus integrations with:
 gem 'devise'
-gem 'cancan' # or cancancan
+gem 'cancancan'
 gem 'draper'
 gem 'pundit'
 ```


javierjulio:
Thanks! ❤️

deivid-rodriguez:
Actually, can you also use `cancancan` in our gemfiles so that we verify there's no regressions here?

You would need to do something like the following: clone the repo, run `bundle`, change `Gemfile.common` to use `cancancan` instead of `cancan`, run `bundle exec rake bundle`, and commit the result.

deivid-rodriguez:
Late to the party, I'll do it! :)

javierjulio:
@deivid-rodriguez sorry haha, let me know when that's ready.

coorasse:
> Actually, can you also use `cancancan` in our gemfiles so that we verify there's no regressions here?
> 
> You would need to do something like the following: clone the repo, run `bundle`, change `Gemfile.common` to use `cancancan` instead of `cancan`, run `bundle exec rake bundle`, and commit the result.

ohhhhhh.....yes

PR #5504, opened 2018-10-07
Switch to sassc-ruby / sassc-rails

deivid-rodriguez:
This is some initial work to fix #5502. Something I've noticed is that this effectively drops support for `sass-rails`, so we might as well reintroduce the `sassc-rails` runtime dependency since we originally switched from `sass-rails` to plain `sass` in order to support `sassc-rails`, but now `sassc-rails` will be the only supported implementation :man_shrugging:.

Closes #5502.
Closes #5226.

PR #5690, opened 2019-02-25
Replaced `sass` with `sassc`

zocoi:


deivid-rodriguez:
@zocoi Thanks! This is already being worked at #5504. I'm closing this, but feel free to leave any feedback you may have in there!

PR #5694, opened 2019-02-28
Update devise

renanbona:
- Remove deprecation method
- Create partial to show devise errors


Gemfile.common
```diff
@@ -8,7 +8,7 @@ gem 'cancancan'
 gem 'pundit'
 gem 'jruby-openssl', '~> 0.10.1', platform: :jruby
 gem 'draper', '~> 3.0'
-gem "devise", '~> 4.4'
+gem "devise", '~> 4.6.1'
```
deivid-rodriguez: Minor nitpick. I would leave `~> 4.6` here. The specific patch level version that we're using is already in the lock files, and the `~> 4.6` alternative conveys the information that `devise` follows semver.

PR #5695, opened 2019-02-28
fix pagination information

jdugarte:
Currently, with the default per page of 30, it's showing this:
Displaying 1 - 909 of 909 in total

This commit changes it to:
Displaying 1 - 30 of 909 in total


jdugarte:
I got stuck in an issue setting up the environment for tests. I did `bundle install`, but I can't do the sample rails app's migrations:
```
$ bin/rake local db:migrate
bundler: command not found: db:migrate
Install missing gem executables with `bundle install`
```
Is this a known issue? I can't seem to find anything in the issues, or a general google search.

deivid-rodriguez:
Hi @jdugarte, sorry about that, it was an (unknown) issue :). I fixed it in #5697.

But in any case, that task should normally not be needed. `bin/rake local s` should automatically create a test app, migrate the DB, and boot it, so you can try out things.

jdugarte:
@deivid-rodriguez I've added some tests in [1bd6b82](https://github.com/activeadmin/activeadmin/pull/5695/commits/1bd6b828dbc10000dae1d7e1c69074ce8e75f5d9), although those tests will also pass in master since the data being fed to the test cases is only arrays (`Kaminari.paginate_array([Post.new] * 81)`), which in [ActiveAdmin::Helpers::Collection#collection_size](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/helpers/collection.rb) returns from `return c.count if c.is_a?(Array)`, for an already paginated array is only a page long, efectively hiding the issue from the tests.

deivid-rodriguez:
Hei @jdugarte. Thanks for adding the initial spec.

I do want this to be a proper regression spec. We don't want to test the behavior that already works, we want to this the behavior that we are changing. Can you change the spec the not be fed up an array but a relation, so that the current issue manifests, and your spec proves the fix?

jdugarte:
@deivid-rodriguez You're right, I've added some more tests

deivid-rodriguez:
@jdugarte All new specs seem to be passing for me without your fix. Can you double check there's actually a bug here? It seems like a too common case to not have been noticed by anyone.

jdugarte:
@deivid-rodriguez I've been trying to build a test case for this, but unfortunately, I don't know enough of the code base and I'm lost. But I do know what's the difference between the path the tests are going through and the path my app goes. Let me try to explain it here, in case you can either build a test case for this or point me in the direction of how to do it myself.

Here's what the test is doing:

1. It creates [5 post records](https://github.com/jdugarte/activeadmin/blob/JD-fix-pagination-information/spec/unit/views/components/paginated_collection_spec.rb#L233).
2. It [paginates them right away](https://github.com/jdugarte/activeadmin/blob/JD-fix-pagination-information/spec/unit/views/components/paginated_collection_spec.rb#L234), creating a collection of actually two elements when the actual collection (`Post.all`) has 5 elements.
3. It [calls the `paginated` block](https://github.com/jdugarte/activeadmin/blob/JD-fix-pagination-information/spec/unit/views/components/paginated_collection_spec.rb#L238), which [calls the `paginated_collection`](https://github.com/jdugarte/activeadmin/blob/JD-fix-pagination-information/spec/unit/views/components/paginated_collection_spec.rb#L17) test helper. This is where I get lost: I haven't been able to clarify the path downwards from here until the `ActiveAdmin::Views::PaginatedCollection#page_entries_info` method, nor upwards from there until, in the regular app, the collection gets paginated.
4. But here's the kicker: once the path gets to `ActiveAdmin::Views::PaginatedCollection#page_entries_info` the `collection` in the test path is indeed paginated but is already narrowed down to the two elements created in step 2 above. So when [the call to `collection_size` is made](https://github.com/jdugarte/activeadmin/blob/master/lib/active_admin/views/components/paginated_collection.rb#L144), which [then asks for `c.count`](https://github.com/jdugarte/activeadmin/blob/JD-fix-pagination-information/lib/active_admin/helpers/collection.rb#L11), the result is 2, which back in `ActiveAdmin::Views::PaginatedCollection#page_entries_info` is added to the `offset` and result in the expected value in the test. So the tests pass.

Now, back in the regular path of the app, the values change:

1. I have 910 records in the table I'm browsing.
2. As I said I can't find the place where this collection is paginated since all I see from my side of the app if the `index` clause in the ActiveAdmin module I created for this model.
3. This third step in the test has no equivalent in the regular path in the app.
4. Now, when `ActiveAdmin::Views::PaginatedCollection#page_entries_info` gets the `collection` in the app path, the collection has indeed a total count of 910 rows, which will be the value returned from [`collection_size`](https://github.com/jdugarte/activeadmin/blob/master/lib/active_admin/views/components/paginated_collection.rb#L144), and [added to `offset`](https://github.com/jdugarte/activeadmin/blob/master/lib/active_admin/views/components/paginated_collection.rb#L144) creates the erroneous page information I'm trying to fix.

What I think might have to be done is rewrite [the `paginated_collection`](https://github.com/jdugarte/activeadmin/blob/JD-fix-pagination-information/spec/unit/views/components/paginated_collection_spec.rb#L17) test helper so we can feed it the whole test collection: the full 5 rows of the `Post` model, not just the result of the previous call ([`Kaminari.paginate_array(Post.all).page(1).per(2)`](https://github.com/jdugarte/activeadmin/blob/JD-fix-pagination-information/spec/unit/views/components/paginated_collection_spec.rb#L234)) which narrows it down to 2 rows.

Have I managed to explain my issue with the test? What do you think I could do?

Thanks,

Jesús.-

jdugarte:
@deivid-rodriguez I found the issue in my app: the Gemfile was including also will_paginate, and its conflict with kaminari messed with the pagination information. I'm sorry I've wasted your time.

deivid-rodriguez:
No problem at all, thanks for following up and letting us know. We'll keep this in mind in case something similar pops up in the future :+1:. I glad you sorted out what was going on in your app!

Closing this!

tylerlu94:
was this fix ever merged?

jdugarte:
@tylerlu94 This is what I did in my app to solve the issue:

```
# config/initializers/kaminari.rb
Kaminari.configure do |config|
  config.page_method_name = :per_page_kaminari
end
```

according to the documentation [here](https://github.com/activeadmin/activeadmin/blob/master/docs/0-installation.md#will_paginate).

erotenberg1:
This is not working for me. I have applied this Kaminari.configure block as well as the solution of [#1833](https://github.com/activeadmin/activeadmin/issues/1833).  The latter was necessary to stop an exception but leads to misnumbering of the pages. The Kaminari.configure block is doing nothing as far as I can tell although for sure it is executing.

PR #5698, opened 2019-03-01
Update capybara gem version

ionut998:
Capybara `3.14` contain numerous bug fixes and new features
https://github.com/teamcapybara/capybara/blob/master/History.md#version-314

### DIFF
--- a/Gemfile.common
+++ b/Gemfile.common
@@ -11,7 +11,7 @@ gem 'draper', '~> 3.0'
 gem "devise", '~> 4.4'
 
 group :test do
-  gem 'capybara', '~> 3.10'
+  gem 'capybara', '~> 3.14'
 
   gem 'simplecov', require: false # Test coverage generator. Go to /coverage/ after running tests
   gem 'cucumber-rails', '~> 1.5', require: false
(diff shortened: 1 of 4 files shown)


javierjulio:
@ionut998 awesome thanks! 👍🏻

@deivid-rodriguez we aren't able to use dependabot on this repo because of our Gemfile setup right?

deivid-rodriguez:
TL;DR It can be done, but it might generate too many PRs because it would generate one PR per updated dependency AND Gemfile (rails 5.0, rails 5.1, and rails 5.2). :man_shrugging:

We would need to remove the `Gemfile.common` file and duplicate it inline in each `Gemfile`. Otherwise dependabot would update `Gemfile.common` for one Gemfile, but leave it "inconsistent" with the other gemfiles that also use it. Ideally dependabot should be able to update all gemfiles in one PR (similarly to our `rake bundle`) task, but I don't think it can do that at the moment.

javierjulio:
Ah yes that would be too much. Might be easier to update manually every once in awhile and keep the Gemfile setup DRY for now. Sounds good thanks!

PR #5699, opened 2019-03-01
Chandler updates

deivid-rodriguez:
Backports the latest chandler updates to 1-4-stable, so we actually release from this branch.


deivid-rodriguez:
We are planning to release AA 2.0 once we land Rails 6 compatibility, so I'll close this.

PR #5713, opened 2019-03-21
Correct Getting Started instructions

ClemensFMN:
The current getting started section does not work for the following two reasons:
- Either you issue a `rails g active_admin:install --skip-users` to make it work without devise (I did not test this), or you include `devise` in your Gemfile. I choose the second option.
- You need to seed the db, otherwise there is no admin user and you cannot login as described below.

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -54,7 +54,6 @@ If you want to use the devise gem for (admin) authentification, add it to
 your Gemfile
 
 ```ruby
-# Gemfile
 gem 'activeadmin'
 gem 'devise'
 ```
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -9,18 +9,19 @@ to implement beautiful and elegant interfaces with very little effort.
 # Getting Started
 
 Active Admin is released as a Ruby Gem. The gem is to be installed within a Ruby
-on Rails application. To install, simply add the following to your Gemfile:
+on Rails application. Active Admin does not provide authentification; this is done 
+by other gems (e.g. devise). To install without any authentification, simply 
+add the following to your Gemfile:
 
 ```ruby
 # Gemfile
 gem 'activeadmin'
-gem 'devise'
 ```
 
 After updating your bundle, run the installer
 
 ```bash
-rails generate active_admin:install
+rails generate active_admin:install --skip-users
 ```
 
 The installer creates an initializer used for configuring defaults used by
@@ -31,26 +32,67 @@ Migrate your db and start the server:
 
 ```bash
 $> rake db:migrate
-$> rake db:seed
 $> rails server
 ```
 
-Visit `http://localhost:3000/admin` and log in using:
+Visit `http://localhost:3000/admin` and voila: You&#8217;re on your brand 
+new Active Admin dashboard.
+
+To register an already existing model, run:
+
+```bash
+$> rails generate active_admin:resource [MyModelName]
+```
+
+This creates a file at `app/admin/my_model_names.rb` for configuring the
+resource. Refresh your web browser to see the interface. In order to CRUD
+items, tweak the parameter `permit_params` in `app/admin/MyModel.rb`.
+
+## Usage with devise
+
+If you want to use the devise gem for (admin) authentification, add it to
+your Gemfile
+
+```ruby
+# Gemfile
+gem 'activeadmin'
+gem 'devise'
+```
+
+Update the bundle and run the installer
+
+```bash
+rails generate active_admin:install
+```
+
+Migrate your db
+
+```bash
+$> rake db:migrate
+```
+
+If you are adding devise with Active Admin, you need to seed the database
+with an admin user for Active Admin (otherwise you probably will already
+have an admin user)
+
+```bash
+$> rake db:seed
+```
+
+and start the server
+
+```bash
+$> rails server
+```
+
+Visit `http://localhost:3000/admin` and log in using (this user has been generated
+by the database seed):
 
 * __User__: admin@example.com
 * __Password__: password
 
-Voila! You&#8217;re on your brand new Active Admin dashboard.
+Adding existing models to Active Admin works as described above.
 
-To register your first model, run:
-
-```bash
-$> rails generate active_admin:resource
-        [MyModelName]
-```
-
-This creates a file at `app/admin/my_model_names.rb` for configuring the
-resource. Refresh your web browser to see the interface.
 
 # Next Steps
 
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -50,7 +50,7 @@ items, tweak `permit_params` in `app/admin/my_model_names.rb`.
 
 ## Usage with devise
 
-If you want to use the devise gem for (admin) authentification, add it to
+If you want to use the devise gem for (admin) authentication, add it to
 your Gemfile
 
 ```ruby
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -14,6 +14,7 @@ on Rails application. To install, simply add the following to your Gemfile:
 ```ruby
 # Gemfile
 gem 'activeadmin'
+gem 'devise'
 ```
 
 After updating your bundle, run the installer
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
https://github.com/activeadmin/activeadmin/blob/8a597c4370b4c9701d339a3a87d11efd2aee7c2d/docs/0-installation.md is more clear I think. Maybe we should look into merging these two pages with the best of each?

ClemensFMN:
From my point of view (I wanted to quickly evaluate activeadmin), it would make sense to have a (kind of landing) page, which shows how to setup activeadmin quickly (and with - let's say - default options) and another page, where all installation options are explained. So having two separate pages makes sense...

This aside, for me it's bit unclear which dependencies activeadmin has; looking at your page, I would understand that you need activeadmin and **one** of the gems devise, cancancan, draper, or pundit?

deivid-rodriguez:
No, it's not like that, so that second page is definitely not clear enough either if it makes it appear like `devise`, `cancancan`, `draper` and `pundit` are different alternatives for the same thing. So we need to improve the other page too. Let's leave `cancancan`, `draper` and `pundit` aside for now and focus on `devise` now, though.

So basically, in your PR description you are right about the different options. You can either:

* Depend on `devise` by including it in your `Gemfile` (or if you're already using `devise`, by doing nothing). In that case, `activeadmin` will detect that and properly configure activeadmin to use it.

* If you're not using or you don't want to use `devise`, you can run the generator with `--skip-users`, and activeadmin will be configured without any authentication.

So, if we want to keep both pages (probably you're right and it's a good idea), let's at least document both of these options. So I would change the current instructions to use `--skip-users` and then add a separate paragraph with something like:

````markdown
If you also want the generator to automatically configure authentication, follow these steps instead:

Add the following to your Gemfile

```ruby
# Gemfile
gem 'activeadmin'
gem 'devise'
```

After updating your bundle, run the installer without the `--skip-users` flag

```bash
rails generate active_admin:install
```
````

**Comment by ClemensFMN** [comment]
Understood. Two minor comments:

- If you do not use devise, you need not sign into the admin page(s) and do not need a `rails db:seed`.
- After you have registered an existing model with active admin (via `rails generate active_admin:resource MyModel`), you need to tweak `permit_params` in `app/admin/MyModel.rb` so that you can actually CRUD MyModel items.

Do you want to change your [proposal](https://github.com/activeadmin/activeadmin/blob/8a597c4370b4c9701d339a3a87d11efd2aee7c2d/docs/0-installation.md) or shall I issue a new PR?

**Comment by deivid-rodriguez** [comment]
> If you do not use devise, you need not sign into the admin page(s) and do not need a rails db:seed.

Yeah, and even if you use devise, you'll only want to reseed if you're adding it together with activeadmin. If you are already using `devise` before adding `activeadmin`, you probably already have test users in your seeds, so you don't want to do it.

I would again add a separate note about seeds explaining that the generator will add a test user to your `seeds.rb` file if you're using devise, and that you need to evaluate whether you want to keep it and run `db:seed`, or remove that bit if you already have test users.

> After you have registered an existing model with active admin (via rails generate active_admin:resource MyModel), you need to tweak permit_params in app/admin/MyModel.rb so that you can actually CRUD MyModel items.

True. We can explain this as well.

Do you want to reuse this PR to consolidate all these improvements?

**Comment by ClemensFMN** [comment]
I consolidated all the points into [this commit](https://github.com/activeadmin/activeadmin/pull/5713/commits/882566dd886e995984e097cb15d740363b4ae4d6) - are you ok with it?

**Comment by rsommerard** [comment]
It could be insteresting to merge this PR :)

File: docs/documentation.md [hunk 1]
```diff
@@ -33,22 +35,63 @@ $> rake db:migrate
 $> rails server
 ```
 
-Visit `http://localhost:3000/admin` and log in using:
+Visit `http://localhost:3000/admin` and voila: You&#8217;re on your brand
```
deivid-rodriguez: Do we need the HTML character here? Also, I would avoid the french word

```suggestion
Visit `http://localhost:3000/admin` and you'll be on your brand
```

docs/documentation.md
```diff
@@ -33,22 +35,63 @@ $> rake db:migrate
...
 
-Voila! You&#8217;re on your brand new Active Admin dashboard.
+```bash
+$> rails generate active_admin:resource [MyModelName]
+```
 
-To register your first model, run:
+This creates a file at `app/admin/my_model_names.rb` for configuring the
+resource. Refresh your web browser to see the interface. In order to CRUD
+items, tweak the parameter `permit_params` in `app/admin/MyModel.rb`.
```
deivid-rodriguez:
```suggestion
items, tweak `permit_params` in `app/admin/my_model_names.rb`.
```

docs/documentation.md
```diff
@@ -33,22 +35,63 @@ $> rake db:migrate
...
-resource. Refresh your web browser to see the interface.
+Migrate your db
+
+```bash
+$> rake db:migrate
+```
+
+If you are adding devise with Active Admin, you need to seed the database
+with an admin user for Active Admin (otherwise you probably will already
+have an admin user)
```
deivid-rodriguez: I would clarify here that the installer adds an admin user to the `seeds.rb` file, and that's it's up to the user whether to reseed the whole thing or manually create the admin user.

docs/documentation.md
```diff
@@ -33,22 +35,63 @@ $> rake db:migrate
...
+gem 'activeadmin'
+gem 'devise'
+```
+
+Update the bundle and run the installer
 
 ```bash
-$> rails generate active_admin:resource
-        [MyModelName]
+rails generate active_admin:install
```
deivid-rodriguez: I don't love the existing style, but let's keep it as it is for now

```suggestion
$> rails generate active_admin:install
```

docs/documentation.md
```diff
@@ -33,22 +35,63 @@ $> rake db:migrate
...
+```
 
-To register your first model, run:
+This creates a file at `app/admin/my_model_names.rb` for configuring the
+resource. Refresh your web browser to see the interface. In order to CRUD
+items, tweak `permit_params` in `app/admin/my_model_names.rb`.
+
+## Usage with devise
+
+If you want to use the devise gem for (admin) authentification, add it to
```
deivid-rodriguez: I missed this instance in the previous review

```suggestion
If you want to use the devise gem for (admin) authentication, add it to
```

docs/documentation.md
```diff
@@ -33,22 +35,63 @@ $> rake db:migrate
...
 
-Visit `http://localhost:3000/admin` and log in using:
+Visit `http://localhost:3000/admin` and you'll be on your brand
+new Active Admin dashboard.
 
-* __User__: admin@example.com
-* __Password__: password
+To register an already existing model, run:
 
-Voila! You&#8217;re on your brand new Active Admin dashboard.
```
deivid-rodriguez: I just noticed that the "Voila!", and the HTML characters were already there, thanks for updating that even if it was not introduced by you :+1:.

docs/documentation.md
```diff
@@ -33,22 +35,63 @@ $> rake db:migrate
...
+resource. Refresh your web browser to see the interface. In order to CRUD
+items, tweak `permit_params` in `app/admin/my_model_names.rb`.
+
+## Usage with devise
+
+If you want to use the devise gem for (admin) authentification, add it to
+your Gemfile
+
+```ruby
+# Gemfile
```
deivid-rodriguez: We already mentioned the Gemfile above so we probably don't need this comment?

```suggestion
```

docs/documentation.md
```diff
@@ -33,22 +35,63 @@ $> rake db:migrate
...
+gem 'activeadmin'
+gem 'devise'
+```
+
+Update the bundle and run the installer
```
deivid-rodriguez: Maybe "Update the bundle" will not be clear for some users? What about

```suggestion
Run `bundle install`, and then run the installer
```

docs/documentation.md
ClemensFMN: :-)

Issue #5549, opened 2018-11-04
Displaying wrong entries count inside table


krthi:
I am trying to add pagination inside table, I added initializer as suggested in this issue as I am using will_paginate gem(https://github.com/activeadmin/activeadmin/issues/1833)

Here is the code for adding the pagination for the table,

```
paginated_collection(histories.page(params[:page]).per(5), download_links: false) do
      table_for collection do
           column ('Status') { |bh| bh.status.humanize }
           column ('Reason') { |bh| bh.reason }
           column ('Date') { |bh| bh.created_at }
      end
 end

```
It renders the records as the 'per' count in the pagination, But the links it displays wrong count as below,

![screen shot 2018-11-04 at 2 29 39 pm](https://user-images.githubusercontent.com/37453468/47962263-3403e780-e040-11e8-8c03-c280ccc3140a.png)

![screen shot 2018-11-04 at 2 31 39 pm](https://user-images.githubusercontent.com/37453468/47962262-323a2400-e040-11e8-8f75-8428d405f6d4.png)

tylerlu94:
Running into this too

jdugarte:
@tylerlu94 This is what I did in my app to solve the issue:

```
# config/initializers/kaminari.rb
Kaminari.configure do |config|
  config.page_method_name = :per_page_kaminari
end
```

according to the documentation [here](https://github.com/activeadmin/activeadmin/blob/master/docs/0-installation.md#will_paginate).

erotenberg1:
solution of @jdugarte is not working for me, I still get the page numbering errors similar to the original post.

I'm initializing with: 
```
Kaminari.configure do |config|
  config.page_method_name = :per_page_kaminari
end
```
and, according to [this link](https://github.com/activeadmin/activeadmin/issues/1833), 
```
 if defined?(WillPaginate)
  module WillPaginate
    module ActiveRecord
      module RelationMethods
        def per(value = nil) per_page(value) end
        def total_count() count end
      end
    end
    module CollectionMethods
      alias_method :num_pages, :total_pages
    end
  end
end
```
I am sure both methods are happening during initialization, but I still get the page number errors similar to the original post.

PR #5720, opened 2019-03-24
Invalid value shown in "to" entries count details.

Premanandh-S:
Fix for issue : https://github.com/activeadmin/activeadmin/issues/5549

Invalid value shown in `to` entries count when using `paginated_collection` method for showing collections with pagination.

### DIFF
--- a/lib/active_admin/views/components/paginated_collection.rb
+++ b/lib/active_admin/views/components/paginated_collection.rb
@@ -141,7 +141,7 @@ module ActiveAdmin
                    model: entries_name,
                    total: total,
                    from: offset + 1,
-                   to: offset + collection_size
+                   to: offset + collection.length
           end
         else
           # Do not display total count, in order to prevent a `SELECT count(*)`.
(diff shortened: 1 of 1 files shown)


tylerlu94:
Running into this issue as well

deivid-rodriguez:
Are you using `will_paginate` in your main application? That sounds like could be causing this.

Premanandh-S:
@deivid-rodriguez Yes, I am using `will_paginate`. We have issue in code level too, you could see that overall count is added to offset that leads showing invalid values in showing from and to entries count.

deivid-rodriguez:
Thank you @Premanandh-S.

I understand that there's is problem here, but I would prefer to fully understand what's going on, and properly study the best way to resolve conflicts between pagination gems before introducing any changes. I plan to work on this right after we release ActiveAdmin 2.0.

javierjulio:
This seems to be resolved by https://github.com/activeadmin/activeadmin/blob/master/docs/0-installation.md#will_paginate as mentioned in the issue.

PR #5721, opened 2019-03-24
Point to rc's of arbre and inherited_resources

deivid-rodriguez:


gemfiles/rails_60.gemfile
```diff
@@ -5,8 +5,7 @@ eval_gemfile(File.expand_path(File.join("..", "Gemfile.common"), __dir__))
 gem "rails", "6.0.0.beta3"
 
 # Use the unreleased version of gems because they are compatible with Rails 6
-gem 'arbre', git: 'https://github.com/activeadmin/arbre'
-gem 'inherited_resources', git: 'https://github.com/activeadmin/inherited_resources'
+gem "arbre", "~> 1.2.a"
```
javierjulio: I went through the related files but I didn't understand why we would remove inherited_resources here and still have it install an RC version while arbre had to be updated and remain?

deivid-rodriguez: The answer lies on our dependencies gemspecs. Inherited resources sets an upper bound on the supported rails version

https://github.com/activeadmin/inherited_resources/blob/ef943a6b6d5d0fdb5ea29adbcfada8b576a79d9b/inherited_resources.gemspec#L26

The version we just released is the first version of the gem released to match the Rails 6 requirement.

So `bundler` will automatically pick it up even if it's a pre-releases, because it's the only way to satisfy all requirements.

`arbre` however, has a very relaxed dependency on Rails

https://github.com/activeadmin/arbre/blob/81db315be69f3ca6a865acff8f6ba935f5fa75e1/arbre.gemspec#L23

So the last stable version release of it is perfectly compatible with Rails 6, and `bundler` will prefer it over a pre-release. Since we want to use the rc release because it fixes a warning with Rails 6, we need to "force it", that why I specify it here.

javierjulio: Interesting so if someone were to run `bundle install` on a new project using ActiveAdmin, they would get the pre-release version of InheritedResources? I didn't know that. Taking both though into consideration.. its starting to make sense now why Bundler would do what you say for InheritedResources although at first glance it seems to me as unexpected. Thanks so much for clarifying! 🙏🏻

deivid-rodriguez: Yeah, if some _using Rails 6_ tries to add `activeadmin`, they will automatically get the rc release of `inherited_resources`, because it's the only possible way the dependencies can be satisfied. Someone using prior Rails will still get the latest stable version of `inherited_resources`.

PR #5722, opened 2019-03-30
Only show form input fields in permit_params

### LINKED ISSUE #3308: permit_params should guide forms?
I may have missed where this was discussed, but I cannot quite understand how the default forms that are created to new and update using ActiveAdmin should include fields that are not on the permit_params list. 

My use case is I have columns in my model that are internally settable, are marked as attr_protected in model (using protected_attributes) and are not in my ActiveAdmin permit_params list. Yet they appear by default on the forms for entry. 

I am thinking this does not make sense, given they will be filtered out anyway for mass assignment protection. So hence I wonder why they appear on the form? Am I missing something? DO they serve a purpose on the form?

### LINKED ISSUE #5390: Only show form input fields in permit_params
### Expected behavior

Specifying a resource with no `form` block should only show fields listed in `permit_params` instead of showing all fields and not being able to save the record.

e.g.

```ruby
ActiveAdmin.register User do
  permit_params :email
end

# /admin/users/1/edit

# should ONLY show `email` field. i.e same as specifying:

form do |f|
  f.inputs do
    f.input :email
  end

  f.actions
end
```

### Actual behavior

It shows all the fields available on the model. Trying to save the user would hit an error.

### How to reproduce

See above

marclerodrigues:
Closes #5390.
Closes #3308.

Adds an option to the resource class named `build_form_with_permitted_params_only` in order to avoid breaking existing applications that depends on the current behavior. Once this is set to `true` the rendered form will only contain the fields specified in the `permit_params` method.


lib/active_admin/resource_dsl.rb
```diff
@@ -65,6 +65,10 @@ def permit_params(*args, &block)
       belongs_to_param = config.belongs_to_param
       create_another_param = :create_another if config.create_another
 
+      if config.build_form_with_permitted_params_only
+        config.permitted_params = block ? block.call : args.flatten
```
deivid-rodriguez: This block needs to be evaluated in the context of a controller, we can have `params` or `current_user` calls here. So I don't think this is going to work.

Mayb you can skip this and instead call `controller.permitted_params` when rendering the form?

deivid-rodriguez: I implemented this suggestion in https://github.com/activeadmin/activeadmin/pull/5722/commits/e5f16cb8dc02a786c196a2b58c3918703db373f8.

docs/5-forms.md
```diff
@@ -42,6 +42,31 @@ form do |f|
...
+The form generated will contain only the fields passed as an argument to the `permit_params` method. This method can also receive a block as argument, the following code will work exactly as the example above:
+
+```ruby
+ActiveAdmin.register Post do
+  config.build_form_with_permitted_params_only = true
+
+  permit_params do
+    permitted = []
+    permitted << :title
+    permitted << :body
```
deivid-rodriguez: We should improve these docs to showcase when this version can be useful over a static list.

deivid-rodriguez:
In other words, how could this break existing applications? It will definitely remove existing form inputs from default forms, but those inputs should be ignored since they're not permitted at the controller level... :thinking:.

javierjulio:
I can see why we'd want to drive the form from params but params can be complex. What happens here with nested params? Does something go wrong or it works or is just ignored?

lib/active_admin/views/pages/form.rb
```diff
@@ -42,7 +42,17 @@ def default_form_path
         def default_form_config
           ActiveAdmin::PagePresenter.new do |f|
             f.semantic_errors # show errors on :base by default
-            f.inputs
+
+            if active_admin_config.build_form_with_permitted_params_only
+              f.inputs do
+                controller.form_params.each do |param|
```
Fivell: what if permit params are defined with nested attributes (accepts_nested_attributes_for) ?

deivid-rodriguez:
Both good questions, thanks! I think this PR needs more work, so I'll try to have a closer look and address your concerns.

ngouy:
👀

javierjulio:
As mentioned in the issue, I don't think this is worth doing even for the simple case of just one level of params. This can get quite complex. I haven't seen any movement on this other than "that would be nice" and I think it's just because it's easier to declare the form and customize as needed.

PR #5724, opened 2019-04-01
Use release candidate version of CanCanCan gem

ionut998:
CanCanCan gem has introduced Rails 6 support in version 3.0.0.rc1
https://github.com/CanCanCommunity/cancancan/blob/feature/3.0.0/CHANGELOG.md#300


javierjulio:
Thanks again! ❤️

PR #5726, opened 2019-04-04
Update gems to use the released version

ionut998:
The following gems have been updated: `arbre`, `cancancan` and `inherited_resources`.

### DIFF
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.3'
 
-  s.add_dependency 'arbre', '~> 1.1', '>= 1.1.1'
+  s.add_dependency 'arbre', '>= 1.2.0'
   s.add_dependency 'formtastic', '~> 3.1'
   s.add_dependency 'formtastic_i18n', '~> 0.4'
   s.add_dependency 'inherited_resources', '~> 1.7'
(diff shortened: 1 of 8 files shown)


activeadmin.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.3'
 
-  s.add_dependency 'arbre', '~> 1.1', '>= 1.1.1'
+  s.add_dependency 'arbre', '>= 1.2.0'
```
deivid-rodriguez: I'm good with bumping this requirement but can we use the more conservative `~> 1.2`?

deivid-rodriguez: I'm unsure whether we should put this in the changelog? I guess it's not necessary.

ionut998: I updated arbre dependency to be `~> 1.2`
I also added a line in the `unreleased` section of the changelog to mention this change

deivid-rodriguez: Thanks!

@javierjulio do you think we should mention this in the change log? I think this shouldn't affect end users at all. `bundler` should be able to work it out, and arbre 1.2 over arbre 1.1 does not have any user facing changes.

javierjulio: Thanks guys! @deivid-rodriguez you're right it shouldn't affect them but since Arbre is pretty important to ActiveAdmin, a mention is helpful and won't hurt so I think it's ok too. 👍🏻

javierjulio:
@ionut998 @deivid-rodriguez thanks so much for your efforts on all these updates! 👏🏻

activeadmin.gemspec
deivid-rodriguez: Ok!

PR #5730, opened 2019-04-10
Bump sqlite3 to 1.4.0

deivid-rodriguez:
Since the problems with previous Rails versions have been fixed.


deivid-rodriguez:
Didn't update the Rails 5.0 Gemfile because a version fixing the issue wasn't released.

PR #5731, opened 2019-04-10
Cucumber better formatter

deivid-rodriguez:
With the previous formatters, we don't get any reason or error messages when failures happen.


deivid-rodriguez:
No idea why jruby 9.1 crashes with the new formatter, so I checked for that in the config. Should I add a comment or is git history good enough for the curious contributor? :)

javierjulio:
If it's only JRuby 9.1 I wouldn't be concerned with it. Despite a release last April, it's not supported:
> JRuby 9.1 Ruby 2.3 compatible supported - till end of 2017
> https://github.com/jruby/jruby/wiki/Roadmap#jruby-9x-current
And since it's targeting Ruby 2.3 that was EOL last month too. Best to have it in the commit message so thanks. 👍🏻

deivid-rodriguez:
Nice, we can drop support for those then after the final 2.0 release.

PR #5732, opened 2019-04-10
Zeitwerk support

deivid-rodriguez:
This is an experiment to see whether activeadmin works better with zeitwerk enabled now.


deivid-rodriguez:
It seems that the latest `zeitwerk` works! And not only that, but also removes the need for some ugly hacks in our specs!

I needed to pull Rails master in order to pick it up, and that required a separate fix in the `jruby` adapter, which I'm using here. We can probably leave this PR hang until the next rails pre, but this is good news! :+1:

deivid-rodriguez:
This is ready now. Also took the chance to make sure everything is passing against the latest patch level versions of supported rubies.

PR #5736, opened 2019-04-11
Add a display_timezone config option

dkniffin:
### Why?

The use-case I'm trying to solve is that I have an admin in US Eastern timezone, who is managing users across the US. Having all times shown in UTC is a poor UX for these admins, so I'd like to display things in EST by default, and override in other places.

The closes solution I could find is documented in this StackOverflow answer: https://stackoverflow.com/a/22791139/1202488

```rb
ActiveAdmin.setup do |config|
  # ...
  config.before_action do
    Time.zone = 'Eastern Time (US & Canada)'
  end
  # ...
end
```

The problem with this solution is it's setting `Time.zone`, which gets used across all controllers.

## What changed?

I added a new config option called `display_timezone`, which defaults to UTC and gets used when displaying dates and times (in conjunction with the existing `localize_format` option.

I also added this option to the initializer generator and documented it.

### DIFF
--- a/lib/generators/active_admin/install/templates/active_admin.rb.erb
+++ b/lib/generators/active_admin/install/templates/active_admin.rb.erb
@@ -167,6 +167,17 @@ ActiveAdmin.setup do |config|
   #
   config.localize_format = :long
 
+  # == Display Date/Time in Timezone
+  #
+  # Sets the timezone to display dates and times in.
+  # For a list of all the valid values, see
+  # https://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html
+  #
+  # This can still be overridden in specific pages with a block as shown here
+  # https://activeadmin.info/6-show-pages.html
+
+  config.display_timezone = "UTC"
+
   # == Setting a Favicon
   #
   # config.favicon = 'favicon.ico'
--- a/lib/active_admin/application_settings.rb
+++ b/lib/active_admin/application_settings.rb
@@ -20,6 +20,9 @@ module ActiveAdmin
     # Set default localize format for Date/Time values
     register :localize_format, :long
 
+    # Set default timezone for Date/Time value display
+    register :display_timezone, "UTC"
+
     # Active Admin makes educated guesses when displaying objects, this is
     # the list of methods it tries calling in order
     # Note that Formtastic also has 'collection_label_methods' similar to this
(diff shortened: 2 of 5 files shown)


dkniffin:
Hmm. I just noticed an issue with this change. the `object.in_time_zone(active_admin_application.display_timezone),` forces it into that timezone, even if it's already been set in the app's page configs. I'll see if I can think of another way to do that.

dkniffin:
Alright, I changed the approach slightly. I think this should work correctly now.

dkniffin:
@deivid-rodriguez Alright, this should be ready for review now

dkniffin:
@deivid-rodriguez bump

dkniffin:
👍Thanks. I will make these updates soon. Maybe later this week.

dkniffin:
Alright, I've updated docs and tests. I will update the changelog too, but I'm actually not sure if this is considered a breaking change or not. I think before it would render the times in the timezone stored in the DB, and now it will show them in UTC by default. That's breaking, right? Should there be a second option or something to opt out of this behavior and just show the times in the DB timezone again? and maybe make that the default?

dkniffin:
@javierjulio This is ready for re-review

dkniffin:
@deivid-rodriguez @javierjulio bump

deivid-rodriguez:
Sorry for the delay @dkniffin. I'll review this on Monday unless someone beats me to it :+1:.

javierjulio:
That's a good question about the breaking change. I'm not sure. In the case where I've used AA it would just display UTC by default. Let's see what input others have there. No problem, on the delay here, thanks for updating!

lib/generators/active_admin/install/templates/active_admin.rb.erb
```diff
@@ -167,6 +167,17 @@ ActiveAdmin.setup do |config|
...
   config.localize_format = :long
 
+  # == Display Date/Time in Timezone
+  #
+  # Sets the timezone to display dates and times in.
+  # For a list of all the valid values, see
+  # https://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html
+  #
+  # This can still be overridden in specific pages with a block as shown here
+  # https://activeadmin.info/6-show-pages.html
```
deivid-rodriguez: Maybe instead of the generic link we can use an example? Like:

```ruby
  column :created_at do |obj|
    obj.created_at.localtime
  end
```

dkniffin:
Sounds good. I'll try to make those changes this week

javierjulio:
I think this is best controlled by existing functionality, a block for column or row. If this is something others would like configured, I'd prefer that the config were a block, likely given the authenticated user if available.

PR #5740, opened 2019-04-13
Ensure application gets reloaded only once

### LINKED ISSUE #5739: Reloads application twice
### Expected behavior

When `Rails.application.reloader.reload!` is invoked, for example because Rails or Spring detected that files have changed, `ActiveAdmin.application.load!` should only be called once because multiple invocations incur unnecessary overhead.

### Actual behavior

`ActiveAdmin.application.load!` is called twice.

### How to reproduce

I became aware of this when running `RubyProf.profile { Rails.application.reloader.reload! }`. The resulting output shows that `load!` is called twice.

I'm also going to send a PR soon that includes a test case.

jscheid:
I've recently used ruby-prof to look at reload performance, like so:

```ruby
require 'ruby-prof'

result = RubyProf.profile do
  Rails.application.reloader.reload!
end
printer = RubyProf::MultiPrinter.new(result)
printer.print(path: "/path/to/output", profile: "reload-profile")
```

Doing so revealed that `ActiveAdmin.application.load!` is called twice. Since that method accounts for a substantial portion of the time it takes to reload our app, this seems quite wasteful.

The reason is that the `prepare` callback is invoked twice, from here:

https://github.com/rails/rails/blob/v5.2.3/activesupport/lib/active_support/reloader.rb#L46

and from here:

https://github.com/rails/rails/blob/v5.2.3/activesupport/lib/active_support/reloader.rb#L59

For the "load" callback alone this wouldn't be a problem because it only calls `load!` if the application is not already loaded. However, the "unload" callback also uses the `prepare` callback, which results in the following order of callbacks:
1. unload
2. load
3. unload
4. load

That defeats the `loaded?` guard because by the time the "load" callback is invoked a second time, the application has been unloaded again.

This PR attempts to solve the problem by using the `class_unload` event instead of the `prepare` event for the "unload" callback. That event is only invoked once per reload.

It also removes the compatibility layer in `reloader.rb`. As far as I can tell, it is no longer needed now that support for Rails 4.x has been dropped.

It also simplifies the logic for registering the "unload" callback because the same callback works regardless of the setting for `reload_classes_only_on_change`.

Closes #5739


deivid-rodriguez:
I love this :heart:.

One thought and two nitpicks.

Maybe we should land the #5732 PR first, in case this conflicts with Zeitwerk. In particular, I think the spec you have added will need to run with the `CLASS_RELOADING` environment variable set (which configures Rails `cache_classes` on our test app), because Zeitwerk explicitly forbids reloading unless that setting is enabled.

As nitpicks, I'd love if you could split the PR into two separate commits, one that removes the unnecessary compatibility layer with the Rails reloader, and another one that fixes the double loading issue.

The other one is that the "integration_spec.rb" file we could maybe be renamed to "rails_integration_spec.rb"

jscheid:
Sure, no problem. I’ll get to it tomorrow or Monday.

jscheid:
@deivid-rodriguez all done. I've also merged your `zeitgeist` branch into my local to see how the two work together. The test seems to pass fine without `CLASS_RELOADING` set. Perhaps because no code has changed before the reload is triggered?

deivid-rodriguez:
This looks good to me! Thanks @jscheid!

Just in case, can you rebase this since there's been a few changes in `zeitwerk` and `rails` since last time?

deivid-rodriguez:
Oh, can you also add a changelog entry for the reloading fix?

jscheid:
@deivid-rodriguez rebased... could you add the change log please? I don't know what version it will be released in.

deivid-rodriguez:
You don't need to know it, put it under the "Unreleased" section. Anyways, I can do it, no problem :smiley:

jscheid:
Ah, now it did break. Good call. Curious, because it definitely didn't the other day (with Zeitgeist merged in). I'll have another look at this tomorrow 😕

deivid-rodriguez:
Great, thanks!

deivid-rodriguez:
I plan to release this with 2.1 by the way, together with "official" Rails 6 support.

jscheid:
@deivid-rodriguez any idea what the codeclimate failure is about? I'm inclined to think it's a bug... I had pushed the same change before with the only difference in the change log, and it didn't complain.

jscheid:
Oh, maybe it didn't run because of the conflict... anyway, do you know off the top of your head how this can be resolved?

deivid-rodriguez:
Since you're adding a new "suite", you need to make sure coverage reports are properly merged together. You should create a new `spec/support/simplecov_reloading_env.rb` file doing something similar to `spec/support/simplecov_regular_env.rb` and `spec/support/simplecov_changes_env.rb`.

jscheid:
@deivid-rodriguez there are already cucumber specs for this type. Were they not considered for code coverage before?

deivid-rodriguez:
Cucumber reloading specs are also part of a separate suite, which is properly configured for coverage report merging. Basically every set of specs/features that run as a separate process needs custom code so that their coverage reports are properly merged.

jscheid:
Ah, I think we're good: https://github.com/activeadmin/activeadmin/blob/master/features/support/simplecov_reload_env.rb

I think problem was just that the new type didn't run in CI. 🤞

jscheid:
You were right, that file was missing.

What a drag... you're lucky you got your PR in before mine 😉 

Should be all good now?

jscheid:
You're welcome, thanks for merging.

Issue #5745, opened 2019-04-15
version upgrade broke `action_item` argument behavior


manubete:
### Expected behavior

When I upgraded activeadmin to version `2.0.0.rc1`, I expected our code to be backwards compatible and not error out on previously used functions and methods.

### Actual behavior

Before I upgraded activeadmin to `2.0.0.rc1`, our calls to `action_item` did not require any arguments before starting the `do ... end` block. Below is an example snippet:

```
  action_item do
    resource = Resource.find(params[:resource_id])	    
    link_to "View Resource", admin_announcement_path(resource)
  end
```

After the upgrade however, we are getting errors that action_item requires an argument or two, such as a label or `:only` filter of sorts. Below is an working version of the new `action_item` :

```
 action_item :view_message do
    resource = Resource.find(params[:resource_id])	    
    link_to "View Resource", admin_announcement_path(resource)
  end
```

Was this an intentional change? Or does anyone suspect anything else going on? We never had errors like this before the upgrade
### How to reproduce

Upgrade active admin to version `2.0.0.rc1` in a code repository that uses similar snippets and run locally

deivid-rodriguez:
Hi @manubete, thanks for reporting this. Can you try 2.0.0.rc2 and see if it fixes it? Thanks!

deivid-rodriguez:
Hi @manubete, so I had a closer look at this.

This seems intentional. `action_item` without a name was deprecated in https://github.com/activeadmin/activeadmin/commit/a73ed3637e6a50e28c33c5b70db0d61d936d94d2#diff-529baad25097a0794f167b102b5afc43, which has been out there since 1.0.0.pre1, and indeed removed in https://github.com/activeadmin/activeadmin/commit/4014a37f0035197fd70c376d3b9e7be7b8c2c7f4#diff-529baad25097a0794f167b102b5afc43, released with 2.0.0.rc1.

Did you not see the deprecations?

deivid-rodriguez:
These deprecations should be displayed on 1.x releases, but in any case I also documented this in the change log in https://github.com/activeadmin/activeadmin/pull/5747/commits/3bf2c6ce105eb5545960e828597332665c732ab0.

deivid-rodriguez:
I double checked that this deprecation is correctly generated on activeadmin 1.x, so closing.

stephenh:
[...]

Issue #5744, opened 2019-04-15
 `action_item :only` broken in version `2.0.0.rc1`


manubete:
### Expected behavior

When I upgraded activeadmin from `1.4.2` to `2.0.0.rc1`, I expected our code to be backwards compatible and not error out on previously used functions and methods.

### Actual behavior

Before I upgraded activeadmin, our calls to certain `action_item`'s were only being called on the show page of a particular resource (say resource.id = 3). Below is a code snippet as an example

```
action_item :only => :show, :if => proc { resource.medium.sms? && !(current_admin_user.role == 'mod') } do
    link_to 'Get Incoming sms resource', show_incoming_text_messages_admin_device_path, :method => :get
  end
```

However, after the upgrade to version `2.0.0.rc1` our code started triggering this action item when one runs the index pages as well. This seems to break the routing on the index, and gives us this error :

`No route matches {:action=>"show_incoming_text_messages", :controller=>"admin/devices"} missing required keys: [:id]`

This call was not happening or erroring before the upgrade. Was this an intentional change on the new versions? Is there something else that might be happening?

### How to reproduce

Upgrade active admin to version `2.0.0.rc1` in a code repository that uses similar snippets and run locally

deivid-rodriguez:
Hi @manubete, thanks for reporting this. Can you try 2.0.0.rc2 and see if it fixes it? Thanks!

deivid-rodriguez:
This sounds like a duplicate of #5745: you need to pass a name argument to `action_item`. Let me know if this is not the case and this issue is actually different.

stephenh:
@deivid-rodriguez yep, we came to the same diagnosis yesterday; apologies for not updating the ticket sooner. Thanks again!

deivid-rodriguez:
No problem, thanks for following up! :+1:

PR #5747, opened 2019-04-23
Cut 2.0.0

deivid-rodriguez:
I think we're ready to cut 2.0.0, but I'd like to wait a bit on #5744 and #5745, to make sure there's no actual regression in 2.0.0.rc2, and both issues were due to the bad arbre release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


javierjulio:
Sounds good, thanks again David!

deivid-rodriguez:
Ok. So no regressions in the end, so let's do this!

deivid-rodriguez:
I released the final version! :+1:

javierjulio:
Awesome thanks David!! 👏🏻

PR #5749, opened 2019-04-25
Check rails 6.0.0.rc1 support

deivid-rodriguez:


deivid-rodriguez:
Rails 6.0.0.rc1 looks good!

PR #5756, opened 2019-05-04
Improved resource generator to list candidate fields for permit_params

coezbek:
This pull request improves the behavior of the activeadmin:resource generator.

It prepopulates the `permit_params` with all fields found in the model for which the generator is run.

It does not list `id`, `updated_at`, `created_at` as these are not likely to be assignable.

The user still needs to uncomment the `permit_params`. This is a precaution against inadvertent mass assignment.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -111,6 +111,9 @@ Layout/SpaceAfterComma:
 Layout/SpaceAroundEqualsInParameterDefault:
   Enabled: true
 
+Layout/SpaceAroundOperators:
+  Enabled: true
+
 Layout/SpaceBeforeBlockBraces:
   Enabled: true
 
--- a/bin/rubocop
+++ b/bin/rubocop
@@ -1,0 +1,7 @@
+#!/usr/bin/env ruby
+
+load File.expand_path("bundle", __dir__)
+
+require "bundler/setup"
+
+load Gem.bin_path("rubocop", "rubocop")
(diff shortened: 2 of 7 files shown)


coezbek:
@deivid-rodriguez Anything I can do to get this merged?

lib/generators/active_admin/resource/templates/admin.rb.erb
```diff
@@ -1,11 +1,12 @@
 ActiveAdmin.register <%= class_name %> do
-# See permitted parameters documentation:
-# https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#setting-up-strong-parameters
+# Uncomment all parameters which should be permitted for assignment
```
deivid-rodriguez: Any reason to invert the order of the document reference and the boilerplate code here?

coezbek: Sorry I did not have time to respond. I felt the permit_params is the most relevant thing here in the template. Did any of this change make it into master? Should I start over with this (that's fine).

deivid-rodriguez: No, I should apologize!

I was editing your PR myself because I changed the template's indentation so it got conflicts. And when doing that I unintentionally force-pushed activeadmin's master branch to your fork, so the PR got closed automatically because it no longer contains any changes. I'll open a separate PR, sorry about that.

PR #5760, opened 2019-05-08
Fix broken link in active_admin.rb template

WaKeMaTTa:
Fixed broken link and improved the description by adding another link, to `active_admin.rb` template, section `Localize Date/Time Format`

### DIFF
--- a/lib/generators/active_admin/install/templates/active_admin.rb.erb
+++ b/lib/generators/active_admin/install/templates/active_admin.rb.erb
@@ -163,7 +163,10 @@ ActiveAdmin.setup do |config|
   #
   # Set the localize format to display dates and times.
   # To understand how to localize your app with I18n, read more at
-  # https://github.com/svenfuchs/i18n/blob/master/lib%2Fi18n%2Fbackend%2Fbase.rb#L52
+  # https://guides.rubyonrails.org/i18n.html
+  #
+  # You can run `bin/rails runner 'puts I18n.t("date.formats")'` to see the
+  # available formats in your application.
   #
   config.localize_format = :long
 


lib/generators/active_admin/install/templates/active_admin.rb.erb
```diff
@@ -162,8 +162,11 @@ ActiveAdmin.setup do |config|
   # == Localize Date/Time Format
   #
   # Set the localize format to display dates and times.
-  # To understand how to localize your app with I18n, read more at
-  # https://github.com/svenfuchs/i18n/blob/master/lib%2Fi18n%2Fbackend%2Fbase.rb#L52
+  # To understand how to localize your app with I18n, read more at:
+  # https://guides.rubyonrails.org/i18n.html
+  #
+  # This are the default formats available:
+  # https://github.com/rails/rails/blob/master/activesupport/lib/active_support/locale/en.yml#L2-L9
```
deivid-rodriguez: This is a bit brittle, becase the defaults or the path could change in the future, or be different across Rails versions. Maybe we can instead tell the user how to find this information for her applications. Something like:

> You can run `bin/rails runner 'puts I18n.t("date.formats")'` to see the default formats for your application.

javierjulio: That's a great suggestion! It wouldn't be hard coded and means it would be specific to the Rails version they are using too. 👌🏻

WaKeMaTTa: Very cool!

WaKeMaTTa:
Changes requested done.

lib/generators/active_admin/install/templates/active_admin.rb.erb
```diff
@@ -163,7 +163,10 @@ ActiveAdmin.setup do |config|
   #
   # Set the localize format to display dates and times.
   # To understand how to localize your app with I18n, read more at
-  # https://github.com/svenfuchs/i18n/blob/master/lib%2Fi18n%2Fbackend%2Fbase.rb#L52
+  # https://guides.rubyonrails.org/i18n.html
+  #
+  # You can run bin/rails runner 'puts I18n.t("date.formats")' to see the
```
deivid-rodriguez:
```suggestion
  # You can run `bin/rails runner 'puts I18n.t("date.formats")'` to see the
```

WaKeMaTTa: I thought about doing it, but I did not know if you would accept it :joy:

WaKeMaTTa:
Done.

lib/generators/active_admin/install/templates/active_admin.rb.erb
deivid-rodriguez: Yeah, I included the backquotes in my comment by I forgot to skip the backquotes, so they didn't appear and instead formatted the command :sweat_smile:

PR #5762, opened 2019-05-24
Fix duplicate namespace when using config.pundit_policy_namespace

kwent:
Hi,

When using `config.pundit_policy_namespace = :active_admin` when the object namespace is already `ActiveAdmin::PagePolicy` results of trying to find `ActiveAdmin::ActiveAdmin::PagePolicy`. Other models inside `ActiveAdmin` are fine. Only no models policy as `Page` are impacted.

This PR is fixing that.

### DIFF
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -57,7 +57,7 @@ module ActiveAdmin
     private
 
     def namespace(object)
-      if ActiveAdmin.application.pundit_policy_namespace
+      if ActiveAdmin.application.pundit_policy_namespace && !object.class.to_s.include?(ActiveAdmin.application.pundit_policy_namespace.to_s.camelize)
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Hi @kwent! Thanks for helping out here. :+1:

Can you detail how this issue can be reproduced? I tried on a fresh application and couldn't see the issue, so I think I'm missing some specific setup to make it happen.

Also, if you could add a regression test, that would be great.

deivid-rodriguez:
I can reproduce it now, actually, no need for steps :). I'd like a test though.

kwent:
Would love to help with a spec but unfortunately i don't have time to set up the environment etc... at the moment. Glad you were able to repro !

deivid-rodriguez:
Closing in favor of #5777. Tried to edit this PR but I don't have permission.

PR #5769, opened 2019-06-11
Fix order of versions for diff of 2.1.0 release

artfuldodger:
Putting the newer version last ensures that we get the expected diff. Otherwise, we see additions and removals switched.


javierjulio:
Noticed we had a build failure so retan the workflow and all checks out. I’ll get this in. Thanks!

PR #5775, opened 2019-06-13
Add a minimal git history to generated test apps

deivid-rodriguez:
It's very common to make experimental modifications to test applications. This makes it very easy to go back to the freshly generated application without having to regenerate it, by running `git checkout .`, or to even keep track of the experimental changes being done using git.

### DIFF
--- a/spec/support/rails_template.rb
+++ b/spec/support/rails_template.rb
@@ -142,3 +142,6 @@ if ENV['RAILS_ENV'] == 'test'
 
   rake "parallel:drop parallel:create parallel:load_schema", env: ENV['RAILS_ENV']
 end
+
+git add: "."
+git commit: "-m 'Bare application'"
--- a/spec/support/rails_template_with_data.rb
+++ b/spec/support/rails_template_with_data.rb
@@ -329,3 +329,6 @@ append_file "db/seeds.rb", "\n\n" + <<-RUBY.strip_heredoc
 RUBY
 
 rake 'db:seed'
+
+git add: '.'
+git commit: "-m 'Bare application with data'"


deivid-rodriguez:
Nothing really changes regarding our testing instructions. The only difference is that if developers navigate to the generated application they will see

```
$ git status
On branch master
nothing to commit, working tree clean

$ git log
commit 49ed0838e18ef5e7893fea1a5f98a2e707fa66fe (HEAD -> master)
Author: David Rodríguez <deivid.rodriguez@riseup.net>
Date:   Thu Jun 13 11:49:41 2019 +0200

    Bare application with data

commit 0abea712b22adbbcba72253c12aa90ed253befe5
Author: David Rodríguez <deivid.rodriguez@riseup.net>
Date:   Thu Jun 13 11:48:58 2019 +0200

    Bare application
```

instead of something like

```
$ git status
On branch master

No commits yet

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.gitignore
	.ruby-version
	README.md
	Rakefile
	app/
	bin/
	config.ru
	config/
	db/
	lib/
	log/
	package.json
	public/
	spec/
	storage/
	test/
	tmp/
	vendor/

nothing added to commit but untracked files present (use "git add" to track)
```

javierjulio:
Ah so the generated application is treated as a separate git repo with its own commits? I wasn’t sure if that was the case.

deivid-rodriguez:
Yeah. Well, that was already the case since rails generators automatically run `git init` unless you pass `--skip-git`. But everything was left uncommitted prior to this PR.

Now we add some initial commits, so any edits made by developers will be more apparent and it will be easier to discard them and go back to the fresh version of the application.

javierjulio:
Sounds good thanks! I’m not at my Mac so I wasn’t able to try it. I’ll do that soon. Thanks again!

PR #5777, opened 2019-06-14
Fix pundit policy retrieving on an edge case

deivid-rodriguez:
When retrieving the policy for an `ActiveAdmin::Page`, and the pundit namespace is `:active_admin`, we need an extra check so that the right policy is properly picked up.

This is a finished version of #5762.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+* Fix pundit policy retrieving for static pages when the pundit namespace is :active_admin. [#5777] by [@kwent]
+
 ## 2.1.0 [☰](https://github.com/activeadmin/activeadmin/compare/v2.0.0..v2.1.0)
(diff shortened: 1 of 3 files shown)


kwent:
Thank you !

deivid-rodriguez:
No problem, thanks to you for the fix!

jensljungblad:
@javierjulio Not sure I understand why this makes sense... I think the previous behavior was the more correct one.

My situation is I have an `AdminUser` model, and ActiveAdmin pundit namespace set to `admin`. This results in AA incorrectly looking for the policy under `app/policies/admin_user_policy.rb` instead of `app/policies/admin/admin_user_policy.rb`. Further I would expect any AA models such as comment and page to be namespaced properly too: `app/policies/admin/active_admin/comment_policy.rb`.

So yeah, I disagree with the problem statement in https://github.com/activeadmin/activeadmin/pull/5762 :)

PR #5783, opened 2019-06-19
Generator cleanup

deivid-rodriguez:
I don't like that we keep a lot of code as inline heredocs in our generators. It doesn't always get properly highlighted, it's hard to keep a consistent style on it, and it requires some odd escaping sometimes. Moving this code to separate files also makes it consistent with other parts of the generators that already use separate files.

This PR implements that, and also adds some other tiny improvements in our generator's code.


javierjulio:
This is a great improvement. Thanks! Love it.

PR #5791, opened 2019-06-24
Use CGI.unescape for non-Latin characters

mabdelfattah:
non-Latin characters are being shown decoded, so this fix to show them correctly.
activeadmin_fix


deivid-rodriguez:
Thanks for this @mabdelfattah. Do you think you could add a spec to cover this fix?

javierjulio:
There is a newer PR addressing this with tests in #7671. Thank you.

Issue #5669, opened 2019-01-23
feature request: ajaxed pagination


jshow:
paginated_collection does not pass the remote: option to Kaminari

PR ⬇️

PR #5667, opened 2019-01-22
Allow :remote option to flow to Kaminari

jshow:
https://github.com/activeadmin/activeadmin/issues/5669

```
posts = member.posts.includes(:replies)

paginated_collection(posts.page(params[:page]).per(10), remote: true, params: {controller: 'admin/members', action: 'my_posts'}, download_links: false) do
  table_for(collection, :class => 'index_table') do
        column 'ID', :id

        column 'Title', :title do |post|
          link_to(post.title, "/admin/posts/#{post.id}")
        end
  end
end
```

PR #5792, opened 2019-06-25
Support remote pagination

jshow:
Closes https://github.com/activeadmin/activeadmin/issues/5669.

Replaces https://github.com/activeadmin/activeadmin/pull/5667.

### DIFF
--- a/docs/8-custom-actions.md
+++ b/docs/8-custom-actions.md
@@ -33,7 +33,7 @@ pointing to the `Admin::PostsController#import_csv` controller action.
 To paginate a table using ajax within a page, the following aproach works with Kaminari
 
 ```ruby
-ActiveAdmin.register Post do
+ActiveAdmin.register Member do
   collection_action :member_posts, method: :get do
     render 'admin/members/member_posts', context: self
   end
@@ -54,7 +54,7 @@ panel "#{member.name} Posts" do
       table_for(collection, :class => 'index_table') do
         column 'ID', :id
 
-        column 'Title', :title do |story|
+        column 'Title', :title do |post|
           link_to(post.title, "/admin/posts/#{post.id}")
         end
       end
--- a/docs/8-custom-actions.md
+++ b/docs/8-custom-actions.md
@@ -27,6 +27,44 @@ ActiveAdmin.register Post do
 end
 ```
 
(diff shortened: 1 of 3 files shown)


docs/8-custom-actions.md
```diff
@@ -30,6 +30,44 @@ end
 This collection action will generate a route at `/admin/posts/import_csv`
 pointing to the `Admin::PostsController#import_csv` controller action.
 
+To paginate a table using ajax within a page, the following aproach works with Kaminari
+
+```ruby
+ActiveAdmin.register Post do
+  collection_action :member_posts, method: :get do
+    render 'admin/members/member_posts', context: self
+  end
```
deivid-rodriguez: Shouldn't this be added to the `Member` admin page?

docs/8-custom-actions.md
```diff
@@ -30,6 +30,44 @@ end
...
+
+  span id: 'member_posts' do
+    render partial: 'admin/members/member_posts'
+  end
```
deivid-rodriguez: I wanted to try this, but I'm not sure in which context this is supposed to be rendered. Is it in the show page? :thinking:

docs/8-custom-actions.md
```diff
@@ -30,6 +30,44 @@ end
...
+# app/views/admin/members/_member_posts.html.arb
+member = Member.find(params[:id])
+
+panel "#{member.name} Posts" do
+  columns do
+    paginated_collection(member.posts.page(params[:page]).per(20), remote: true, params: {controller: 'admin/members', action: 'member_posts'}, param_name: :page, download_links: false) do
+      table_for(collection, :class => 'index_table') do
+        column 'ID', :id
+
+        column 'Title', :title do |story|
```
deivid-rodriguez: This block variable should be `post` I guess?

docs/8-custom-actions.md
jshow: the above partial (for which an example is below) is an #index - render an index of all `posts` for a `member`

jshow:
thanks for your comments @deivid-rodriguez

docs/8-custom-actions.md
deivid-rodriguez: But I then you need to wrap the `span` helper with an `index do ... end` block, right?

PR #5794, opened 2019-06-29
Update `status_tag` to support separate `nil` and `false` boolean cases

javierjulio:
When using the `status_tag` component with boolean fields, both `nil` and `false` are treated the same and have the same output but without a way to modify that other than overriding the component.

My suggestion is to split those 2 conditions so for the `nil` case it would now return "Unset" but the locale will keep a display value of "No" for that locale key. That **will maintain backwards compatibility** so when `status_tag` is called with `nil` or `false` it will continue to display the same "No" label but provide the option to customize the label for the `nil` case only, leaving the `false` one as is.

## Todo List
- [x] Improve test case and description
- [x] Look into [updating the documentation](https://github.com/activeadmin/activeadmin/search?l=Markdown&q=status_tag&type=) to include a note on the "unset" case
 - [x] Add a note to the changelog?
- [x] Add “no” CSS class for unset case to preserve same output
- [x] Test CSS class output for unset case
- [x] Update all other locale files with new "unset" key

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Minor
 
+* The `status_tag` component now supports different labels for `false` and `nil` boolean cases through the locale. Both default to display "No" for backwards compatibility. [#5794] by [@javierjulio]
 * Add Macedonian locale. [#5710] by [@violeta-p]
 
 ### Bug Fixes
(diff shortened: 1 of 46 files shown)


deivid-rodriguez:
@javierjulio This looks like a great improvement to me. Only consideration is whether "unset" or something like that would convey the intended meaning better than "unknown"?

javierjulio:
@deivid-rodriguez thanks! Yes, from the perspective of ActiveAdmin using "unset" would be much better. Using the term "unknown" makes more sense to me as a user of the admin and what I would want displayed/seen. Thanks!

Let me know if you have any thoughts on whether this needs a changelog and/or doc update (link in description). Give me some time and I'll get this updated.

deivid-rodriguez:
@javierjulio I think both a changelog entry and a docs note would be awesome!

javierjulio:
@deivid-rodriguez ok great, thanks! I've made the "unset" changes. I'll circle back and do the changelog and docs as well.

javierjulio:
Awesome thanks David! 🙏🏻

PR #5797, opened 2019-07-08
Cleaner test application setup

deivid-rodriguez:
This a some work I made for #5563. The idea was to be able to have a separate Gemfile including `turbolinks`, run tests against it, and be able to write a regression spec for #5563 against that Gemfile. I managed to run tests against a turbolinks-enabled application, but I'm not sure why the bug didn't reproduce there either.

The previous setup didn't work for the previously mentioned purpose because the test application folder name was based off the Rails version, not the gemfile name, so the different test applications for each Gemfile would collide if they used the same Rails version. This doesn't happen with the new setup and we could potentially run tests against two different Gemfiles using the same Rails version.

Since I could make having this turbolinks gemfile useful to catch the bug in #5563, this PR is not really necessary, but I guess it makes the setup more generic, so we could merge it in case we need something like this in the future.


deivid-rodriguez:
I introduced a few extra changes here:

* Track coverage for the test application's code. I did this because the coverage check failed, and it seems like a good way to ensure that we don't leave dead code in there.

* Add tasks to regenerate the test application "forcefully", because many times I find myself running `rm -rf tmp/rails/rails52 && bin/rake setup`, and the new task `bin/rake setup:force` provides a handy shortcut for it.

spec/rails_helper.rb
```diff
@@ -2,9 +2,9 @@
 
 ENV['RAILS_ENV'] = 'test'
 
-require "active_admin/dependency"
+require_relative "../tasks/test_application"
 
-require_relative "../tmp/rails/rails-#{ActiveAdmin::Dependency.rails_version}/config/environment"
+require "#{ActiveAdmin::TestApplication.new.full_app_dir}/config/environment.rb"
```
javierjulio: I had a concern here possibly with the defaults but both would be ones we'd want anyway since this is always running in the test environment (along with the `features/support/env.rb` file) so all good.

deivid-rodriguez:
It doesn't matter since it's just about the dev environment, but let's merge it!

deivid-rodriguez:
Thanks @javierjulio! :heart:

javierjulio:
No problem, anytime! Thanks for making regular improvements with this. Always appreciated.

PR #5800, opened 2019-07-08
fix CSVBuilder not respecting global ":humanize_name" setting

### LINKED ISSUE #5799: CSV Export not using global humanize_name correctly
### Expected behavior

Set `config.csv_options = { humanize_name: false }` in the initializer.
Expected all CSV exports to use actual column names.

### Actual behavior

Column names still used humanized names.

### How to reproduce

Set `config.csv_options = { humanize_name: false }` in the initializer and download any CSV file.

HappyKadaver:
fix CSVBuilder not respecting default ":humanize_name" setting in "ActiveAdmin.application.csv_options"

change two test cases to allow for extra options in CSVBuilder as long as it assumes any options directly passed to the initializer
fixes #5799

If someone can tell me how I can set ActiveAdmin.application.csv_options for a single spec without affecting other specs I will add some tests for this.

### DIFF
--- a/features/index/format_as_csv.feature
+++ b/features/index/format_as_csv.feature
@@ -79,6 +79,38 @@ Feature: Format as CSV
       | title | body |
       | Hello, World | (.*) |
 
+  Scenario: With humanize_name option turned off globally
+    Given a configuration of:
+    """
+      ActiveAdmin.application.csv_options = { humanize_name: false }
+      ActiveAdmin.register Post do
+      end
+    """
+    And a post exists
+    When I am on the index page for posts
+    And I follow "CSV"
+    Then I should download a CSV file with "," separator for "posts" containing:
+      | id  | title | body | published_date | position | starred | foo  | created_at | updated_at |
+      | (.*)| (.*)  | (.*) | (.*)           | (.*)     | (.*)    | (.*) | (.*)       | (.*)       |
+
+  Scenario: With humanize_name option turned off globally and enabled locally
+    Given a configuration of:
+    """
+      ActiveAdmin.application.csv_options = { humanize_name: false }
+      ActiveAdmin.register Post do
+        csv humanize_name: true do
+          column :title
+          column :body
+        end
+      end
+    """
+    And a post exists
+    When I am on the index page for posts
+    And I follow "CSV"
+    Then I should download a CSV file with "," separator for "posts" containing:
+      | Title | Body |
+      | (.*)  | (.*) |
+
   Scenario: With CSV option customization
     Given a configuration of:
     """
--- a/lib/active_admin/csv_builder.rb
+++ b/lib/active_admin/csv_builder.rb
@@ -33,7 +33,7 @@ module ActiveAdmin
     def initialize(options = {}, &block)
       @resource = options.delete(:resource)
       @columns = []
-      @options = options
+      @options = ActiveAdmin.application.csv_options.merge options
       @block = block
     end
 
@@ -44,7 +44,6 @@ module ActiveAdmin
     def build(controller, csv)
       @collection  = controller.send :find_collection, except: :pagination
       columns      = exec_columns controller.view_context
-      options      = ActiveAdmin.application.csv_options.merge self.options
       bom          = options.delete :byte_order_mark
       column_names = options.delete(:column_names) { true }
       csv_options  = options.except :encoding_options, :humanize_name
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
@HappyKadaver We have setup for that in https://github.com/activeadmin/activeadmin/blob/97e48ece25b5ea534183f47acc68dda8f9e58e76/features/index/format_as_csv.feature, so you can probably add a scenario in there.

features/index/format_as_csv.feature
```diff
@@ -79,6 +79,38 @@ Feature: Format as CSV
...
+    Given a configuration of:
+    """
+      ActiveAdmin.application.csv_options = { humanize_name: false }
+      ActiveAdmin.register Post do
+      end
+    """
+    And a post exists
+    When I am on the index page for posts
+    And I follow "CSV"
+    And I should download a CSV file with "," separator for "posts" containing:
```
deivid-rodriguez: s/And/Then/

features/index/format_as_csv.feature
```diff
@@ -79,6 +79,38 @@ Feature: Format as CSV
...
+        csv humanize_name: true do
+          column :title
+          column :body
+        end
+      end
+    """
+    And a post exists
+    When I am on the index page for posts
+    And I follow "CSV"
+    And I should download a CSV file with "," separator for "posts" containing:
```
deivid-rodriguez: s/And/Then/ for readability.

lib/active_admin/csv_builder.rb
```diff
@@ -42,7 +44,7 @@ def column(name, options = {}, &block)
     def build(controller, csv)
       @collection  = controller.send :find_collection, except: :pagination
       columns      = exec_columns controller.view_context
-      options      = ActiveAdmin.application.csv_options.merge self.options
+      options      = @options.clone
```
deivid-rodriguez: I don't think you need `clone` here?

```suggestion
      options      = self.options
```

should do the trick?

lib/active_admin/csv_builder.rb
```diff
@@ -32,7 +32,9 @@ def self.default_for_resource(resource)
 
     def initialize(options = {}, &block)
       @resource = options.delete(:resource)
-      @columns, @options, @block = [], options, block
+      @columns = []
+      @options = ActiveAdmin.application.csv_options.merge options
+      @block = block
```
deivid-rodriguez: I introduced the change to make this assignment sequential instead of parallel separately in #5803, so I would need you to rebase on top of that.

features/index/format_as_csv.feature
HappyKadaver: Changed last "And" to "Then" now it reads like a sentence if you ignore the configuration block.

lib/active_admin/csv_builder.rb
```diff
@@ -44,10 +44,9 @@ def column(name, options = {}, &block)
     def build(controller, csv)
       @collection  = controller.send :find_collection, except: :pagination
       columns      = exec_columns controller.view_context
-      options      = ActiveAdmin.application.csv_options.merge self.options
-      bom          = options.delete :byte_order_mark
-      column_names = options.delete(:column_names) { true }
-      csv_options  = options.except :encoding_options, :humanize_name
```
deivid-rodriguez: I think leaving the above three lines untouched and use the accessor should work and makes the fix more clear (move the csv option merge to the initializer because it's needed before `build` is called).

lib/active_admin/csv_builder.rb
```diff
@@ -44,7 +44,7 @@ def column(name, options = {}, &block)
     def build(controller, csv)
       @collection  = controller.send :find_collection, except: :pagination
       columns      = exec_columns controller.view_context
-      options      = ActiveAdmin.application.csv_options.merge self.options
+      options      = @options
```
deivid-rodriguez: We don't need this line at all. `options` (used below) already means reading the instance variable `@options`, right?

HappyKadaver: You are right. I totally missed the attr_reader for options

HappyKadaver:
:rofl: I hope I didn't take to much of your time. You have a really nice testing setup. The actual tests on the Changelog did catch me off guard but now we should have it. :laughing:

javierjulio:
Looks great. 👌🏻 Thanks to you both for finishing this! ❤️ We'll get this in now.

PR #5802, opened 2019-07-10
Make sure show page title is escaped

### LINKED ISSUE #5693: Result of guessed display name not escaped
### Expected behavior

I have a User model that has a method `#full_name` and no method `#display_name`. Since `#full_name` is mostly a concatenation of user-supplied first and last name fields, it may contain adversarial HTML code. The result is also not `html_safe`.

I expect the result of `#full_name` to be used automatically as the title the show page, and to be automatically escaped, as happens to any string that is not `html_safe` in Rails.

### Actual behavior

The result of `#full_name` is used automatically in the show page, but it is not escaped.

If I override the title by using

```ruby
  show title: :full_name do
    ...
  end
```

.. it is escaped.

### How to reproduce

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  if ENV['ACTIVE_ADMIN_PATH']
  else
  end

  gem 'rails', '5.2.1.1'

  gem 'sass-rails', '5.0.7'
  gem 'sqlite3', '1.3.13', platform: :mri
  gem 'activerecord-jdbcsqlite3-adapter', "52.0", platform: :jruby
  gem 'jruby-openssl', '0.10.1', platform: :jruby
end

require 'active_record'

  create_table :active_admin_comments, force: true do |t|
    t.string :namespace
    t.text :body
    t.string :resource_type
    t.bigint :resource_id
  end

    t.string :full_name
  end
end

require 'action_view/railtie'
require 'active_admin'

  config.root = __dir__

  config.eager_load = false
  Rails.logger  = config.logger
end

end

end

ActiveAdmin.setup do |config|
end

Rails.application.initialize!

  content do
    'Test Me'
  end
end

ActiveAdmin.register User do
end

  ActiveAdmin.routes(self)
end

require 'minitest/autorun'
require 'rack/test'
require 'rails/test_help'

  def test_admin_users
    john = User.create! full_name: '<a href="https://somewhere-nasty.com/">John Doe</a>'
    get admin_user_url(john)
    assert_response :success
    assert_match 'John Doe', response.body
    refute_match '<a href="https://somewhere-nasty.com/">John Doe</a>', response.body
  end

  private

  def app
    Rails.application
  end
end
```

deivid-rodriguez:
This PR changes the `display_name` method (used to render the show page title, for example) to automatically escape HTML.

Fixes #5693.

Note that this PR changes the way we fixed https://github.com/activeadmin/activeadmin/pull/5284 by escaping instead of sanitizing, just like we did in https://github.com/activeadmin/activeadmin/pull/5299.

Escaping fixes issues like #5693 better than `sanitize` because the default Rails sanitizer only removes certain tags, not all, and it leads to behavior more similar to Rails defaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Bug Fixes
 
 * Fix pundit policy retrieving for static pages when the pundit namespace is :active_admin. [#5777] by [@kwent]
+* Fix show page title not being properly escaped if title's content included HTML. [#5802] by [@deivid-rodriguez]
 
 ## 2.1.0 [☰](https://github.com/activeadmin/activeadmin/compare/v2.0.0..v2.1.0)
 
@@ -468,6 +469,7 @@ Please check [0-6-stable] for previous changes.
 [#5758]: https://github.com/activeadmin/activeadmin/pull/5758
 [#5777]: https://github.com/activeadmin/activeadmin/pull/5777
 [#5794]: https://github.com/activeadmin/activeadmin/pull/5794
+[#5802]: https://github.com/activeadmin/activeadmin/pull/5802
 
 [@5t111111]: https://github.com/5t111111
 [@aarek]: https://github.com/aarek
--- a/features/show/page_title_with_html.feature
+++ b/features/show/page_title_with_html.feature
@@ -1,0 +1,11 @@
+Feature: Show - Page Title with HTML
+
+  Page Title is escaped
+
+  Scenario: Set an HTML string as the title
+    Given a post with the title "<a href='https://somewhere-nasty.com/'>John Doe</a>" written by "Jane Doe" exists
+    And a show configuration of:
+    """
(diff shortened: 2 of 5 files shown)


deivid-rodriguez:
@mvz Can you confirm that this fixes your issue in a reasonable way?

@markstory Are you ok with this alternative solution to the problem you fixed in #5284?

markstory:
@deivid-rodriguez I no longer work on the application that was using ActiveAdmin but HTML escaping is generally a better idea than sanitizing :+1:

mvz:
@deivid-rodriguez yes, this looks like the right fix to me, thanks!

deivid-rodriguez:
Thanks @timoschilling!

javierjulio:
Great update @deivid-rodriguez! Thanks to everyone here for pitching in. 👍🏻❤️

PR #4068, opened 2015-08-11
Skip optimization if decorated collection does not support it.

varyonic:
Fixes #4059

PR #5811, opened 2019-07-24
Make count query optimisation work when using decorate_with

irmela:
#3848 introduced a great performance improvement for pagination total_count. After #4068 this optimization was not available when using `decorate_with`.

This is another approach of #5389 with minimal change. Please let me know in case I missed something.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Enhancements
+
+* Make optimization to not use expensive COUNT queries also work for decorated actions. [#5811](https://github.com/activeadmin/activeadmin/pull/5811) by [@irmela]
+
 ## 2.3.1 [☰](https://github.com/activeadmin/activeadmin/compare/v2.3.0..v2.3.1)
 
 ### Bug Fixes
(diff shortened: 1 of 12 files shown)


deivid-rodriguez:
Is it possible to add a test?

irmela:
@deivid-rodriguez I'm not really sure what would be a good test case for that. I could add a unit test:

```ruby
# spec/unit/resource_controller/decorators_spec.rb

it 'delegates offset method' do
  expect(applied).to respond_to(:offset)
end
```

But I don't think think provides much value. I would appreciate some guidance on how to best test this change 🙏

deivid-rodriguez:
Yeah, I see how a proper test can be difficult. I guess ideally we should test that when using `decorate_with` on a resource, pagination produces no `SELECT COUNT` queries. We could leverage https://github.com/civiccc/db-query-matchers for that, but it still seems tricky.

I'll look into this manually and consider merging without a test.

One question that comes to mind at first sight is whether we can undo #4068 if we include this patch.

irmela:
> One question that comes to mind at first sight is whether we can undo #4068 if we include this patch.

I will look into that.

irmela:
I tried to investigate if the issue fixed by #4068 still exists with this change but it feels a little like #4068 got merged without any deeper understanding of why the issue occurred. That makes it super hard to reproduce so I didn't really come of which a answer to the question if we can revert the change (` && collection.respond_to?(:offset)`).

deivid-rodriguez:
You mentioned that the regression this PR is fixing was created after #4068 was merged. So if after reverting #4068 the regression is still fixed, and no tests are broken, I would undo it.

irmela:
@deivid-rodriguez I reverted #4068 and tests are passing.

deivid-rodriguez:
Thanks @irmela! I'll have a look at this soon :+1:.

deivid-rodriguez:
I added a test for this change! Setting this as ready.

PR #5815, opened 2019-07-25
Detach previous batch_action click handler before attaching new one

sgara:


deivid-rodriguez:
Thanks @sgara for wrapping this up! CI is failing though, can you have a look?

sgara:
Could those be flaky tests? I can see that they failed on different steps than first run now.
Haven’t finished replicating failing envs locally yet, still in progress

deivid-rodriguez:
They seem indeed flaky, but created by this PR, because we have no flakies in master...

It could be some order dependent issue, caused by some of the specs in there not resetting some state, maybe. If you pass `--order random:<seed>` with the same seed as in the CI, hopefully you should be able to replicate.

sgara:
Ok i was wondering if they were known flaky indeed. 
I have been able to reproduce it locally. The behavior is a bit strange, it looks like the batch action confirmation is being fired twice, most likely because of another turbolinks/click handler related issue I guess. Still looking into it

deivid-rodriguez:
Great, thanks so much for working on this!

app/assets/javascripts/active_admin/lib/batch_actions.es6
```diff
@@ -3,7 +3,7 @@ const onDOMReady = function() {
   // Use ActiveAdmin.modal_dialog to prompt user if
   // confirmation is required for current Batch Action
   //
-  $('.batch_actions_selector li a').click(function(event){
+  $('.batch_actions_selector li a').off('click').on('click', function(event){
```
javierjulio: @sgara while you are testing this, could you try rewriting this in the delegated event handling style to see if it fixes the behavior in Turbolinks but also does not affect non-Turbolinks usage?

```js
$(document).on('click', '.batch_actions_selector li a', function(event){
```

It really should be using that style like we do elsewhere in this file. At least I'd prefer that it do that rather than the off/on switch here. I'm just not sure if that will ultimately resolve the Turbolinks issue.

sgara: I wasn’t fond of off/on either 😅. I’ll give it a try

sgara: Tested delegated style but couldn't get it working at all, don't really know why though. I moved `off` on its own to clear both click and complete handlers, and I figured this would be much easier to read.

sgara:
Got the whole thing and tests working. Currently, this adds turbolinks to all tests by default but ideally I guess we should be able to enable/disable it at will. However, doubling the number of CI runs would drastically slow feedback on pull requests. Do you have any preference on how to handle that?

deivid-rodriguez:
Hei @sgara, yeah I agree.

I think we could add a new gemfile, say `turbolinks.gemfile`, that uses the latest version of Rails and `turbolinks`, and that should be enough. That way we get some tests in place for an application with turbolinks, but we don't need to double the CI matrix.

I actually worked on that when I first reviewed your PR. Do you want me to add it here?

sgara:
Would be great!

deivid-rodriguez:
Great, I'll take that next. Thanks for your patience!

sgara:
👋 hi! Just checking if there’s anything else I should do here :)

deivid-rodriguez:
Hi @sgara! Thanks for the ping!

No, not really, it's now on top of my TODO list to consider adding a separate CI run where the app includes turbolinks. If I end up doing that, this PR might need some tweaking and I'll get to you for that, otherwise I'll merge it as it is.

deivid-rodriguez:
Hi again @sgara. I added a separate gemfile to test against turbolinks and everything seems to be working fine. As a matter of fact, we probably don't need the additional cucumber scenario because a a few exisiting tests actually fail when scenarios are run against a turbolinks application.

Let me get other PRs in that affect our development environment first, and then I'll either update this PR or open a separate one including your fix and my additional changes :+1:.

sgara:
Alright thanks for the update. Ping me if I can be of some help ;)

deivid-rodriguez:
Superseded by #5842.

PR #5816, opened 2019-07-25
Fix filtering by nested resource

deivid-rodriguez:
After enabling a sample nested resource in #5814, I found a crash in the users page, when filtering by post.

I guess this kind of filtering is pretty uncommon (it always results in a single record), maybe it shouldn't be included by default?

Anyways, this should fix the crash, which by the way looks like this:

```
    And I press "Filter"                                                              # features/step_definitions/web_steps.rb:24
      undefined method `user' for #<Post:0x000055a5adf3d8d0>
      Did you mean?  super (ActionView::Template::Error)
      ./lib/active_admin/resource/routes.rb:113:in `public_send'
      ./lib/active_admin/resource/routes.rb:113:in `route_instance_params'
      ./lib/active_admin/resource/routes.rb:79:in `instance_path'
      ./lib/active_admin/resource/routes.rb:19:in `route_instance_path'
      ./lib/active_admin/view_helpers/auto_link_helper.rb:29:in `auto_url_for'
      ./lib/active_admin/view_helpers/auto_link_helper.rb:15:in `auto_link'
      ./lib/active_admin/view_helpers/display_helper.rb:78:in `pretty_format'
      ./lib/active_admin/filters/active_sidebar.rb:31:in `block (8 levels) in block'
(...)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug Fixes
 
 * Fix CSVBuilder not respecting `ActiveAdmin.application.csv_options = { humanize_name: false }` setting. [#5800] by [@HappyKadaver]
+* Fix crash when displaying current filters after filtering by a nested resource. [#5816] by [@deivid-rodriguez]
 
 ## 2.2.0 [☰](https://github.com/activeadmin/activeadmin/compare/v2.1.0..v2.2.0)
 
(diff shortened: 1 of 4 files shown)


javierjulio:
In regards to the question in the description, I think it would be better that ActiveAdmin didn't render all/any filters by default, instead it would have to be specified explicitly in the file. I've had trouble in the past leaving the defaults so I've been in the habit of always specifying filters on all pages.

PR #5587, opened 2018-11-23
Get ready for 1.4.1 release

deivid-rodriguez:
Let's try to get used to releasing stuff early! I plan to release @tiagotex's bug fix as 1.4.1.

PR #5817, opened 2019-07-26
Fix `belongs_to` specification

deivid-rodriguez:
I had this change locally, but I forgot to push it. It was supposed to be part of #5587 

It's necessary to make filtering by post in the user's index page (using the fix in #5816), otherwise, the applied filter cannot be properly displayed.


javierjulio:
Oh I see you removed review request. I’ll leave this approved for you but if you need me to review again just let me know!

deivid-rodriguez:
Yeah, I found some errors so I removed the request and the label. They should be fixed now.

javierjulio:
Great thanks! I'll get this in now.

PR #5822, opened 2019-08-01
Hide "Last" link when `pagination_total` is false

deivid-rodriguez:
I was having a look at #5811 and I noticed that when using `pagination_total: false`, a "Last" link is still displayed, and it has the same link as the "Next" link.

In this case, we don't have the total pages information so displaying an incorrect "Last" link is misleading.

The solution is a bit verbose, but I'm unsure how to do it otherwise. What I did was to copy the default activeadmin pagination theme, but skip the "Last" link and the the `_last_page.html.erb` partial. And when `pagination_total` is false, use the alternative "countless" theme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Fix CSVBuilder not respecting `ActiveAdmin.application.csv_options = { humanize_name: false }` setting. [#5800] by [@HappyKadaver]
 * Fix crash when displaying current filters after filtering by a nested resource. [#5816] by [@deivid-rodriguez]
+* Fix pagination when `pagination_total` is false to not show a "Last" link, since it's incorrect because we don't have the total pages information. [#5822] by [@deivid-rodriguez]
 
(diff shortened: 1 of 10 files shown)


tiagotex:
@deivid-rodriguez I found the same problem and started debugging it on my local version (didn't see that this was already fixed on master), so I solved this in a different way.

I change Kaminari to stop showing first page link on second page pagination and last page link on penultimate page pagination. https://github.com/kaminari/kaminari/pull/1005

This indirectly fixes this problem. If the PR gets merged would you like me to remove this workaround?

deivid-rodriguez:
Yeah, it seems to indirectly fix the problem indeed. Thanks! And sorry if you had to work extra because this fix was not yet released :sweat_smile:.

If this is changed in `kaminari`, yeah, it'd be great to remove this workaround, because it's also quite ugly... But we'll have to wait a while for that, until the full range of `kaminari` versions we support includes your fix.

PR #5823, opened 2019-08-06
Improve support for renamed resources in belongs_to associations

### LINKED ISSUE #3647: optional belongs_to for renamed active_admin resources creates incorrect url helpers

Kris-LIBIS:
This pull request fixes #3647 and improves the reliability of the url helpers in general when resources are renamed. Renaming resources is typically required when models are encapsulated in modules.

### DIFF
--- a/docs/3-index-pages/index-as-table.md
+++ b/docs/3-index-pages/index-as-table.md
@@ -193,6 +193,13 @@ controller do
 end
 ```
 
+You can also define associated objects to include outside of the
+`scoped_collection` method:
+
+```ruby
+includes :publisher
+```
+
 Then it's simple to sort by any Publisher attribute from within the index table:
 
 ```ruby
--- a/lib/active_admin/views/index_as_table.rb
+++ b/lib/active_admin/views/index_as_table.rb
@@ -192,6 +192,13 @@ module ActiveAdmin
     # end
     # ```
     #
+    # You can also define associated objects to include outside of the
+    # `scoped_collection` method:
+    #
+    # ```ruby
+    # includes :publisher
+    # ```
+    #
     # Then it's simple to sort by any Publisher attribute from within the index table:
     #
     # ```ruby


Kris-LIBIS:
@deivid-rodriguez All the tests seem to pass, but one appears to time-out during the coverage report generation. I'm not sure what I can do about that. Could you have a look at that please?

deivid-rodriguez:
@Kris-LIBIS Sorry, that's a flaky we have in master. I restarted the job and now the lack of coverage should be true correct. Is there some obvious dead code in here, or parts that we should be testing?

I plan to have a look at this soon, since this is a long standing issue I'd like to see fixed, thanks for doing this!

deivid-rodriguez:
@Kris-LIBIS I think I found a solution that doesn't require copying copy from Rails. I'll amend this PR tomorrow if that's ok with you!

Kris-LIBIS:
@deivid-rodriguez Yes, please do. I'm not happy with the copying of the Rails code, but that is what I inherited from the old PR. There may be a catch, though as the PolymorphicRoutes module was only added to Rails in version 5.1.3, if I'm not mistaken. We may have to let the copied code in and only enable it for Rails < 5.1.3.

Not being familiar with the Rails internals, I was wondering if there might be a more generic or better way to get hold of the class-name mapping as I did in the get_resource_name method?

deivid-rodriguez:
@Kris-LIBIS I'm so sorry I closed this accidentally when trying to update your branch. See #5826 for the updated PR.

Kris-LIBIS:
@deivid-rodriguez That is OK. Your updated PR looks very good. I'll wait with deleting the fork until the PR is published in a release as my current project depends on the change.

deivid-rodriguez:
@Kris-LIBIS If you're using your branch in production, make sure you restore the old reference (https://github.com/activeadmin/activeadmin/commit/8f115d2c7b9b452b15ea4d3bc05dd800a353467b) before your next deploy because I accidentally messed it up. Or, better than that, use my branch instead so you can also give it a try :)

PR #5826, opened 2019-08-14
Fix renamed resources and optional `belongs_to`

### LINKED ISSUE #3647: optional belongs_to for renamed active_admin resources creates incorrect url helpers

deivid-rodriguez:
I still want to have another look since I think there might be a bug in the resource <-> model matching part when a resource is renamed different in different namespaces.

But I think it's looking good!

Supersedes #5823.
Fixes #3647.
Closes #5483.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix CSVBuilder not respecting `ActiveAdmin.application.csv_options = { humanize_name: false }` setting. [#5800] by [@HappyKadaver]
 * Fix crash when displaying current filters after filtering by a nested resource. [#5816] by [@deivid-rodriguez]
 * Fix pagination when `pagination_total` is false to not show a "Last" link, since it's incorrect because we don't have the total pages information. [#5822] by [@deivid-rodriguez]
+* Fix optional nested resources causing incorrect routes to be generated, when renamed resources (through `:as` option) are involved. [#5826] by [@ndbroadbent], [@Kris-LIBIS] and [@deivid-rodriguez]
 
 ## 2.2.0 [☰](https://github.com/activeadmin/activeadmin/compare/v2.1.0..v2.2.0)
 
@@ -485,6 +486,7 @@ Please check [0-6-stable] for previous changes.
 [#5802]: https://github.com/activeadmin/activeadmin/pull/5802
 [#5816]: https://github.com/activeadmin/activeadmin/pull/5816
 [#5822]: https://github.com/activeadmin/activeadmin/pull/5822
+[#5826]: https://github.com/activeadmin/activeadmin/pull/5826
 
 [@5t111111]: https://github.com/5t111111
 [@aarek]: https://github.com/aarek
@@ -554,3 +556,5 @@ Please check [0-6-stable] for previous changes.
 [@panasyuk]: https://github.com/panasyuk
 [@jscheid]: https://github.com/jscheid
 [@violeta-p]: https://github.com/violeta-p
+[@ndbroadbent]: https://github.com/ndbroadbent
+[@Kris-LIBIS]: https://github.com/Kris-LIBIS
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
I fixed the mentioned issue, and now this should be ready to merge. I want to do some squashing and add @ndbroadbent's authorship to the commits, but this is good for an initial review.

javierjulio:
@deivid-rodriguez sorry, I'm not sure I can help much here. I'm not familiar with the nested/belongs_to feature. I don't use it.

@Fivell sorry to bother, would you have some time to review this? GitHub is suggesting you as a recent editor of these files.

@Kris-LIBIS are you able to look at this and try it out too? Would help to get your review if you can. Thanks!

deivid-rodriguez:
No problem @javierjulio!

@ndbroadbent (author of the PR that first got this working) and @senid231 (author of the original issue) might also be helpful reviewers.

Fivell:
hello @deivid-rodriguez  @javierjulio , 
 As I can remember this  Bug report has some more examples to check than we have in this PR,  https://gist.github.com/senid231/0dcfc8b89eb0234ba5cb5ac66fec5be7 
I think its better to run it against this branch before merging

deivid-rodriguez:
Thanks @Fivell, I didn't notice that. I'll run that bug report template against this branch and see if it's worth adding some extra coverage :+1:.

Kris-LIBIS:
IMHO I think I have covered that use case by namespacing the Blog::User in the current feature. What the original author's (@ndbroadbent) test case did not cover was when both parent and child of the belong to are renamed and that is what my test case and @senid231 's test case cover. But of course, there is no harm in running the test in the gist. I have good hopes that it will pass with this fix.

deivid-rodriguez:
Some heads up here. I run the bug report template and it fails 2 tests against this PR (against 4 failures against master).

I have coded an extra fix for those 2 failures, but I still want to add coverage for it inside our specs.

deivid-rodriguez:
Hopefully it's ok now.

I didn't add more coverage per se, what I did was to modify the previous cucumber feature to look more like the bug report template. After I did that, the same issues the bug template was showing reproduced.

First one, not being able to infer the correct parent class to retrieve records from:

```
    And I click "Author Articles"                                 # features/step_definitions/index_views_steps.rb:1
      undefined method `find' for nil:NilClass (NoMethodError)
      ./lib/active_admin/resource_controller/data_access.rb:74:in `scoped_collection'
      ./lib/active_admin/resource_controller/data_access.rb:57:in `find_collection'
      ./lib/active_admin/resource_controller/data_access.rb:44:in `collection'
      ./lib/active_admin/resource_controller/streaming.rb:12:in `index'
      ./features/step_definitions/index_views_steps.rb:2:in `/^I click "(.*?)"$/'
      features/renamed_resource.feature:66:in `And I click "Author Articles"'
```

Fixed that by passing the correctly renamed class to inherited_resources's `belongs_to` helper.

After that, another issue showed up where the parent class was not being properly "translated" before calling the polymorphic route helpers:

```
    And I click "Author Articles"                                 # features/step_definitions/index_views_steps.rb:1
      undefined method `new_admin_user_article_path' for #<Admin::ArticlesController:0x000055b6ee43c7b8>
      Did you mean?  new_admin_article_path
                     new_admin_user_password_path
                     new_admin_author_article_path
                     new_admin_user_session_path (ActionView::Template::Error)
      ./lib/active_admin/resource_controller/polymorphic_routes.rb:12:in `polymorphic_path'
```

Fixed that it by looking at the parent renamed resource to do the mapping so that the polymorphic helpers generate the correct route.

Kris-LIBIS:
I checked it in my application and it still works. 
@deivid-rodriguez I would prefer if you would reinstate my namespaced parent test case in the features though as apparently it is a different case. Just to make sure it will not break in the future.

deivid-rodriguez:
I didn't like the previous scenario because it was quite confusing (it renamed models as already existing models), and it also failed to detect at least two issues the bug report template detected. And I don't think namespaced models are directly involved in this issue anyways?

deivid-rodriguez:
I'll land this now. Thanks everyone, hopefully `belongs_to` is a bit friendlier now :)

PR #5827, opened 2019-08-14
Docs site local improvements

deivid-rodriguez:
I booted the documentation site locally and I was getting unstyled pages and a few broken links. This does not happen in production (I guess the plugins are applied by default), but this helps for developing locally.

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,4 +1,3 @@
 source 'https://rubygems.org'
 
 gem 'github-pages'
-gem 'jekyll-redirect-from'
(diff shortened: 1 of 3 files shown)


javierjulio:
I reran the build as one failed due to a timeout error.

deivid-rodriguez:
Thanks @javierjulio. I restarted it again and now it passed.

Hopefully 9.2.8.0 will be better! https://www.jruby.org/2019/08/12/jruby-9-2-8-0.html.

javierjulio:
No problem, thank you! This is a great change. ❤️

PR #5830, opened 2019-08-18
Test against Rails 6 final

deivid-rodriguez:


javierjulio:
Oh wow, finally! 🤣❤️

deivid-rodriguez:
Any thoughs on bumping the ransack dependency?

I tend to think we should because it might be more handy for users. Users using older versions of ransack, and upgrading to Rails 6, will get a broken `active_admin` and it could take a while to figure out that you also need to upgrade ransack.

Of course, the compromise here is that users pinned to previous versions of ransack will get resolution conflicts when upgrading.

javierjulio:
@deivid-rodriguez I think that's a good idea. I know we've done it before and is helpful. What do you think about including the bump update in the changelog and maybe a link to the ransack PR that fixes the Rails final release issue with a new release?

deivid-rodriguez:
Yeah, let's do that. Let's get this one in, and I'll add that on a separate PR.

javierjulio:
Awesome thanks David! I’ll get this in now.

PR #5837, opened 2019-08-29
Bump base rubies and rubygems in CI

deivid-rodriguez:
Skipping upgrading to 2.6.4 due to https://bugs.ruby-lang.org/issues/16136.


deivid-rodriguez:
Ready!

PR #5839, opened 2019-09-02
Rails 6 by default

deivid-rodriguez:
This PR changes our development environment to use Rails 6 by default.


deivid-rodriguez:
It should be ready now. Took the chance to upgrade some development dependencies too.

Gemfile
```diff
@@ -22,7 +22,8 @@ group :docs do
   gem 'kramdown' # Markdown implementation (for yard)
 end
 
-gem "rails", "~> 5.2.3"
-gem "activerecord-jdbcsqlite3-adapter", "~> 52.0", platform: :jruby
+gem "rails", "~> 6.0.0"
+gem "activerecord-jdbcsqlite3-adapter", "~> 60.0.rc1", platform: :jruby
+gem "sqlite3", "~> 1.4", platform: :mri
```
javierjulio: The commit says this is still required by Rails 5.x which we have in each gemfile for that major version set but if the default Gemfile is now Rails 6, why do we need it here? I was just curious.

deivid-rodriguez: This is not really necessary, but I think it makes things clearer because I tried updating `sqlite3` in all gemfiles unconditionally, and Rails 5 failed.

Previously the `sqlite` dependency was specified in `Gemfile.common`, which all of our Gemfiles use. Since now we're specifying a different `sqlite` dependency per Gemfile, I had to move the dependency away from `Gemfile.common`, and instead specify it explicitly in every Gemfile.

javierjulio: Interesting, I would think its unnecessary too even on updating but its no issue for me. I was just curious as to the reasoning. Thanks David! ❤️

PR #5841, opened 2019-09-04
Update CircleCI builds to test against ruby 2.6.4

andjosh:


deivid-rodriguez:
Hi @andjosh! 

Due to a bug in ruby (https://bugs.ruby-lang.org/issues/16136#change-81425), we won't be upgrading our CI until 2.6.5 is released.

But thanks so much for your contribution anyways!

andjosh:
@deivid-rodriguez Thanks!
We ran into that issue on our own system and I only opened this to test that out. Great to hear you already know about it!

PR #5845, opened 2019-09-06
Run RSpec against ruby-head

deivid-rodriguez:
This is just an experiment to see if the current issue with MRI 2.6.4 we're getting in #5837 and #5841 also reproduces against ruby-head.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ version: 2.1
 .install_build_deps: &install_build_deps
   run:
     name: Install build dependencies so that sassc-rails extension builds
-    command: sudo apt install make gcc g++
+    command: apt install make gcc g++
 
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
Bug has been fixed in ruby-head, and it should be backported to 2.6. So we should be able to upgrade smoothly once 2.6.5 is released: https://bugs.ruby-lang.org/issues/16136.

I still think it _might_ be interesting to run one built matrix entry against ruby-head to detect these issues early and help ruby-core. In this case, we could've prevented the bug from being shipped with ruby 2.6.4.

deivid-rodriguez:
Our 2.7 support seems pretty good at the moment. Only a single test failure, which seems like some kind of Encoding incompatibility. I [reported it to the csv gem](https://github.com/ruby/csv/issues/105).

PR #5846, opened 2019-09-06
Fix CircleCI dependency caching

deivid-rodriguez:
I noticed that we haven't been properly caching dependencies in CI for a while, so CI is slower. Hopefully this PR will fix it.


deivid-rodriguez:
This worked beautifully. I restarted the previous passing build, and every "install dependencies" step was immediate and properly used caching. We're back to an ~15 minutes build :+1:.

PR #5847, opened 2019-09-06
Make jruby hangs appear earlier

deivid-rodriguez:
The `parallel_tests` gem has this feature where it will output a dot every 60 seconds to avoid CIs that fail when no new output has not been displayed for a while.

Our jruby build is hanging randomly and when that happens, this feature makes it take hours for the job to fail.

In this commit, I'm essentially disabling this feature, by giving the heartbeat a high value, so that we actually get these failures earlier.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,7 @@ version: 2.1
   run:
     name: Run Cucumber features
     command: |
-      COVERAGE=true PARALLEL_TEST_PROCESSORS=4 bin/parallel_cucumber features/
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
This is ready, although I can't check that it really works until another hang happens.

javierjulio:
No problem! We'll get this in and see how it does. I've noticed the same issue too. Thanks so much for tackling this! ❤️ I'll get this in now.

deivid-rodriguez:
Let's see how it goes! :eyes:

PR #5856, opened 2019-09-13
Get ready for 2.3.1 release

deivid-rodriguez:
Releasing 2.3.1 to unpin ransack so that users can skip https://github.com/activerecord-hackery/ransack/issues/1039.


deivid-rodriguez:
It should be all good, thanks again! :heart:

PR #5858, opened 2019-09-13
Try potential fix for jruby hangs

deivid-rodriguez:
I want to try out [this simplecov PR](https://github.com/colszowka/simplecov/pull/746) in case it helps with the jruby hangs.

The PR doesn't mention anything about jruby, but the sympthoms are very similar: it always hangs right before the last parallel process finishes, no matter whether it's the cucumber or the RSpec suite.

So, I guess we can try it.

### DIFF
--- a/Gemfile.common
+++ b/Gemfile.common
@@ -16,7 +16,7 @@ group :test do
   gem 'capybara', '~> 3.14'
   gem 'db-query-matchers', '0.9.0'
 
-  gem 'simplecov', require: false # Test coverage generator. Go to /coverage/ after running tests
+  gem 'simplecov', '0.17.1', require: false # Test coverage generator. Go to /coverage/ after running tests
   gem 'cucumber-rails', '~> 1.5', require: false
   gem 'cucumber'
   gem 'database_cleaner'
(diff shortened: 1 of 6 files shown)


javierjulio:
Sounds like a good idea, thanks!

deivid-rodriguez:
This looks very promising since I run CI 7 times and no hangs so far. I needed to add [an extra fix to the jruby activerecord adapter](https://github.com/jruby/activerecord-jdbc-adapter/pull/1032).

deivid-rodriguez:
Oh well. So:

* Simplecov PR was merged and released.
* Jruby PR I'm not sure it's the right way/place to fix the crash.

I reverted the jruby PR for now, and upgraded to simplecov 0.17.1.

And we're still getting [a crash on jruby](https://github.com/activeadmin/activeadmin/pull/5858) :(

In my opinion, a crash is better than hanging, and since it's happening in the same place as before (right before generating the last coverage report), maybe it was also happening before, but was being hidden by the hang.

So I'd say let's upgrade.

I'll restart the failing job now.

deivid-rodriguez:
@javierjulio Seems like we got the hang again in https://app.circleci.com/jobs/github/activeadmin/activeadmin/22913, let's merge this one and see how it goes.

deivid-rodriguez:
Merging to accelerate other PRs.

PR #5855, opened 2019-09-13
Add webpacker compatibility with config switch and installation generator

sgara:
After reading a few issues and related discussions (see #5829, #5551 and #5048), and willing to switch to webpacker myself, I wanted to propose a first step toward moving to webpacker. This is of course open to discussion. The goal here is to be able to simple switch to loading active_admin assets from webpack instead of sprockets. Also introduced a gemfile to generate development app running webpacker although there is no js nor css at the moment. 

To me, the next step would be to create a npm package from sources present in this repo already. Note that a few changes may be required, I'm not sure we can have one set of assets compatible with both webpack and sprockets. But that can be discussed separately I guess

PR #5868, opened 2019-09-27
Configure rollup to bundle javascript outside of sprockets

sgara:
Move javascript source files outside of `app/assets/javascript` and configure rollup to bundle them back into `app/assets/javascript`. As JS files are already bundled, no need to keep `sprockets-es6` dependency.

⚠️ this requires rollup build to be run before merging PRs. That would be great to automate this - or at least add a step in ci build to ensure the bundle is up-to-date


deivid-rodriguez:
Hei @sgara, this is looking very good to me.

Regarding having to generate the JS for `sprockets`, I think the upsides of doing this beat the slight annoyance of having to run `yarn build` for development under sprockets. I think we can add one more `lint` check similar to the ones we already have that runs `yarn build` and makes sure no source control changes are generated.

So, from your plan, this PR would get the first two bullet points done, right? Do you think it's worth splitting the first bullet point into its own PR?

* [x] Rework js files to remove the use of ActiveAdmin global constant that isn't necessary.
* [x] Setup npm packaging and rollup (if rollup works for you of course, tell me if there another bundling option you'd prefer). JS assets would be compiled when pushing to the repo, and the generated bundle then used by sprockets.
* [ ] Release the npm package including assets ready to be imported by webpack.
* [ ] Add that use_webpacker switch when everything else is ready (#5855).

Also, I think we would need to also migrate css to be consumed by sprockets in the same way, or to keep usage of sprockets unconditional for css for the time being. Right?

sgara:
Actually, only the second point is being handled here. Reworking js files is almost ready but I needed `import` to be available before doing this nicely - which is not the case with `sprockets-es6`. So (previously) point one will have its own PR as well.

Plan has slightly changed to :
- [x] Setup npm packaging and rollup (if rollup works for you of course, tell me if there another bundling option you'd prefer). JS assets would be compiled when pushing to the repo, and the generated bundle then used by sprockets.
- [ ] Rework js files to remove the use of ActiveAdmin global constant that isn't necessary.
- [ ] Release the npm package including assets ready to be imported by webpack.
- [ ] Add that use_webpacker switch when everything else is ready (#5855).

Adding a `lint` check is a great idea, I'll look into it before merging the PR. 

As for css, we can keep them in place, and export them through npm as well. This way, we keep compatibility with both `sprockets` and `webpacker` which seems nice. I like the idea of still being able to use scss from sprockets instead of bundling them into css beforehand.

deivid-rodriguez:
Oh, of course, thanks for the info! :+1: 

A couple of questions, just out of curiosity.

* Is there any reason for using `yarn` here over `npm`?
* Could we use `webpacker` to do the same thing `rollup` is doing now?

sgara:
Well, rails itself is using yarn so I assumed developers would be already used to it. That was pretty much the only reason to chose it over npm. 
Webpack would do the job fairly easily as well. I just go inspired by active storage in fact but have no preference whatsoever. If you do prefer webpack I can make the change

deivid-rodriguez:
Following Rails sounds good to me, leave it as it is. So I think once we add the `lint` check for the bundled JS being in sync, we can proceed with this PR :+1:.

sgara:
Lint added 👍

rollup.config.js
```diff
@@ -0,0 +1,56 @@
...
+        content:stripIndent`
+          //= require jquery3
+          //= require jquery-ui/widgets/datepicker
+          //= require jquery-ui/widgets/dialog
+          //= require jquery-ui/widgets/sortable
+          //= require jquery-ui/widgets/tabs
+          //= require jquery-ui/widget
+          //= require jquery_ujs
+          //= require_self
+        `.trim()
```
deivid-rodriguez: Using a license header plugin for this job is really smart, hehehe :+1:. I don't think we need the `trim()` call though? Also, maybe we add `+ "\n"` instead to make the output a little bit prettier (I'm used to a blank line after requires)?

tasks/lint.rake
```diff
@@ -138,6 +159,11 @@ namespace :lint do
     FixmeLinter.new.run
   end
 
+  desc "Check that javascript bundle is up to date"
+  task :javascript_bundle do
+    JavascriptBundleLinter.new.run
+  end
+
```
deivid-rodriguez: Thanks for adding the lint, I was thinking instead of doing something along the lines of the `spec/gemfiles_spec.lint.rb`. The reason is that, although it works, I think it's a bit artificial to use the `Linter` class here, for example, I don't like that calling `applicable_files` has the side effect of runninng `yarn build`, or that running the linter generates source control changes.

What about

#### spec/js_bundle_spec.lint.rb

```ruby
require "open3"

RSpec.describe "JS bundle sanity" do
  it "is up to date" do
    current_bundle = File.read("app/assets/javascripts/active_admin/base.js")

    output, status = Open3.capture2e("yarn build -o tmp/bundle.js")
    expect(status).to be_success, output

    new_bundle = File.read("tmp/bundle.js")

    msg = "Javascript bundle is out of date. Please run `yarn build` and commit changes."

    expect(current_bundle).to eq(new_bundle), msg
  end
end
```

deivid-rodriguez:
@javierjulio Do you have any thoughts on the approach we are taking here? Does it look good to you?

tasks/lint.rake
sgara: Makes complete sense to me indeed

javierjulio:
Looks good so far! Should be fine going with webpacker, just curious how it will work with sprockets. Thanks!

rollup.config.js
```diff
@@ -0,0 +1,59 @@
...
+    commonjs(),
+    babel(),
+    uglify(uglifyOptions),
+    license({
+      banner: {
+        commentStyle: 'none',
+        // Add rails-style imports for sprockets usage
+        content:stripIndent`
+          /**
+           * Warning: This file is auto-generated, do not modify. Instead, make your changes in 'app/javascript/active_admin/' and run \`yarn build\`
```
deivid-rodriguez: :+1:

CONTRIBUTING.md
```diff
@@ -34,6 +34,12 @@ Now install the development dependencies:
 bundle install
 ```
 
+To update javascript bundle run (add `-w` flag for watch mode):
```
javierjulio: Since we now require Node and Yarn for development, can you please include a paragraph with that and links to each on how to install? While it’s obvious to us it won’t be to others so we should help them out.

deivid-rodriguez: Perhaps we can provide a yarn binstub similar to the one provided with Rails applications, and recommend to run that. Like

```ruby
#!/usr/bin/env ruby

begin
  exec "yarnpkg", *ARGV
rescue Errno::ENOENT
  $stderr.puts "Yarn executable was not detected in the system."
  $stderr.puts "Download Yarn at https://yarnpkg.com/en/docs/install"
  exit 1
end
```

sgara: 👍 New rails apps with webpacker do include a `yarn` binstub so I just copied it over here.
Added some documentation as well. `rollup` is being installed through yarn `devDependencies` but it's still a good thing to mention it indeed. 
Also updated calls to use the binstub.

javierjulio: Looks great, thanks!

javierjulio:
@deivid-rodriguez did you have anything else? For me everything looks good to go. Also the Tidelift check didn't complete. Is that something we can run manually to confirm before merging?

bin/yarn
```diff
@@ -0,0 +1,11 @@
...
+APP_ROOT = File.expand_path('..', __dir__)
+Dir.chdir(APP_ROOT) do
+  begin
+    exec "yarnpkg", *ARGV
+  rescue Errno::ENOENT
+    $stderr.puts "Yarn executable was not detected in the system."
+    $stderr.puts "Download Yarn at https://yarnpkg.com/en/docs/install"
+    exit 1
+  end
+end
```
deivid-rodriguez: Only small comment from me is that in my suggestion I avoided setting the `APP_ROOT` constant because it felt a bit confusing to me (we're not an app), and I don't really expect developers to run this binstub from the root of the repo.

I'm happy with the current version though if you disagree.

deivid-rodriguez:
Just a tiny comment on the binstub from my side, not important at all.

Also @sgara, if you're up for a bit of git history cleanup, I generally prefer no merge commits, and squashing "amending commits" into their "parent" (like squash "Add missing blank line in contributing doc" into "Update documentation to explicit yarn/rollup/node requirements").

Sorry if I'm too picky, I can apply these tweaks for you if necessary.

deivid-rodriguez:
> Also the Tidelift check didn't complete. Is that something we can run manually to confirm before merging?

@javierjulio Not sure why that wasn't triggered, the Tidelift check scans for vulnerable issues in dependencies, like vulnerable, unmaintained, or with incompatible licenses. Unfortunately, I don't think there's a CLI tool to check this locally yet.

javierjulio:
@deivid-rodriguez ok cool thanks! Good to know. If it keeps happening I'll ignore for now. Perhaps on the next force push it'll rerun. 🙏🏻

sgara:
@deivid-rodriguez not that I really expect developers to check binstubs content, I do get your point. Yarn binstub updated 👍 
I usually squash-merge PRs entirely to avoid this issue but I guess that's just me being lazy ^^ I rewrote history to keep only meaningful commits, tell me if there is any more commits you'd like me to squash.

sgara:
Damn it! Fixed ;)

PR #5873, opened 2019-10-02
Remove redundant \_

udayan28:
Remove warning of regex character class.
`\w` have already included `_`.
cf. https://ruby-doc.org/core-2.6.3/Regexp.html

```
$ ruby -w lib/active_admin/order_clause.rb
lib/active_admin/order_clause.rb:6: warning: character class has duplicated range: /^([\w\_\.]+)(->'\w+')?_(desc|asc)$/
```


javierjulio:
Thank you! 👌🏻

PR #5874, opened 2019-10-02
Improved german translations

juril33t:
Added translations that were in [en.yml](https://github.com/activeadmin/activeadmin/blob/master/config/locales/en.yml) but not in de.yml


javierjulio:
Thank you! ❤️

PR #5877, opened 2019-10-02
Drop rails 5.0 and 5.1 support

deivid-rodriguez:
It seems that since the release of Rails 6.0, the Rails core team has dropped support for Rails 5.0 and Rails 5.1. See https://rubyonrails.org/security/ or https://edgeguides.rubyonrails.org/maintenance_policy.html.

So we should follow their lead.


javierjulio:
Looks great thanks! We'll need to update the branch settings to only require the 52/60 CI builds.

deivid-rodriguez:
I updated the required checks, so this should be good to go now :+1:.

johnaAr555 (approved):

PR #5884, opened 2019-10-11
Bump rubocop to 0.75.0

deivid-rodriguez:
I needed to remove

```yaml
Rails:
  Enabled: true
```

from the configuration, because whereas before (when Rails cops lived in the core repo) it seemed to mean "Allow manually enabling specific Rails cops", now (when Rails cops live as a separate extension) it apparently means "Enable every Rails cop".


javierjulio:
@deivid-rodriguez is this ready?

deivid-rodriguez:
Yes, it is! :+1:

PR #5887, opened 2019-10-14
Use $sidebar-with to calculate right-margin of #main_content

guigs:
Here are the steps to reproduce the issue and the result after this fix.

1. Create a new rails project with activeadmin and a simple Article model, the layout looks like this:

activeadmin-1

2. Edit `app/assets/stylesheets/active_admin.scss` 

Add `$sidebar-width: 350px;` before `@import "active_admin/mixins";`

The result layout is like this:

activeadmin-2

3. Apply the fix from this PR and  the `#main_content` right margin will be adjusted to take into account the custom value of `$sidebar-width` and the layout will look like this:

activeadmin-3


deivid-rodriguez:
Thank you! This makes sense to me! Can you copy a couple of screenshots of the before and after here, so people can easily the visual changes when they come from the changelog? This improves the visual alignment for non standard widths, if I understand correctly?

guigs:
Yep, that is correct. 
I'll try to add the screenshots on Friday.

guigs:
@deivid-rodriguez Screenshots added.

app/assets/stylesheets/active_admin/structure/_main_structure.scss
```diff
@@ -15,7 +15,7 @@
     width: 100%;
 
     #main_content{
-      margin-right: 300px;
+      margin-right: $sidebar-width + $section-padding * 2;
```
javierjulio: This should be:
```suggestion
      margin-right: $sidebar-width + ($section-padding * 2);
```

deivid-rodriguez: `*` should always have higher precedence than `+`, right?

deivid-rodriguez:
Thanks for documenting the PR with such detail @guigs!

deivid-rodriguez:
Can you also add a changelog entry under "Bug Fixes" following the existing changelog format?

app/assets/stylesheets/active_admin/structure/_main_structure.scss
javierjulio: Yes but this improves readability here as its easy to miss.

deivid-rodriguez: Sure!

javierjulio:
Awesome thanks @guigs!

PR #5888, opened 2019-10-16
Rework js files to leverage es6 imports

sgara:
Not a crucial PR on the road to webpack, but now that we use proper es6 imports, I figured we could clean things up a bit. Also renamed js files for consistency as some used hypens, others used underscores (I picked hypens). This also removes the ugly global ActiveAdmin variable :)

After this one, I think we can setup package distribution through npm. I can start it and give over the ownership afterward if you want.

- [x] setup npm packaging and rollup
- [ ] **rework js files to remove the use of ActiveAdmin global constant**
- [ ] **release the npm package including assets ready to be imported by webpack**
- [ ] add use_webpacker switch


app/assets/javascripts/active_admin/base.js
```diff
@@ -345,184 +427,81 @@
...
+  function _inheritsLoose(subClass, superClass) {
+    subClass.prototype = Object.create(superClass.prototype);
+    subClass.prototype.constructor = subClass;
+    subClass.__proto__ = superClass;
+  }
+  var TableCheckboxToggler = function(_CheckboxToggler) {
+    _inheritsLoose(TableCheckboxToggler, _CheckboxToggler);
     function TableCheckboxToggler() {
-      return _ActiveAdmin$Checkbox.apply(this, arguments) || this;
+      return _CheckboxToggler.apply(this, arguments) || this;
```
deivid-rodriguez: I think the rename of this parameter unrelated to the commit including it, so I would split the rename to a previous commit.

deivid-rodriguez:
Regarding the `npm` package, it sounds good to me, whatever it's easier to you works :+1:.

Thanks for doing this work!

app/assets/javascripts/active_admin/base.js
sgara: This comes from changing `ActiveAdmin. CheckboxToggler = ...` to `export default CheckboxToggler;` which is in this same commit. That's on the compiled side, so no, for this one, it's one the right commit :)

deivid-rodriguez: Right! :+1:

sgara:
@deivid-rodriguez - @javierjulio also maybe? - do you guys already have an npm account? I'll need your username or email to invite you once I finish setting things up

deivid-rodriguez:
I don't think I have an `npm` account. I guess I should register one so that you can invite me, right?

sgara:
Yep, you can create one here: https://www.npmjs.com/signup

deivid-rodriguez:
Done! I used the email present in activeadmin's gemspec!

sgara:
I have minor changes to do on scss files (like using relative imports) before publishing the package. So we can merge this one and open yet another PR with remaining changes for publication

javierjulio:
@sgara yes, I have an npm account. Can you please invite jjfutbol? Thanks!

javierjulio:
Thanks! ❤️

PR #5889, opened 2019-10-17
Fix missing Pundit policy causing extra queries

### LINKED ISSUE #5199: Pundit + joins/includes loads whole table before paginating
Enable the Pundit adapter and have a default policy with the default `Scope#resolve` method. Then in a resource:

```
controller do
  def scoped_collection
    super.includes(:association)
  end
end
```

### Expected behavior

A single query with pagination should be executed.

### Actual behavior

First a query is run that loads the whole table, and then another query with pagination.

This is obviously an issue with big tables, as the index page for the resource takes much longer than it should to load.

```begin
  require 'bundler/inline'
rescue LoadError => e
  raise e
end

gemfile(true) do
  source 'https://rubygems.org'

  if ENV['ACTIVE_ADMIN_PATH']
  else
    gem 'activeadmin', github: 'activeadmin/activeadmin', require: false, ref: 'f28a4e0fb315c5a6a3fb78f423b19562246160a0'
  end

  gem 'rails', '~> 4.2.10'

  gem 'sass-rails'
  gem 'sqlite3', platform: :mri
  gem 'pundit'
  # gem 'activerecord-jdbcsqlite3-adapter',
  #     git: 'https://github.com/jruby/activerecord-jdbc-adapter',
  #     branch: 'rails-5',
  #     platform: :jruby
end

require 'active_record'

  end

    t.string :full_name
  end

  create_table :comments, force: true do |t|
    t.references :user
  end
end

require 'action_view/railtie'
require 'active_admin'

  config.root = __dir__

  config.eager_load = false
  Rails.logger  = config.logger
end

end

  has_many :comments
end

  belongs_to :user
end

class AdminPolicy
  attr_reader :user, :record

  def initialize(user, record)
    @user = user
    @record = record
  end

  def index?
    true
  end

  def scope
    Pundit.policy_scope!(user, record.class)
  end

  class Scope
    attr_reader :user, :scope

    def initialize(user, scope)
      @user = user
      @scope = scope
    end

    def resolve
      scope
    end
  end
end

ActiveAdmin.setup do |config|
  config.pundit_default_policy = 'AdminPolicy'
end

Rails.application.initialize!

  content do
    'Test Me'
  end
end

ActiveAdmin.register User do
  controller do
    def scoped_collection
      super.includes(:comments)
    end
  end
end

  ActiveAdmin.routes(self)
end

require 'minitest/autorun'
require 'rack/test'
require 'rails/test_help'

  def test_admin_users
    User.create full_name: 'John Doe'
    query_count = 0
    callback = lambda { |_name, _start, _finish, _id, payload|
      query_count += 1 if payload[:sql] == 'SELECT "users".* FROM "users"'
    }
    ActiveSupport::Notifications.subscribed callback, "sql.active_record" do
      get admin_users_url
    end
    assert_equal 0, query_count
  end

  private

  def app
    Rails.application
  end
end
```

filipkis:
`PunditAdapter` was checking for policy scope using `Pundit.policy_scope!` which throws an error that calls `.inspect` on the object it was called with. This is fine when `scoped_collection` is just a `ActiveRecord` class, however when users modify the `scoped_collection` (e.g. to use `includes` or `joins`) then the `.inspect` would cause unnecessary query execution.

This PR fixes it by calling `policy_scope` alternative that doesn't throw error, but returns nil instead.

Fixes  #5199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Use filter label when condition has a predicate. [#5886] by [@ko-lem]
 * Fix error when routing with array containing symbol. [#5870] by [@jwesorick]
+* Fix additional loading with modified scoped_collection and missing Pundit policy. [#5199] by [@filipkis]
 
 ### Removals
 
@@ -528,6 +529,7 @@ Please check [0-6-stable] for previous changes.
 [#5877]: https://github.com/activeadmin/activeadmin/pull/5877
 [#5886]: https://github.com/activeadmin/activeadmin/pull/5886
 [#5870]: https://github.com/activeadmin/activeadmin/pull/5870
+[#5199]: https://github.com/activeadmin/activeadmin/pull/5199
 
 [@5t111111]: https://github.com/5t111111
 [@aarek]: https://github.com/aarek
@@ -554,6 +556,7 @@ Please check [0-6-stable] for previous changes.
 [@eikes]: https://github.com/eikes
 [@f1sherman]: https://github.com/f1sherman
 [@faucct]: https://github.com/faucct
+[@filipkis]: https://github.com/filipkis
 [@Fivell]: https://github.com/Fivell
 [@glebtv]: https://github.com/glebtv
 [@gonzedge]: https://github.com/gonzedge
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -20,26 +20,30 @@ module ActiveAdmin
     def scope_collection(collection, action = Auth::READ)
       # scoping is appliable only to read/index action
       # which means there is no way how to scope other actions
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
Thank you for fixing this @filipkis!

The original issue report includes a nice test to showcase the problem using our bug report template. Could you adapt that into a test that we can add to our suite so that we don't regress here?

filipkis:
I will try. But any hint on similar test cases (i.e. ones using templatesI
already looked at the Pundit unit tests, but those don't do anything with
the templates I think) will help me do this sooner as I currently don't
have that much time to figure out how exactly are the specs setup in
activeadmin.

javierjulio:
I don't believe we still need this due to the changes made later on in #7144 and #7479. If this is still applicable, let me know and I will reopen. Thank you.

PR #5893, opened 2019-10-23
Use relative import and parameterise image path in scss files

sgara:
Last tweaks to publish npm package, and ensure compatibility with both sprockets and webpacker. Mainly use relative imports and allow changing imagePath on client side. By default, everything stays as is, no required changes in current usage (defaulting for sprockets).


app/assets/stylesheets/active_admin/components/_tables.scss
```diff
@@ -33,7 +33,7 @@ table.index_table {
     }
 
     &.sortable a {
-      background: image-url('active_admin/orderable.png') no-repeat 0 4px; padding-left: 13px;
+      background: url("#{$imagePath}/orderable.png") no-repeat 0 4px; padding-left: 13px;
```
javierjulio: How does this work with precompilation in Rails with sprockets? Are you sure that the paths get updated? My main concern is that sprockets is looking for that `image-url` method so I don't think this will work as expected.

sgara: Hum that's something I completely oversaw indeed. 
I'll try to run precompilation on my test app and keep you updated. In case it doesn't work, do you have any idea how to handle that? I'm thinking of adding a `sed` or something like that when copying files to `src/`, or have different "flavours" of those 2 files

javierjulio: I'm not actually sure how to handle that. It seems sassc-rails just registers a sass function called `asset-url` (the `image-url` is just an alias) so we'd need something here for that. My first thought is to check the webpacker docs and any example Rails apps using webpacker for CSS as well. I'd start there and see what you find out. Thanks!

javierjulio: What I'd love to do though is just replace the images with inline SVG using data URLs. These are simple icons so it would work and then we'd be able to use the SCSS in both without issue as we would no longer be dependent on external assets. Another nice bonus is users wouldn't be precompiling any assets here other than the CSS and JS files.

sgara: Using data uri is a brilliant idea! Since those icons were really, really low-def converting them was pretty much impossible, so we should find completely new icons in SVG, and the ones I found were way bigger than current files (eg. https://www.iconfinder.com/icons/211614/arrow_b_down_icon). 
I used current png files converted in base64, and used them as data uris to keep it simple for now. I that okay for you?

app/assets/stylesheets/active_admin/mixins/_variables.scss
```diff
@@ -2,6 +2,11 @@
 // They can be overridden by prepending your own
 // to 'app/assets/stylesheets/active_admin.scss'.
 
+// Images
+$menu-arrow-light-icon-url: 'data:image/png;base64,R0lGODlhBwAEAKIAAL6+vry8vIiIiJWVlf///3t7ewAAAAAAACH5BAEAAAUALAAAAAAHAAQAAAMLWLol80MoF5mQKgEAOw==';
```
javierjulio:
```suggestion
$menu-arrow-light-icon-url: 'data:image/png;base64,R0lGODlhBwAEAKIAAL6+vry8vIiIiJWVlf///3t7ewAAAAAAACH5BAEAAAUALAAAAAAHAAQAAAMLWLol80MoF5mQKgEAOw==' !default;
```

app/assets/stylesheets/active_admin/mixins/_variables.scss
```diff
@@ -2,6 +2,11 @@
...
+$menu-arrow-dark-icon-url: 'data:image/png;base64,R0lGODlhBwAEAKIAAG1tbWxsbElJSVBQUP///0JCQgAAAAAAACH5BAEAAAUALAAAAAAHAAQAAAMLWLol80MoF5mQKgEAOw==';
```
javierjulio:
```suggestion
$menu-arrow-dark-icon-url: 'data:image/png;base64,R0lGODlhBwAEAKIAAG1tbWxsbElJSVBQUP///0JCQgAAAAAAACH5BAEAAAUALAAAAAAHAAQAAAMLWLol80MoF5mQKgEAOw=='  !default;
```

app/assets/stylesheets/active_admin/mixins/_variables.scss
```diff
@@ -2,6 +2,11 @@
...
+$orderable-icon-url: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAcAAABGCAYAAAAAVo4aAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAH5JREFUeNpi3LhlOwMU1AExGxDXwARYoHQLEFdD2cxAXAliMKFJgEAFEHfBJEHGMKLhMpgkTsAEdch/NNwCk2xCdiEQtML4LEgCf6EubUX3Cgh0oNvJ+P//f7wOGpUclRwYSZb41CyidNbB8giNM+9oXhmVHHm5bJjUSAABBgDKKiwMMUxPwgAAAABJRU5ErkJggg==';
```
javierjulio:
```suggestion
$orderable-icon-url: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAcAAABGCAYAAAAAVo4aAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAH5JREFUeNpi3LhlOwMU1AExGxDXwARYoHQLEFdD2cxAXAliMKFJgEAFEHfBJEHGMKLhMpgkTsAEdch/NNwCk2xCdiEQtML4LEgCf6EubUX3Cgh0oNvJ+P//f7wOGpUclRwYSZb41CyidNbB8giNM+9oXhmVHHm5bJjUSAABBgDKKiwMMUxPwgAAAABJRU5ErkJggg=='  !default;
```

deivid-rodriguez:
Nice catches @javierjulio, thanks!

app/assets/stylesheets/active_admin/components/_tables.scss
```diff
@@ -33,7 +33,7 @@ table.index_table {
     }
 
     &.sortable a {
-      background: image-url('active_admin/orderable.png') no-repeat 0 4px; padding-left: 13px;
+      background: url($orderable-icon-url) no-repeat 0 4px; padding-left: 13px;
```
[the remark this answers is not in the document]
sgara: Checked 👍

app/assets/stylesheets/active_admin/_header.scss
```diff
@@ -75,12 +75,12 @@
       }
 
       &.has_nested > a {
-        background: image-url('active_admin/nested_menu_arrow.gif') no-repeat calc(100% - 7px) 50%;
+        background: url($menu-arrow-light-icon-url) no-repeat calc(100% - 7px) 50%;
```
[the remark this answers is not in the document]
sgara: Checked! 👍

sgara:
@javierjulio nice catches! Added `default!` and checked for remaining asset usage, we're good on this. Also added a note on contributing and a linter to check for sass-rails usage in scss files so we're sure one doesn't slip through.

tasks/lint.rake
```diff
@@ -138,6 +155,13 @@ namespace :lint do
     FixmeLinter.new.run
   end
 
+  desc "Check for sass-rails directives to ensure webpacker compatibility"
+  task :sass_rails do
+    puts ""
```
deivid-rodriguez: I guess you forgot to add the logging here?

PR #5894, opened 2019-10-23
Publish @activeadmin/activeadmin package to npm

sgara:
Update package information (name, url...), version and exported files, ready to be published to npm.

Version 4.2.0-alpha-3 has been published, with changes from #5893 and #5888 included and tested with #5855, both with webpacker and sprockets 👍


sgara:
@deivid-rodriguez @javierjulio PR updated and published to npm for testing purpose. You have both been invited to npm organization

deivid-rodriguez:
Hi @sgara! Thanks so much for doing this, it seemed something quite hard for me version you started this and now it looks like we're almost there! :tada:

I'm sorry I haven't yet get to trying this out, but regarding versioning, is the idea to sync versioning so that everytime we push version x.y.z of the gem to rubygems.org we also push version x.y.z of the npm package to npmjs.org?

sgara:
Yeah! Almost!!! I thought it would be a 2 weeks projects... haha!

My initial thought was to sync them both but in the end, it might be too troublesome for nothing. I just need a starting version, can go back to 1.0 or even 0.1, as you prefer really

deivid-rodriguez:
So, my doubts are about compatibility between the `npm` package and the `gem` package.

Sppecially now that we'll support both `webpacker` and `sprockets`, if I understand this right, whenever we make changes to the JS they will need an updated version provided with the `gem` package for `sprockets` usage, and another version provided with the `npm` package for `webpacker` usage.

So I was thinking whenever we release the `gem` package, we release the `npm` package at the same time with the exact same version and expect `webpacker` users to use the exact same version of both.

So, questions:

* Does this make sense?
* How does `Rails` handle this?

sgara:
It totally makes sense and that’s how Rails itself handles it as far as I know. 
It’s just that from my experience it leads to some troubles and confusion (eg. would you release an npm update even though js has not been updated?) although it does make things simpler in general. 
It’s entirely opinion based, both solutions having pros and cons. 
I’d say we should go the rails way, and as long as rules are clear (such as yes, we do release even though js hasn’t change) everything should go well.

deivid-rodriguez:
Yeah, some releases to the `npm` package that would include no changes at all other than the version bump, but I still feel it's simpler that way.

So, we would need some new tooling (`bin/rake release_all`?, enhace `rake release`?), that automatically makes sure that `package.json` and `lib/bundler/version.rb` are in sync, and then releases both packages.

sgara:
I’m not really familiar with gem release so I’ll need your input here :)
What is the current source of truth for versioning? Is it `version.rb` or gemspec? Either way, reading version from either of them looks quite troublesome. Rails is solving this by using a `RAILS_VERSION` at the root of the repo, would you mind doing something similar? Adding a `VERSION` file.

deivid-rodriguez:
Yeah, sure! The source of truth for our purposes would be the `version.rb` file. I think adding a `VERSION` file (or maybe an `.activeadmin-version` file, consistent with the `.ruby-version` file naming used by ruby version managers) at the root of the repo is a good solution, yeah.

In case it helps, I worked on another gem in the past that used something similar. We didn't push a `npm` package to `npmjs.org`, but we did have a `package.json` file versioned in sync with a bunch of `version.rb` files. You can have a look at https://github.com/decidim/decidim/blob/a41d0fd3c7b76591f2a5bdeb465f780482c9c6df/Rakefile#L59-L63, https://github.com/decidim/decidim/blob/a41d0fd3c7b76591f2a5bdeb465f780482c9c6df/Rakefile#L24-L28, or https://github.com/decidim/decidim/blob/3c663f6858be3576486a3084928205f81eab38c5/lib/decidim/gem_manager.rb.

Also, if you feel adding this tooling is beyond the tasks you feel like taking, just let me know and I'll assign this task to myself (although I might need input from you for the `npm` release part :sweat_smile:).

sgara:
@deivid-rodriguez @javierjulio should be functional enough. I hooked into `release` task not to introduce confusion for current and future maintainers and tried to keep things as simple as possible. Rebased on master and published v2.6.0 using newly created `rake activeadmin:npm_push` task. I'm having some doubt for the namespace (`activeadmin`, `release`, `activeadmin:release`...?) but I didn't want to pollute external namespaces either. 

**nb.** I chose `VERSION` over `.activeadmin-version` because the later is generally used for external tools versioning whereas we're keeping internal version here. Feel free to comment if you disagree.

deivid-rodriguez:
Thanks @sgara, it definitely looks functional enough to me :purple_heart:. Both the namespacing you used and using `VERSION` sound good.

I'm currently a bit trapped with other projects, give me a few days and I'll get here and to your other PR.

Thanks so much for doing all this excellent work, and for your patience.

tasks/release.rake
```diff
@@ -1,6 +1,42 @@
 require "chandler/tasks"
 
+root    = File.expand_path("..", __dir__)
+version = File.read("#{root}/VERSION").strip
+
+# See https://github.com/rails/rails/blob/master/tasks/release.rb#L26
```
deivid-rodriguez: This link is now outdated, we should use a permalink if linking to Rails code. In any case, the ruby-like version to npm-like version is incorrect like they point out in the source code comment.

I think we can use a more correct version of the conversion like

```ruby
def npmify(version)
  segments = version.split(".")

  split_point = segments.index{|segment| segment.to_s =~ /[a-z]/}

  release_segments, prerelease_segments = segments[0...split_point], segments[split_point..-1]

  [release_segments.join("."), prerelease_segments.join(".")].join("-")
end
```

which gives

```
irb(main):016:0> npmify "5.0.1"
=> "5.0.1"
irb(main):017:0> npmify "5.0.1.1"
=> "5.0.1.1"
irb(main):018:0> npmify "5.0.0.rc1"
=> "5.0.0-rc1"
```

tasks/release.rake
```diff
@@ -1,6 +1,42 @@
...
+        sh "npm version #{npm_version} --no-git-tag-version"
+      else
+        raise "You must have npm installed to release ActiveAdmin."
+      end
+    end
+  end
+
+  desc 'Publish npm package'
+  task :npm_push do
+    npm_tag = /[a-z]/.match?(version) ? "pre" : "latest"
```
deivid-rodriguez: As a followup to the other comment, here we should use `npm`'s criteria for prereleases (which strictly conforms to [semantic versioning](https://semver.org/#spec-item-9). So instead check whether the version string includes a hyphen or not.

tasks/release.rake
deivid-rodriguez: Actually, scratch that. Things like "5.1.1.1" are invalid according to semver since semver doesn't define a TINY (4th) segment. So it's undefined what we should do with that, and we don't use that numbering in activeadmin anyways.

So let's go with the `Rails` version. Only thing I would tweak is to only convert the dot separating the prerelease part from the release part to a hyphen, because the result looks less weird and more similar to the "ruby version". So `==` instead of `>=`:

```
npm_version = version.gsub(/\./).with_index { |s, i| i == 2 ? "-" : s } 
```

sgara: As [commented in rails repo](https://github.com/rails/rails/blob/0d0c30e534af7f80ec8b18eb946aaa613ca30444/tasks/release.rb#L27) versions like `5.0.0.beta1.1` should be transformed to `5.0.0-beta1-1`, so I guess we should keep `>=` but I'm no semver expert and haven't seen any such version so far so it's really up to you

deivid-rodriguez: Yeah, that's the case I want to change. `5.0.0-beta1-1` looks very weird to me, and `5.0.0-beta1.1` is perfectly semver compliant and less ambiguos in my opinion since it has a single hyphen separating the release part from the prerelease part.

tasks/release.rake
deivid-rodriguez: To clarify this, the current check is perfectly fine, but it seems more clear to me, since we are pushing to `npm`, to use the `npm_version` and the `npm` criteria. But it's just another nitpick, doesn't fix anything since the current check is correct.

javierjulio:
I’ll take a look at this today. Thanks!

deivid-rodriguez:
Sweet! I'll release 2.7.0 tomorrow to make sure everything is good with the new synced release process.

deivid-rodriguez:
On second thought I'll release 2.6.1, since I don't think we're shipping any new features here. The `release` tooling is just for us, maintainers, and the `npm` package is not yet expected to be usable.

javierjulio:
After reviewing unreleased changes, a patch release sounds good. 👍🏻

PR #5896, opened 2019-10-24
Setup dependabot

deivid-rodriguez:
I think our config at `inherited_resources` is working well, so I think we can try the same approach here.


deivid-rodriguez:
This one should be ready now.

We should get a few dependabot PR's automerged after merging this PR.

deivid-rodriguez:
Dependabot setup worked but we need to remove the "require an approving review" merge check from the master branch, so that dependabot can automerge its development dependency updates. We still want approvals before merging, it's just not automated requirement. This is the same we already did in inherited_resources.

PR #5929, opened 2019-11-06
Display multiple flash messages in separate elements

### LINKED ISSUE #5279: Allow multiple flash messages from the same type
Currently ActiveAdmin only allows to one flash message per type, but sometimes we have the need to show more than one message per type. 

Previously we concatenate strings from multiple alerts but this his hard to read and far from a good UI, so we decided to monkey patch ActiveAdmin to allow multiple flash messages.

In `lib/active_admin/views/pages/base.rb#build_flash_messages `

```ruby
def build_flash_messages
  div class: 'flashes' do
-   flash_messages.each do |type, message|
-     div message, class: "flash flash_#{type}"
+   flash_messages.each do |type, messages|
+     [*messages].each do |message|
+       div message, class: "flash flash_#{type}"
+     end
    end
  end
end
```

This allows us to optionally pass an array to flash.

Would this be a feature that ActiveAdmin should support? If so I can create a PR to add it.

mirelon:
closes #5279

### DIFF
--- a/spec/unit/views/pages/base_spec.rb
+++ b/spec/unit/views/pages/base_spec.rb
@@ -11,4 +11,17 @@ RSpec.describe ActiveAdmin::Views::Pages::Base do
   it "defines default main content" do
     expect(NewPage.new.main_content).to eq 'Please implement NewPage#main_content to display content.'
   end
+
+  describe "build_flash_messages" do
+    class PageWithFlashMessages < ActiveAdmin::Views::Pages::Base
+      def flash_messages
+        { alert: 'Alert message', notice: ['First notice message', 'Second notice message'] }
+      end
+    end
+
+    it "shows all flash messages" do
+      div = PageWithFlashMessages.new.send :build_flash_messages
+      expect(div.get_elements_by_tag_name('div').map(&:content)).to contain_exactly('Alert message', 'First notice message', 'Second notice message')
+    end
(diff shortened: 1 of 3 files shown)


spec/unit/views/pages/base_spec.rb
```diff
@@ -11,4 +11,17 @@ class NewPage < ActiveAdmin::Views::Pages::Base
...
+
+  describe "build_flash_messages" do
+    class PageWithFlashMessages < ActiveAdmin::Views::Pages::Base
+      def flash_messages
+        {alert: 'Alert message', notice: ['First notice message', 'Second notice message']}
+      end
+    end
+
+    it "shows all flash messages" do
+      div = PageWithFlashMessages.new.send :build_flash_messages
```
mirelon: Before:

    <div class="flashes">
      <div class="flash flash_alert">Alert message</div>
      <div class="flash flash_notice">First notice messageSecond notice message  </div>
     </div>

After:

    <div class="flashes">
      <div class="flash flash_alert">Alert message</div>
      <div class="flash flash_notice">First notice message</div>
      <div class="flash flash_notice">Second notice message</div>
    </div>

javierjulio:
Yes just a conflict with the changelog but it’s ready. Let us know if you can update. Thanks!

PR #5931, opened 2019-11-08
Convert namespace to sym to prevent duplicate namespaces

westonganger:
If namespace `:foo` is passed in one place and `"foo"` is passed in another it will try to register the namespace twice. This causes an error for duplicate root routes.

With this PR we convert the passed namespace to sym to prevent duplicate namespaces

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+* Convert namespace to sym to prevent duplicate namespaces such as :foo and 'foo'. [#5931] by [@westonganger]
 * Use filter label when condition has a predicate. [#5886] by [@ko-lem]
 * Fix error when routing with array containing symbol. [#5870] by [@jwesorick]
 * Fix error when there is a model named `Tag` and `meta_tags` have been configured. [#5893] by [@micred], [@FabioRos] and [@deivid-rodriguez]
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Thanks! This makes sense to me. I wrote a small regression unit test for you:

```diff
diff --git a/spec/unit/application_spec.rb b/spec/unit/application_spec.rb
index 7a26814de..8433f4f55 100644
--- a/spec/unit/application_spec.rb
+++ b/spec/unit/application_spec.rb
@@ -162,6 +162,15 @@ RSpec.describe ActiveAdmin::Application do
       }.to raise_error("found")
     end
 
+    it "should admit both strings and symbols" do
+      expect {
+        application.namespace "admin" do |ns|
+          expect(ns).to eq application.namespaces[:admin]
+          raise "found"
+        end
+      }.to raise_error("found")
+    end
+
     it "should not pollute the global app" do
       expect(application.namespaces).to be_empty
       application.namespace(:brand_new_ns)
```

Can you add it?

Also, can you add a change log entry [following the existing format](https://github.com/activeadmin/activeadmin/blob/master/CONTRIBUTING.md#add-a-changelog-entry) explaining the fix?

westonganger:
I have made the requested changes.

Thanks for writing the spec for me, appreciate that.

Issue #221, opened 2011-06-23
Support multiple belongs_to


knoopx:
_Edit: Original title: Specifing multiple belongs_to does not seem to work_

The following:

```
ActiveAdmin.register Movie do
  belongs_to :provider, :optional => true
  belongs_to :actor, :optional => true
end
```

only generates routes for the `actor`, `provider` is completely ignored

emilebosch:
Well 'ive managed to work around it by just making my own polymorphic controller. And adding the routes myself. Makes total sense. Its a monkey patch but hey. at least it works.

app/controllers/admin/comments_controller.b
```
module Admin
  class CommentsController < ApplicationController
    def create
      commentable.comments.create!(comments_params[:comment])
      redirect_to :back
    end

    def comments_params
      params.permit!
    end

    def commentable
      return House.find_by_slug(params[:house_id]) if params[:house_id]
      return Person.find(params[:person_id]) if params[:person_id]
      raise "Can't find commentable"
    end
  end
end
``` 
routes

```
  ActiveAdmin.routes(self)
  namespace :admin do
    resources :people do
      resources :comments
    end
    resources :houses do
      resources :comments
    end
  end
```

jnettome:
:+1: and thank you @Zequez, good solution :)

zerocool4u2:
Well, working around this i came to this solutions
ActiveAdmin.register Pickup do
	controller do
		belongs_to :client, :warehouse, optional: true
	end
end

and then add the routes, 

namespace :admin do
	resources :clients do
		resources :pickups
	end
	resources :warehouses do
		resources :pickups
	end
end

this gives me support for both nested routes and the resource itself

admin/pickups/....
admin/clients/pickups/...
admin/warehouses/pickups

if you don't want the route for "pickups" i don't know how could you do this, with menu false option at least you can remove it from the navigation bar, but for my case i need it, also the optional: true, allows you to have the belongs without one of the parents, you should work around the forms as well, i'll gonna work with the route to show the corresponding dropdowns depending where i'm to create new pickups in each case

gabrieltong:
+1

seanlinsley:
Github now allows you to add a reaction to comments. If someone wants to add a +1 to this ticket, please do it via a reaction to the first comment on the ticket instead of adding a new comment.

You can also subscribe to the ticket so you receive notifications if anything new happens.

[...]

PR #5938, opened 2019-11-18
Improve multiple belongs_to hack

leio10:
This PR slightly modifies polymorphic routes generation. When mapping the route parameters, instead of checking each class of the given parameters against the `parent` class, it does it against the configured `belongs_to` resource class. This has sense because if this check is successful it won't use the `parent` but the configured `belongs_to` resource.

As explained in #221, when you define several `belongs_to` for a resource:
```ruby
ActiveAdmin.register Movie do
  belongs_to :provider, optional: true
  belongs_to :actor, optional: true
end
```
And then add the missing routes:
```
namespace :admin do
    resources :movies
    resources :providers do
        resources :movies
    end
    resources :actors do
        resources :movies
    end
end
```
Everything seems to work well, but when browsing the movies for the provider 123, some links are generated as `/actors/123/movies` instead of `/providers/123/movies`.

This change fixes this issue and **keeps the same behavior when having only one `belongs_to` declaration**.

Also, I've added some tests to avoid regression, even when multiple `belongs_to` is still not supported, since it can be useful for many people, and could be a little step further toward supporting this :smiley: .

### DIFF
--- a/spec/unit/resource_controller/polymorphic_routes_spec.rb
+++ b/spec/unit/resource_controller/polymorphic_routes_spec.rb
@@ -3,26 +3,103 @@ require 'rails_helper'
 RSpec.describe ActiveAdmin::ResourceController::PolymorphicRoutes, type: :controller do
   let(:klass) { Admin::PostsController }
 
-  shared_context 'with post config' do
-    before do
-      load_resources { post_config }
+  %w(polymorphic_url polymorphic_path).each do |method|
+    describe method do
+      let(:add_extra_routes) {}
+      let(:params) { {} }
 
-      @controller = klass.new
+      before do
+        load_resources { post_config }
 
-      get :index
-    end
-  end
+        add_extra_routes
 
-  context 'polymorphic routes' do
-    include_context 'with post config' do
-      let(:post_config) { ActiveAdmin.register Post }
-      let(:post) { Post.create! title: "Hello World" }
-    end
+        @controller = klass.new
 
-    %w(polymorphic_url polymorphic_path).each do |method|
-      describe method do
-        it 'arrays wtih action names' do
+        get :index, params: params
+      end
+
+      context 'without belongs_to' do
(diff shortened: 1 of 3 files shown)


spec/unit/resource_controller/polymorphic_routes_spec.rb
```diff
@@ -3,28 +3,105 @@
...
+            end
+
+            it 'works with a user as parent' do
+              expect(controller.send(method, [:admin, user, post])).to include("/admin/users/#{user.id}/posts/#{post.id}")
+            end
+          end
+        end
+      end
+
+      context 'with multiple belongs_to (not fully suported yet)' do
```
deivid-rodriguez:
```suggestion
      context 'with multiple belongs_to (not fully supported yet)' do
```

leio10:
Hi @deivid-rodriguez, thanks for the review!

I did what you asked for, but now seems that JS linting is failing for some strange reason :face_with_head_bandage:

leio10:
Ok, @deivid-rodriguez, it was my mistake. I got confused with the Gherkin error, which is actually ignored. The real issue was that my changelog entry wasn't pointing to the right URL. I'm impressed that you have a test for that. :heart_eyes: :raised_hands:

deivid-rodriguez:
Thanks @leio10! Still `gherkin-lint` failing is unexpected, trying to fix it in #5941.

javierjulio:
@deivid-rodriguez I see you pushed the latest master changes to get this to pass. Thanks again! Just wanted to confirm this was ready in case you may have needed to do something else?

deivid-rodriguez:
Yeah, I fixed changelog conflicts through github's UI in order to not bother @leio10 anymore. It's ready (I would "Squash and merge" to get rid of the merge commit).

javierjulio:
Oh don't you worry, I'll be getting rid of that merge commit. 😁 Thanks again!

PR #5941, opened 2019-11-20
Run `gherkin-lint` through `yarn`

deivid-rodriguez:
We switched from `npm` to `yarn` in #5868, but we didn't migrate the `rake lint` task to use yarn as well. Also took the chance to upgrade `gherkin-lint`.

@leio10 discovered that `gherkin-lint` is currently failing in CI in #5938. Surprisingly, it's not making the CI fail, but it seems to me that it's due to a bug in `npx` not transmitting the status code of the executable is running. Since we're migrating away from `npx`, it doesn't really matter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "rollup-plugin-uglify": "^6.0.3"
   },
   "scripts": {
-    "lint": "gherkin-lint features",
+    "gherkin-lint": "gherkin-lint",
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Should be ready now!

javierjulio:
Awesome thanks David! 🙌🏻

PR #5950, opened 2019-11-28
Fix render_blank_slate for specific resources

andiskiy:
Override method render_blank_slate for specific resources as well as `new_model`, `edit_model` & `delete_model`

### DIFF
--- a/lib/active_admin/views/pages/index.rb
+++ b/lib/active_admin/views/pages/index.rb
@@ -110,9 +110,10 @@ module ActiveAdmin
         end
 
         def render_blank_slate
-          blank_slate_content = I18n.t("active_admin.blank_slate.content", resource_name: active_admin_config.plural_resource_label)
+          localizer = ActiveAdmin::Localizers.resource(active_admin_config)
+          blank_slate_content = localizer.t('blank_slate.content', resource_name: active_admin_config.plural_resource_label)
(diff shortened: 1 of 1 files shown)


lib/active_admin/views/pages/index.rb
```diff
@@ -110,9 +110,10 @@ def find_index_renderer_class(klass)
         end
 
         def render_blank_slate
-          blank_slate_content = I18n.t("active_admin.blank_slate.content", resource_name: active_admin_config.plural_resource_label)
+          localizer = ActiveAdmin::Localizers.resource(active_admin_config)
+          blank_slate_content = localizer.t('blank_slate.content', resource_name: active_admin_config.plural_resource_label)
```
deivid-rodriguez: Our i18n linter is saying that we're missing translations for this string. But the underlying problem seems that the string is incorrect? Isn't is missing the `activeadmin.` namespace?

andiskiy: Did you mean that there should be like that: `localizer.t('active_admin.blank_slate.content', resource_name: active_admin_config.plural_resource_label)` ?

deivid-rodriguez:
I'm so sorry @andiskiy I never answered your last question, too many things to deal with. Let me have a quick look now.

javierjulio:
This could use an update as there is a merge conflict.

PR #5956, opened 2019-12-02
Export ModalDialog component to re-enable client side usage

### LINKED ISSUE #5953: Customer JS calling ActiveAdmin.modal_dialog no longer working after upgrade
Hi,

I can't see any docs on how to use `ActiveAdmin.modal_dialog` in custom JS after the upgrading from 2.3.1 to 2.5.0. Looking through the commits I can see the JS has undergone significant changes which would explain the breakage but can't see the new way to use the `modal_dialog` function.

We have a `app/assets/javascripts/active_admin.js` file which is calling `ActiveAdmin.modal_dialog`, this is now raising an `ActiveAdmin is not defined` error.

What is the new way to do this? Our app uses webpacker so I'm happy to switch this JS to webpacker if thats the way to go now, the only sprockets JS we have in our app is some custom active admin code.

Thanks

sgara:
This fixes #5953 and possible other components usage in client code.

NB. case has changed from `modal_dialog` to `ModalDialog`. We could export it with both names but I think that would be unnecessary legacy code.

### DIFF
--- a/app/assets/javascripts/active_admin/base.js
+++ b/app/assets/javascripts/active_admin/base.js
@@ -10,9 +10,10 @@
 //= require jquery_ujs
 //= require_self
 
-(function(factory) {
-  typeof define === "function" && define.amd ? define([ "jquery", "jquery-ui/ui/widgets/datepicker", "jquery-ui/ui/widgets/dialog", "jquery-ui/ui/widgets/sortable", "jquery-ui/ui/widgets/tabs", "jquery-ui/ui/widget", "jquery-ujs" ], factory) : factory();
-})(function() {
+(function(global, factory) {
+  typeof exports === "object" && typeof module !== "undefined" ? factory(exports, require("jquery"), require("jquery-ui/ui/widgets/datepicker"), require("jquery-ui/ui/widgets/dialog"), require("jquery-ui/ui/widgets/sortable"), require("jquery-ui/ui/widgets/tabs"), require("jquery-ui/ui/widget"), require("jquery-ujs")) : typeof define === "function" && define.amd ? define([ "exports", "jquery", "jquery-ui/ui/widgets/datepicker", "jquery-ui/ui/widgets/dialog", "jquery-ui/ui/widgets/sortable", "jquery-ui/ui/widgets/tabs", "jquery-ui/ui/widget", "jquery-ujs" ], factory) : (global = global || self, 
+  factory(global.ActiveAdmin = {}));
+})(this, function(exports) {
   "use strict";
   $.fn.serializeObject = function() {
     return this.serializeArray().reduce(function(obj, item) {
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Rebasing on top of latest master should fix the failing job.

sgara:
Looks like rollup is adding a whitespace, can't find the reason why and can't find a way to remove it (except manually, but that would become a recurring issue...). I ignored the generated file from linters as, well, it's auto-generated anyway :)

deivid-rodriguez:
Yup, agreed :+1:.

sgara:
@deivid-rodriguez rebased, I think this one is good to go.
It does fix #5953 and future webpacker usage :
- with sprockets, it adds a global `ActiveAdmin` containing all components, eg.
```js
var dialog = ActiveAdmin.ModalDialog
```
- with webpacker, clients will be able to 
```js
import { ModalDialog } from '@activeadmin/activeadmin';
```

kernow:
@deivid-rodriguez @sgara this PR resolves the issue I was having in #5953, thank you for fixing this so quickly. Any idea when this fix will be released as a gem?

deivid-rodriguez:
Thanks! As soon as it's merged I'll release it :+1:.

Only thing is: I didn't expect either these components to be used in client's code. I think this is the _only_ component for which [we document usage in client's code](https://github.com/activeadmin/activeadmin/blob/master/docs/9-batch-actions.md#batch-action-forms), so I tend to think nobody is using other components.

Should we only export only this one component?

If somebody is actually using other components, I'd like to here about that, export those components and document for the rest of the world how they're using them?

What do you think @sgara? Do you think this makes sense or maybe it's better to not bother and export everything :man_shrugging:

sgara:
Glad to hear it @kernow. 
Totally missed this piece of documentation @deivid-rodrigez!
Makes sense to only export this one then. Re exporting another component isn’t that hard in case someone else needs it.

deivid-rodriguez:
@sgara Sorry to bother you one last time. Could you add a changelog entry?

app/assets/javascripts/active_admin/base.js
```diff
@@ -506,4 +507,8 @@
     return $("#active_admin_content .tabs").tabs();
   };
   $(document).ready(onDOMReady$2).on("page:load turbolinks:load", onDOMReady$2);
+  exports.ModalDialog = ModalDialog;
```
javierjulio: Wouldn't we need an additional export for the camelcase format since others would still be using it?

deivid-rodriguez: Right, I saw the `N.B.` in the PR description but I think @javierjulio is right. We _could_ export it and warn and suggest the better alternative when it's used so users start changing their code without it being broken.

javierjulio: I didn't think about that. That's a great idea. Can we export `modal_dialog` but have it display a console warning (deprecation message) if used?

sgara: That was my first idea, and that's precisely what I was referring to by "legacy code". Seeing the lack of reactions on the issue and the fact that people having the issue would land on the issue and this PR, I thought it would be unnecessary 🤷‍♂ 
Added back the legacy export with console warning

deivid-rodriguez: Yeah, I see, sorry for the back and forth :pray:.

I can see how people that already got bit by this issue won't really care about it since the upgrade already broke them. But I think this can help people who haven't yet upgraded and will jump directly to the next version that won't break their applications and show them a warning instead.

PR #5957, opened 2019-12-03
Use default ActionView options instead of default Formtastic options for DateRangeInput

### LINKED ISSUE #5955: Date Range fields broken after 2.5.0 release
### Expected behavior

Date Range fields broken after 2.5.0 release.
I believe that is related to https://github.com/activeadmin/activeadmin/pull/5867

If you configure:
<pre>
filter :month, as: :date_range
</pre>

Before 2.5.0 it creates the following html:
`<div class="date_range input optional stringish filter_form_field filter_date_range" id="q_month_input"><label class="label">Mês</label>
<input size="12" class="datepicker hasDatepicker" maxlength="10" placeholder="De" value="" type="text" name="q[month_gteq]"  id="q_month_gteq">
<input size="12" class="datepicker hasDatepicker" maxlength="10" placeholder="Até" value="" type="text" name="q[month_lteq]"  id="q_month_lteq"></div>`

id="q_month_gteq"
id="q_month_lteq"

After 2.5.0:
`<div class="date_range input optional stringish filter_form_field filter_date_range" id="q_month_input"><label for="q_month" class="label">Mês</label>
<input class="datepicker hasDatepicker" value=""  id="q_month"  type="text" name="q[month_gteq]">
<input class="datepicker hasDatepicker" value="" id="q_month"  type="text" name="q[month_lteq]"></div>`

id="q_month"
id="q_month"

### Actual behavior

The date picker does not work anymore because the IDs now are equal in both range fields.
And also it lost other html options like placeholder, maxlength.

All my date/datetime fields are broken like this example.

mirelon:
fixes #5955

TODO:

- [x] Changelog
- [x] Regression test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * Display multiple flash messages in separate elements. [#5929] by [@mirelon]
 
+### Bug Fixes
+
+* Use default ActionView options instead of default Formtastic options for DateRangeInput [#5957] by [@mirelon]
+
 ## 2.5.0 [☰](https://github.com/activeadmin/activeadmin/compare/v2.4.0..v2.5.0)
 
 ### Enhancements
@@ -528,6 +532,7 @@ Please check [0-6-stable] for previous changes.
 [#5931]: https://github.com/activeadmin/activeadmin/pull/5931
 [#5938]: https://github.com/activeadmin/activeadmin/pull/5938
 [#5929]: https://github.com/activeadmin/activeadmin/pull/5929
+[#5957]: https://github.com/activeadmin/activeadmin/pull/5957
 
 [@5t111111]: https://github.com/5t111111
 [@aarek]: https://github.com/aarek
--- a/lib/active_admin/inputs/filters/date_range_input.rb
+++ b/lib/active_admin/inputs/filters/date_range_input.rb
@@ -33,7 +33,7 @@ module ActiveAdmin
             class: "datepicker",
             maxlength: 10,
             placeholder: placeholder,
-            value: current_value ? current_value.strftime("%Y-%m-%d") : "" }.merge(input_html_options)
+            value: current_value ? current_value.strftime("%Y-%m-%d") : "" }.merge(options[:input_html] || {})
         end
 
         def gt_input_placeholder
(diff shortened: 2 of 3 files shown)


mirelon:
@deivid-rodriguez what do you think about the regression test? Shouldn't all the attributes be tested? (placeholder, size, maxlength etc.)

mirelon:
I understand this is just a hotfix. Maybe there should be a totally different design of the class hierarchy - CompositeInput, which has multiple inputs (with defaults from Formtastic).
Or make Formtastic introduce DateRangeInput

deivid-rodriguez:
@mirelon Your fix, regression spec, and ideas for a better future class hierarchy look good to me. Thanks for fixing this! :purple_heart:

@kivanio Can you double check that this fix restores the correct behavior in your application?

kivanio:
@deivid-rodriguez It is not exacly the same as before but I can confirm it works.

The only diference is:
before:
`<label class="label">Mês</label>`
now:
`<label for="q_month" class="label">Mês</label> `

all:
`<div class="date_range input optional stringish filter_form_field filter_date_range" id="q_month_input"><label for="q_month" class="label">Mês</label>
<input size="12" class="datepicker hasDatepicker" maxlength="10" placeholder="De" value="" type="text" name="q[month_gteq]" id="q_month_gteq">
<input size="12" class="datepicker hasDatepicker" maxlength="10" placeholder="Até" value="" type="text" name="q[month_lteq]" id="q_month_lteq"></div>`

deivid-rodriguez:
Thanks for trying this out @kivanio.

It sounds like we should explicitly exclude `for` here like it's been done until now, since it doesn't really make sense for a "composite input" like here, right @mirelon?

mirelon:
Before 2.5.0 there was a method `input_html_options` with explicit keys not containing id. That's why there was no "for" attribute.
In 2.5.0 there is a method `input_html_options_for`.
Labelling works like this:
```
        def label_html_options
          {
            :for => input_html_options[:id],
            :class => ['label'],
          }
        end```

Because `input_html_options` is not implemented in DateRangeInput, the Formtastic default method takes place with id=q_month.

I think about implementing both methods: `input_html_options` (with no arguments) and `input_html_options_for` (with two argument)

deivid-rodriguez:
Thanks @mirelon, it looks good to me. Only thing maybe I would add a extra test to make sure that the `:for` input is not generated in this case.

budiman789:
Budiman
panaragan
Tulang bawang tengah 34593
Tulang bawang barat
Lampung
indonesia
+6285313074789

Pada tanggal Sab, 7 Des 2019 21:46, David Rodríguez <
notifications@github.com> menulis:

mirelon:
I added the test and rebased the branch

PR #6047, opened 2020-01-28
i18n: Rename 'destroy' => 'delete'; Fix grammar

vfonic:
I realized that in most places "destroy" has been renamed to "delete", which sounds more "human", "destroy" sounds "robotic". :)

I found one message, that my client noticed yesterday, and decided to translate it as well.

I also fixed one "typo" where singular was used instead of plural.


deivid-rodriguez:
This looks good, but you need to update some tests too. Thanks!

vfonic:
😱 Good catch! Sorry I missed that!

I've added another commit, but I guess the best would be to have these two squashed into a single commit.

PR #6066, opened 2020-01-31
bumped bumped all the versions for the gems

jesusalc:
Updated versions of gems to more recent ones. Not sure if it will work. I could use some guidance.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ group :development , :test do
   gem 'cancancan'
   gem 'pundit'
   gem 'jruby-openssl', '~> 0.10.1', platform: :jruby
+  gem 'jquery-rails'
 
   gem 'draper', '~> 3.1'
   gem "devise", "~> 4.7"
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -17,16 +17,16 @@ Gem::Specification.new do |s|
 
   s.extra_rdoc_files = %w[CHANGELOG.md CODE_OF_CONDUCT.md CONTRIBUTING.md README.md]
 
-  s.required_ruby_version = '>= 2.4'
+  s.required_ruby_version = '>= 2.5'
 
   s.add_dependency 'arbre', '~> 1.2', '>= 1.2.1'
-  s.add_dependency 'formtastic', '~> 3.1'
-  s.add_dependency 'formtastic_i18n', '~> 0.4'
-  s.add_dependency 'inherited_resources', '~> 1.7'
-  s.add_dependency 'jquery-rails', '~> 4.2'
-  s.add_dependency 'kaminari', '~> 1.0', '>= 1.0.1'
+  s.add_dependency 'formtastic', '>= 3.1'
+  s.add_dependency 'formtastic_i18n', '>= 0.4'
+  s.add_dependency 'inherited_resources', '>= 1.7'
+  s.add_dependency 'jquery-rails', '>= 1.0'
+  s.add_dependency 'kaminari', '~> 1.0', '>= 1.1.1'
   s.add_dependency 'railties', '>= 5.2', '< 6.1'
-  s.add_dependency 'ransack', '~> 2.1', '>= 2.1.1'
-  s.add_dependency 'sassc-rails', '~> 2.1'
+  s.add_dependency 'ransack', '>= 2.1', '<= 2.3.2'
+  s.add_dependency 'sassc-rails', '>= 2.1', '<= 2.1.2'
   s.add_dependency 'sprockets', '>= 3.0', '< 4.1'
 end
--- a/lib/active_admin/version.rb
+++ b/lib/active_admin/version.rb
@@ -1,3 +1,3 @@
 module ActiveAdmin
-  VERSION = '2.6.0'
+  VERSION = '2.6.1'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,14 +3,14 @@ PATH
   specs:
     activeadmin (2.6.0)
       arbre (~> 1.2, >= 1.2.1)
-      formtastic (~> 3.1)
-      formtastic_i18n (~> 0.4)
-      inherited_resources (~> 1.7)
-      jquery-rails (~> 4.2)
-      kaminari (~> 1.0, >= 1.0.1)
+      formtastic (>= 3.1)
+      formtastic_i18n (>= 0.4)
+      inherited_resources (>= 1.7)
+      jquery-rails (>= 1.0)
+      kaminari (~> 1.0, >= 1.1.1)
       railties (>= 5.2, < 6.1)
-      ransack (~> 2.1, >= 2.1.1)
-      sassc-rails (~> 2.1)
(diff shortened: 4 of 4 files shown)


deivid-rodriguez:
Hi!

This PR doesn't make much sense to me, so I'll close it. If you have a particular reason to make any of the changes in this PR, please open a separate PR explaining why you think we need the changes.

Thank you!

intuivo:
Dear @deivid-rodriguez, you really don't know how to read, or your EGO is way in the sky, the PR said: " I could use some guidance." Your response is so negative, and not inviting, plus not helpful at all. You failed to be a good lead.

deivid-rodriguez:
I'm sorry if I gave that impression, you're right, I should've read better and provided better feedback before closing. I'll review the changes more thoroughly to explain why I think the changes are not needed.

Gemfile
```diff
@@ -9,6 +9,7 @@ group :development , :test do
   gem 'cancancan'
   gem 'pundit'
   gem 'jruby-openssl', '~> 0.10.1', platform: :jruby
+  gem 'jquery-rails'
```
deivid-rodriguez: Currently `jquery-rails` is a runtime dependency of activeadmin (it's listed as `add_runtime_dependency` in the gemspec). At the end of this file, there's a `gemspec path: "."` call, which automatically picks up the dependencies from the gemspec. So adding this is not needed, and it will result in the gem being added twice.

activeadmin.gemspec
```diff
@@ -17,16 +17,16 @@ Gem::Specification.new do |s|
 
   s.extra_rdoc_files = %w[CHANGELOG.md CODE_OF_CONDUCT.md CONTRIBUTING.md README.md]
 
-  s.required_ruby_version = '>= 2.4'
+  s.required_ruby_version = '>= 2.5'
```
deivid-rodriguez: This would mean we drop support for users running ruby 2.4. We don't want to do that yet because ruby 2.4 is still a supported version.

activeadmin.gemspec
```diff
@@ -17,16 +17,16 @@ Gem::Specification.new do |s|
...
   s.add_dependency 'arbre', '~> 1.2', '>= 1.2.1'
-  s.add_dependency 'formtastic', '~> 3.1'
-  s.add_dependency 'formtastic_i18n', '~> 0.4'
-  s.add_dependency 'inherited_resources', '~> 1.7'
-  s.add_dependency 'jquery-rails', '~> 4.2'
-  s.add_dependency 'kaminari', '~> 1.0', '>= 1.0.1'
+  s.add_dependency 'formtastic', '>= 3.1'
+  s.add_dependency 'formtastic_i18n', '>= 0.4'
+  s.add_dependency 'inherited_resources', '>= 1.7'
+  s.add_dependency 'jquery-rails', '>= 1.0'
```
deivid-rodriguez: Generally we prefer to use the `~>` operator. Specifying `~> x.y` means, higher than or equal to `x.y`, but also lower than `x+1.0`. This is good, because normally major versions include breaking changes, so we don't want to unrestrict the dependency in that way, because a major release of one of our dependencies could break `activeadmin`. This is why we normally use `~> x.y` rather than just `>= x.y`.

activeadmin.gemspec
```diff
@@ -17,16 +17,16 @@ Gem::Specification.new do |s|
...
+  s.add_dependency 'kaminari', '~> 1.0', '>= 1.1.1'
```
deivid-rodriguez: In this case, you're bumping the minimum version of `kaminari`, efectively dropping support for versions between 1.0.1 and 1.1.1. In general, to be as flexible as possible with the range of dependencies we support, we don't do that unless we have a reason.

activeadmin.gemspec
```diff
@@ -17,16 +17,16 @@ Gem::Specification.new do |s|
...
   s.add_dependency 'railties', '>= 5.2', '< 6.1'
-  s.add_dependency 'ransack', '~> 2.1', '>= 2.1.1'
-  s.add_dependency 'sassc-rails', '~> 2.1'
+  s.add_dependency 'ransack', '>= 2.1', '<= 2.3.2'
```
deivid-rodriguez: In this case, we don't want to do this either because if `ransack` release a bug fix release named `2.3.3`, `activeadmin` wouldn't be able to use it.

activeadmin.gemspec
```diff
@@ -17,16 +17,16 @@ Gem::Specification.new do |s|
...
+  s.add_dependency 'sassc-rails', '>= 2.1', '<= 2.1.2'
```
deivid-rodriguez: Same thing as in the case of `ransack`, we don't want to restrict upper bounds like that.

lib/active_admin/version.rb
```diff
@@ -1,3 +1,3 @@
 module ActiveAdmin
-  VERSION = '2.6.0'
+  VERSION = '2.6.1'
```
deivid-rodriguez: We do this when we want to release a new patch level version of activeadmin. Don't worry about bumping this when you make PRs to gems, maintainers take care of this :+1:.

Gemfile
jesusalc: I have implemented this gems all together. Adding gem here was what worked, The other way around did not wok when I did that.

activeadmin.gemspec
jesusalc: Yes, for Rails 6 and 7 leaving 2.4 was breaking to install.

activeadmin.gemspec
jesusalc: I also tried that, and bundler would install the older versions, breaking the rest of the installation. Causing version trouble down the line.

activeadmin.gemspec
jesusalc: Same reason as before, leaving a very wide open range of gem acceptance was not doing the trick the resolver would fall in the lowest and break the rest of the bundle.

activeadmin.gemspec
jesusalc: I have think about this one, perhaps: 
` s.add_dependency 'ransack', '>= 2.1'`

activeadmin.gemspec
jesusalc: I have to think about this one, perhaps: 
`  s.add_dependency 'sassc-rails', '>= 2.1'`

Gemfile
deivid-rodriguez: This file is not used at runtime, so I don't see how it could work. Could you show a sample application to prove your point?

activeadmin.gemspec
deivid-rodriguez: We have already dropped support separately so this should no longer be a problem.

activeadmin.gemspec
deivid-rodriguez: I think we made this change, or similar, separately, so it should no longer be a problem.

activeadmin.gemspec
deivid-rodriguez: What's the problem with the current specification?

PR #6048, opened 2020-01-29
Improve release tasks

deivid-rodriguez:
I was about to release 2.6.1, but ended up making a few changes to further automate our release process.

This PR defines:

```
rake release:prepare_major       # Prepare a major release
rake release:prepare_minor       # Prepare a minor release
rake release:prepare_patch       # Prepare a patch release
rake release:prepare_premajor    # Prepare a prerelease for the next major release
rake release:prepare_preminor    # Prepare a prerelease for the next minor release
rake release:prepare_prepatch    # Prepare a prerelease for the next patch release
rake release:prepare_prerelease  # Prepare a prerelease
```

To make all needed code changes to get ready to release a release of a specific level.

PR #6071, opened 2020-02-01
Further improve release tasks

deivid-rodriguez:
This PR addesses @javierjulio's comment in https://github.com/activeadmin/activeadmin/pull/6048#pullrequestreview-350450901, by reverting all partial changes if any exception is raised during a release task.


deivid-rodriguez:
Yeah, I had the same issue while developing the tasks so I thought undoing the changes automatically on error would be useful, because I had to do that several times manually.

PR #6078, opened 2020-02-03
Move `npm_push` task inside `ReleaseManager`

deivid-rodriguez:
I managed to push both versions now, but needed one final fix. Decided to fix it, release, and then propose the fix to get out of this loop :sweat_smile:.


javierjulio:
No problem at all. I know it's hard to test these tasks without actually doing a full release. Thanks for catching and following up! 👍🏻

Issue #6079, opened 2020-02-03
Datalist filters


JSRossiter:
### Did you find a bug?

There doesn't seem to be any way to use a [datalist input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) as a filter and they would be perfectly suited for this. The error is

```
undefined method `map' for #<Proc:0x00007fe58fcd3bc0>
```

which comes from `raw_collection.map { |o| [send_or_call(label_method, o), send_or_call(value_method, o)] }` in https://github.com/justinfrench/formtastic/blob/f94423144eee14ccc03a64267fcafd583d3cdf85/lib/formtastic/inputs/base/collections.rb#L58

But I think the fix is creating a datalist_input.rb similar to select_input.rb in lib/active_admin/inputs/filters/

I was able to create a datalist input by passing the collection in without using a proc, however this only works in a development environment as that code is executed during application load.

### Expected behavior

Be able to use the datalist input as a filter

### How to reproduce

Add this filter to an `ActiveAdmin.register` block for a model that belongs to a User, or try to replace any similar select filter.
```
filter :user_id_equals, label: 'User ID', as: :datalist, collection: -> { User.all }
```

I'd be willing to attempt a PR for this if it's something that you're open to working on. Thanks.

javierjulio:
Thank you for opening a PR for this. Some updates for it are needed.

PR #6081, opened 2020-02-04
Add DatalistInput filter #6079

JSRossiter:
Fixes https://github.com/activeadmin/activeadmin/issues/6079

Changes here are pretty simple, I just copied `SelectInput` and removed code relating to multiple select. I tested with the bug template by adding the below, and also with my existing rails 5.2 app.

```
ActiveAdmin.register User do
  filter :id, as: :datalist, collection: -> { [1, 2, 3] }
end
```


zorab47:
What automated tests could be added to prevent future regressions?

JSRossiter:
Added tests using the select filter examples as a template

javierjulio:
I found it curious that this isn't implemented by other libraries (Tailwind, Flowbite, etc.) so I wonder if best to leave out. I wouldn't be opposed to bringing it in but would like to see an example of how it's used. It seems it would be limited to an input of type text as while other types are supported, browser support varies. This is also missing CSS styles. Note that the default branch has switched to Tailwind CSS.

PR #6086, opened 2020-02-07
Add flashes wrapper to logged_out view

irmela:
When logged in we have flash notifications wrapped inside a `<div class="flashes">` wrapper. When logged out there is no wrapper so the styling applied to `.flashes` does not get applied. This PR add the wrapper to `active_admin_logged_out.html.erb`.

### DIFF
--- a/app/views/layouts/active_admin_logged_out.html.erb
+++ b/app/views/layouts/active_admin_logged_out.html.erb
@@ -24,9 +24,11 @@
 <div id="wrapper">
 
   <div id="content_wrapper">
-    <% flash_messages.each do |type, message| %>
-      <%= content_tag :div, message, class: "flash flash_#{type}" %>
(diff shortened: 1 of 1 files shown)


javierjulio:
@irmela did this change do something different style-wise on that page after you made the update? I ask because after searching in master I don't see any CSS selectors with `flashes` other than one in the print stylesheet that just hides the flash messages if the page is printed.

irmela:
@javierjulio Did change did not add/change any styling. It only aligns the markup for flash messages for logged_in the same as for logged_out. Besides having consistent markup, the other motivation is that I have custom styling in my application and therefore needed the markup to be the same.

javierjulio:
Great thanks for confirming! Yes I think its worthwhile to do so they remain consistent for now at least.

PR #6115, opened 2020-02-28
Support multiple belongs_to optional parents

### LINKED ISSUE #221: Support multiple belongs_to

Looooong:
This PR resolves #221 by allowing multiple symbols to be passed to `belongs_to`, which `optional` must be `true`. Example:

```ruby
ActiveAdmin.register Project
ActiveAdmin.register User
ActiveAdmin.register Ticket do
  belongs_to :project, :user, optional: true
end
```

These changes include: resource DSL, page DSL, breadcrumb, router, routes and polymorphic routes. Tests are added or updated for the corresponding modifications.

### DIFF
--- a/lib/active_admin/page.rb
+++ b/lib/active_admin/page.rb
@@ -84,10 +84,11 @@ module ActiveAdmin
       @page_actions = []
     end
 
-    def belongs_to(target, options = {})
-      @belongs_to = Resource::BelongsTo.new(self, target, options)
-      self.navigation_menu_name = target unless @belongs_to.optional?
-      controller.send :belongs_to, target, options.dup
+    def belongs_to(targets, options = {})
+      @belongs_to = Resource::BelongsTo.new(self, targets, options.dup)
+      # Required `belongs_to` "should" have only one target
+      self.navigation_menu_name = targets[0] if @belongs_to.required?
+      controller.send :belongs_to, *targets, options.dup
     end
 
     def belongs_to_config
--- a/lib/active_admin/resource.rb
+++ b/lib/active_admin/resource.rb
@@ -135,11 +135,19 @@ module ActiveAdmin
       controller.instance_methods.map(&:to_sym) & ResourceController::ACTIVE_ADMIN_ACTIONS
     end
 
-    def belongs_to(target, options = {})
-      @belongs_to = Resource::BelongsTo.new(self, target, options)
+    def belongs_to(targets, options = {})
+      @belongs_to = Resource::BelongsTo.new(self, targets, options.dup)
       self.menu_item_options = false if @belongs_to.required?
-      options[:class_name] ||= @belongs_to.resource.resource_class_name if @belongs_to.resource
-      controller.send :belongs_to, target, options.dup
+
+      # When the app is first booted, if this resource is loaded before parent resource,
+      #   Resource::BelongsTo::TargetNotFound is certain to be raised
+      suppress Resource::BelongsTo::TargetNotFound do
+        if @belongs_to.target_names.count == 1
+          options[:class_name] ||= @belongs_to.targets[0].resource_class_name
+        end
+      end
+
+      controller.send :belongs_to, *targets, options.dup
     end
 
     def belongs_to_config
--- a/lib/active_admin/resource/belongs_to.rb
+++ b/lib/active_admin/resource/belongs_to.rb
@@ -3,6 +3,7 @@ require 'active_admin/resource'
 module ActiveAdmin
   class Resource
     class BelongsTo
+      class NotSupported < StandardError; end
 
       class TargetNotFound < StandardError
         def initialize(key, namespace)
@@ -14,23 +15,26 @@ module ActiveAdmin
       # The resource which initiated this relationship
       attr_reader :owner
 
-      # The name of the relation
-      attr_reader :target_name
+      # The name of the relations
+      attr_reader :target_names
 
-      def initialize(owner, target_name, options = {})
+      def initialize(owner, target_names, options = {})
+        if target_names.size > 1 && !options[:optional]
+          raise NotSupported, "Belongs to multiple parents must be optional"
+        end
+
         @owner = owner
-        @target_name = target_name
+        @target_names = target_names
         @options = options
       end
 
-      # Returns the target resource class or raises an exception if it doesn't exist
-      def target
-        resource or raise TargetNotFound.new (@options[:class_name] || @target_name.to_s.camelize), namespace
-      end
-
-      def resource
-        namespace.resources[@options[:class_name]] ||
-          namespace.resources[@target_name.to_s.camelize]
+      # Returns the target resource classes or raises an exception if one of the targets doesn't exist
+      def targets
+        target_names.map do |target_name|
+          namespace.resources[@options[:class_name]] ||
+            namespace.resources[target_name.to_s.camelize] ||
+            raise(TargetNotFound.new((@options[:class_name] || target_name.to_s.camelize), namespace))
+        end
       end
 
       def namespace
@@ -46,7 +50,7 @@ module ActiveAdmin
       end
 
       def to_param
-        :"#{@target_name}_id"
+        target_names.map { |target_name| :"#{target_name}_id" }
       end
     end
   end
--- a/spec/unit/belongs_to_spec.rb
+++ b/spec/unit/belongs_to_spec.rb
@@ -3,14 +3,16 @@ require 'rails_helper'
 RSpec.describe ActiveAdmin::Resource::BelongsTo do
   before do
     load_resources do
-      ActiveAdmin.register User
-      ActiveAdmin.register Post do belongs_to :user end
+      category_config
+      user_config
+      post_config
     end
   end
 
   let(:namespace) { ActiveAdmin.application.namespace(:admin) }
+  let(:category_config) { ActiveAdmin.register Category }
   let(:user_config) { ActiveAdmin.register User }
-  let(:post_config) { ActiveAdmin.register Post do belongs_to :user end }
+  let(:post_config) { ActiveAdmin.register(Post) { belongs_to :user } }
   let(:belongs_to) { post_config.belongs_to_config }
 
   it "should have an owner" do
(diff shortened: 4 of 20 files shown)


Looooong:
Ping @deivid-rodriguez . I would be happy if this PR is reviewed soon, as our project depends on it.

deivid-rodriguez:
Yes, this is at the top of my ActiveAdmin list now. Hope to get to it soon.

deivid-rodriguez:
I'm so sorry for the delay @Looooong, you must be tired of waiting :disappointed:. I started reviewing this, and will have some feedback for you tomorrow. But it's not going to be major concerns since it's looking great!

lib/active_admin/resource/belongs_to.rb
```diff
@@ -14,23 +15,26 @@ def initialize(key, namespace)
...
-        resource or raise TargetNotFound.new (@options[:class_name] || @target_name.to_s.camelize), namespace
-      end
-
-      def resource
-        namespace.resources[@options[:class_name]] ||
-          namespace.resources[@target_name.to_s.camelize]
+      # Returns the target resource classes or raises an exception if one of the targets doesn't exist
+      def targets
+        target_names.map do |target_name|
+          namespace.resources[@options[:class_name]] ||
```
deivid-rodriguez, line 34: I'm just getting started with this so bear with me if I don't make much sense, but. The `:class_name` option used to refer to the (previously uniq) target. Since now we can have multiple targets, shouldn't the `:class_name` be configurable _per target_? So maybe something like the following would work better?

```ruby
belongs_to :category, author: { class_name: "User" }, optional: true 
```

spec/unit/belongs_to_spec.rb
```diff
@@ -3,14 +3,16 @@
...
+      user_config
+      post_config
     end
   end
 
   let(:namespace) { ActiveAdmin.application.namespace(:admin) }
+  let(:category_config) { ActiveAdmin.register Category }
   let(:user_config) { ActiveAdmin.register User }
-  let(:post_config) { ActiveAdmin.register Post do belongs_to :user end }
+  let(:post_config) { ActiveAdmin.register(Post) { belongs_to :user } }
```
deivid-rodriguez, line 15: Ugg, this style is ugly indeed, I'm going to check if `rubocop` can forbid it.

lib/active_admin/resource.rb
```diff
@@ -135,11 +135,19 @@ def defined_actions
       controller.instance_methods.map(&:to_sym) & ResourceController::ACTIVE_ADMIN_ACTIONS
     end
 
-    def belongs_to(target, options = {})
-      @belongs_to = Resource::BelongsTo.new(self, target, options)
+    def belongs_to(targets, options = {})
+      @belongs_to = Resource::BelongsTo.new(self, targets, options.dup)
```
deivid-rodriguez, line 139: Do the `options` get mutated now or why are we dupping them?

lib/active_admin/page.rb
```diff
@@ -84,10 +84,11 @@ def clear_page_actions!
       @page_actions = []
     end
 
-    def belongs_to(target, options = {})
-      @belongs_to = Resource::BelongsTo.new(self, target, options)
-      self.navigation_menu_name = target unless @belongs_to.optional?
-      controller.send :belongs_to, target, options.dup
+    def belongs_to(targets, options = {})
+      @belongs_to = Resource::BelongsTo.new(self, targets, options.dup)
```
deivid-rodriguez, line 88: Same question about mutating the options.

lib/active_admin/resource.rb
Looooong: Yes, it get mutated.

lib/active_admin/resource/belongs_to.rb
Looooong: I have noticed that too. However, I haven't seen anyway to specify `class_name` for each optional resource in the `inhertied_resources` API that we are passing the `@options` to. That's why I leave it the way it is. I can support specifying individual `class_name` for each resource in ActiveAdmin domain if you like.

Looooong: Indirectly related to this problem, I think that `inherited_resources` API has to be redesigned to be more intuitive. For example, this code declares that `task` is nested under `project`:

```ruby
belongs_to :project
belongs_to :task
# The same as
belongs_to :project do
  belongs_to :task
end
```

I propose that we change this behaviour to allow specifying multiple resources in the same level as well as specifying options for each of them.

```ruby
class CommentsController < InheritedResources::Base
  # /other_projects/1/comments
  belongs_to :project, optional: true, route_collection_name: 'other_projects'
  # /tasks/1/comments
  belongs_to :task, optional: true, class_name: 'Todo'

  # Nesting behaviour
  belongs_to :manager do
    # /managers/1/projects/1/comments
    belongs_to :project, optional: true
    # /managers/1/tasks/1/comments
    belongs_to :tasks, optional: true
  end
end
```

lib/active_admin/page.rb
Looooong: Same.

lib/active_admin/resource/belongs_to.rb
deivid-rodriguez: Yes, that's very confusing to me too! I think your proposal makes much more sense, I believe we can could the behaviour on the next major version.

Regarding the first issue, you're right too... We need to extend the `inherited_resources` API to allow options to be passed _per target_. Would you be willing to work on that?

I stumped with this issue while trying to add a sample multiple optional `belongs_to` to our sample application. I believe it's quite common and people will request it as soon as we release this...

lib/active_admin/resource.rb
deivid-rodriguez: mmm, but if it's the `controller` mutating them, we are already passing a mutated hash to it, so I guess this one shouldn't be needed.

lib/active_admin/page.rb
deivid-rodriguez: Let's leave it as it is if not necessary.

lib/active_admin/resource.rb
Looooong: I remember the reason why, the controller mutates `class_name` option, which interferes with `BelongsTo`. For multiple optional resources, we don't want `class_name` to be set, because it will mess up the `targets` array.

deivid-rodriguez: But the controller has its own copy of this hash, right? I just tried to run the specs without passing mutated hashes the `BelongsTo.new` and they all passed so if that's really need I think we need some extra coverage somewhere.

Looooong: Let me check if this is a problem without duplicating the hash.

lib/active_admin/resource/belongs_to.rb
Looooong: Unfortunately, I will stop working on Rails or any Ruby related project soon so I won't be able to work on `inherited_resources` API change. On the other hand, since we want to set multiple `belongs_to` resources in the same level, they should be `optional` by default. Also, I doubt that this change will be pretty complex.

lib/active_admin/resource.rb
Looooong: I haven't tested it yet, but I think that if you don't duplicate the `options` hash and have the code reloaded the second time, then the `BelongsTo#targets` will be incorrect.

lib/active_admin/resource/belongs_to.rb
deivid-rodriguez: Too bad, good luck with your next projects with other tools anyways :)

Just to clarify, you won't be able to work on either the API change you proposed nor in my proposal for this issue (`belongs_to :category, author: { class_name: "User" }, optional: true`), right? If that's the case, no big deal, I had a look at `intherited_resources` and as you point out, it should be fairly simple, you did all the hard work already :purple_heart:

lib/active_admin/resource.rb
deivid-rodriguez: I'll like into that, thanks!

lib/active_admin/resource/belongs_to.rb
Looooong: I took a look at your proposal and I found that it is not as easy as it look. Let's take your example code for demonstration:

```ruby
belongs_to :category, author: { class_name: "User" }, optional: true
```

The call like this will result in:
+ `targets == [:category]`
+ `options == { author: { class_name: "User" }, optional: true }`

If we want to support this behaviour, we need to know with key in the options is the target, and which keys is the real option. We can whitelist the option keys, but if `inherit_resources` API is changed, someone must know to whitelist the option keys in ActiveAdmin API as well. I find this absurd.

There are other alternatives to your proposal:

The first one looks ugly to me:

```ruby
belongs_to :category, { author: { class_name: "User" } }, optional: true
```

The second one would need to redesign the entire API:

```ruby
belongs_to :category, optional: true
belongs_to :author, class_name: "User", optional: true
```

What do you think?

deivid-rodriguez: Hi @Looooong. Shit, you're right :anguished:

Should we introduce an `optional_belongs_to` helper that automatically assumes `optional: true`? I believe that would work?

Looooong: Then we must ensure that there is only one call to `belongs_to` or `optional_belongs_to` in the resource. Sounds good?

deivid-rodriguez: I think that makes sense? Currently, multiple `belongs_to` in a resource doesn't really work in any way, right? Or does it lead to the nesting behavior of the inherited_resources API? Sorry, it's been a while and I don't remember this well.

Looooong: Yes, multiple `belongs_to` doesn't work in ActiveAdmin. We can take this chance to enforce that there is only a single `belongs_to` in resource file. Multiple `belongs_to` will cause the later one override all the `belongs_to` configurations before it. About the nesting behaviour, I'm not so sure about that.

deivid-rodriguez: Sounds like a good plan @Looooong

Looooong: @deivid-rodriguez I'm sorry to be the bearer of bad news, but I find myself in the position where I cannot allocate more time to work on this PR. Would you mind taking over this PR for me?

deivid-rodriguez: Of course, thanks for letting me know, and thanks for getting this started!

RobertLowe:
Nesting would be really nice

```
  # Nesting behaviour
  belongs_to :manager do
    # /managers/1/projects/1/comments
    belongs_to :project, optional: true
    # /managers/1/tasks/1/comments
    belongs_to :tasks, optional: true
  end
```

rosendi:
> Nesting would be really nice
> 
> ```
>   # Nesting behaviour
>   belongs_to :manager do
>     # /managers/1/projects/1/comments
>     belongs_to :project, optional: true
>     # /managers/1/tasks/1/comments
>     belongs_to :tasks, optional: true
>   end
> ```

@deivid-rodriguez Almost 8 years have passed away from the first issue about nested belongs_to. How do you handle it in real projects? Only 2 levels?

corlissc:
@deivid-rodriguez, @Looooong any estimated time-frame on getting this functionality merged and released?

Thank you so much for your effort. Hopefully this capability can be released soon. I know I'm not alone in needing this capability.

javierjulio:
There are many merge conflicts here to resolve to have to be reconsidered. I don't use the belongs_to feature and don't intend to support it. If others find value in this and can update it to resolve conflicts and get tests passing, we can reconsider. Thank you.

PR #6122, opened 2020-03-09
Add ruby 2.7 to CI

deivid-rodriguez:
I'm starting to review #5855, but tests don't seem to pass locally.

I want to rule out that it's related to my ruby version, and start testing against ruby 2.7 in CI, since I'm at it.


deivid-rodriguez:
This is ready @javierjulio!

PR #6123, opened 2020-03-09
Test against latest formtastic master

deivid-rodriguez:
I want to help getting a new release of formtastic out there. I want to make sure everything is ok with activeadmin.


deivid-rodriguez:
Ready @javierjulio!

PR #6124, opened 2020-03-10
Sort files to be loaded

jiikko:
hi.

I'm trying to make sure on CI that missing comments of annotate.
https://rubygems.org/gems/annotate

```shell
# for example
bundle exec annotate --routes
git diff --exit-code config/routes.rb # CI fail if exit(1)
```

However, there was a difference in the output of rake routes between macOS and CircleCI.
As a result of investigation, it was caused by using `Dir.[]` In `ActiveAdmin.routes()`.
https://bugs.ruby-lang.org/issues/8709

With this fix, the output of rake routes will be the same in all environments to sort the result of `Dir.[]`.

Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Enhancements
+
+* Make source `ActiveAdmin.routes` provides routes in a consistent order. [#6124] by [@jiikko]
+
 ### Dependency Changes
 
 * Support for ruby 2.4 has been removed. [#6198] by [@deivid-rodriguez]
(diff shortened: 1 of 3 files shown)


leonid-shevtsov:
💯 - I have this same fix monkeypatched in multiple projects.

I also guard the patch with `Rails.env.development? || Rails.env.test?` to avoid production issues. But I don't think that that _would_ be any production issues with sorting the files.

IgorDobryn:
Any update on this? 
Met this issue just recently. Will need to apply the monkeypatch to keep behaviour consistent between dev and CI(test) envs.

deivid-rodriguez:
Hei!

I'm good with doing this as long as we add a test that ensures that we don't remove it in the future because of not being necessary :+1:.

jiikko:
@deivid-rodriguez 
thanks for response. I will try to write a test.

jiikko:
@deivid-rodriguez 
I have written a test.
could do you please review?

tf:
This change breaks my app in so many ways. I was working under the assumption that I can override settings of admin resources defined in a gem in admin resources in the main application. This change now makes the load order order depend on resource file path. That only made the error visible when running in production: In development gems come from `/home/foo/.rvm/...` and the app from `/home/foo/my_project/...` - so admins from gems are loaded first. In production gems come from `/home/app/my_project/vendor/...` and are thus loaded after the admin resources defined in in `app/admins`.

If consistent file order is important, I'd suggest to sort the result of `Dir[]`, but still honor the order of the load path.

deivid-rodriguez:
Can you propose a PR and some tests to illustrate the problem being fixed?

tf:
@deivid-rodriguez I've created #7488 to address the issue.

PR #6133, opened 2020-03-12
Ensure changelog has well-defined main sections

deivid-rodriguez:
And make sure it stays like that in the future.


deivid-rodriguez:
@javierjulio The typo being fixed here was merged as part of #5855, but I think we can still keep the changelog check to prevent future similar typos.

Rebased, it should be ready :+1:.

PR #6148, opened 2020-03-19
Update Tidelift referral link

phated:
@deivid-rodriguez last one, I believe. ref #6146


brobertsaz:
Wow, is @phated playing with Rails these days?

deivid-rodriguez:
Thanks so much @phated!! :purple_heart:

PR #6149, opened 2020-03-20
Fix scopes renderer when resource has only optional scopes and their conditions are false

### LINKED ISSUE #5578: conditional scopes output strange output in index page
After we upgraded to 1.4.0 we see that conditional scopes are outputting some strange debug info in the place where scopes should be. 
```
scope 'test_scope', if: -> { false } do |scope|
    scope.where(...)
  end 
```

This outputs:
screenshot 2018-11-20 11 02 00

Looooong:
This PR resolves #5578 .

### DIFF
--- a/features/step_definitions/index_scope_steps.rb
+++ b/features/step_definitions/index_scope_steps.rb
@@ -40,3 +40,7 @@ Then "I should see an empty group with the scope {string}" do |name|
   name = name.tr(" ", "").underscore.downcase
   expect(page).to     have_css ".scopes .scope-default-group .#{name}"
 end
+
+Then "I should see empty scopes" do
+  expect(page.find(".scopes").text).to be_empty
+end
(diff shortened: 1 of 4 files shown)


Looooong:
@deivid-rodriguez Maybe I'm over-testing this. One scenario should be enough to cover the issue.

Looooong:
I have updated the scenario.

features/step_definitions/index_scope_steps.rb
```diff
@@ -36,7 +36,11 @@
   expect(page).to     have_css ".scopes .scope-group-#{group} .#{name2}"
 end
 
-Then"I should see an empty group with the scope {string}" do |name|
+Then "I should see an empty group with the scope {string}" do |name|
```
deivid-rodriguez: Sorry, I forgot to mention this. I landed this separately, so if you could rebase on top of latest master to remove this bit from the final diff, that'd be great :+1:.

Looooong: I will do it.

PR #6193, opened 2020-04-12
Update streaming.rb

bhargavd:
Following error I am getting
ArgumentError - wrong number of arguments (given 1, expected 0):
   () Users/bhargav/.rvm/gems/ruby-2.4.4/bundler/gems/activeadmin-85ab3f5961e1/lib/active_admin/resource_controller/streaming.rb:11:in `index'

### DIFF
--- a/lib/active_admin/resource_controller/streaming.rb
+++ b/lib/active_admin/resource_controller/streaming.rb
@@ -8,7 +8,7 @@ module ActiveAdmin
     #
     module Streaming
 
-      def index
+      def index(arg=nil)
         super do |format|
           format.csv { stream_csv }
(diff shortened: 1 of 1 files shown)


lib/active_admin/resource_controller/streaming.rb
```diff
@@ -8,7 +8,7 @@ class ResourceController < BaseController
     #
     module Streaming
 
-      def index
+      def index(arg=nil)
```
sunny, line 11: Another way of specifying that it accepts arguments without having to name them would be:

```suggestion
      def index(*)
```

javierjulio:
If this is still an issue we'll reopen.

PR #6195, opened 2020-04-13
Update filters disabled error to include specific action

javawizard:
`ActiveAdmin::Filters::Disabled` is raised both when attempting add a filter and when attempting to remove a filter when filters are disabled. Change the message to suit both use cases.

### DIFF
--- a/lib/active_admin/filters/resource_extension.rb
+++ b/lib/active_admin/filters/resource_extension.rb
@@ -4,7 +4,7 @@ module ActiveAdmin
 
     class Disabled < RuntimeError
       def initialize
-        super "Can't remove a filter when filters are disabled. Enable filters with 'config.filters = true'"
+        super "Can't add or remove a filter when filters are disabled. Enable filters with 'config.filters = true'"
       end
     end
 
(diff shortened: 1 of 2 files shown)


javawizard:
Is this change significant enough to warrant an entry in `CHANGELOG.md`? If so, is it a bugfix or an enhancement?

deivid-rodriguez:
Hi! Sorry for the delay, I'd go with considering this a bug fix!

lib/active_admin/filters/resource_extension.rb
```diff
@@ -3,7 +3,7 @@ module Filters
 
     class Disabled < RuntimeError
       def initialize
-        super "Can't remove a filter when filters are disabled. Enable filters with 'config.filters = true'"
+        super "Can't add or remove a filter when filters are disabled. Enable filters with 'config.filters = true'"
```
deivid-rodriguez, line 6: I was thinking we could stop overriding `initialize`, and do `class Disabled < RuntimeError; end`, and then pass the correct message for each of the cases instead?

javierjulio:
That's a worthwhile suggestion so at least the message is specific to the action taking place. @javawizard I'd accept that change if you still want to get this in. Thank you.

Issue #5430, opened 2018-06-20
Associations are persisted before validation of main record


tkalliom:
### Expected behavior

When updating a record with changes to its children that make the record fail validation, no changes are saved to the database.

### Actual behavior

The invalid association change is persisted, and then the user gets redirected to a form displaying a validation error.

### How to reproduce

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'activeadmin', ENV['ACTIVE_ADMIN_PATH'] ? {path: ENV['ACTIVE_ADMIN_PATH'], require: false} : {github: 'activeadmin/activeadmin', require: false}
  gem 'rails', '~> 5.1.0'
  gem 'sass-rails'
  gem 'sqlite3', platform: :mri
  gem 'activerecord-jdbcsqlite3-adapter', "~> 51.0", platform: :jruby
end

require 'active_record'

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :agents, force: true do |t|
    t.string :full_name
  end

  create_table :agents_clients, force: true do |t|
    t.bigint :agent_id, null: false
    t.bigint :client_id, null: false
  end

  create_table :clients, force: true do |t|
    t.string :name
  end
end

require 'action_controller/railtie'
require 'action_view/railtie'
require 'active_admin'

class TestApp < Rails::Application
  config.root = __dir__
  secrets.secret_token = 'secret_token'
  secrets.secret_key_base = 'secret_key_base'

  config.eager_load = false
  Rails.logger = config.logger = Logger.new($stdout)
end

class ApplicationController < ActionController::Base
  include Rails.application.routes.url_helpers
end

class Client < ActiveRecord::Base
  has_and_belongs_to_many :agents
  validates :agents, presence: true
end

class Agent < ActiveRecord::Base
  has_and_belongs_to_many :clients
end

ActiveAdmin.setup do |config|
  config.authentication_method = false
  config.current_user_method = false
end

Rails.application.initialize!

ActiveAdmin.register Client do
  permit_params agent_ids: []
  
  form do |f|
    f.input :agents, as: :check_boxes
  end
end

Rails.application.routes.draw do
  ActiveAdmin.routes(self)
end

require 'minitest/autorun'
require 'rack/test'

class HabtmTest < ActionDispatch::IntegrationTest
[...]

PR #6202, opened 2020-04-22
run update_resource inside a transaction to avoid autosaving relationships through assign_attributes when the record is invalid

arbesulo:
fixes https://github.com/activeadmin/activeadmin/issues/5430

When submitting a form with invalid attributes, some associations are autosaved via `assign_attributes` before calling `save_resource`. As a consequence, the user sees a validation error and is not aware that the changes to the associations got persisted.

This PR runs everything inside an `ActiveRecord::Transaction` so that everything is rolled back when the model is not valid.


arbesulo:
I need some help with the regression tests... Can anybody point me in the right direction?

legendado:
Tested on real project. Works fine :+1:

arbesulo:
closing due to lack of help... I hope this works already on the stable branches

deivid-rodriguez:
Thanks @arbesulo, sorry nobody was able to help out. I created an updated copy of this PR, this seems useful work, so by keeping an open PR maybe some will eventually get to ir.

deivid-rodriguez:
By the way, this seems like a Rails issue? Is there any open report about this?

arbesulo:
@deivid-rodriguez honestly I haven't checked if there is any open issue in rails related to this.

Also I'm no longer in the project where this issue arised, so I don't know if the current version still has it or not.

PR #6210, opened 2020-05-04
Fix es and es-MX sign_in and sign_up translation:

roramirez:
This change fix the use the sign_up and sign_in meaning. In the last
version of these files it's was using the reverse meaning.

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -144,7 +144,7 @@ es:
         submit: "Reenviar instrucciones de confirmación"
       links:
         sign_up: "Registrarse"
-        sign_in: "Ingresar"
+        sign_in: "Iniciar Sesión"
         forgot_your_password: "¿Olvidó su contraseña?"
         sign_in_with_omniauth_provider: "Conéctate con %{provider}"
         resend_unlock_instructions: "Reenviar instrucciones de desbloqueo"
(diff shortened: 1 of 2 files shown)


config/locales/es.yml
```diff
@@ -143,8 +143,8 @@ es:
         title: "Reenviar instrucciones de confirmación"
         submit: "Reenviar instrucciones de confirmación"
       links:
-        sign_up: "Ingresar"
-        sign_in: "Registrarse"
+        sign_up: "Registrarse"
+        sign_in: "Ingresar"
```
javierjulio, line 147: @roramirez I realize this was using `"Ingresar"` before but overall it seems we use `"Iniciar Sesión"` consistently. What do you think about changing that here to the following?
```suggestion
        sign_in: "Iniciar Sesión"
```

roramirez: I'm seems good to change this string. It's more clearly to use this way  in spanish

javierjulio: Great thanks!

deivid-rodriguez:
CI failure seems unrelated, so this can be merged I believe?

javierjulio:
@deivid-rodriguez thanks for the heads up! I didn't know it ended up failing. Yes I think so. I checked CI and re-ran the build to ensure its all good and it passed. 👍🏻

deivid-rodriguez:
Thanks!!

PR #6225, opened 2020-05-14
Move sassc to being a development dependency ONLY

alex-hall:
Sassc has quite a [few](https://github.com/sass/sassc-ruby/issues/179) open issues about the subject of portability, 
And often times it’s recommended to just re-install the gem and rebuild
Native extensions. However, for build/CI processes, it’s unacceptable to wait 4+ minutes on a gem
that you've already built in a previous step or don't even rely on. 

This PR moves the sassc dependency to being a development ONLY dependency as it should not be required during runtime.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -17,7 +17,6 @@ PATH
       kaminari (~> 1.0, >= 1.0.1)
       railties (>= 5.2, < 6.1)
       ransack (~> 2.1, >= 2.1.1)
-      sassc-rails (~> 2.1)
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Hei @alex-hall, I'm addressing this at #6584. This solution is incomplete because it is only moving the dependency but internal code still depends on `sassc`. Anyways, thanks a lot for your contribution. I'll amend my commit in #6584 to give you credit too and I apologize for taking this long to address this!

PR #6232, opened 2020-05-18
Fix filter_columns_for_large_association and filter_method_for_large_association examples

ndbroadbent:
I was just checking out the new `maximum_association_filter_arity`, `filter_columns_for_large_association`, `filter_method_for_large_association` options from https://github.com/activeadmin/activeadmin/pull/5548, which look like they will be a huge help! I think `maximum_association_filter_arity` should definitely be set to a sane maximum by default.

I copied these lines into my `config/initializers/active_admin.rb` and uncommented them, and saw that there was a syntax error. (`,` should be `=`). So I just thought I would send a PR to fix that!


deivid-rodriguez:
Thanks! :heart:

PR #6237, opened 2020-05-18
Ruby 2.7 warnings

deivid-rodriguez:
Fixing a bunch of ruby 2.7 warnings here.

Opening a draft PR to check CI.


deivid-rodriguez:
This should be ready now @javierjulio. No more warnings on ruby 2.7 from our own code, only some dependencies are left.

PR #6244, opened 2020-05-20
Disallow extra spacing even for alignment

deivid-rodriguez:
I'd like to enable the `AlllowForAlignment: false` setting for the `Style/ExtraSpacing` cop in our rubocop configuration.

The rationale is explained below but I understand this is a matter of taste so if other maintainers disagree I'll immediately close this without any bike-shedding :)

Some people like the following code because it's prettier. I don't, because it makes diffs harder to read. Sometimes it forces you to change a bunch of lines just to keep the prettiness when you just want to really fix a single line. In my opinion, this hides meaningful changes inside stylistic ones.

For example, say I want to fix a bug caused by a typo in one of the keys in a big hash.  I used `deployments` but the actual key is `deployment`.

```ruby
{
  :clean       => true,
  :deployments => true,
  :frozen      => true,
  :"no-cache"  => true,
  :"no-prune"  => true,
  :path        => "vendor/bundle",
  :shebang     => "ruby27",
  :system      => true,
  :without     => "development",
  :with        => "development",
}
```

If using the "pretty alignment" style, I need to edit all lines in the hash just to make the fix, in order to keep the style, so the diff would look like:

```diff
{
- :clean       => true,
- :deployments => true,
- :frozen      => true,
- :"no-cache"  => true,
- :"no-prune"  => true,
- :path        => "vendor/bundle",
- :shebang     => "ruby27",
- :system      => true,
- :without     => "development",
- :with        => "development",
+ :clean      => true,
+ :deployment => true,
+ :frozen     => true,
+ :"no-cache" => true,
+ :"no-prune" => true,
+ :path       => "vendor/bundle",
+ :shebang    => "ruby27",
+ :system     => true,
+ :without    => "development",
+ :with       => "development",
}
```

In my opinion this is both inconvenient, and also hides the fix. With a no spacing style, it's simpler:

```diff
{
  :clean => true,
- :deployments => true,
+ :deployment => true,
  :frozen  => true,
  :"no-cache" => true,
  :"no-prune" => true,
  :path => "vendor/bundle",
  :shebang => "ruby27",
  :system => true,
  :without => "development",
  :with => "development",
}
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -55,6 +55,7 @@ Layout/EndOfLine:
   Enabled: true
 
 Layout/ExtraSpacing:
+  AllowForAlignment: false
   Enabled: true
 
 Rails/FilePath:
--- a/Gemfile
+++ b/Gemfile
@@ -56,7 +56,7 @@ group :lint do
 end
 
 group :docs do
-  gem 'yard'     # Documentation generator
+  gem 'yard' # Documentation generator
   gem 'kramdown' # Markdown implementation (for yard)
 end
 
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -1,19 +1,19 @@
 require File.join(__dir__, "lib", "active_admin", "version")
 
 Gem::Specification.new do |s|
-  s.name          = 'activeadmin'
-  s.license       = 'MIT'
-  s.version       = ActiveAdmin::VERSION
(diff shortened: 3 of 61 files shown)


deivid-rodriguez:
Thanks @javierjulio :heart_eyes:

javierjulio:
You’re welcome!

PR #6249, opened 2020-05-24
Allow using PORO decorators

### LINKED ISSUE #6248: Allow using PORO decorators
## Motivation

We were removing Draper from our projects template in favor of PORO presenter since we were filling decorators were used in a none-optimal way, plus we wanted to have fewer dependencies and a lower learning curve for newcomers. During that process, we found out how often we used decorators on ActiveAdmin pages and how useful they could be.

We also found that ActiveAdmin has a hard dependency on Draper making it very complex to develop presenters/decorators that conform with the expected API (this issue was very helpful https://github.com/activeadmin/activeadmin/issues/3600)

## Implementation

I think most of the issues with decorators and the need to delegate to the underlying ActiveRecord Relation come from decorating in an early stage. Since the decorator's methods are meant to be used just at the presentation layer I believe we can decorate the collection closer to the point where the methods will be used and just work with the plain ActiveRecord Relation before that point.

I'll open a PR with this implementation change, I'm still trying to figure out a better way to get the decorating method to the Index Page but it's more to check out if it's something desirable. I tested it locally and automated tests are green but let me know if there's anything else I can do to assure this works properly!

brunvez:
Closes #6248

### DIFF
--- a/docs/11-decorators.md
+++ b/docs/11-decorators.md
@@ -8,11 +8,6 @@ Active Admin allows you to use the decorator pattern to provide view-specific
 versions of a resource. [Draper](https://github.com/drapergem/draper) is
 recommended but not required.
 
-To use decorator support without Draper, your decorator must support a variety
-of collection methods to support pagination, filtering, etc. See
-[this github issue discussion](https://github.com/activeadmin/activeadmin/issues/3600)
-and [this gem](https://github.com/kiote/activeadmin-poro-decorator) for more details.
-
 ## Example usage
 
 ```ruby
@@ -38,6 +33,22 @@ ActiveAdmin.register Post do
     column :title
     column :image
     actions
+  end
+end
+```
+
+You can pass any decorator class as an argument to `decorate_with`
+as long as it accepts the record to be decorated as a parameter in
+the initializer, and responds to all the necessary methods.
+
+```ruby
+# app/decorators/post_decorator.rb
+class PostDecorator
+  attr_reader :post
+  delegate_missing_to :post
+
+  def initialize(post)
+    @post = post
   end
 end
 ```
--- a/lib/active_admin/resource_controller/decorators.rb
+++ b/lib/active_admin/resource_controller/decorators.rb
@@ -52,14 +52,9 @@ module ActiveAdmin
         @cache = {}
 
         def self.wrap(decorator)
-          collection_decorator = find_draper_collection_decorator(decorator)
-
-          if collection_decorator
-            name = "#{collection_decorator.name} of #{decorator} + ActiveAdmin"
-            @cache[name] ||= wrap! collection_decorator, name
-          else
-            decorator
-          end
+          collection_decorator = find_collection_decorator(decorator)
+          name = "#{collection_decorator.name} of #{decorator} + ActiveAdmin"
+          @cache[name] ||= wrap! collection_decorator, name
         end
 
         private
@@ -68,16 +63,18 @@ module ActiveAdmin
           ::Class.new parent do
             delegate :reorder, :page, :current_page, :total_pages, :limit_value,
                      :total_count, :total_pages, :offset, :to_key, :group_values,
-                     :except, :find_each, :ransack
+                     :except, :find_each, :ransack, to: :object
 
             define_singleton_method(:name) { name }
           end
         end
 
-        def self.find_draper_collection_decorator(decorator)
-          return unless Dependency.draper? && decorator && decorator <= ::Draper::Decorator
-
-          decorator.collection_decorator_class
+        def self.find_collection_decorator(decorator)
+          if decorator.respond_to?(:collection_decorator_class)
+            decorator.collection_decorator_class
+          else
+            CollectionDecorator
+          end
         end
       end
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Enhancements
 
 * Make source `ActiveAdmin.routes` provides routes in a consistent order. [#6124] by [@jiikko]
+* Allow using PORO decorators. [#6249] by [@brunvez]
 
 ### Dependency Changes
 
@@ -582,6 +583,7 @@ Please check [0-6-stable] for previous changes.
 [#6099]: https://github.com/activeadmin/activeadmin/pull/6099
 [#6124]: https://github.com/activeadmin/activeadmin/pull/6124
 [#6198]: https://github.com/activeadmin/activeadmin/pull/6198
+[#6249]: https://github.com/activeadmin/activeadmin/pull/6249
 
 [@5t111111]: https://github.com/5t111111
 [@aarek]: https://github.com/aarek
@@ -594,6 +596,7 @@ Please check [0-6-stable] for previous changes.
 [@bartoszkopinski]: https://github.com/bartoszkopinski
 [@blocknotes]: https://github.com/blocknotes
 [@bolshakov]: https://github.com/bolshakov
+[@brunvez]: https://github.com/brunvez
 [@buren]: https://github.com/buren
 [@chancancode]: https://github.com/chancancode
 [@chrp]: https://github.com/chrp
--- a/features/decorators.feature
+++ b/features/decorators.feature
@@ -26,6 +26,25 @@ Feature: Decorators
     And I should see "A very unique post"
     And I should see "No"
 
+  Scenario: Index page with PORO decorator
+    Given a configuration of:
+    """
+      ActiveAdmin.register Post do
+        decorate_with PostPoroDecorator
+
+        index do
+          column(:id)
+          column(:title)
+          column(:decorator_method)
+          column(:starred)
+        end
+      end
+    """
+    When I am on the index page for posts
+    Then I should see "A method only available on the PORO decorator"
+    And I should see "A very unique post"
+    And I should see "No"
+
   Scenario: Show page with decorator
     Given a configuration of:
     """
--- a/features/index/batch_actions.feature
+++ b/features/index/batch_actions.feature
@@ -60,6 +60,23 @@ Feature: Batch Actions
     Then I should see a flash with "Successfully deleted 2 posts"
     And I should see 3 posts in the table
 
+  Scenario: Use default (destroy) batch action on a PORO decorated resource
+    Given 5 posts exist
+    And an index configuration of:
+    """
+      ActiveAdmin.register Post do
+        decorate_with PostPoroDecorator
+      end
+    """
+    When I check the 2nd record
+    And I check the 4th record
+    And I follow "Batch Actions"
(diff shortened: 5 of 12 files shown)


brunvez:
@deivid-rodriguez @javierjulio any thoughts on this? It'd really help our team to break the dependency with Draper

deivid-rodriguez:
@brunvez Thanks for this! :tada:

 'm currently working on upgrading the whole development & runtime stack to ruby 2.7. That's alomost done now. After that I'll work on #6115, and this one is next on my list.

brunvez:
Hey @deivid-rodriguez sorry to poke again but I was wondering if there's anything I can do to move this PR forward 😄

deivid-rodriguez:
Hei @brunvez! 

Since #6115 will require a bit more work, I'll skip directly to reviewing this PR. Hope to do it very soon!

lib/active_admin/resource_controller/decorators.rb
```diff
@@ -77,9 +77,9 @@ def self.wrap!(parent, name)
         # Draper::CollectionDecorator was introduced in 1.0.0
         # Draper::Decorator#collection_decorator_class was introduced in 1.3.0
         def self.find_collection_decorator(decorator)
-          if Dependency.draper? '>= 1.3.0'
+          if Dependency.draper?('>= 1.3.0') && decorator <= Draper::Decorator
```
deivid-rodriguez, line 80: I removed all this stuff because we don't really need it. Can you rebase and adapt the PR? I guess it's a matter of saying

```suggestion
          if Dependency.draper? && decorator <= Draper::Decorator
```

and restoring the final branch.

brunvez: There we go, by removing the if here the expected interface for the decorator class changed to also require `collection_decorator_class`. But I think is a good tradeoff to have the code here cleaner.

deivid-rodriguez: I wasn't meaning to suggest changing the interface for PORO decorators. I just meant to adapt the code to the lack of versions checks for `draper, with:

```ruby
def self.find_collection_decorator(decorator)
  if Dependency.draper? && decorator <= Draper::Decorator
    decorator.collection_decorator_class
  else
    decorator
end
```

Do you think we should require PORO decorators to define `collection_decorator_class`?

By the way, this remainded me that we need docs for this feature at `docs/11-decorators.md`.

deivid-rodriguez: I proposed #6351. With that PR I believe we don't need any changes here in this file, nor to add a requirement for PORO decorators to define `collection_decorator_class`.

brunvez: Cool, yeah on a second thought I think making PORO define the `collection_decorator_class` to just return `self` might not be too elegant. Do you want to merge that refactor first and I'll add docs on how to use PORO presenters?

deivid-rodriguez: Yeah! :+1:

javierjulio: @brunvez we just merged in the change so you are good to rebase. Thanks!

docs/11-decorators.md
```diff
@@ -42,6 +37,33 @@ ActiveAdmin.register Post do
...
+  end
+
+  def self.decorate(record_or_collection, *_)
+    if record_or_collection.is_a?(Enumerable)
+      record_or_collection.map { |individual_record| new(individual_record) }
+    else
+      new(record_or_collection)
+    end
+  end
+end
```
deivid-rodriguez, line 61: Something I don't love about this is that each decorator needs to add this code, which is as you point out later very generic. I feel ActiveAdmin should just work without having to add this. So I took a stab at it by copying the basic functionality from draper's collection presenter. I came up with this:

```diff
--- a/lib/active_admin/resource_controller/decorators.rb
+++ b/lib/active_admin/resource_controller/decorators.rb
@@ -48,18 +48,38 @@ module ActiveAdmin
         end
       end
 
+      class CollectionDecorator
+        # @return the collection being decorated.
+        attr_reader :object
+
+        # @return [Class] the decorator class used to decorate each item, as set by {#initialize}.
+        attr_reader :decorator_class
+
+        array_methods = Array.instance_methods - Object.instance_methods
+        delegate :==, :as_json, *array_methods, to: :decorated_collection
+
+        def initialize(object, with:)
+          @object = object
+          @decorator_class = with
+        end
+
+        class << self
+          alias_method :decorate, :new
+        end
+
+        def decorated_collection
+          @decorated_collection ||= object.map { |item| decorator_class.new(item) }
+        end
+      end
+
       class Wrapper
         @cache = {}
 
         def self.wrap(decorator)
-          collection_decorator = find_draper_collection_decorator(decorator)
+          collection_decorator = find_draper_collection_decorator(decorator) || CollectionDecorator
 
-          if collection_decorator
-            name = "#{collection_decorator.name} of #{decorator} + ActiveAdmin"
-            @cache[name] ||= wrap! collection_decorator, name
-          else
-            decorator
-          end
+          name = "#{collection_decorator.name} of #{decorator} + ActiveAdmin"
+          @cache[name] ||= wrap! collection_decorator, name
         end
 
         private
```

With that in place, I believe we don't require any other changes whatsoever to support PORO decorators, and people can just plugin their own decorators, regardless of whether they are draper decorators or other types.

What do you think?

brunvez: I like the idea of allowing people to use decorators without needing to add any custom code. My only concern would be that we are forcing the use of `#new` with just one parameter and not making it very flexible. So if I wanted to initialize my presenter from many different sources, I'd like to use something like a Factory Method like:

```ruby
class PostPresenter
  def self.wrap(post, *_)
    self.for_admin(post)
  end

  def self.for_admin(post)
    # ...
  end
end
```

But after writing that code I guess what I'd rather do is an `AdminPostPresenter` class instead.

In short, I agree with the change. I'll implement it and change the docs accordingly if that's alright.

deivid-rodriguez: Right.

Maybe instead of explicitly checking for `draper`, we could use `decorator.collection_decorator_class` if it's defined. As long as we document that expected interface `.decorate(<collection>, with: <decorator_class>)`, people should be able to override that to fit their needs.

deivid-rodriguez: Yes, go ahead!

Oh, I forgot I also needed to add `, to: :object` add the end of the `Wrapper.wrap` delegations so that it uses the standard Rails delegation instead of the `draper` shortcut.

brunvez: Yeah, no worries I got that 😄 . BTW with this change, we wouldn't need to delegate or call those methods at all. Since wrapping would be done on the render level all we need are the array methods to iterate. Though I think small changes are better

deivid-rodriguez: My idea is to keep the current wrapping at the same level, and ideally just touch code paths that don't affect the `draper` integration at all. So ideally revert all changes at `lib/active_admin/resource_controller.rb`, `lib/active_admin/resource_controller/data_access.rb`, `lib/active_admin/views/pages/index.rb`, and restore the removed specs.

Maybe delaying the decoration is a good thing, but I'd like to minimize the chance of regressions because of the addition of this feature.

brunvez: Agreed, I made the proposed changes. Let me know if you have any other comments, happy to implement them.

deivid-rodriguez: Thanks for your work! I'll do another review pass tomorrow, but should be essentially ready for my side :)

docs/11-decorators.md
```diff
@@ -42,6 +37,40 @@ ActiveAdmin.register Post do
...
+  attr_reader :post
+  delegate_missing_to :post
+
+  def initialize(post)
+    @post = post
+  end
+end
+```
+
+You can also customize the way a collection is wrapped by implementing
```
deivid-rodriguez, line 55: After a second thought, and although I believe using duck typing rather that checking from draper directly is fine, I think I wouldn't document this unless there's a good use case for it. The one you mentioned in https://github.com/activeadmin/activeadmin/pull/6249#discussion_r449672211 is probably best implemented as a separate decorator and our decorator support requires decorators to be instantiated by passing the decorated object as the single parameter anyways.

lib/active_admin/resource_controller/decorators.rb
```diff
@@ -48,18 +48,45 @@ def collection_decorator
...
-            name = "#{collection_decorator.name} of #{decorator} + ActiveAdmin"
-            @cache[name] ||= wrap! collection_decorator, name
-          else
-            decorator
-          end
+          collection_decorator = find_collection_decorator(decorator)
+          name = if collection_decorator <= CollectionDecorator
+                   "#{collection_decorator.name} of #{decorator}"
+                 else
+                   "#{collection_decorator.name} of #{decorator} + ActiveAdmin"
```
deivid-rodriguez, line 87: Are the two branches here needed? I guess the `+ ActiveAdmin` part makes sense either way since we're adding an extra layer on top of the raw collection decorator.

lib/active_admin/resource_controller/decorators.rb
```diff
@@ -48,18 +48,45 @@ def collection_decorator
         end
       end
 
+      class CollectionDecorator
```
deivid-rodriguez, line 51: This class is essentially copied from Draper, so I would add a commennt on top of it about what it does, and include a permalink to the current version of this class inside Draper for proper attribution.

docs/11-decorators.md
brunvez: Agreed, it felt very weird documenting that part

brunvez:
> One final round of comments, but this is awesome and exciting! Could you also add a CHANGELOG entry? Don't forget to add links to PR and your github profile at the bottom :)

More than happy to! I hope I got all comments but lmk if anything else pops into your head

lib/active_admin/resource_controller/decorators.rb
```diff
@@ -48,18 +48,47 @@ def collection_decorator
...
+        delegate :==, :as_json, *array_methods, to: :decorated_collection
+
+        def initialize(object, with:)
+          @object = object
+          @decorator_class = with
+        end
+
+        def self.name
+          'ActiveAdmin::CollectionDecorator'
+        end
```
deivid-rodriguez, line 74: I think it's not a good practice to override `self.name` for non anonymous classes. But I agree that the description looks better with this name. So maybe we can move the class to its own `lib/active_admin/collection_decorator.rb` and nest it directly under `ActiveAdmin` so that it has this better name.

brunvez: Gladly! Is it alright if I require the new file form `lib/active_admin/resource_controller.rb` or did you have anywhere else in mind?

deivid-rodriguez: Sure!

PR #6281, opened 2020-06-09
Final ruby 2.7 warnings

deivid-rodriguez:
This PR bumps the remaining development dependencies so that activeadmin specs run with 0 warnings on ruby 2.7.


deivid-rodriguez:
This should be ready now. Finally no warnings, so I'm in peace now with moving on to something else :)

features/support/rails.rb
```diff
@@ -1,5 +1,5 @@
 require 'cucumber/rails/application'
-require 'cucumber/rails/action_controller'
+require 'cucumber/rails/action_dispatch'
```
javierjulio, line 2: If you have a moment, I was just curious why we had to change this?

deivid-rodriguez: Oh, sorry, I thought I had explained this in the commit message, I added the explanation now to the commit too.

The explanation is that the file was renamed in the `cucumber-rails` gem. We didn't need some components of `cucumber-rails`, and some were causing issues for us so [I started doing what we're doing now](https://github.com/activeadmin/activeadmin/commit/ce0df9179e829faa2e7fcd2e125f07715cbd9e1f). Since that's not the "official way" of requiring `cucumber-rails`, I don't think it makes sense to complain about anything to the gem, so I catched up with the rename.

deivid-rodriguez:
Thanks @javierjulio, what I think we can do now that Github allows you to convert draft PRs to ready PRs and back is to make a draft PR is changes are not ready, and other than that, if CI passes, assume it's ready.

features/support/rails.rb
javierjulio: That makes sense, thanks!

javierjulio:
Ok great thanks!

PR #6308, opened 2020-06-16
Fix open redirect security issue

### LINKED ISSUE #6307: Security issue: open redirect
### Did you find a bug?

There is a security issue with ActiveAdmin.

The issue is what's known as an open redirect. It's caused because we rely on the Referrer header when using `redirect_back` without validating its contents. This means an attacker could redirect a user to a malicious website.

Here's an example exploit. An attacker modifies the Referrer header to point to a malicious website that looks like your app's login page. The user thinks they've been logged out so they enter their login credentials again, thereby giving their password to the attacker. The malicious website then redirects them back to the real application, and the user is none the wiser. If the application uses OAuth, the attacker will also receive the OAuth code which gives them access to the user's Google or Facebook account (with whatever permissions your app was granted).

### Expected behavior

Expected behavior is that the gem does not redirect to whatever host the user provides.

The fix here is to either validate the Referrer header or simply not trust it at all. Headers are technically user input which is always considered untrustworthy for security purposes. Rails actually provides this validation for free, you just need to pass `allow_other_hosts: false` in the call to `redirect_back`, like this:

```
      def redirect_backwards_or_to_root
        redirect_back fallback_location: active_admin_root, allow_other_hosts: false
      end
```

### Actual behavior

The gem will redirect to any url provided by the user in the Referrer header, including other (potentially malicious) websites.

You can see an example of the redirect being used here in the `ActiveAdmin::BaseController::Authorization` module.

https://github.com/activeadmin/activeadmin/blob/3bc0b5a5472926de3c27821cdf8009177fe458b9/lib/active_admin/base_controller/authorization.rb#L112

I've monkey patched it in my codebase but it would be better to fix it in the gem. I'm happy to make a PR for this.

### How to reproduce

1. In Burp Suite, make a request to an admin route while unauthenticated and intercept the request
2. Add the header `Referer: https://example.com/foobar`
3. Forward the modified request to the app
Expected behavior: App does **not** redirect to `https://example.com/foobar`
Actual behavior: App **does** redirect to `https://example.com/foobar`

williamcodes:
This PR closes #6307 by disallowing backward redirects to other hosts (which are potentially malicious). This eliminates the [open redirect](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html) attack vector as identified by OWASP in [CWE-601](https://cwe.mitre.org/data/definitions/601.html).

### DIFF
--- a/lib/active_admin/base_controller/authorization.rb
+++ b/lib/active_admin/base_controller/authorization.rb
@@ -119,7 +119,7 @@ module ActiveAdmin
       end
 
       def redirect_backwards_or_to_root
-        redirect_back fallback_location: active_admin_root
+        redirect_back fallback_location: active_admin_root, allow_other_hosts: false
       end
 
     end
--- a/lib/active_admin/orm/active_record/comments.rb
+++ b/lib/active_admin/orm/active_record/comments.rb
@@ -64,21 +64,21 @@ ActiveAdmin.after_load do |app|
         def create
           create! do |success, failure|
             success.html do
-              redirect_back fallback_location: active_admin_root
+              redirect_back fallback_location: active_admin_root, allow_other_hosts: false
             end
             failure.html do
               flash[:error] = I18n.t 'active_admin.comments.errors.empty_text'
-              redirect_back fallback_location: active_admin_root
+              redirect_back fallback_location: active_admin_root, allow_other_hosts: false
             end
           end
 
           def destroy
             destroy! do |success, failure|
               success.html do
-                redirect_back fallback_location: active_admin_root
+                redirect_back fallback_location: active_admin_root, allow_other_hosts: false
               end
               failure.html do
-                redirect_back fallback_location: active_admin_root
+                redirect_back fallback_location: active_admin_root, allow_other_hosts: false
               end
             end
           end


deivid-rodriguez:
Hi @williamcodes!

In my opinion, if you believe this "insecure by default" default should be toggled to "secure by default", you should raise an issue at the Rails project so that they change the default in the next major version. If Rails decides to change their default, I'm happy to follow them in the next ActiveAdmin major version.

Regardless of that, this PR is clearly needing some tests since I believe you made a typo in the option name (it's `allow_other_host`, not `allow_other_hosts`), and nothing failed. Well, only the check telling you to add coverage failed :laughing:.

williamcodes:
> Hi @williamcodes!
> 
> In my opinion, if you believe this "insecure by default" default should be toggled to "secure by default", you should raise an issue at the Rails project so that they change the default in the next major version. If Rails decides to change their default, I'm happy to follow them.
> 
> Regardless of that, this PR is clearly needing some tests since I believe you made a typo in the option name (it's `allow_other_host`, not `allow_other_hosts`), and nothing failed. Well, only the check telling you to add coverage failed 😆.

Hi  @deivid-rodriguez!

It looks like [the original RFC called for the secure default](https://github.com/rails/rails/issues/33577) but they left it insecure for backward compatibility. I've [raised an issue](https://github.com/rails/rails/issues/39643) to make it secure by default in the next major release.

Perhaps in the meantime, ActiveAdmin could make it configurable the way it is for Rails. At the moment, the only way to eliminate the open redirect is by monkey patching. If I update the PR with test coverage and leave the default insecure but make it configurable, will you merge?

deivid-rodriguez:
> It looks like the original RFC called for the secure default but they left it insecure for backward compatibility. I've raised an issue to make it secure by default in the next major release.

Thanks!

> Perhaps in the meantime, ActiveAdmin could make it configurable the way it is for Rails. At the moment, the only way to eliminate the open redirect is by monkey patching. If I update the PR with test coverage and leave the default insecure but make it configurable, will you merge?

Yes, I'm definitely willing to make this configurable.

saiqulhaq:
I think we should add "depreciation warning" to the log file

williamcodes:
@deivid-rodriguez how do you want the configuration to be set? Perhaps an environment variable called `SECURE_AGAINST_OPEN_REDIRECT_ATTACKS`?

williamcodes:
@saiqulhaq how would you want to handle the warning? Perhaps something like this on boot:
```
if ENV["SECURE_AGAINST_OPEN_REDIRECT_ATTACKS"].nil?
    Deprecation.warn "SECURE_AGAINST_OPEN_REDIRECT_ATTACKS will default to true in next major release."
end
```

saiqulhaq:
> @saiqulhaq how would you want to handle the warning? Perhaps something like this on boot:
> 
> ```
> if ENV["SECURE_AGAINST_OPEN_REDIRECT_ATTACKS"].nil?
>     Deprecation.warn "SECURE_AGAINST_OPEN_REDIRECT_ATTACKS will default to true in next major release."
> end
> ```

I think we have to give link to the documentation page too, e.g.: "SECURE_AGAINST_OPEN_REDIRECT_ATTACKS will default to true in next major release. see: https://xxx "

deivid-rodriguez:
Let's way to see how the Rails conversation goes, and then we can figure this out here.

deivid-rodriguez:
This seems fixed in Rails by https://github.com/rails/rails/commit/77e5f91f420a22c972d4abc78d2369555d4f56b6, so it's no longer an issue since Rails 7. Correct?

javierjulio:
If we need to support this, I'm happy to reopen or accept a new PR. We would just need an optional config for this case and I think that would suffice.

PR #6315, opened 2020-06-18
Improve Japanese translation

rn0rno:
I feel that some Japanese translations are unnatural to use.
This PR improves that. Thanks.


deivid-rodriguez:
Thanks!!

PR #6318, opened 2020-06-21
Allow formtastic 4

deivid-rodriguez:
And test against formtastic 4.0.0.rc1.


danielweil:
How can we use this gem version? I have put the github in the Gemfile and the branch or ref but in the end, formtastic still happens to stay to same.

deivid-rodriguez:
You need to add a `formtastic` dependency like the one I added in our development Gemfile, and point to `activeadmin` master:

```ruby
gem "activeadmin", github: "activeadmin/activeadmin"
gem "formtastic", "~> 4.0.rc1"
```

And then run `bundle update activeadmin` should do the trick.

danielweil:
Thank you very much!

PR #6341, opened 2020-06-25
missing scope/all italian translation key added

fuzziness:

### DIFF
--- a/config/locales/it.yml
+++ b/config/locales/it.yml
@@ -1,4 +1,20 @@
 it:
+  activerecord:
+    models:
+      comment:
+        one: "Commento"
+        other: "Commenti"
+      active_admin/comment:
+        one: "Commento"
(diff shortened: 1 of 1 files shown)


config/locales/it.yml
```diff
@@ -42,6 +42,8 @@ it:
       current_scope: "Contesto selezionato:"
       current_filters: "Filtri attivi:"
       no_current_filters: "Nessuno"
+    scopes:
+      all: "Tutti"
```
javierjulio, line 46: @fuzziness just to maintain consistency, can you please move this `scopes:` block so it is listed before the `search_status:` one?

fuzziness: fixed & added other missing

deivid-rodriguez:
Thanks!!!

PR #6342, opened 2020-06-26
Use display name fallback if blank display name result

javierjulio:
The `display_name` helper only uses the display name fallback if no matching method is defined on a model from the configured `display_name_methods` list. The output of the display name fallback is the model human name and the primary key. So with a given `User` model and a numeric primary key, you would get `User #1` as an example output.

If the model implements a method from that `display_name_methods` list but it returns a blank value, **this has a side effect where an association link is generated without text so it's not seen or clickable**. I think it would help to use our display name fallback in this specific case since it is a good default, rather than be all or nothing. The `display_name_methods` list is a global configuration that cannot be modified per resource.

When working on this change, the test case I modified demonstrated the exact behavior I ran into. The anchor (link) HTML is generated but without a value so a user wouldn't see anything in the browser. While the test shows this is expected behavior, I think this would be an improvement for this specific scenario.

This change would not have any effect on cases where there is a custom display name because as long as it returns a present value, it will always be used. This is a small improvement to rely on a good default, our display name fallback, when there is no present value.

Let me know your thoughts and any suggestions for changes here.

### DIFF
--- a/lib/active_admin/view_helpers/display_helper.rb
+++ b/lib/active_admin/view_helpers/display_helper.rb
@@ -24,7 +24,14 @@ module ActiveAdmin
       # Attempts to call any known display name methods on the resource.
       # See the setting in `application.rb` for the list of methods and their priority.
       def display_name(resource)
-        ERB::Util.html_escape(render_in_context(resource, display_name_method_for(resource))) unless resource.nil?
+        unless resource.nil?
+          result = render_in_context(resource, display_name_method_for(resource))
+          if result.to_s&.strip&.present?
+            ERB::Util.html_escape(result)
+          else
+            ERB::Util.html_escape(render_in_context(resource, DISPLAY_NAME_FALLBACK))
+          end
+        end
       end
 
       # Looks up and caches the first available display name method.
--- a/lib/active_admin/views/pages/show.rb
+++ b/lib/active_admin/views/pages/show.rb
@@ -40,13 +40,7 @@ module ActiveAdmin
         protected
 
         def default_title
-          title = display_name(resource)
-
-          if title.blank?
-            title = "#{active_admin_config.resource_label} ##{resource.id}"
-          end
-
-          title
+          display_name(resource)
         end
 
         module DefaultMainContent
--- a/spec/unit/view_helpers/display_helper_spec.rb
+++ b/spec/unit/view_helpers/display_helper_spec.rb
@@ -192,12 +192,12 @@ RSpec.describe ActiveAdmin::ViewHelpers::DisplayHelper do
       expect(value).to eq "right"
     end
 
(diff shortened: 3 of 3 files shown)


deivid-rodriguez:
As discussed privately, my concern is that this introduces the inconsistency of potentially showing the value of different model attributes in the same column. So, in case it is intended that the value of the chosen display attribute is empty, then this would be a strict improvement. However, in the case where this is an unintended data problem, the current behaviour would make it easier to identify and fix, whereas the new behaviour would make a model whose name is `User #1` indistinguishable from a model with empty name.

So, although I don't currently have a great idea, I'd like us to at least apply some visual hint in the case where this "extra fallback" has been applied. I'll try to think about something more concrete.

javierjulio:
I think a visual hint is a great idea. Some other suggestions could be to also log a warning and/or use a configuration to opt-in to this change.

sunny:
Perhaps `(User #1)`?

deivid-rodriguez:
I don't think there's a "correct" solution involving changing only the wording of the name, because users should be free to name models whatever they like without causing conflicts with this, including the `(User #1)` name.

I mean, we could probably use some virtually impossible name like `THIS MODEL HAS EMPTY NAME SO WE'RE SHOWING THIS DUMMY TEXT SO THAT YOU CAN STILL CLICK THIS LINK AND NAVIGATE TO THE MODEL PAGE`. But a visual hint to explain this feels better, maybe a clickable "info" icon or something.

sunny:
A quick fix could also be to replace blank values by a non-breaking space. This would, at least, make the link appear and be clickable. This would help show visually that the association is present but the name is empty.

E.g.:

image

This would not be ideal since the link is very small, though.

deivid-rodriguez:
Well, that's definitely better than what we have now, and it doesn't bring any ambiguity. Thoughts @javierjulio?

javierjulio:
Thanks for the suggestion! Hmm I don't think so because by implementing that then we are actually giving a present value of `&nbsp;` vs just an empty space as now. I don't think others would even find the link with such a small, invisible click area. If we do that, I'd prefer to put effort towards hopefully a suitable fix.

Just to reiterate, there are two key items here. One is that the display name fallback is a good default overall but is not used in all cases. Second is that by just adding a column on a model, that does not require a value but just so happens to be in the `config.display_name_methods` list, causes this blank display name issue, even if the new field is not anyway related for the admin. I'm torn really because the fallback works great so to me it makes sense to utilize that somehow since in our admin usage, we are used to the default display output.

deivid-rodriguez:
@javierjulio Sorry for dropping the ball on this, I'll have another look as soon as possible.

javierjulio:
@deivid-rodriguez haha I was just updating to keep it current with CI/lint changes since then, so no worries, thanks!

PR #6358, opened 2020-07-09
Steps to be able to auto manage our changelog

deivid-rodriguez:
Changelog is a pain point for contributing to activeadmin, since it requires adding entries in a specific format, and it's not always straightforward to get it right. Also, as with almost every other repo, the changelog is the number 1 cause of merge conflicts, so it complicates rebasing.

Also, it's also a pain for maintainers because since we sometimes forget to ask contributors to add changelog entries for user visible changes, we need to go through all changes before each release and add the missing stuff.

In future PRs I plan to enforce a workflow were changelog entries are automatically inferred from PR titles & labels, so that contributors don't need to add changelog entries at all, and maintainers only need to add the proper tags and titles to them. Other than that, the changelog is completely automanaged.

This PR is a first step towards this, adding `bin/rake changelog:add_references` that given an unreleased section in the changelog, it automatically edits it to add the proper user references and pull request references at the proper place.

I'm not documenting this task because the idea is not that contributors or ourselves run it, but that our CI scripts run it after PRs are merged and automatically update the changelog.


deivid-rodriguez:
@javierjulio Maybe this PR if lacking a bit of context without further work. Do you prefer that I build the full proposal for auto managing our changelog directly on top of this PR?

javierjulio:
@deivid-rodriguez no, I think this looks great. 👍🏻 I wasn't sure if it was ready though. I don't mind getting this in as is or if you prefer to do one large PR. Just let me know and I can review.

If I follow, this just adds a rake task that can format and sort the references and then be able to add a new version section, right?

The thinking is CI would run the add_references task and auto commit changes to the changelog file on merge?

While we automate more which is good, we need to be mindful how PR titles are written. I can't help but shake the feeling after working on a change, more is learned and a better description comes up. Perhaps a note in the PR template that the PR title is used as the changelog entry?

deivid-rodriguez:
> @deivid-rodriguez no, I think this looks great. 👍🏻 I wasn't sure if it was ready though. I don't mind getting this in as is or if you prefer to do one large PR. Just let me know and I can review.

Go ahead and review it!

> If I follow, this just adds a rake task that can format and sort the references and then be able to add a new version section, right?

The rake task, given an unreleased section without references (just with `[#xxxx]` and `[@yyyy]`), it will add the proper references (`[#xxxx]: https://...`, `[@yyyy]: https://...`) at the bottom of the changelog. The idea is to use [release-drafter](https://github.com/release-drafter/release-drafter) to upload a draft release after every PR. Then write the body of the draft release to the unreleased section of the changelog, then add the proper references, and then commit the updated changelog.

> While we automate more which is good, we need to be mindful how PR titles are written. I can't help but shake the feeling after working on a change, more is learned and a better description comes up. Perhaps a note in the PR template that the PR title is used as the changelog entry?

My idea is that us maintainers reword the PR title as needed if necessary right before merging.

javierjulio:
Nice, release-drafter looks really cool! We'll have to update the PR template anyway to remove the changelog line but might as well replace it with that note to keep the PR title current/accurate. That can be done separately. This all looks good to me, so let's get it in and continue. Thanks David!

deivid-rodriguez:
Awesome! Tomorrow I'll propose the final solution including the updates to the PR template. Thanks so much Javier!! :heart:

javierjulio:
No problem, happy to help. Thank you! ❤️

PR #6363, opened 2020-07-14
fix: Streaming CSV export.

schwern:
Streaming CSV exports was not working either in development (with `config.disable_streaming_in = []`)
nor on production Heroku. Adding a Last-Modified header made it work as is [recommended elsewhere](https://medium.com/table-xi/stream-csv-files-in-rails-because-you-can-46c212159ab7).

### DIFF
--- a/lib/active_admin/resource_controller/streaming.rb
+++ b/lib/active_admin/resource_controller/streaming.rb
@@ -20,6 +20,7 @@ module ActiveAdmin
       def stream_resource(&block)
         headers['X-Accel-Buffering'] = 'no'
         headers['Cache-Control'] = 'no-cache'
+        headers["Last-Modified"] = Time.current.httpdate
 
         if ActiveAdmin.application.disable_streaming_in.include? Rails.env
           self.response_body = block['']


lib/active_admin/resource_controller/streaming.rb
```diff
@@ -20,6 +20,7 @@ def index
       def stream_resource(&block)
         headers['X-Accel-Buffering'] = 'no'
         headers['Cache-Control'] = 'no-cache'
+        headers["Last-Modified"] = Time.zone.now.ctime.to_s
```
sunny, line 23: Minor style suggestions that, I think, should be equivalent:

```suggestion
        headers['Last-Modified'] = Time.current.ctime
```

schwern: Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified neither are quite correct. The format is slightly wrong and it should be in UTC. `Time.now.httpdate`. I will fix that.

Fivell:
@deivid-rodriguez  I also noticed this when we updated rack, here is the issue https://github.com/rack/rack/issues/1628

currently we lock rack to older version to keep streaming export works, according to issue it is related wirh  etag middleware, and 'Last-Modified' header does fix this

Fivell (approved):

deivid-rodriguez:
Hi @Fivell and @schwern! Fix looks good. It'd  be nice to add a nice explaination of the issue to the commit message and the PR description. This commit message is really nice: https://github.com/alphagov/e-petitions/commit/46413a7127859e2d2d38a5e6f55ce3ec4504781f.

Something I'd also like to confirm is why our cucumber features didn't break when we started using rack 2.2 in CI. Are streaming CSV exports not covered? If that's the case, we should get that covered.

Fivell:
@deivid-rodriguez  csv exports are covered, more than that user can receive file when export is completed, the thing is that streaming mechanism is not working and from user point of view it looks like download starting when all required data already exported, so for big files user waits big amount of time and nothing happening

deivid-rodriguez:
I see @Fivell that makes sense, thanks.

From an end user point of view, this sounds really hard to test, we would need to start transferring a big file, and make sure that the download progress dialog or something like that appears immediately? Seems too much.

So let's do this as it is. I'll rebase, reword this, and merge it later today!

Fivell:
@deivid-rodriguez  thanks

deivid-rodriguez:
@schwern I don't think I have permissions to push to this branch, so could you:

* Rebase.

* Change the commit message to something like
  > 
  >    Fix streaming CSV export
  >    
  >    Streaming was not working either in development (disable_streaming off)
  >    nor on production Heroku. Adding a Last-Modified header made it work.
  >    
  >    The problem would only manifest when using rack 2.2 or higher, and was
  >    caused by https://github.com/rack/rack/pull/1416, where the `Rack::ETag`
  >    middleware changed it's behaviour to buffer the response and calculate
  >    an ETag digest even though the `Cache-Control: no-cache` header is set.
  >    This breaks the CSV streaming responses so to avoid that we use the hack
  >    outlined in rack/rack#1619 and set a Last-Modified header which triggers
  >    the `skip_caching?` condition in the `Rack::ETag` middleware.
  >   
  >    The credit for this commit message goes to
  >    https://github.com/alphagov/e-petitions/commit/46413a7127859e2d2d38a5e6f55ce3ec4504781f.


* Add a small unit test like:

  ```diff
  diff --git a/features/step_definitions/format_steps.rb b/features/step_definitions/format_steps.rb
   index 21b52609c..7649d2083 100644
  --- a/features/step_definitions/format_steps.rb
  +++ b/features/step_definitions/format_steps.rb
  @@ -22,11 +22,13 @@ end
   # Check first rows of the displayed CSV.
   Then /^I should download a CSV file with "([^"]*)" separator for "([^"]*)" containing:$/ do |sep, resource_name, table|
     body = page.driver.response.body
  -  content_type_header, content_disposition_header = %w[Content-Type Content-Disposition].map do |header_name|
  +  content_type_header, content_disposition_header, last_modified_header = %w[Content-Type Content-Disposition Last-Modified].map do |header_name|
       page.response_headers[header_name]
     end
     expect(content_type_header).to eq "text/csv; charset=utf-8"
     expect(content_disposition_header).to match /\Aattachment; filename=".+?\.csv"\z/
  +  expect(last_modified_header).to_not be_nil
  +  expect(Date.strptime(last_modified_header, "%a, %d %b %Y %H:%M:%S GMT")).to be_a(Date)
 
     csv = CSV.parse(body, col_sep: sep)      table.raw.each_with_index do |expected_row, row_index|
    ```

deivid-rodriguez:
Closing in favour of #6451.

PR #6368, opened 2020-07-21
Fix several Arabic translations

### LINKED ISSUE #6367: Typo in Arabic translation for login submit text
There's a typo in Arabic translation for login submit button text here:

https://github.com/activeadmin/activeadmin/blob/5a6518f63615bcc8ad3df8068e91bc1111e67840/config/locales/ar.yml#L105

The correct translation is 'تسجيل'.

mshalaby:
This PR fixes a typo in submit button text

Closes #6367

### DIFF
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -97,12 +97,12 @@ ar:
       password:
         title: "كلمة المرور"
       sign_up:
-        title: "تسجيل الدخول"
-        submit: "تسجيل الدخول"
-      login:
-        title: "دخول"
-        remember_me: "تذكرني"
+        title: "تسجيل"
         submit: "تسجيل"
+      login:
+        title: "تسجيل الدخول"
+        remember_me: "تذكرني"
+        submit: "تسجيل الدخول"
       reset_password:
         title: "هل نسيت كلمة المرور؟"
         submit: "استرجاع كلمة المرور"
@@ -117,7 +117,7 @@ ar:
         submit: "إعادة ارسال تعليمات التأكيد"
       links:
         sign_up: "التسجيل"
-        sign_in: "دخول"
+        sign_in: "تسجيل الدخول"
         forgot_your_password: "هل نسيت كلمة المرور؟"
         sign_in_with_omniauth_provider: "تسجيل الدخول بـ %{provider}"
         resend_unlock_instructions: "إعادة إرسال تعليمات تنشيط الحساب"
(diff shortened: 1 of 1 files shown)


config/locales/ar.yml
```diff
@@ -102,7 +102,7 @@ ar:
       login:
         title: "دخول"
         remember_me: "تذكرني"
-        submit: "تسحيل"
+        submit: "تسجيل"
```
javierjulio, line 105: When I translated the original and the new version for this change, both return "Sign Up" as the English translation, yet its for the **login** locale.

I noticed the correction you made is the same as the second word in the `sign_up.submit` locale. Translating that full value, "تسجيل الدخول" actually returns "Sign In". 😳

mshalaby: @javierjulio I appreciate your taking the time to check this.

Regarding the original word "تسحيل". It's not an actual word in Arabic that's why Google translate shows results for the corrected word instead as indicated by the note on the left saying "Showing translation for تسجيل".

As for the issue you spotted. You're right. I was focused on fixing the typo and completely missed noticing it.
It seems the locales for "sign up" and "sign in" have been misplaced and nobody noticed before since Admin users don't sign up. Also using 'تسجيل' (record / register) in conjunction with the title "دخول" (login / entry) would implicitly mean "تسجيل الدخول" (i.e. to sign in), just more concise.

So, to correct this I suggest we use:
- `sign_up.title`: "تسجيل" and `sign_up.submit`: "تسجيل"
- `login.title`: "تسجيل الدخول" and `login.submit`: "تسجيل الدخول"

I'll update the PR to reflect that if it's fine with you.

javierjulio: @mshalaby you're very welcome. I love seeing these locale contributions come in. We really appreciate you and others contributing them!

You don't have too but the suggested change looks great to me as long as you are ok with updating, let's do it. We can get that change in. Thank you!

mshalaby: Thanks @javierjulio!

I've pushed the fixes we mentioned.

While doing so I noticed another unrelated issue.
The word order is reversed for `new_model` translation:
https://github.com/activeadmin/activeadmin/blob/7b45fbf429542ebd5ddb137a499cedc21ce20773/config/locales/ar.yml#L11
The correct order should be:
```yml
new_model: "%{model} جديد"
```
Should I create a separate issue for that? or just push another commit to fix it here?

javierjulio: You're welcome! Up to you, either is fine for me since it's a small change. Thank you! ❤️

mshalaby: Pushed the fix. You're very welcome @javierjulio :pray:

Issue #6206, opened 2020-04-29
ActiveAdmin crashes on start if there's no app/assets directory


Ecco:
### Problem

1. Create a new Rails 6 app using Webpacker. For instance, `rails new FooBar --skip-sprockets --webpack=vue`
2. Remove the `app/assets` directory
3. Run `rails server`. It works!
4. Add activeadmin to the Gemfile and run `bundle install`
5. (Optional) Edit `config/initializers/active_admin.rb` to use webpack
5. Run `rails server`. It crashes with an error.

### Expected behavior

It shouldn't crash. The "app/assets" directory only makes sense if the app uses sprockets, and ActiveAdmin supports Webpacker as an alternative :)

### More info

Here's the error message
```
/Users/ecco/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/sprockets-rails-3.2.1/lib/sprockets/railtie.rb:105:in `block in <class:Railtie>': Expected to find a manifest file in `app/assets/config/manifest.js` (Sprockets::Railtie::ManifestNeededError)
But did not, please create this file and use it to link any assets that need
to be rendered by your app:

Example:
  //= link_tree ../images
  //= link_directory ../javascripts .js
  //= link_directory ../stylesheets .css
and restart your server
```

Ecco:
This might be related to #6201.

Ecco:
#### My investigation

1. This is due to unconditionally [requiring sassc-rails](https://github.com/activeadmin/activeadmin/blob/3756c83ab720b0474bb73c477317d55b7948c866/lib/active_admin.rb#L11), which in turn will [require `sprockets/railtie`](https://github.com/sass/sassc-rails/blob/8d0462d54b5b5dd84cb1df31823c3afaebb64534/lib/sassc/rails/railtie.rb#L4).
2. Even if this gem isn't loaded, the engine references `config.assets` unconditionnaly too [here](https://github.com/activeadmin/activeadmin/blob/3756c83ab720b0474bb73c477317d55b7948c866/lib/active_admin/engine.rb#L10) and [there](https://github.com/activeadmin/activeadmin/blob/3756c83ab720b0474bb73c477317d55b7948c866/lib/active_admin/engine.rb#L13).

Commenting those 3 lines fixes the problem, but of course isn't the right solution 😄 

#### Note regarding point #1025 

[...]

PR #6370, opened 2020-07-22
Issue 6206

hcatlin:
As reported in #6206, a Rails application that doesn't have an "app/assets" folder and is using Webpacker entirely without Sprockets will report an error on server startup. 

The reason for the error is that ActiveAdmin was including Sprockets and sassc-rails into the runtime and causing Sprockets to get initialized without the user intending this.

Here, several changes were needed to stop the automatic loading and requirement for sassc as a Ruby Gem.

This is in two commits, the first showing the failing test changes, and the second addresses the issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@
 
 * Fix filter_columns_for_large_association and filter_method_for_large_association examples. [#6232] by [@ndbroadbent]
 
+### Bug Fixes
+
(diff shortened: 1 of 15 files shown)


deivid-rodriguez:
@hcatlin I'm sorry, I just addressed the exact same issue you're addressing here in #6584. I missed your PR, but I'll amend my commit to add your attribution since it's essentially the same solution. If you want to double check #6584 you're more than welcome. Also apologies for taking this long to get to this.

PR #6371, opened 2020-07-23
Enhance release tasks to automatically generate our changelog

deivid-rodriguez:
Initially I was expecting to automatically update the changelog after every PR, so it's always up to date. I also planned to use [release-drafter](https://github.com/release-drafter/release-drafter) to publish release drafts after every PR to GitHub.

But in the end I realized that all that is a bit overkill, and hard to get right.

My intention was to ease the creation of the changelog for every release, avoiding having to manually triage every merged PR. So, what I did was to enhance our release tasks to including autogenerating the changelog too, using the criteria we defined in https://github.com/activeadmin/activeadmin/pull/6358 (i.e., include merged PRs with their titles, as long as they have a proper label).

After this PR, I'll use the new release tasks to release 2.8.0.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,14 +12,11 @@ Before submitting your PR make sure that:
 * Your feature branch is up-to-date with `master` (if not - rebase it), and does not include merge commits.
 * Related commits are squashed together, and unrelated commits are splitted apart.
 * Your PR includes a regression test if it fixes a bug.
-* You add an entry to the [Changelog] if the new code introduces user-observable changes. See [changelog entry format].
 
 Before expecting a review for your PR make sure that all github checks are passing, but feel free to ask for early feedback if you need guidance.
 
 [CONTRIBUTING file]: https://github.com/activeadmin/activeadmin/blob/master/CONTRIBUTING.md
 [good commit messages]: https://chris.beams.io/posts/git-commit/
 [includes keywords to automatically close issues]: https://help.github.com/en/articles/closing-issues-using-keywords
-[Changelog]: https://github.com/activeadmin/activeadmin/blob/master/CHANGELOG.md
-[changelog entry format]: https://github.com/activeadmin/activeadmin/blob/master/CONTRIBUTING.md#add-a-changelog-entry
 
 -->
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -174,12 +174,12 @@ met.
 
 Maintainers need to do the following to push out a release:
 
-* Make sure all pull requests are in and that changelog is current
 * Switch to the master branch and make sure it's up to date.
 * Make sure you have [chandler] properly configured. Chandler is used to
   automatically submit github release notes from the changelog right after
   pushing the gem to rubygems.
 * Run one of `bin/rake release:prepare_{prerelease,prepatch,patch,preminor,minor,premajor,major}`, push the result and create a PR.
+* Review and merge the PR. The generated changelog in the PR should include all user visible changes you intend to ship.
 * Run `bin/rake release` from the target branch once the PR is merged.
 
 [chandler]: https://github.com/mattbrictson/chandler#2-configure-credentials
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,22 +115,6 @@ checks as the rest of the project. `bin/rake lint` will give you feedback in
 this regard. You can check & fix style issues by running each linter
 individually. Run `bin/rake -T lint` to see the available linters.
 
-### Add a changelog entry
-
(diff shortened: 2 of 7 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,14 +12,11 @@ Before submitting your PR make sure that:
 * Your feature branch is up-to-date with `master` (if not - rebase it), and does not include merge commits.
 * Related commits are squashed together, and unrelated commits are splitted apart.
 * Your PR includes a regression test if it fixes a bug.
-* You add an entry to the [Changelog] if the new code introduces user-observable changes. See [changelog entry format].
```
javierjulio, line 15: It would help to inform contributors that the PR title will be used in the changelog since its auto generated now. Perhaps a new line here to point that out or add on to existing/related line?

If I understand right, the PR needs to have an appropriate label now which is great. Perhaps a line to have them pick one of the `type: ...` labels they think is a best fit, at least to start/requirement to create PR?

CONTRIBUTING.md
```diff
@@ -115,22 +115,6 @@ checks as the rest of the project. `bin/rake lint` will give you feedback in
...
-
-* One line per PR describing your fix or enhancement.
-* Entries end with a dot, followed by "[#pr-number] by [@github-username]".
-* Entries are added under the "Unreleased" section at the top of the file, under
-  the "Bug Fixes" or "Enhancements" subsection.
-* References to github usernames and pull requests use [shortcut reference
-  links].
-* Your github username reference definition is included in the correct
-  alphabetical position at the bottom of the file.
-
```
javierjulio, line 133: In the **Shipping a release** section of this document, does anything there need to be updated to account for the overall changes in this PR? Since I haven't shipped a release before I wasn't sure if something changed or additional we'd need to document there for myself or other maintainers.

javierjulio:
Not to be cheeky but you need to pick a `type: ...` label for this PR! 😏 That made me realize though, should only user facing changes (in other words goes into the changelog) get a label? And thus only certain PRs get included in the changelog rather than all? I've seen this done differently. I didn't want to assume that's what you were going for here. I'd be fine with either.

.github/PULL_REQUEST_TEMPLATE.md
deivid-rodriguez: I was thinking of adding something about this, but:

* Regarding the PR title, we already have "The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.". I think that guideline might be good enough? Also, we can always change the title ourselves before merge, after merge, or even locally after running the changelog generation tasks. So it's not as crucial as I expected it to be when I planned to generate commits to update the changelog automatically. That said, I don't mind adding something to mention this, in case it could motivate people to write more descriptive titles.

* Regarding labels, I don't think contributors can assign labels to PRs. You mean that they mention it in the PR description or something like that?

CONTRIBUTING.md
deivid-rodriguez: The new changelog generation logic was added to the `bin/rake release:prepare_<minor|major|patch>` tasks, so those tasks now also automatically generate the changelog for the release. I tweaked bullet points in that section to catch up with that :+1:.

deivid-rodriguez:
> Not to be cheeky but you need to pick a type: ... label for this PR! smirk That made me realize though, should only user facing changes (in other words goes into the changelog) get a label? And thus only certain PRs get included in the changelog rather than all? I've seen this done differently. I didn't want to assume that's what you were going for here. I'd be fine with either.

@javierjulio This is exactly right! My personal best practice is to only include user visible changes in the changelog, since only those are relevant for end users. Since you can't just blindly include all merged PRs in the changelog if you follow this practice, it requires manual triaging of all merged PRs before every release, which is quite a bit of work. That's what this new workflow intends to solve: do this triaging at PR creation/merge time, and no more work is needed when releasing. So yeah, only PRs intented to be included in the changelog need to get a label, for the others there's no need to bother.

.github/PULL_REQUEST_TEMPLATE.md
javierjulio: I didn't know contributors could not set a label. I tried on other repos I've contributed on and it doesn't allow me. So the idea for a release is you would have to go through each PR that was merged since the last release and add the appropriate label if needed?

javierjulio: Rather based on your other comment, when we review PRs we add the appropriate label then?

deivid-rodriguez: Yes, only if they include user-visible changes. Otherwise no need to label. I'd like to make this labelling mental process part of the review process. If we do that, we don't need to worry about the changelog anymore.

javierjulio: Ok sounds good thanks! I hope we remember. 😆

PR #6385, opened 2020-08-03
Send `active_admin/print.scss` to the proper webpacker directory

danielricecodes:
1. Slightly tweaks the webpacker generator.  It was sending `print.scss` into the wrong directory.
2. Adds a spec for this file where there was none.
3. Slightly DRY's up the spec, too.


danielricecodes:
I'm going to close this since the tests are failing.  Webpacker's directory structure does not make intuitive sense.

PR #6391, opened 2020-08-11
Update mdl gem so we can update kramdown for important security fix

javierjulio:
We [received an important security alert for `kramdown` but Dependabot was not able to create an automated update](https://github.com/activeadmin/activeadmin/network/alert/Gemfile.lock/kramdown/open) because the `mdl` gem version we use depends on a much older version of `kramdown`. The change here is a result of just running `bundle update mdl` at the project root. Other gems were updated in the process. The `kramdown` version automatically updated to the latest version 2.3.0 which has the important security fix.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,7 +56,7 @@ group :lint do
   gem 'rubocop', '0.89.1'
   gem 'rubocop-rspec', '~> 1.30'
   gem 'rubocop-rails', '~> 2.3'
-  gem 'mdl', '0.6.0'
+  gem 'mdl', '0.11.0'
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Nice, thanks! :purple_heart: 

I'm always bothered with dependabot creating these security alerts for development dependencies, but it's always good to update our dependencies :)

javierjulio:
You're welcome! It was good to do this as it was a part I'm not that familiar with. After this maybe I'll do a pass on the development gems to see what's very old and could use an update.

PR #6393, opened 2020-08-14
Fix comment layout so regardless of size, each is aligned and spaced evenly

Ivanov-Anton:
height between comments is too small

before
Screenshot 2020-08-14 at 16 23 05

after
Screenshot 2020-08-14 at 16 24 13

### DIFF
--- a/app/assets/stylesheets/active_admin/components/_comments.scss
+++ b/app/assets/stylesheets/active_admin/components/_comments.scss
@@ -2,9 +2,9 @@
 .comments {
 
   .active_admin_comment {
-    clear: both;
+    @include clearfix;
     margin-top: 10px;
-    margin-bottom: 40px;
+    margin-bottom: 20px;
     max-width: 700px;
 
     .active_admin_comment_meta {


Fivell:
@javierjulio hello, also noticed it, what do you think

javierjulio:
@Fivell this won't fix the issue because its dependent on how long the author name is. So for this specific case, an author name spanning 2 lines, the update will address it but not if its more. Ideally, none of this should be using floats for layout. To address the issue regardless of size and while maintaining float layout, we should apply our clearfix mixin on the `.active_admin_comment` class and lower the margin bottom.

app/assets/stylesheets/active_admin/components/_comments.scss
```diff
@@ -4,7 +4,7 @@
   .active_admin_comment {
     clear: both;
```
javierjulio, line 5: This change should fix the issue more reliably across different sized comments:
```suggestion
    @include clearfix;
```

app/assets/stylesheets/active_admin/components/_comments.scss
```diff
@@ -4,7 +4,7 @@
...
     margin-top: 10px;
-    margin-bottom: 40px;
+    margin-bottom: 50px;
```
javierjulio, line 7: We can lower this value to just be a nice, consistent spacing between comment containers.
```suggestion
    margin-bottom: 20px;
```
Since the containers were not properly cleared beforehand, this wasn't having the right effect as shown by the screenshots shared.

javierjulio:
To help demonstrate what I mean by the issue being dependent on the number of text lines, here is a screenshot with the original style changes proposed but with an additional (3rd) line of text for the author name display:
Screen Shot 2020-08-17 at 6 30 54 PM

and then afterwards by replacing that with the clearfix mixin and lowering the bottom margin so its applied evenly:
Screen Shot 2020-08-17 at 6 31 49 PM

Fivell:
@javierjulio  thanks !

deivid-rodriguez (approved):
Awesome, thanks so much!

PR #6401, opened 2020-08-24
Get `base.js` file in sync with its sources

deivid-rodriguez:
Also, fix the root cause of it falling out of sync, namely, some lint specs not being run.


deivid-rodriguez:
I don't think so, the change was only a refactoring, and even if it was a regression, it's still unreleased. So no labels :+1:.

PR #6402, opened 2020-08-24
Configure and enforce eslint

deivid-rodriguez:
When we first added our js bundle in order to migrate to webpack, `eslint` and `eslint-plugin-import` were added as development dependencies. However, they were never used or configured.

In this PR, I configure them to keep our current code style as much as possible.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,6 @@
+app/assets/**/*.js
+vendor/**/*.js
+lib/generators/**/*.js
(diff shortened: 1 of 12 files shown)


deivid-rodriguez:
This PR is built on top of #6401 because it was introducing changes in the same area as that PR. Once #6401 is merged, I can rebase this.

deivid-rodriguez:
This one should be ready as well @javierjulio!

PR #6403, opened 2020-08-24
Replace `rollup-plugin-uglify` with `rollup-plugin-terser`

deivid-rodriguez:
Because `rollup-plugin-uglify` depends on a vulnerable version of `serialize-javascript` (https://github.com/advisories/GHSA-hxcc-f52p-wc94).

This closes a github security alert about our JS dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "rollup-plugin-babel": "^4.3.3",
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-node-resolve": "^5.2.0",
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
This PR is built on top of #6401 because it requires a rebuild of the js sources, and would introduce unrelated changes otherwise. Once #6401 is merged, I can rebase this.

deivid-rodriguez:
This one should be ready. No user-facing changes since the generated JS is exactly the same.

PR #6406, opened 2020-08-25
Bump simplecov to 0.19.0

deivid-rodriguez:
Supersedes #6396.

It also bumps code climate's test-reporter to the latest version.


deivid-rodriguez:
Simplecov changed its internal format a few releases ago, and that broke code climate's integration. This will be fixed in future releases of simplecov & codeclimate's test reporter, but for now, using a small script to convert to the old format does the trick just fine.

PR #6416, opened 2020-08-31
removed unused method

Fivell:
ActiveAdmin::Filters::FormtasticAddons#ransacker? is not needed


Fivell:
just testing where it can be used

deivid-rodriguez:
CI shoudn't get in the middle if you rebase now.

PR #6417, opened 2020-08-31
Automatically create a release branch from release scripts

deivid-rodriguez:
Current scripts tell you to start off the default branch, and release scripts don't do any branch switching, so they would commit to the default branch and tell you to push after that. We don't want that, but push to a non default branch and create a PR from it.

### DIFF
--- a/tasks/release_manager.rb
+++ b/tasks/release_manager.rb
@@ -129,6 +129,8 @@ class ReleaseManager
   end
 
   def prepare_version(version)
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I was preparing the next release and noticed that our release scripts commit directly to the default branch. They should create a branch instead, so this PR does that.

deivid-rodriguez:
Thanks, and no need to apologize at all!

PR #6419, opened 2020-08-31
Pin base node-browsers images in CI

deivid-rodriguez:
CircleCI changed something that broke all of our builds, so I'm pinning ourselves to the last working version.

We should probably look into migrating to Github Actions, CircleCI has deprecated their base docker images, and there seems to be no jruby alternative, and no `browser` variants provided in their new images.


deivid-rodriguez:
Back to green now.

PR #6422, opened 2020-09-01
Fix input string filter no rendering dropdown input when its column name ends with a ransack predicate 

### LINKED ISSUE #6413: filter :src_name_in, as: :string I dont have drop down for selecting “Contains”, “Equals” ...
Because my field ends with "in" my filter does not work correct. My filter is now displayed as field with the text type without drop down.

I tried this option: `filter :src_name_in, as: :string, filters: [:starts_with, :equals, :contains, :ends_with] `
I tried this option: `filter :src_name_in, as: :string`
I tried this option: `filter :src_name_in`
I tried this option: `filter :src_name_in, filters: [:contains, :equals, :starts_with, :ends_with]`

Fivell:
fixes  #6413

### DIFF
--- a/features/sti_resource.feature
+++ b/features/sti_resource.feature
@@ -7,10 +7,10 @@ Feature: STI Resource
     And a configuration of:
     """
       ActiveAdmin.register Publisher do
-        permit_params :first_name, :last_name, :username, :age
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
I'll review this ASAP.

deivid-rodriguez (approved):
Looks great, thanks for the fix!

PR #6443, opened 2020-09-10
Check if cuprite works well with latest chrome

deivid-rodriguez:
It seems like [cuprite](https://github.com/rubycdp/cuprite) works well with the latest chrome, so we can keep our docker images up to date. I still kept them locked because I really dislike third party changes breaking our CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -256,7 +256,7 @@ jobs:
 
   ruby25rails52:
     docker:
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
It seems like [cuprite](https://github.com/rubycdp/cuprite) works well with the latest chrome, so we can keep our docker images up to date. I still kept them locked because I really dislike third party changes breaking our CI.

deivid-rodriguez:
@javierjulio This one should be ready. Just making sure specs work with the latest version of chrome. Apparently `apparition` has issues with it.

javierjulio (approved):
Ok sounds good, thanks!

PR #6451, opened 2020-09-15
Fix streaming CSV export

deivid-rodriguez:
Streaming was not working either in development (disable_streaming off) nor on production Heroku. Adding a Last-Modified header made it work.

The problem would only manifest when using rack 2.2 or higher, and was caused by https://github.com/rack/rack/pull/1416, where the `Rack::ETag` middleware changed it's behaviour to buffer the response and calculate an ETag digest even though the `Cache-Control: no-cache` header is set. This breaks the CSV streaming responses so to avoid that we use the hack outlined in rack/rack#1619 and set a Last-Modified header which triggers the `skip_caching?` condition in the `Rack::ETag` middleware.

This credit of this commit message goes to https://github.com/alphagov/e-petitions/commit/46413a7127859e2d2d38a5e6f55ce3ec4504781f.

This PR supersedes #6363.

### DIFF
--- a/features/step_definitions/format_steps.rb
+++ b/features/step_definitions/format_steps.rb
@@ -22,11 +22,13 @@ end
 # Check first rows of the displayed CSV.
 Then /^I should download a CSV file with "([^"]*)" separator for "([^"]*)" containing:$/ do |sep, resource_name, table|
   body = page.driver.response.body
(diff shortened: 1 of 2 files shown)


javierjulio (approved):
Thank you!

PR #6460, opened 2020-09-18
Fix `permitted_param` generation for `belongs_to` when `:param` is used

deivid-rodriguez:
When playing around with our sample application, I noticed that you couldn't create post taggings due to this bug.


javierjulio (approved):
Looks good, thanks! ❤️

PR #6461, opened 2020-09-19
Support id_column and index_column in table_for

c960657:
`id_column` and `index_column` are not only useful in index tables but also in `table_for`. Move these two methods to the parent case, so they are available for any table.

I renamed the `i18n` option for `table_for` to `resource_class` (with fallback to `i18n` for BC), because with this change, the specified model is used not only for i18n purposes but also for determining the primary key.


docs/12-arbre-components.md
```diff
@@ -149,14 +151,15 @@ The `column` method can take a title as its first argument and data
 (`:your_method`) as its second (or first if no title provided). Column also
 takes a block.
 
-### Internationalization
+The `index_column` nethoe takes an optional offset parameter to allow a developer
+to set the starting number for the index (default is 1).
```
deivid-rodriguez, line 155: The `index_column` method is already documented in the `index_as_table.rb` source file, which generates https://activeadmin.info/3-index-pages/index-as-table.html. Perhaps we could move all new documentation there, and link there from here, stating that all options used for "table indexes" can also be used for `table_for`. What do you think?

features/belongs_to.feature
```diff
@@ -27,7 +27,7 @@ Feature: Belongs To
     """
       ActiveAdmin.register User
       ActiveAdmin.register Post do
-        belongs_to :user
+        belongs_to :author, class_name: "User", param: "user_id", route_name: "user"
```
deivid-rodriguez, line 30: Are these and the other changes in this file related to this PR?

lib/active_admin/views/components/table_for.rb
```diff
@@ -11,7 +11,8 @@ def build(obj, *attrs)
         options = attrs.extract_options!
         @sortable = options.delete(:sortable)
         @collection = obj.respond_to?(:each) && !obj.is_a?(Hash) ? obj : [obj]
-        @resource_class = options.delete(:i18n)
+        @resource_class = options.delete(:resource_class)
+        @resource_class ||= options.delete(:i18n) # Deprecated
```
deivid-rodriguez, line 15: Let's print a deprecation warning in this case, so that people can notice this and change their code.

lib/active_admin/views/index_as_table.rb
```diff
@@ -283,26 +283,6 @@ def selectable_column
...
-        def id_column
-          raise "#{resource_class.name} has no primary_key!" unless resource_class.primary_key
-          column(resource_class.human_attribute_name(resource_class.primary_key), sortable: resource_class.primary_key) do |resource|
-            if controller.action_methods.include?("show")
-              link_to resource.id, resource_path(resource), class: "resource_id_link"
-            elsif controller.action_methods.include?("edit")
-              link_to resource.id, edit_resource_path(resource), class: "resource_id_link"
-            else
-              resource.id
-            end
```
deivid-rodriguez, line 302: I see that you changed all the above code with `auto_link resource, resource.id`. But the `auto_link` helper does a bit more stuff like checking authorization. Is this a separate bug fix?

spec/unit/views/components/table_for_spec.rb
```diff
@@ -99,6 +99,54 @@
...
+
+      it "should create a table header based on the primary key" do
+        expect(table.find_by_tag("th").first.content).to eq "Id"
+      end
+
+      it "should create a table row for each element in the collection" do
+        expect(table.find_by_tag("tr").size).to eq 4 # 1 for head, 3 for rows
+      end
+
+      [1, 2, 4].each_with_index do |content, index|
```
deivid-rodriguez, line 119: Could you change this to be an array of strings and remove the `to_s` below? I think it makes it more clear that these mean content and not indexes.

deivid-rodriguez (commented):
I just made a few comments, but this PR looks very good!

features/belongs_to.feature
c960657: Yes, they are needed. I am not too familiar with `belongs_to`, but AFAICT the arguments are necessary for this to work at all (because the class name, `User`, does not match the attribute name, `author`). But I don't understand why the tests did not fail before. I'll try to figure out what is going on.

docs/12-arbre-components.md
c960657: Good idea. I have added a link to that page instead.

lib/active_admin/views/index_as_table.rb
c960657: I could not just move the existing code. For example, if I create a table of posts on the User page, the links for Post id: 123 would be to `/user/123`. So in order to generate the right links, I would have to get the resource config using. This logic is already implemented in `auto_link`, so I chose to just use that. The additional authorisation checks seem like a free bonus.

I took me a while to recall this conclusion, so I had to do some mental backtracking, and I reverted the change in the process. Now I have re-implemented the old solution with the necessary changes but without using auto_link. Let me know, what you prefer.

c960657: Arg! I cannot figure out how to get the `active_admin_config` instance inside `ActiveAdmin::Views::TableFor`. Can I ask the audience? 😅

docs/12-arbre-components.md
deivid-rodriguez: Great, thanks!

features/belongs_to.feature
deivid-rodriguez: Did you figure it out @c960657?

lib/active_admin/views/index_as_table.rb
deivid-rodriguez: The new version looks good to me!

docs/12-arbre-components.md
```diff
@@ -138,25 +138,23 @@ uses `column` to build the fields to show with the table.
...
   column "Received On",     :created_at
   column "Details & Notes", :payment_details
   column "Amount",          :amount_in_dollars
 end
 ```
 
-The `column` method can take a title as its first argument and data
-(`:your_method`) as its second (or first if no title provided). Column also
-takes a block.
+`column`, `index_column`, and `id_column` works like for [index tables](3-index-pages/index-as-table.md).
```
deivid-rodriguez, line 150:
```suggestion
`column`, `index_column`, and `id_column` work like for [index tables](3-index-pages/index-as-table.md).
```

I think "work" is grammatically correct?

deivid-rodriguez (commented):
Just a very small grammar comment, and also curious about why specs didn't fail before without the `belongs_to` options.

If you can also rebase while fixing the grammar nit, that'd be great.

And sorry for the long delay :pray:.

c960657:
> And sorry for the long delay 🙏.

Likewise :-)

> Just a very small grammar comment, and also curious about why specs didn't fail before without the `belongs_to` options.

The difference is caused by the change from `resource_path` to `config.route_instance_path`. The former is somehow able to extract the User from the Post parameter even without the `belongs_to` parameters, but the latter isn't. There is a lot of magic going on here, and I cannot really grasp it.

kwent:
We also need `index_column`

javierjulio:
Sorry, the `index_column` method has been removed. I will accept a PR for the `id_column` method update though. Thank you.

PR #6462, opened 2020-09-20
Drop git in gemspec

utkarsh2102:
Hello 👋🏻,

Thanks for your work on this! ❤️ 

We found that `activeadmin` relies on `git ls-files` to produce a list of files, which could also be achieved using a pure Ruby alternative. Using `git` in gemspec, in general, is a bit problematic.
The rationale behind that can be found here: https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#gemspec-git-rationale

This PR drops the usage of `git` and uses `Dir` instead w/o changing the output:
```
irb(main):001:0> Dir["LICENSE", "{app,config/locales,docs,lib,vendor}/**/{.*,*}"].reject { |f| File.directory?(f) }.sort == `git ls-files LICENSE app config/locales docs lib vendor -z`.split("\x0").sort
=> true
irb(main):002:0> 
```

Also adds the `Packaging` extension to help the downstream maintainers as well! 😄 

Let me know what you think? 💫

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,7 @@ end
 group :lint do
   # Code style
   gem "rubocop", "0.92.0"
+  gem "rubocop-packaging", "~> 0.5"
   gem "rubocop-rspec", "~> 1.30"
   gem "rubocop-rails", "~> 2.3"
   gem "mdl", "0.11.0"
--- a/Gemfile
+++ b/Gemfile
@@ -52,7 +52,7 @@ end
 group :lint do
   # Code style
   gem "rubocop", "0.92.0"
-  gem "rubocop-packaging", "~> 0.5"
+  gem "rubocop-packaging", "0.5.0"
   gem "rubocop-rspec", "~> 1.30"
   gem "rubocop-rails", "~> 2.3"
   gem "mdl", "0.11.0"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,7 @@ inherit_mode:
     - Include
 
 require:
+  - rubocop-packaging
   - rubocop-rails
   - rubocop-rspec
 
@@ -87,6 +88,18 @@ Layout/ParameterAlignment:
 Layout/IndentationStyle:
   Enabled: true
   EnforcedStyle: spaces
+
+Packaging/BundlerSetupInTests:
+  Enabled: true
+
+Packaging/GemspecGit:
+  Enabled: true
+
+Packaging/RequireHardcodingLib:
+  Enabled: true
+
+Packaging/RequireRelativeHardcodingLib:
+  Enabled: true
 
 Rails/FilePath:
   Enabled: true
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -417,6 +417,8 @@ GEM
       unicode-display_width (>= 1.4.0, < 2.0)
     rubocop-ast (0.5.0)
       parser (>= 2.7.1.5)
+    rubocop-packaging (0.5.0)
+      rubocop (~> 0.89)
     rubocop-rails (2.8.1)
       activesupport (>= 4.2.0)
       rack (>= 1.1)
@@ -512,6 +514,7 @@ DEPENDENCIES
   rspec-rails
   rspec-support!
   rubocop (= 0.92.0)
+  rubocop-packaging (= 0.5.0)
   rubocop-rails (~> 2.3)
   rubocop-rspec (~> 1.30)
   simplecov (= 0.19.0)
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
     "application patterns to make it simple for developers to implement " \
     "beautiful and elegant interfaces with very little effort."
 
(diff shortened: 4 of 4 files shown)


utkarsh2102:
I don't entirely understand why the `lint_and_docs` pipeline fails. It passes for me locally:
```
➜  activeadmin git:(drop-git-in-gemspec) rspec spec/gemspec_spec.lint.rb --format documentation
Run options:
  include {:focus=>true}
  exclude {:changes_filesystem=>true}

All examples were filtered out; ignoring {:focus=>true}

Randomized with seed 15312

gemspec sanity
  has no warnings
  succeeds

Finished in 1.02 seconds (files took 0.14446 seconds to load)
2 examples, 0 failures

Randomized with seed 15312
```

@deivid-rodriguez, any idea? 🙈

deivid-rodriguez:
Hi @utkarsh2102! Thanks for this!

The problem is that on CI, `bundler` installs gems to `vendor/bundle`, and since in this PR you're changing our gemspec to ship everything inside `vendor`, it's trying to package everything in there, and finding errors.

So, immediate fix for this PR would be to be more specific and say `vendor/assets` instead, for example.

As separate improvements/things to consider:

* We should problem move away from installing to `vendor/bundle` in CI since we don't intend to vendor those gems code.
* We could commit some bundler configuration to enforce the exact same environment in developers' machines and CI to avoid this "works on my machine, but not on CI" issues.
* This issue shows the difficulties of changing this by default in `bundler` and the need to provide a pure ruby alternative for `git ls-files` that still respects the `.gitignore` file. We don't want people to push a gem with a `vendor/bundle` folder full of external gems code by mistake.

utkarsh2102:
Hello! 👋🏻 

> The problem is that on CI, `bundler` installs gems to `vendor/bundle`, and since in this PR you're changing our gemspec to ship everything inside `vendor`, it's trying to package everything in there, and finding errors.
> 
> So, immediate fix for this PR would be to be more specific and say `vendor/assets` instead, for example.

Eeeks, fixed! The CI is now greeeen!  🌮 

> As separate improvements/things to consider:
> 
>     * We should probably move away from installing to `vendor/bundle` in CI since we don't
>       intend to vendor those gems code.

Right. Maybe we can use something like `tmp/`? Or do you have something else in mind?

>     * We could commit some bundler configuration to enforce the exact same environment
>       in developers' machines and CI to avoid this "works on my machine, but not on CI" issues.

That'd be ideal!

>     * This issue shows the difficulties of changing this by default in `bundler` and the need to 
>       provide a pure ruby alternative for `git ls-files` that still respects the `.gitignore` file. We
>       don't want people to push a gem with a `vendor/bundle` folder full of external gems code
>       by mistake.

Right! I'm looking at different things but we basically want something like `fast_ignore` (a part of it) inside bundler to make this the default way. But we're not there yet. For a lot of people, `Rake::FileList` seems to work because a lot of people don't really use edge cases like `!` in their `.gitignore` which `Rake::FileList` fails to parse. So one way would be to patch `Rake::FileList` to completely respect the `.gitignore` file.

Or maybe I can come up with something during my Christmas or New Year break. Somewhat unsure at this point though. But I'll try to make that happen. It'd need a bit of research from my end to put something together that replicates `git` but in pure Ruby.

Gemfile
```diff
@@ -52,6 +52,7 @@ end
 group :lint do
   # Code style
   gem "rubocop", "0.91.0"
+  gem "rubocop-packaging", "~> 0.5"
```
deivid-rodriguez, line 55: Not a big deal at all, since gems are locked in the `Gemfile.lock` file anyways, but the current criteria is to use `~> x.y` for gems that have reached 1.0, and an exact version for gem still in the 0.x phase. Could you use

```suggestion
  gem "rubocop-packaging", "0.5.0"
```

just to keep it consistent?

deivid-rodriguez (commented):
This looks good to me other than the minor thing I mentioned. I'll address the other improvements I raised separately :+1:.

Gemfile
utkarsh2102: Most certainly! I just committed the suggestion!

deivid-rodriguez:
@javierjulio Just to give you some context here, @utkarsh2102 is a member of the [Debian Ruby Team](https://wiki.debian.org/Teams/Ruby) and does a bunch of work on packaging ruby software for Debian. As his [GSoC project](https://wiki.debian.org/SummerOfCode2020/ApprovedProjects/UpstreamDownstreamCooperationInRuby), he built [rubocop-packaging](https://github.com/utkarsh2102/rubocop-packaging), which implements checkers for several best practices upstream developers can follow to make OS ruby packagers' lives easier.

javierjulio:
Awesome! Thanks for letting me know.

utkarsh2102:
I've rebased the changes to remove the merge conflicts so it should be good to go now! 🌮 
Additionally, I also want to drop `git` from `inherited_resources`, hope you're open to that as well? 😄

deivid-rodriguez (approved):
Looks good to me! Sorry for not looping back earlier!

javierjulio (approved):
Thanks! ❤️

deivid-rodriguez:
>  Additionally, I also want to drop git from inherited_resources, hope you're open to that as well? smile

:+1:

PR #5860, opened 2019-09-15
Enable `Layout/Tab` rubocop cop

deivid-rodriguez:

PR #6471, opened 2020-09-24
Enable some layout cops

deivid-rodriguez:
When working on a different improvement I noticed several layout coding style normalizations that I'd like to make.

I could've added some more but I decided to stop here.

Let me know if you agree with this coding guidelines.

Supersedes https://github.com/activeadmin/activeadmin/pull/5860.


javierjulio (approved):
Much better thanks!

PR #6474, opened 2020-09-25
Use a consistent location for gems locally and in CI

deivid-rodriguez:
The idea is to eliminate "passes on my machine, but fails in CI" issues.

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,0 +1,2 @@
+---
+BUNDLE_PATH: "vendor/bundle"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,7 +45,6 @@ jobs:
       - name: Configure bundler
         run: |
           echo "BUNDLE_GEMFILE=Gemfile" >> $GITHUB_ENV
-          echo "BUNDLE_PATH=$(pwd)/vendor/bundle" >> $GITHUB_ENV
 
       - name: Enable coverage tracking
         run: echo "COVERAGE=true" >> $GITHUB_ENV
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
This is definitely not very "standard practice" but it seems nice to me to have the same configuration locally and in CI. When I first created it, it was because of some RuboCop problems that would only reproduce in CI but I guess it could affect other situations.

deivid-rodriguez:
This is definitely not very "standard practice" but it seems nice to me to have the same configuration locally and in CI. When I first created it, it was because of some RuboCop problems that would only reproduce in CI but I guess it could affect other situations.

deivid-rodriguez:
I was rebasing old PRs and fixed this one. @javierjulio Is this ok with you?

deivid-rodriguez:
I don't recall the specifics, but RuboCop was inspecting some files under `vendor/bundle` that it was not supposed to inspect, and I could not reproduce locally, because Bundler does not install to `vendor/bundle` by default, but it does do that in CI.

In any case, this is not a big deal, I can close it.

javierjulio:
I looked into how we run Rubocop now and noticed its as a rake task. Do you know why we run it that way? I'm used to running it as a separate CI process. I wonder if that would help to ensure it reviews the right files by running through the CLI interface. Would you be open to [moving Rubocop into its own GitHub workflow using reviewdog/action-rubocop](https://github.com/reviewdog/action-rubocop)?

deivid-rodriguez:
I think the original issue was fixed by adding some exclusions to the RuboCop configuration, I just had some trouble initially reproducing it. In other words, RuboCop is working just fine now.

Regarding this change, it's no big deal, this might be unified automatically in the future because future versions of Bundler will automatically install to a path local to the application, not globally. So I'll close this.

Regarding this `review dog/action-rubocop`, I'm open to try it. I think the reason we run it as part of a rake task is that we also run other checks together with it. But I think it's nice to get RuboCop issues as inline comments, so I would be happy to try it.

PR #6476, opened 2020-09-25
Fix tiny formatting issue in changelog generator

deivid-rodriguez:
If a PR title include trailing spaces, the generated CHANGELOG entry has an extra space before the final dot:

```
* Fix input string filter no rendering dropdown input when its column name ends with a ransack predicate . [#6422] by [@Fivell]
```

This kind of formatting little issues can be easily corrected by changing the PR title, but in this particular case Github doesn't display this trailing space, so it's hardly impossible to notice. We can fix it when generating the changelog entry instead by striping any trailing whitespace from the title.

### DIFF
--- a/tasks/changelog.rb
+++ b/tasks/changelog.rb
@@ -29,7 +29,7 @@ class Changelog
       lines << ""
 
       pulls.sort_by(&:merged_at).reverse_each do |pull|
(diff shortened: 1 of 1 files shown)


javierjulio (approved):
Thanks!

PR #6480, opened 2020-09-27
Ignore any blank flash messages 

ndbroadbent:
Hello! I have a very small annoyance that I would like to fix in ActiveAdmin. This is also related to using the Devise gem for authentication.

I don't like the default `already_authenticated` error for Devise, which shows the flash alert: "You are already signed in."
I wanted to turn this off so that people would just get silently redirected to the dashboard, and I think most people would quickly figure out what happened without the need to show an error. This is also what GitHub is doing when you visit: https://github.com/login

I was looking for a way to disable these messages, and it's surprisingly difficult. There are a lot of solutions posted on StackOverflow, involving various overrides, but [the most common and straightforward solution](https://stackoverflow.com/a/5762835/304706) is to replace the error translation with an empty string. (I also tried changing this to `null`, but that gives an i18n "missing translation" error.) 

This works fine in my app, since I change my view to check for `present?`. But in ActiveAdmin, this shows an empty flash message with an empty red bar at the top of the screen:
 
Screen Shot 2020-09-27 at 2 39 41 PM

Screen Shot 2020-09-27 at 2 39 47 PM

(The `nil` is needed at the end of the block is to prevent the hash from getting rendered by default when there are no messages.)

### DIFF
--- a/app/views/layouts/active_admin_logged_out.html.erb
+++ b/app/views/layouts/active_admin_logged_out.html.erb
@@ -34,7 +34,7 @@
   <div id="content_wrapper">
     <div class="flashes">
       <% flash_messages.each do |type, message| %>
-        <%= content_tag :div, message, class: "flash flash_#{type}" %>
+        <%= content_tag(:div, message, class: "flash flash_#{type}") if message.present? %>
       <% end %>
     </div>
     <div id="active_admin_content">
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Thanks @ndbroadbent! Shouldn't this be fixed in `devise` though? If someone configures an empty message, why would `devise` generate a flash message at all?

deivid-rodriguez:
I had another look at this and I believe it should've been fixed in devise by https://github.com/heartcombo/devise/pull/5148. Let me know if this is not the case so that we can report this upstream.

Thanks!

ndbroadbent:
Just confirming that I've updated to the latest versions of devise (`4.8.0`) and activeadmin (`2.9.0`), and this issue with the empty flash alerts is no longer happening (so this PR is no longer needed.) Thanks!

PR #6482, opened 2020-09-27
Fix Comments controller destroy declaration

bliof:
The destroy was defined in the create method 🤪


deivid-rodriguez:
Oh wow :scream:.

I wonder if the default action is actually working fine, since nobody has actually reported any issues while deleting comments that I recall :thinking:.

javierjulio:
@bliof I tried the update branch feature here but seeing the same code coverage error elsewhere so I think the issue is I can't rebase on my side. Can you please rebase or push an empty commit to re-run CI? I will approve and merge this change once build passes. Thank you.

javierjulio (approved):

javierjulio (approved):

PR #6484, opened 2020-09-28
Use bundler 2.2.0.rc.1

deivid-rodriguez:
There's a bug in the `rake release` task that's [fixed in the latest bundler rc release](https://github.com/rubygems/rubygems/pull/3785) and it's preventing me from releasing. I get this:

```
$ bin/rake release
activeadmin 2.8.1 built to pkg/activeadmin-2.8.1.gem.
Tagged v2.8.1.
Untagging v2.8.1 due to error.
rake aborted!
Couldn't git push. `git push  --tags' failed with the following output:

To https://github.com/activeadmin/activeadmin
 * [new tag]             v2.8.1 -> v2.8.1
 ! [rejected]            v2.7.0 -> v2.7.0 (already exists)
error: failed to push some refs to 'https://github.com/activeadmin/activeadmin'
hint: Updates were rejected because the tag already exists in the remote.
```

I think I could also fix it by messing with my local tags, but I'd like to try the bundler rc release too.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ version: 2.1
 .install_bundler: &install_bundler
   run:
     name: Install a specific bundler version
-    command: gem install bundler -v 2.1.4
+    command: gem install bundler -v 2.2.0.rc.1
 
 .install_build_deps: &install_build_deps
   run:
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
This actually revealed an issue with the "inline Gemfile in our bug report template, which had incompatible dependencies. Since bundler 2.2 will resolve Gemfiles for all platforms by default, it detected the issue.

This one should be ready!

javierjulio (approved):
Awesome thanks!

PR #6487, opened 2020-09-28
Add example of using `default_main_content` in show pages

### LINKED ISSUE #6486: Docs improvement: Custom show pages should include an example with `default_main_content`
None of the custom show page examples includes `default_main_content`.

I think it's a common case where a user like the default content from AA, but wants to add something extra.

sjieg:
See #6486 
Added an example using `default_main_content` function.

Closes #6486

### DIFF
--- a/docs/6-show-pages.md
+++ b/docs/6-show-pages.md
@@ -79,3 +79,15 @@ ActiveAdmin.register Book do
   end
 end
 ```
+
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (approved):
Looks good to me!

javierjulio (approved):
❤️

PR #6491, opened 2020-10-01
Update 0-installation.md - Advise not to require_tree, as this will load all activeadmin js on all pages

sorenwiz:
Added

If you do not want to require the javascript for ActiveAdmin on all pages,
then make sure that you don't have `require_tree .` in your application.js

### DIFF
--- a/docs/0-installation.md
+++ b/docs/0-installation.md
@@ -48,6 +48,9 @@ The generator adds these core files, among others:
 * `app/assets/stylesheets/active_admin.scss`
 * `config/initializers/active_admin.rb`
 
+If you do not want to require the javascript for ActiveAdmin on all pages,
+then make sure that you don't have `require_tree .` in your application.js
+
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (commented):
Thanks!

Not particularly against this but isn't this more of a generic "Rails advice"?

docs/0-installation.md
```diff
@@ -48,6 +48,9 @@ The generator adds these core files, among others:
 * `app/assets/stylesheets/active_admin.scss`
 * `config/initializers/active_admin.rb`
 
+If you do not want to require the javascript for ActiveAdmin on all pages,
+then make sure that you don't have `require_tree .` in your application.js
+
```
deivid-rodriguez, line 53: We now have #6676 which includes a note for this case in the installer itself, so I'll close this in favour of that change, since I didn't get any feedback here.

PR #6523, opened 2020-10-24
Add ability to override "Remove" button text on has_many forms

### LINKED ISSUE #5283: Override text of "Remove" button from has_many form builder for specified resource
There doesn't seem to be a straightforward way to change the text of the "Remove" button (to, in my case, something like "Hide") in the has_many form builder, .i.e., f.has_many, on a single page/model and not site-wide.  I just want to confirm that is the case.  If so, was there a specific reason this was left out?  I can submit a PR to make it changeable if that helps.

littleforest:
Closes #5283 

* Adds new option on a has_many form to override the "Remove" button text
* Adds a new feature spec to test this functionality
* Updates documentation

### DIFF
--- a/docs/5-forms.md
+++ b/docs/5-forms.md
@@ -112,6 +112,7 @@ ActiveAdmin.register Post do
     f.inputs do
       f.has_many :comment,
                  new_record: 'Leave Comment',
+                 remove_record: 'Remove Comment',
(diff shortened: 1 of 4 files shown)


deivid-rodriguez (approved):
This looks good to me!

javierjulio (approved):
Thank you!

Issue #3802, opened 2015-02-19
Should we differentiate authorization for new and create?


bamorim:
(This is more like a discussion than an issue. :smile:)

Well, I'm using Pundit for my authorization layer and I have some issues (not critical, easy workaround).

The question is whether we should have different actions for new and create.
Let me exemplify:

A Theme has many Subjects
An User can be the coordinator of a given Theme
User can create a Subject only for the Theme he is the coordinator.

If I were impementing my own admin, I'd create a Pundit Policy like this:

```ruby
class SubjectPolicy < ApplicationPolicy
  def new?
    # Every user should be able to reach the new page
    true
  end

  def create?
    # However, the user can only create subjects belonging to themes he is the coordinator
    record.theme.coordinator_id == user.id
  end
end
```

But, ActiveAdmin uses the ```create?``` for showing the new form as well for creating the subject itself.

The workaround I did was checking whether the theme was present:

```ruby
  def create?
    user.admin? or
      record.theme.nil? or
      record.theme.coordinator_id == user.id
  end
```

What you guys think? If you'd agree I can make some effort to create a PR for this.

Thank you for your attention. :smile:

timoschilling:
double of / related to #3405

timoschilling:
Feel free to create a PR

bamorim:
@timoschilling Okay, I'll create a Pull Request, but do you think that should be backwards compatible?

I can try to call new? and if it's not present then call create?
Or maybe that should be a configuration.

What do you think?

timoschilling:
The next release will be a 1.0.0, so we can break thinks.

finishingmove:
I just think that `new?` and `create?` should not be connected in *any* way by default. The same for `edit?` and `update?`. It would be easy enough for the end-users to reproduce the current behavior if they wanted it, but the way it is now, it does more bad than good.

gkop:
I have a (much) more specific related issue, which is that the Pundit adapter is not aware of `belongs_to` when initializing the policy; see https://github.com/activeadmin/activeadmin/blob/d01155f73a0e2cc27d712792270357e2b3759d05/lib/active_admin/pundit_adapter.rb#L34 , because it doesn't passing the "parent" record in the belongs to association, at `subject.new`, I am unable to configure authorization in Pundit in context of the parent record.

wvteijlingen:
@bamorim 

I'm running into exactly the same issue. What is the status on this? Was there any progress made on a PR?

Issue #3405, opened 2014-09-05
PunditAdapter doesn't execute authorize


finishingmove:
It seems the PunditAdapter doesn't check for authorization on `create` and  `update`, from what I can tell.

Somewhat related to: https://github.com/activeadmin/activeadmin/issues/3068

PR #6535, opened 2020-11-10
Allow different new and edit rules in authorization adapters

### LINKED ISSUE #6529: Pundit adapter treats new and create the same
Pundit has separate permissions for `new?` and `create?`. Often these are the same, but there are many cases where you would want it to be different: for instance, you'd like to allow any user access to the new item _form_, but users should only be able to create items within their departments.

```ruby
class ItemPolicy < ApplicationPolicy
  def new?
    user_signed_in?
  end

  def create?
    user_signed_in? && user.departments.include? record.department
  end
end
```

In Active Admin, the "New Item" button would never show because `create?` would never return true. I believe the resolution is to use `new?` instead of `create?` to determine whether to show the New form ([assumedly here](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/resource/action_items.rb#L67) with an additional case in the [adapter](https://github.com/activeadmin/activeadmin/blob/b90cdb634f00d0ab37ed01a249693ad3a87bf181/lib/active_admin/pundit_adapter.rb#L49)).

This seems like a rather important feature of the pundit adapter, though, and I'm wondering how people have even got on without this? I see this was mentioned in 2014 (#3405) and 2015 (#3802) but both times it seems like it may have been misunderstood. Am _I_ missing something?

timwis:
This pull request allows authorisation adapters to specify different logic for `:new` vs `:create` and `:edit` vs `:update`, for the use cases described in #6529, #3802, and #3405, for example:

```ruby
class PostPolicy < ApplicationPolicy
  def new?
    true
  end

  def create?
    record.category.name != "Announcements" || user.admin
  end
end
```

It does so in a (hopefully) backwards-compatible way: prior to this PR, ActiveAdmin checked whether the user was authorised for `Auth::CREATE` before showing the "New Post" button. Now, ActiveAdmin checks whether the user is authorised for `Auth::NEW`. By default, though, `Auth::NEW` maps to the `:create` symbol, just like `Auth::CREATE` does, so any custom authorisation adapters using the `:create` symbol should continue to work fine (as the tests do).

The _Pundit_ adapter, however, _does_ make this differentiation, since pundit specifically lets you specifiy different logic for `new?` vs `create?` (and edit/update). I believe this will be backwards compatible anyway, though, since the [default application policy generated by `pundit:install`](https://github.com/varvet/pundit/blob/master/lib/generators/pundit/install/templates/application_policy.rb) has `new?` simply calling `create?` anyway. Likewise, [ActiveAdmin recommends an application policy](https://activeadmin.info/13-authorization-adapter.html#using-the-pundit-adapter) that does the same.

Therefore the only users for whom this would be breaking are those who use pundit but do not have `new?` or `edit?` methods defined in their policies nor their `ApplicationPolicy`, despite them being defined by default (they will have had to remove them or not use the default `ApplicationPolicy`).

Posting as draft to get feedback. I've still got to get the test to work. I'm under the impression this will be backwards compatible, particularly given that recommended pundit policies have `new?` calling `create?` anyway, but would love to hear if there's any chance this won't be.

Closes #6529

### DIFF
--- a/features/authorization.feature
+++ b/features/authorization.feature
@@ -15,7 +15,7 @@ Feature: Authorizing Access
 
         when normalized(Post)
           case action
-          when ActiveAdmin::Auth::UPDATE, ActiveAdmin::Auth::DESTROY
+          when :update, :destroy
             false
           else
             true
--- a/features/authorization.feature
+++ b/features/authorization.feature
@@ -15,7 +15,7 @@ Feature: Authorizing Access
 
         when normalized(Post)
           case action
-          when :update, :destroy
+          when :edit, :update, :destroy
             false
           else
             true
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -44,7 +44,9 @@ module ActiveAdmin
     def format_action(action, subject)
       # https://github.com/varvet/pundit/blob/master/lib/generators/pundit/install/templates/application_policy.rb
       case action
+      when Auth::NEW then :new?
       when Auth::CREATE then :create?
+      when Auth::EDIT then :edit?
       when Auth::UPDATE then :update?
       when Auth::READ then subject.is_a?(Class) ? :index? : :show?
       when Auth::DESTROY then subject.is_a?(Class) ? :destroy_all? : :destroy?
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -44,10 +44,6 @@ module ActiveAdmin
     def format_action(action, subject)
       # https://github.com/varvet/pundit/blob/master/lib/generators/pundit/install/templates/application_policy.rb
       case action
-      when Auth::NEW then :new?
-      when Auth::CREATE then :create?
-      when Auth::EDIT then :edit?
-      when Auth::UPDATE then :update?
       when Auth::READ then subject.is_a?(Class) ? :index? : :show?
       when Auth::DESTROY then subject.is_a?(Class) ? :destroy_all? : :destroy?
       else "#{action}?"
--- a/features/comments/commenting.feature
+++ b/features/comments/commenting.feature
@@ -251,7 +251,7 @@ Feature: Commenting
       """
         class NoNewComments < ActiveAdmin::AuthorizationAdapter
           def authorized?(action, subject = nil)
-            if action == :create && subject == ActiveAdmin::Comment
+            if (action == :new || action == :create) && subject == ActiveAdmin::Comment
               false
             else
               true
--- a/lib/active_admin/authorization_adapter.rb
+++ b/lib/active_admin/authorization_adapter.rb
@@ -4,9 +4,9 @@ module ActiveAdmin
   # Default Authorization permissions for Active Admin
   module Authorization
     READ = :read
-    NEW = :create
+    NEW = :new
     CREATE = :create
-    EDIT = :update
+    EDIT = :edit
     UPDATE = :update
     DESTROY = :destroy
   end
--- a/lib/active_admin/authorization_adapter.rb
+++ b/lib/active_admin/authorization_adapter.rb
@@ -4,7 +4,9 @@ module ActiveAdmin
   # Default Authorization permissions for Active Admin
   module Authorization
     READ = :read
+    NEW = :create
     CREATE = :create
+    EDIT = :update
     UPDATE = :update
     DESTROY = :destroy
   end
--- a/lib/active_admin/base_controller/authorization.rb
+++ b/lib/active_admin/base_controller/authorization.rb
@@ -7,9 +7,9 @@ module ActiveAdmin
       ACTIONS_DICTIONARY = {
         index: ActiveAdmin::Authorization::READ,
         show: ActiveAdmin::Authorization::READ,
-        new: ActiveAdmin::Authorization::CREATE,
+        new: ActiveAdmin::Authorization::NEW,
         create: ActiveAdmin::Authorization::CREATE,
-        edit: ActiveAdmin::Authorization::UPDATE,
+        edit: ActiveAdmin::Authorization::EDIT,
         update: ActiveAdmin::Authorization::UPDATE,
         destroy: ActiveAdmin::Authorization::DESTROY
       }
--- a/lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb
+++ b/lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb
@@ -36,7 +36,7 @@ module ActiveAdmin
 
           text_node paginate @comments
 
-          if authorized?(ActiveAdmin::Auth::CREATE, ActiveAdmin::Comment)
+          if authorized?(ActiveAdmin::Auth::NEW, ActiveAdmin::Comment)
             build_comment_form
           end
         end
--- a/lib/active_admin/resource/action_items.rb
+++ b/lib/active_admin/resource/action_items.rb
@@ -65,7 +65,7 @@ module ActiveAdmin
       # Adds the default New link on index
       def add_default_new_action_item
         add_action_item :new, only: :index do
-          if controller.action_methods.include?("new") && authorized?(ActiveAdmin::Auth::CREATE, active_admin_config.resource_class)
+          if controller.action_methods.include?("new") && authorized?(ActiveAdmin::Auth::NEW, active_admin_config.resource_class)
             localizer = ActiveAdmin::Localizers.resource(active_admin_config)
             link_to localizer.t(:new_model), new_resource_path
           end
@@ -75,7 +75,7 @@ module ActiveAdmin
       # Adds the default Edit link on show
       def add_default_edit_action_item
         add_action_item :edit, only: :show do
-          if controller.action_methods.include?("edit") && authorized?(ActiveAdmin::Auth::UPDATE, resource)
+          if controller.action_methods.include?("edit") && authorized?(ActiveAdmin::Auth::EDIT, resource)
             localizer = ActiveAdmin::Localizers.resource(active_admin_config)
             link_to localizer.t(:edit_model), edit_resource_path(resource)
           end
--- a/lib/active_admin/view_helpers/auto_link_helper.rb
+++ b/lib/active_admin/view_helpers/auto_link_helper.rb
@@ -29,7 +29,7 @@ module ActiveAdmin
           authorized?(ActiveAdmin::Auth::READ, resource)
           url_for config.route_instance_path resource, url_options
         elsif config.controller.action_methods.include?("edit") &&
-          authorized?(ActiveAdmin::Auth::UPDATE, resource)
+          authorized?(ActiveAdmin::Auth::EDIT, resource)
           url_for config.route_edit_instance_path resource, url_options
         end
       end
--- a/lib/active_admin/views/index_as_table.rb
+++ b/lib/active_admin/views/index_as_table.rb
@@ -377,7 +377,7 @@ module ActiveAdmin
           if controller.action_methods.include?("show") && authorized?(ActiveAdmin::Auth::READ, resource)
             item localizer.t(:view), resource_path(resource), class: "view_link #{options[:css_class]}", title: localizer.t(:view)
           end
-          if controller.action_methods.include?("edit") && authorized?(ActiveAdmin::Auth::UPDATE, resource)
+          if controller.action_methods.include?("edit") && authorized?(ActiveAdmin::Auth::EDIT, resource)
             item localizer.t(:edit), edit_resource_path(resource), class: "edit_link #{options[:css_class]}", title: localizer.t(:edit)
           end
           if controller.action_methods.include?("destroy") && authorized?(ActiveAdmin::Auth::DESTROY, resource)
--- a/lib/active_admin/views/pages/index.rb
+++ b/lib/active_admin/views/pages/index.rb
@@ -112,7 +112,7 @@ module ActiveAdmin
 
         def render_blank_slate
           blank_slate_content = I18n.t("active_admin.blank_slate.content", resource_name: active_admin_config.plural_resource_label)
-          if controller.action_methods.include?("new") && authorized?(ActiveAdmin::Auth::CREATE, active_admin_config.resource_class)
+          if controller.action_methods.include?("new") && authorized?(ActiveAdmin::Auth::NEW, active_admin_config.resource_class)
             blank_slate_content = [blank_slate_content, blank_slate_link].compact.join(" ")
           end
           insert_tag(view_factory.blank_slate, blank_slate_content)
--- a/spec/support/templates/policies/post_policy.rb
+++ b/spec/support/templates/policies/post_policy.rb
@@ -5,7 +5,7 @@ class PostPolicy < ApplicationPolicy
   end
 
   def create?
-    record.category.name != "Announcements" || user.is_a?(User::VIP)
+    record.category.nil? || record.category.name != "Announcements" || user.is_a?(User::VIP)
   end
 
   def update?
--- a/spec/support/templates/policies/post_policy.rb
+++ b/spec/support/templates/policies/post_policy.rb
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 class PostPolicy < ApplicationPolicy
+  def new?
+    true
+  end
+
+  def create?
+    record.category.name != "Announcements" || user.admin
+  end
+
   def update?
     record.author == user
   end
--- a/spec/support/templates/policies/post_policy.rb
+++ b/spec/support/templates/policies/post_policy.rb
@@ -5,7 +5,7 @@ class PostPolicy < ApplicationPolicy
   end
 
   def create?
-    record.category.name != "Announcements" || user.admin
+    record.category.name != "Announcements" || user.is_a?(User::VIP)
   end
 
   def update?
--- a/spec/unit/authorization/controller_authorization_spec.rb
+++ b/spec/unit/authorization/controller_authorization_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe "Controller Authorization", type: :controller do
   end
 
   it "should authorize the new action" do
-    expect(authorization).to receive(:authorized?).with(auth::CREATE, an_instance_of(Post)).and_return true
+    expect(authorization).to receive(:authorized?).with(auth::NEW, an_instance_of(Post)).and_return true
     get :new
     expect(response).to be_successful
   end
--- a/spec/unit/cancan_adapter_spec.rb
+++ b/spec/unit/cancan_adapter_spec.rb
@@ -13,6 +13,7 @@ RSpec.describe ActiveAdmin::CanCanAdapter do
 
         def initialize(user)
           can :read, Post
+          can :create, Post
           cannot :update, Post
         end
       end
@@ -28,6 +29,11 @@ RSpec.describe ActiveAdmin::CanCanAdapter do
     it "should initialize the ability stored in the namespace configuration" do
       expect(auth.authorized?(:read, Post)).to eq true
       expect(auth.authorized?(:update, Post)).to eq false
+    end
+
+    it "should treat :new ability the same as :create" do
+      expect(auth.authorized?(:new, Post)).to eq true
+      expect(auth.authorized?(:create, Post)).to eq true
     end
 
     it "should scope the collection with accessible_by" do
--- a/spec/unit/pundit_adapter_spec.rb
+++ b/spec/unit/pundit_adapter_spec.rb
@@ -30,7 +30,8 @@ RSpec.describe ActiveAdmin::PunditAdapter do
     let(:application) { ActiveAdmin::Application.new }
     let(:namespace) { ActiveAdmin::Namespace.new(application, "Admin") }
     let(:resource) { namespace.register(Post) }
-    let(:auth) { namespace.authorization_adapter.new(resource, double) }
+    let(:user) { User.new }
+    let(:auth) { namespace.authorization_adapter.new(resource, user) }
     let(:default_policy_klass) { DefaultPolicy }
     let(:default_policy_klass_name) { "DefaultPolicy" }
 
@@ -45,10 +46,12 @@ RSpec.describe ActiveAdmin::PunditAdapter do
 
     it "should allow differentiating between new and create" do
       expect(auth.authorized?(:new, Post)).to eq true
+      expect(auth.authorized?(ActiveAdmin::Auth::NEW, Post)).to eq true
 
-      announcement_category = Category.create! name: "Announcements"
-      announcement_post = Post.create! title: "Big announcement", category: announcement_category
+      announcement_category = Category.new(name: "Announcements")
+      announcement_post = Post.new(title: "Big announcement", category: announcement_category)
       expect(auth.authorized?(:create, announcement_post)).to eq false
+      expect(auth.authorized?(ActiveAdmin::Auth::CREATE, announcement_post)).to eq false
     end
 
     it "should scope the collection" do
--- a/spec/unit/pundit_adapter_spec.rb
+++ b/spec/unit/pundit_adapter_spec.rb
@@ -43,6 +43,14 @@ RSpec.describe ActiveAdmin::PunditAdapter do
       expect(auth.authorized?(:update, Post)).to eq false
     end
 
+    it "should allow differentiating between new and create" do
+      expect(auth.authorized?(:new, Post)).to eq true
+
+      announcement_category = Category.create! name: "Announcements"
+      announcement_post = Post.create! title: "Big announcement", category: announcement_category
+      expect(auth.authorized?(:create, announcement_post)).to eq false
+    end
+
     it "should scope the collection" do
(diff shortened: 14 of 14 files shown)


lib/active_admin/pundit_adapter.rb
```diff
@@ -46,7 +46,9 @@ def retrieve_policy(subject)
     def format_action(action, subject)
       # https://github.com/varvet/pundit/blob/master/lib/generators/pundit/install/templates/application_policy.rb
       case action
+      when Auth::NEW then :new?
```
timwis, line 49: I believe I've figured out why the test is failing. `pundit_adapter_spec.rb:54` calls `authorized?(:create, announcement_post)`, which runs through this case statement with the `action` of `:create`, which matches on both `Auth::NEW` and `Auth::CREATE`. It takes the first match here and runs the `new?` method instead of the `create?` method.

It would work fine if active admin and its tests always called `authorized?` with the `Auth::METHOD` constant format, but it seems interchanged, at least in the tests.

I'd love input on how to resolve this?

timwis:
Hey, just checking in to see if anyone has a suggestion on that case statement? Feels like this is 95% of the way there!

timwis:
Hiya 👋 just checking in to see if anyone has had a chance to look at this, or give any steer as to whether this will eventually be accepted?

deivid-rodriguez:
Hi @timwis, I'm sorry for not responding, I'm currently buried with other work, sorry :(. Just trying to leave a message here so that you know that I'll eventually get to this, I just don't know when...

timwis:
Hey @deivid-rodriguez, just checking in to see if you've had a moment to review?

timwis:
Hey @deivid-rodriguez any chance I could hire you to review this?

lib/active_admin/pundit_adapter.rb
deivid-rodriguez: Hei @timwis!

I think we basically want to fallback to the default case branch, the one falling back to `#{action}?`, so I think removing the conditions where you want just that might do the trick?

```diff
diff --git a/lib/active_admin/pundit_adapter.rb b/lib/active_admin/pundit_adapter.rb
index d32f14504..c4d24570d 100644
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -46,10 +46,6 @@ module ActiveAdmin
     def format_action(action, subject)
       # https://github.com/varvet/pundit/blob/master/lib/generators/pundit/install/templates/application_policy.rb
       case action
-      when Auth::NEW then :new?
-      when Auth::CREATE then :create?
-      when Auth::EDIT then :edit?
-      when Auth::UPDATE then :update?
       when Auth::READ then subject.is_a?(Class) ? :index? : :show?
       when Auth::DESTROY then subject.is_a?(Class) ? :destroy_all? : :destroy?
       else "#{action}?"
```

deivid-rodriguez:
@timwis This seems like a good solution to me, I left a comment on the case statement that might do the trick :+1:.

deivid-rodriguez:
> Hey @deivid-rodriguez any chance I could hire you to review this?

Well, if you can support me that'd definitely be awesome. I'm trying to make a living from maintaining ruby open source but it gets pretty hard. Whatever you planned to offer me to get this reviewed, feel free to use GitHub sponsors or any other means to get it to me once we're done here and you're happy with my review :)

lib/active_admin/pundit_adapter.rb
timwis: Thanks @deivid-rodriguez, good idea! I tried this, but because I used backwards-compatible constant-symbol mappings in `authorization_adapter.rb`, it ended up checking the wrong action anyway. `authorization_adapter.rb` set `NEW = :create` and `EDIT = :update` for backwards compatibility. This section was intended to override that for pundit.

But apparently [CanCan considers `:new` and `:create` to be aliases](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/Action-Aliases.md) (and edit/update), so the backwards compatibility is already covered for CanCan — I've added a test to verify that.

The only thing it doesn't cover is backwards compatibility for custom authorization adapters.

I've just pushed changes (including a few minor improvements to the tests for consistency). The unit tests are passing now, but the cucumber tests fail because their custom authorization adapters don't restrict `:new`.

Need to think of a better way to provide backwards compatibility which still allows this to work 🤔

timwis: The options I've thought of so far are:

1. Leave `NEW = :create` in the AuthorizationAdapter. Then, in the PunditAdapter, somehow check whether the `action` passed to `authorized?` is a constant, and what the constant was. I'm thinking like you might check the name of a class instance's class, but my guess is this isn't possible in Ruby, and it probably just passes the value (`:create`)
2. Have discrete `NEW = :new` in the AuthorizationAdapter. Then, in the AuthorizationAdapter class, somehow add a method for mapping/normalising the actions before passing them to `authorized?`. By default, `:new` gets mapped to `:create` as if it's an alias, but in the pundit adapter, we don't do that (we'd extend the mapping/normalising method to be no-op). Having poked around, though, I'm not sure this is possible either as ActiveAdmin just instantiates the class and calls its `authorized?` method with global constants.
3. Go through the ActiveAdmin codebase and swap out the use of `ActiveAdmin::Auth::<action>` with something that can be overridden by our adapter subclass. Perhaps something like `ActiveAdmin.application.namespace(:admin).authorization_adapter.class::NEW`, which would allow us to set `NEW = :create` in the parent AuthorizationAdapter class but override it with `NEW = :new` in the PunditAdapter subclass.

So far option 3 seems the most promising! Will try it out tomorrow morning. Let me know if you've any other ideas.

timwis: Just thought of a fourth option:

```diff
  def format_action(action, subject)
    case action
-   when Auth::CREATE then :create?
+   when Auth::CREATE then subject.is_a?(Class) ? :new? : :create?
-   when Auth::UPDATE then :update?
+   when Auth::UPDATE then subject.changed? ? :edit? : :update?
    when Auth::READ then subject.is_a?(Class) ? :index? : :show?
    when Auth::DESTROY then subject.is_a?(Class) ? :destroy_all? : :destroy?
    else "#{action}?"
    end
  end
```

Just wondering whether `.changed?` covers all edge cases 🤔

timwis: Okay @deivid-rodriguez I've hit a wall after about 6 hours today. I'm quite confident we could actually get this down to a _one-line change_:

```diff
  def format_action(action, subject)
    case action
-   when Auth::CREATE then :create?
+   when Auth::CREATE then subject.is_a?(Class) ? :new? : :create?
    when Auth::UPDATE then :update?
    when Auth::READ then subject.is_a?(Class) ? :index? : :show?
    when Auth::DESTROY then subject.is_a?(Class) ? :destroy_all? : :destroy?
    else "#{action}?"
    end
  end
```

Because I don't think the use case for this PR really applies to edit/update.

This _looks_ like it should work, given how index/show and destroy_all/destroy work below. Unfortunately, it doesn't: when you browse to the `/new` page, `authorized?` is called with an _instantiated_ resource, rather than the Class.

I've tracked this down to:
- [inherited_resources/actions.rb:#new](https://github.com/activeadmin/inherited_resources/blob/master/lib/inherited_resources/actions.rb#L19) calls the `build_resource` method instead of the `collection` method like it uses for `#index`
- This calls [data_access.rb:build_resource](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/resource_controller/data_access.rb#L120), which appears to instantiate the resource

I'm not sure why the authorization call on `/new` would use an instantiated object instead of the class. It uses the class to determine whether to show the "New" button, but an instantiated object to access the page. If there's a way we can get it to use the class when accessing the page, this PR can be incredibly simple!

But I've dug about as far as I can :/ Any idea?

Fivell:
@timwis  calrify please why it is not suitable to check record for persistence instead ,  like 
```
class PostPolicy < ApplicationPolicy
 
  def create?
    if record.persisted?
       something
    else
     something else
    end
  end
end
```

I'm using this for years and never though it is a problem

timwis:
@Fivell Hm, why would `create?` ever be called for a record that is persisted? Wouldn't that call `update?`? When differentiating between whether the user can (A) view the `/new` page and (B) create a given record, neither scenario would have `#persisted?` return `true`.

Fivell:
@timwis  my fault, you're right, I rechecked and it was sort of singleton resources only

lib/active_admin/pundit_adapter.rb
deivid-rodriguez: Hi @timwis, I'm not too sure about what the problem is.

I looked at the failing specs with the change I suggested before, and the failures seem expected, due to how the policy was setup. The unit spec you added fails because the `create?` method of the policy is correctly invoked, but ends up calling `.admin` on the test double that represents the user, which didn't respond to that. Similarly, some cucumber features failed for the same reason: they were calling `create?` on a record without category, and thus crashing.

Changing the unit spec and the policy as follows gets everything passing:

```diff
diff --git a/lib/active_admin/pundit_adapter.rb b/lib/active_admin/pundit_adapter.rb
index d32f14504..c4d24570d 100644
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -46,10 +46,6 @@ module ActiveAdmin
     def format_action(action, subject)
       # https://github.com/varvet/pundit/blob/master/lib/generators/pundit/install/templates/application_policy.rb
       case action
-      when Auth::NEW then :new?
-      when Auth::CREATE then :create?
-      when Auth::EDIT then :edit?
-      when Auth::UPDATE then :update?
       when Auth::READ then subject.is_a?(Class) ? :index? : :show?
       when Auth::DESTROY then subject.is_a?(Class) ? :destroy_all? : :destroy?
       else "#{action}?"
diff --git a/spec/support/templates/policies/post_policy.rb b/spec/support/templates/policies/post_policy.rb
index 4b26840fd..9490eaa4e 100644
--- a/spec/support/templates/policies/post_policy.rb
+++ b/spec/support/templates/policies/post_policy.rb
@@ -4,7 +4,7 @@ class PostPolicy < ApplicationPolicy
   end
 
   def create?
-    record.category.name != "Announcements" || user.admin
+    record.category.nil? || record.category.name != "Announcements" || user.admin
   end
 
   def update?
diff --git a/spec/unit/pundit_adapter_spec.rb b/spec/unit/pundit_adapter_spec.rb
index c36e88beb..e02739f52 100644
--- a/spec/unit/pundit_adapter_spec.rb
+++ b/spec/unit/pundit_adapter_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe ActiveAdmin::PunditAdapter do
     let(:application) { ActiveAdmin::Application.new }
     let(:namespace) { ActiveAdmin::Namespace.new(application, "Admin") }
     let(:resource) { namespace.register(Post) }
-    let(:auth) { namespace.authorization_adapter.new(resource, double) }
+    let(:auth) { namespace.authorization_adapter.new(resource, double(:admin => false)) }
     let(:default_policy_klass) { DefaultPolicy }
     let(:default_policy_klass_name) { "DefaultPolicy" }
 ```

I also played around with a simple application with a similar policy that always whitelisted `new?` but had some custom behavior on `create?` and things seemed good.

Can you explain the issue my suggested approach has? As I understand it, my approach should be independent from this `Auth` mapping in `authorization.rb`.

File: features/authorization.feature [hunk 2]
```diff
@@ -15,7 +15,7 @@ Feature: Authorizing Access
 
         when normalized(Post)
           case action
-          when ActiveAdmin::Auth::UPDATE, ActiveAdmin::Auth::DESTROY
+          when :update, :destroy
```
ngouy, line 18: is it intended?

lib/active_admin/pundit_adapter.rb
ngouy: @timwis 
> I'm not sure why the authorization call on /new would use an instantiated object instead of the class

My guess is that the controller method "new" instantiate a resource and authorize it before rendering

The workaround here is that if you are using a scoped_collection, your instantiated resource should be compliant with the `:create?` policy, because your scoped_collectin would probably leverage your policy scope (I'm still confused on why it's not the case by default)

TLTR : a user belongs to a company, the create new user would ask for the `new?` policy
but a direct hit on /new would instanciate an user instance within the company scope, and your create? policy which would be something like `my_admin_user.company == new_user_record.company` would be valid

ngouy: btw send a PR to scope new resources build (new and create methods) from auth scopes
https://github.com/activeadmin/activeadmin/pull/6884

So when you hit /new and the policy receives the instantiated resource, even if it hits the :create policy it should do the job, as your resource would be scoped by it (and thus probably make your create? pass). Not fixing the issue we have here but it can ease it a bit depending on use cases.

BTW I think, even if my or is not merged, that it's a good illustration of why we should avoid the changed? calls alongside new? to check if the user tries to create? or "render the new page/form". A new resource can have changed attribute even before being presented to any forms, depending on what you have in your controller#new

deivid-rodriguez: I think all policies should assume they can be called with an instantiated resource so checking whether the subject of the authorization `is_a?` class is not a good solution. For example, from #6884, we can assume that the `new` action will instantiate a new empty object but with the proper foreign key pointing to its parent already filled, and it makes sense to authorize that object.

@timwis How should we move this forward? I think a first step would be to get this green, then we can figure out what backwards compatibility issues we need to address.

timwis: Sure thing, will work on it this weekend

ngouy: same here will try to close #6884 this weekend

timwis: Okay @deivid-rodriguez I've implemented the changes you recommended and tests are passing. Note that, since I removed the backwards compatibility in the constants in `ActiveAdmin::Authorization`` (`NEW` is now `:new` not `:create`), removing those `when` lines probably wasn't necessary, but they were already redundant to begin with since the `else` case covers all four.

In order to get the tests to pass, I had to modify the custom authorisation adapters in the feature specs. It was _these_ that I was worried about: if any users have custom adapters that only handle `:create` and `:update` and then have `else true`. But at least we don't have to worry about CanCan users, per my previous comment's link.

deivid-rodriguez: Thanks @timwis! Alright, I had another look and I do understand the backwards compatibility concerns better now. How about this approach:

```diff
diff --git a/lib/active_admin/authorization_adapter.rb b/lib/active_admin/authorization_adapter.rb
index 171cd0a2a..7fd9d6421 100644
--- a/lib/active_admin/authorization_adapter.rb
+++ b/lib/active_admin/authorization_adapter.rb
@@ -3,7 +3,9 @@ module ActiveAdmin
   # Default Authorization permissions for Active Admin
   module Authorization
     READ = :read
+    NEW = :new
     CREATE = :create
+    EDIT = :edit
     UPDATE = :update
     DESTROY = :destroy
   end
diff --git a/lib/active_admin/base_controller/authorization.rb b/lib/active_admin/base_controller/authorization.rb
index fff86b96a..b38db2df6 100644
--- a/lib/active_admin/base_controller/authorization.rb
+++ b/lib/active_admin/base_controller/authorization.rb
@@ -35,7 +35,7 @@ module ActiveAdmin
       # @return [Boolean]
       #
       def authorized?(action, subject = nil)
-        active_admin_authorization.authorized?(action, subject)
+        active_admin_authorization.authorized?(action_to_permission(action), subject)
       end
 
       # Authorize the action and subject. Available in the controller
@@ -96,7 +96,11 @@ module ActiveAdmin
       # @return [Symbol] The permission name to use.
       def action_to_permission(action)
         if action && action = action.to_sym
-          Authorization::ACTIONS_DICTIONARY[action] || action
+          if active_admin_authorization.respond_to?(:action_to_permission)
+            active_admin_authorization.action_to_permission(action) || action
+          else
+            Authorization::ACTIONS_DICTIONARY[action] || action
+          end
         end
       end
 
diff --git a/lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb b/lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb
index d900b293e..8af0621fc 100644
--- a/lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb
+++ b/lib/active_admin/orm/active_record/comments/views/active_admin_comments.rb
@@ -35,7 +35,7 @@ module ActiveAdmin
 
           text_node paginate @comments
 
-          if authorized?(ActiveAdmin::Auth::CREATE, ActiveAdmin::Comment)
+          if authorized?(ActiveAdmin::Auth::NEW, ActiveAdmin::Comment)
             build_comment_form
           end
         end
diff --git a/lib/active_admin/resource/action_items.rb b/lib/active_admin/resource/action_items.rb
index 4b439b7ee..45bb83236 100644
--- a/lib/active_admin/resource/action_items.rb
+++ b/lib/active_admin/resource/action_items.rb
@@ -64,7 +64,7 @@ module ActiveAdmin
       # Adds the default New link on index
       def add_default_new_action_item
         add_action_item :new, only: :index do
-          if controller.action_methods.include?("new") && authorized?(ActiveAdmin::Auth::CREATE, active_admin_config.resource_class)
+          if controller.action_methods.include?("new") && authorized?(ActiveAdmin::Auth::NEW, active_admin_config.resource_class)
             localizer = ActiveAdmin::Localizers.resource(active_admin_config)
             link_to localizer.t(:new_model), new_resource_path
           end
@@ -74,7 +74,7 @@ module ActiveAdmin
       # Adds the default Edit link on show
       def add_default_edit_action_item
         add_action_item :edit, only: :show do
-          if controller.action_methods.include?("edit") && authorized?(ActiveAdmin::Auth::UPDATE, resource)
+          if controller.action_methods.include?("edit") && authorized?(ActiveAdmin::Auth::EDIT, resource)
             localizer = ActiveAdmin::Localizers.resource(active_admin_config)
             link_to localizer.t(:edit_model), edit_resource_path(resource)
           end
diff --git a/lib/active_admin/view_helpers/auto_link_helper.rb b/lib/active_admin/view_helpers/auto_link_helper.rb
index 3e1d1506f..f78c350f3 100644
--- a/lib/active_admin/view_helpers/auto_link_helper.rb
+++ b/lib/active_admin/view_helpers/auto_link_helper.rb
@@ -28,7 +28,7 @@ module ActiveAdmin
           authorized?(ActiveAdmin::Auth::READ, resource)
           url_for config.route_instance_path resource, url_options
         elsif config.controller.action_methods.include?("edit") &&
-          authorized?(ActiveAdmin::Auth::UPDATE, resource)
+          authorized?(ActiveAdmin::Auth::EDIT, resource)
           url_for config.route_edit_instance_path resource, url_options
         end
       end
diff --git a/lib/active_admin/views/index_as_table.rb b/lib/active_admin/views/index_as_table.rb
index 948569e41..a653e5f95 100644
--- a/lib/active_admin/views/index_as_table.rb
+++ b/lib/active_admin/views/index_as_table.rb
@@ -376,7 +376,7 @@ module ActiveAdmin
           if controller.action_methods.include?("show") && authorized?(ActiveAdmin::Auth::READ, resource)
             item localizer.t(:view), resource_path(resource), class: "view_link #{options[:css_class]}", title: localizer.t(:view)
           end
-          if controller.action_methods.include?("edit") && authorized?(ActiveAdmin::Auth::UPDATE, resource)
+          if controller.action_methods.include?("edit") && authorized?(ActiveAdmin::Auth::EDIT, resource)
             item localizer.t(:edit), edit_resource_path(resource), class: "edit_link #{options[:css_class]}", title: localizer.t(:edit)
           end
           if controller.action_methods.include?("destroy") && authorized?(ActiveAdmin::Auth::DESTROY, resource)
diff --git a/lib/active_admin/views/pages/index.rb b/lib/active_admin/views/pages/index.rb
index f9f12a7fe..6a49316a6 100644
--- a/lib/active_admin/views/pages/index.rb
+++ b/lib/active_admin/views/pages/index.rb
@@ -111,7 +111,7 @@ module ActiveAdmin
 
         def render_blank_slate
           blank_slate_content = I18n.t("active_admin.blank_slate.content", resource_name: active_admin_config.plural_resource_label)
-          if controller.action_methods.include?("new") && authorized?(ActiveAdmin::Auth::CREATE, active_admin_config.resource_class)
+          if controller.action_methods.include?("new") && authorized?(ActiveAdmin::Auth::NEW, active_admin_config.resource_class)
             blank_slate_content = [blank_slate_content, blank_slate_link].compact.join(" ")
           end
           insert_tag(view_factory.blank_slate, blank_slate_content)
```

By doing this, you can define your own adapter class with its own permission to action mapping instead of the default one, by implementing the `action_to_permission` method. For example, if you want to define separate rules for each action in your , you can do:

```ruby
class PunditAdapterWithPerActionRules < ActiveAdmin::PunditAdapter
  def action_to_permission(action)
    action
  end
end

Comment (timwis) [reply lib/active_admin/pundit_adapter.rb]: Ah! I was trying something very similar with my #2 suggestion above, but in AuthorizationAdapter instead of Authorization. I like the extension approach, too. I'll play around with it and push a commit when I get it working. Thanks for diving in.

**Comment by javierjulio** [comment]
@timwis can you please rebase this and push? I will accept this change if its backwards compatible with Pundit and suggested ActiveAdmin defaults since I think others should be able to change the `new` from `create` and `edit` from `update` if needed. Thank you.

**Comment by timwis** [comment]
Hi @javierjulio! Thanks for picking this up. I've just rebased. It's been a few years since I've been in the weeds of this PR, so I'm just going to refamiliarse myself with the codebase and the changes.

**Comment by javierjulio** [comment]
@timwis that sounds great, thank you! I really appreciate you following up. Once you are ready with that case and further testing, we will get this in.

**Comment by timwis** [comment]
Okay @javierjulio, I think I've got my head re-wrapped around it. So the only case where this would not be backwards compatible would seem to be where the user is using a custom authorisation adapter (not the built-in pundit or cancan one) with an open-ended `else true` (aka deny list approach).

For example, in `features/authorization.feature`:

```ruby
def authorized?(action, subject = nil)
  case subject

  when normalized(Post)
    case action
    when :update, :destroy # <-- this is missing :edit
      false
    else
      true
    end
  end
end
```

An example we'd be more likely to come across, though, is from the [Setting up your own AuthorizationAdapter](https://activeadmin.info/13-authorization-adapter.html#setting-up-your-own-authorizationadapter) page:

```ruby
def authorized?(action, subject = nil)
  case subject
  when normalized(Post)
    # Only let the author update and delete posts
    if action == :update || action == :destroy # <-- this is also missing :edit
      subject.author == user
    else
      true
    end
  else
    true
  end
end
```

In both these cases, the only real consequence would be that the "Edit" button would be visible, and they'd be able to get to the edit form, but they wouldn't be able to save it.

@deivid-rodriguez's suggestion proposes that we allow authorisation adapters to implement their own mapping of actions (e.g. `:edit`, `:update`) to permissions (e.g. `ActiveAdmin::Authorization::EDIT`, `ActiveAdmin::Authorization::UPDATE`). Unless I'm misunderstanding, this would not address the issue above unless we went a step further and changed the _default_ mapping of `:new` to `ActiveAdmin::Authorization::CREATE` and `:edit` to `...::UPDATE` in the `ACTIONS_DICTIONARY` (base_controller/authorization.rb), and then fixed the mapping in the Pundit Authorization Adapter.

But this would carry its own trade-offs: mainly in terms of readability/confusion, for example, you'd expect this test to pass:

```ruby
it "should authorize the new action" do
  expect(authorization).to receive(:authorized?).with(auth::NEW, an_instance_of(Post)).and_return true
  get :new
  expect(response).to be_successful
end
```

But it fails because `authorized?` _actually_ gets called with `auth::CREATE` because of the default mapping. So we'd end up needing to add comments to the counter-intuitive areas like that (admittedly few).

I'll give this some more thought tomorrow, but wanted to share where I'd got to. I've got a feeling this will be fine because the worst case scenario should be that a user sees a new/edit button but then can't actually save the form (because create/update are denied), but would certainly appreciate any other thoughts. Thanks for reading.

timwis:
Hey @javierjulio, just checking in to see if there's anything else I can do to progress this?

javierjulio:
@timwis no, this is it, thank you! I appreciate the detail as I went through it all now. I agree, I think it would be fine as is, all considering.

javierjulio (approved):
Thanks!

ce07c3:
Strangely, this broke some functionality for us, and we used the `ActiveAdmin::PunditAdapter`, but we also used namespaces. I haven't been able to figure this out yet, but it's also not a big issue to fix in our code base. :)

PussInDasBoot:
This caused an issue for us, using `ActiveAdmin::PunditAdapter`, all `/new` routes were resulting in access_denied. Routing the `new?` method to point to `create?` in our ApplicationPolicy (like the default) did not fix it, has anyone else seen this issue?

ce07c3:
I had that issue, too. I haven't had time to troubleshoot. I prefer the way it works now, so it was just anecdotal that the migration here wasn't backwards compatible. :)

timwis:
@PussInDasBoot I'm sorry to hear that! Really thought we'd identified all the edge cases. Hope it hasn't caused you too much trouble. Would you be up for sharing your ApplicationPolicy? And are you using the default/built-in Pundit adapter, or a custom one?

PR #6536, opened 2020-11-11
Pick up upstream fixes in devise templates

munen:
Without this fix, the URL to unlock will yield "Unlock token is invalid".

Updated according to the devise upstream view: https://github.com/heartcombo/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/app/views/devise/mailer/unlock_instructions.html.erb

### DIFF
--- a/app/views/active_admin/devise/mailer/unlock_instructions.html.erb
+++ b/app/views/active_admin/devise/mailer/unlock_instructions.html.erb
@@ -1,7 +1,7 @@
 <p>Hello <%= @resource.email %>!</p>
 
-<p>Your account has been locked due to an excessive amount of unsuccessful sign in attempts.</p>
+<p>Your account has been locked due to an excessive number of unsuccessful sign in attempts.</p>
 
 <p>Click the link below to unlock your account:</p>
 
-<p><%= link_to 'Unlock my account', unlock_url(@resource, unlock_token: @resource.unlock_token) %></p>
+<p><%= link_to 'Unlock my account', unlock_url(@resource, unlock_token: @token) %></p>
--- a/app/views/active_admin/devise/mailer/reset_password_instructions.html.erb
+++ b/app/views/active_admin/devise/mailer/reset_password_instructions.html.erb
@@ -2,7 +2,7 @@
 
 <p>Someone has requested a link to change your password, and you can do this through the link below.</p>
 
-<p><%= link_to 'Change my password', edit_password_url(@resource, reset_password_token: @resource.reset_password_token) %></p>
+<p><%= link_to 'Change my password', edit_password_url(@resource, reset_password_token: @token) %></p>
 
 <p>If you didn't request this, please ignore this email.</p>
 <p>Your password won't change until you access the link above and create a new one.</p>


app/views/active_admin/devise/mailer/unlock_instructions.html.erb
```diff
@@ -1,7 +1,7 @@
 <p>Hello <%= @resource.email %>!</p>
 
-<p>Your account has been locked due to an excessive amount of unsuccessful sign in attempts.</p>
+<p>Your account has been locked due to an excessive number of unsuccessful sign in attempts.</p>
 
 <p>Click the link below to unlock your account:</p>
 
-<p><%= link_to 'Unlock my account', unlock_url(@resource, unlock_token: @resource.unlock_token) %></p>
+<p><%= link_to 'Unlock my account', unlock_url(@resource, unlock_token: @token) %></p>
```
javierjulio, line 7: Looks like Devise was using the former previously but got lost in the mix when we last updated this file. Both updates were back in 2013/2014. Based on this, the same change would be necessary for the reset password instructions email as well.

javierjulio (commented):
Thanks! I looked into why this was unnoticed for so long so I commented with background info for other reviewers. I think it's a safe change. Would need same for the reset password email.

app/views/active_admin/devise/mailer/unlock_instructions.html.erb
deivid-rodriguez: Indeed! We essentially need to backport https://github.com/heartcombo/devise/commit/e947a9cbec77a06340aebb1517168e7e6cb2e6ad. Weird nobody noticed this before :open_mouth:.

munen: > Weird nobody noticed this before 😮.

I haven’t debugged why, but the weird thing was that the unlock link worked for me in `development`, but it didn’t in `production`. Maybe that contributed to the issue being unseen for a while. But I was surprised, too, and almost couldn’t believe it when I stumbled upon it^^

Still, happy to find some low hanging fruit(;

deivid-rodriguez: I guess people have been fixing their templates, but not contributing the fix upstream? :man_shrugging:

Could you fix the other templates as well?

munen: I have added the changes for the reset password token, but haven't tested them. This is the second of the three files from the devise upstream change (https://github.com/heartcombo/devise/commit/e947a9cbec77a06340aebb1517168e7e6cb2e6ad). The third file (confirmation instructions) doesn't seem to be overwritten by ActiveAdmin, so there's no need to change it.

javierjulio: @munen thanks for updating! If you have a moment can you please rebase? I'll approve and get this in next.

deivid-rodriguez: I rebased this yesterday @javierjulio, it should be ready now!

javierjulio: Thanks! I'll get this in now.

javierjulio (approved):

PR #6541, opened 2020-11-17
RSpec 3.10 was released

deivid-rodriguez:
We can lock to the final release now.


javierjulio (approved):
Nice, thanks! I don't remember, what was the reason for locking the version?

deivid-rodriguez:
I was to be able to use fixes for keyword args on ruby 2.7, otherwise running specs was really verbose on warnings.

PR #6544, opened 2020-11-18
Sprockets 4.0.2 was released

deivid-rodriguez:
We can lock to the final release now.

The reason we were testing using a (previously unreleased) ref is that it included fixes the keyword parameter warnings.


javierjulio (approved):
Awesome thanks! ❤️

PR #6545, opened 2020-11-18
Bump DB adapters on jruby

deivid-rodriguez:
Maybe spec hangs have been fixed by these releases :man_shrugging:.


deivid-rodriguez:
Interestingly we seem to be getting actual failures on jruby as of current master rather than hangs, which seems like a good thing since it's much easier to troubleshoot.

deivid-rodriguez:
I did look into the failures but they are really cryptic. Cucumber seems to have had quite a lot of movement lately, we just upgraded to cucumber 5 in #6547, maybe they have better errors there :man_shrugging:.

Anyways, this should be ready!

javierjulio (approved):
No problem at all. We understand it's tough. Thank you for trying! ❤️

PR #6547, opened 2020-11-19
Bump cucumber-rails to 2.2.0

deivid-rodriguez:
We're only using a subset of `cucumber-rails` functionality, so I needed to update the file where we require the bits that we needed to catch up with some internal refactoring they made.


javierjulio (approved):
Awesome thanks David! ❤️

PR #6548, opened 2020-11-20
Support for Rails 6.1

deivid-rodriguez:


nathan-appere:
`railties` version in the gemspec probably need to be updated too.

deivid-rodriguez:
Nice catch, yeah! I'll handle this soon.

kwent:
Blocker for me to upgrade to 6.1 ! Thanks for your hard work guys

manas11:
> Blocker for me to upgrade to 6.1 ! Thanks for your hard work guys

Yeah, same here. Thank you for upgrading this

javierjulio (approved):
Looks great, thanks! 👏🏻

deivid-rodriguez:
Ok, I'm releasing this thing now!

deivid-rodriguez:
Oh, I forgot we already merged the PR removing the sprockets dependency, and I wanted to ship that as 3.0.

Should we take our chances and release that as 2.9 for simplicity?

javierjulio:
Yes I think we should ship a 2.9 release instead and leave the breaking change as a separate release.

deivid-rodriguez:
So you mean 2.9 with only Rails 6.1 support? If we do that I need to either create a stable branch from before we merged the sprockets PR and backport this PR there, or revert the sprockets PR on master and then release.

What I was thinking is that the sprockets change we concluded it was 99% backwards compatible. Only if an application is not explicitly depending on sprockets and relying on activeadmin providing the dependency things could break, and that should be very very rare. So I figured we could cut master as it is.

javierjulio:
Ah yes sorry I realized that after I commented. Hmm I don't think it's an issue. If you prefer to do a 2.9 release with the sprockets dependency related change I'm ok with that. 👍🏻

deivid-rodriguez:
Ok, I'll do that then.

amirnateghi198 (commented):
Update your account information for

PR #6563, opened 2020-11-26
Stabilize CI

deivid-rodriguez:
While working on https://github.com/activeadmin/activeadmin/pull/6561, I noticed that the flakies we have (which are more frequent under Github Actions), only happen in javascript scenarios.

I think these are just a few scenarios, so it shouldn't make things too much slower and hopefully, it will stabilize the CI.

NOTE: I also included a couple other improvements in here that I made while investigating this.


javierjulio (approved):
Thanks! 👏🏻❤️

PR #6564, opened 2020-11-26
Removes duplicate private declaration

brunoarueira:

### DIFF
--- a/lib/active_admin/error.rb
+++ b/lib/active_admin/error.rb
@@ -45,8 +45,6 @@ module ActiveAdmin
       raise new exception
     end
 
-    private
-
     def self.database_error_classes
       @classes ||= []
     end
--- a/lib/active_admin/resource/routes.rb
+++ b/lib/active_admin/resource/routes.rb
@@ -42,8 +42,6 @@ module ActiveAdmin
         config[:route_collection_name] == config[:route_instance_name]
       end
 
-      private
-
       class RouteBuilder
         def initialize(resource)
           @resource = resource
--- a/lib/active_admin/resource_controller/decorators.rb
+++ b/lib/active_admin/resource_controller/decorators.rb
@@ -57,8 +57,6 @@ module ActiveAdmin
           @cache[name] ||= wrap! collection_decorator, name
         end
 
-        private
-
         def self.wrap!(parent, name)
           ::Class.new parent do
             delegate :reorder, :page, :current_page, :total_pages, :limit_value,
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -192,3 +192,6 @@ Layout/SpaceInsideHashLiteralBraces:
 
 Layout/SpaceInsideParens:
   Enabled: true
+
+Lint/UselessAccessModifier:
+  Enabled: true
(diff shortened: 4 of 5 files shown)


javierjulio (approved):
Thanks! ❤️

deivid-rodriguez (commented):
Thanks @brunoarueira!

Can you enable the `Lint/UselessAccessModifier` rubocop cop? Just to double check more instances of this issue and make sure it doesn't happen again in the future. It'd be a matter of explicitly enabling the cop in `.rubocop.yml` and then run `rubocop` and autocorrect whatever it finds.

brunoarueira:
@deivid-rodriguez Sure, I'll do and update asap!

deivid-rodriguez:
Thanks!

brunoarueira:
@deivid-rodriguez done!

lib/active_admin/error.rb
```diff
@@ -45,8 +45,6 @@ def self.capture
       raise new exception
     end
 
-    private
```
deivid-rodriguez, line 48: Interesting. I think the intention here was to make the `database_error_classes` method private. Should we add `private_class_method :database_error_classes` here to make that happen?

lib/active_admin/resource/routes.rb
```diff
@@ -42,8 +42,6 @@ def route_uncountable?
         config[:route_collection_name] == config[:route_instance_name]
       end
 
-      private
```
deivid-rodriguez, line 45: In this case, it seems like the intention was to make the `RouteBuilder` class private, but it doesn't work like that. Should we add `private_constant :RouteBuilder` to make that actually happen?

lib/active_admin/resource_controller/decorators.rb
```diff
@@ -57,8 +57,6 @@ def self.wrap(decorator)
           @cache[name] ||= wrap! collection_decorator, name
         end
 
-        private
```
deivid-rodriguez, line 60: Same here, it seems like the code wanted to make `wrap!` private, but it doesn't work like that.

deivid-rodriguez (commented):
It turns out that the duplicate `private` you removed was the only case of a duplication. The other cases seem to be trying to make methods/constants private in an incorrect way.

I guess we could leave this PR as it is, but open a new ticket to actually make these things private? Or maybe just change it here, I just don't want to break anybody's apps, but it seems very unlikely that anybody is actually using these.

@javierjulio What do you think?

deivid-rodriguez:
@brunoarueira Sorry, not sure how to proceed here, a simple PR got complicated :laughing:

deivid-rodriguez (approved):
I'd say let's stay on the safe side and just merge this PR as it is now?

brunoarueira:
> @brunoarueira Sorry, not sure how to proceed here, a simple PR got complicated

@deivid-rodriguez Yeah, when I see those changes proposed by rubocop, I thought it isn't right. Bu anyway, I can make the requested changes at another pull request if you prefer.

javierjulio:
No problem at all! Yes I think this is good. The points at which `private` was removed, it wouldn't have applied anyway since the code there has the same public access level. I think it's good to merge as is. Thanks!

javierjulio (approved):

deivid-rodriguez:
Thanks @brunoarueira!!

PR #6569, opened 2020-11-26
Stop persisting test app between jobs

deivid-rodriguez:
In Github Actions, persisting data between jobs isn't as easy/efficient as it is on CircleCI. The main reason we did this was because generating a test application was very slow under jruby, so by generating it only once per Rails version on a "fast ruby", we can skip that.

In this PR I'm speeding up test application generation under jruby. With this speed up, I think we can generate the test application for every matrix combination.

This should make migrating to Github Actions more straightforward.

This has the extra advantage of potentially detecting issues with generating the test app under jruby, which is something we didn't test before, and could potentially help a contributor working under jruby.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ version: 2.1
 .create_test_app: &create_test_app
   run:
     name: Create test app
-    command: COVERAGE=true PARALLEL_TEST_PROCESSORS=4 bin/rake setup
+    command: COVERAGE=true bin/rake setup
 
 .restore_test_reporter: &restore_test_reporter
   attach_workspace:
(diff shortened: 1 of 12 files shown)


javierjulio:
Sounds good thanks. Should hopefully makes things isolated and a bit more predictable.

deivid-rodriguez:
@javierjulio Since CircleCI is completely stuck now, I'll merge this without a green CI. Then merge #6561 and then we can completely switch off CircleCI. Ok?

deivid-rodriguez:
Also, even if CircleCI hasn't run here, we know everything is green because #6561 (which migrates the CI to Github Actions) is rebased on top of this PR :+1:.

javierjulio:
Yes that sounds good! I'll approve as is and let you merge in. Thanks!

javierjulio (approved):

PR #6583, opened 2020-12-03
Fix `has_many` syntax in forms documentation

krzcho:
has_many needs plural

### DIFF
--- a/docs/5-forms.md
+++ b/docs/5-forms.md
@@ -110,7 +110,7 @@ ActiveAdmin.register Post do
       end
     end
     f.inputs do
-      f.has_many :comment,
+      f.has_many :comments,
                  new_record: 'Leave Comment',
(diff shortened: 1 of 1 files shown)


javierjulio (approved):

deivid-rodriguez (approved):
Thanks!

krzcho:
Actually I am now not sure if it was a typo or not :D
For sure I needed plural (as my model had many) but the description above claims "You can create forms with nested models using the has_many method, even if your model uses has_one:" - so this code could have been an example of that (however not the best one as normally post has many comments)

deivid-rodriguez:
Oh wow, nice catch. That's a weird detail of the API... Should we add an example of the `has_one` usage too?

krzcho:
Nah :)

PR #6584, opened 2020-12-03
Remove sassc and sprockets runtime dependencies

### LINKED ISSUE #6201: Should the sprockets dependency be removed?
**Background**

Hey guys, first of all let me say I appreciate all te amazing work on this gem! It's a great addition to the Rails toolset.
Recently, like many of us, I'm looking into moving over to Webpacker and with #5855 you guys now also support it. But when switching,  Sprockets is still installed due to the dependency in the Gemspec. It's not major, but when a Gem is not used, I'd rather not have it installed at all :)

**Reproduction**

1. Set the `config.use_webpacker` config
2. Run `rails g active_admin:webpacker`
3. Remove `gem 'sprockets'` from your project Gemfile
4. Run `bundle update`

**Actual**

Sprockets 4 is installed and added to your Gemfile.lock

**Expected**

Sprockets to not be used at all in the rails project.

**Cause**

In the Gemspec file there is still a dependency to Sprockets: https://github.com/activeadmin/activeadmin/blob/master/activeadmin.gemspec
```
  s.add_dependency 'sprockets', '>= 3.0', '< 4.1'
```

**Possible solution**

Could this dependency be removed? Should it be removed?
The dependency kind of depends on the config setting, if you installed AA normally, then it's dependent on Sprockets, but with the `:webpacker` installation, it's dependent on Webpacker.

Should the generator maybe add the required gem to the Gemfile?
Or should there be 2 gems: `activeadmin` and `activeadmin_webpacker`
Or should a AA version 3 be release dropping support for Sprockets, having a dependency on Webpacker?

_Just braindumping some possible solutions, but I don't have enough knowledge about the whole of AA to know if they're valid options._

### LINKED ISSUE #6206: ActiveAdmin crashes on start if there's no app/assets directory

### LINKED ISSUE #6222: Make the sassc-rails dependency optional
When using webpacker, the `sassc-rails` gem is pulled in even though it is never going to be used.

This is sort of similar to #6201, except that sassc-rails is *not* pulled by Rails already, so it's more visible.

deivid-rodriguez:
I think now that we support apps using webpacker for assets and js, it makes
sense to remove these dependencies, since these apps don't need them.

This could break people's apps that do need this but are relying on activeadmin
for providing this dependency, so I think we could do it for ActiveAdmin 3.

Thoughts?

Closes #6201.
Closes #6222.
Closes #6206.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ group :development, :test do
   gem "activerecord-jdbcsqlite3-adapter", "~> 60.0", platform: :jruby
 
   gem "sprockets-rails"
+  gem "sassc-rails", "~> 2.1"
 
   gem "formtastic", "~> 4.0.rc1"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,6 @@ PATH
       kaminari (~> 1.0, >= 1.2.1)
       railties (>= 5.2, < 6.1)
       ransack (~> 2.1, >= 2.1.1)
-      sprockets (>= 3.0, < 4.1)
 
 GEM
   remote: https://rubygems.org/
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,6 @@ PATH
       kaminari (~> 1.0, >= 1.2.1)
       railties (>= 5.2, < 6.1)
       ransack (~> 2.1, >= 2.1.1)
-      sassc-rails (~> 2.1)
       sprockets (>= 3.0, < 4.1)
 
 GEM
@@ -478,6 +477,7 @@ DEPENDENCIES
   rubocop-packaging (= 0.5.1)
   rubocop-rails (~> 2.3)
   rubocop-rspec (~> 1.30)
+  sassc-rails (~> 2.1)
   simplecov (= 0.20.0)
   sprockets-rails
   sqlite3 (~> 1.4)
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -27,6 +27,5 @@ Gem::Specification.new do |s|
   s.add_dependency "kaminari", "~> 1.0", ">= 1.2.1"
   s.add_dependency "railties", ">= 5.2", "< 6.1"
   s.add_dependency "ransack", "~> 2.1", ">= 2.1.1"
-  s.add_dependency "sassc-rails", "~> 2.1"
   s.add_dependency "sprockets", ">= 3.0", "< 4.1"
 end
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -27,5 +27,4 @@ Gem::Specification.new do |s|
   s.add_dependency "kaminari", "~> 1.0", ">= 1.2.1"
   s.add_dependency "railties", ">= 5.2", "< 6.1"
   s.add_dependency "ransack", "~> 2.1", ">= 2.1.1"
-  s.add_dependency "sprockets", ">= 3.0", "< 4.1"
 end
--- a/gemfiles/rails_52/Gemfile
+++ b/gemfiles/rails_52/Gemfile
@@ -16,6 +16,7 @@ group :development, :test do
   gem "activerecord-jdbcsqlite3-adapter", "~> 52.0", platform: :jruby
 
   gem "sprockets-rails"
+  gem "sassc-rails", "~> 2.1"
 
   gem "formtastic", "~> 4.0.rc1"
 end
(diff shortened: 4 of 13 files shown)


javierjulio:
I think this is ok as other gems do this so the user has a choice and thus more flexibility on integration. I don't have a strong opinion here on this change or an alternative since in my experience this is used with Rails so one of those dependencies is always included.

deivid-rodriguez:
@javierjulio Yes, that's the case for `sprockets`, which is still pulled by rails, so removing it will make no difference. We could technically keep the `sprockets` line in the gemspec and it would be the same thing, but I thought since we are doing this, we could go all the way.

What makes a difference is removing the `sassc-rails` dependency, because this dependency is no longer pulled by `rails` by default, so by removing we allow `webpacker` users to avoid its installation altogether. And this is a good thing because compiling its extension takes a looooooooooong time, so I can see how it can be frustating to spend a long time installing a dependency that you don't even need. There was another reason concerning the stability of the `sassc` gem (I had to upgrade it in fccc567614e11c2fb1d68c29b7e0e9d316fa2582 because it was segfaulting in CI), but the be honest the latest version seems to be working just fine.

So, my plan here would be to aim to ship activeadmin 3.0 with this change, all the other minor breaking changes that we have pending, plus #6115 which I plan to work on tomorrow. What do you think?

javierjulio:
Sorry, I think there might have been a misunderstanding as I wasn't clear on the last part of my comment as by "one of those dependencies is always included" I was thinking of sass-rails and webpacker, not sprockets and sassc as you mentioned.

What you covered I know and I think it makes sense. That flexibility will benefit all parties. I don't see how this could impact much in a negative way since the default `Gemfile` for Rails still includes `sass-rails` v6 as a minimum which means `sassc-rails` is included anyway as the former is now a lightweight wrapper around the latter. So if you are going the default route for CSS that's all there, but if not you can easily remove it and just use webpacker with this change for ActiveAdmin and not be pulling in an unnecessary dependency.

Is there any reason or way to handle the case where neither are set and to have an explicit error that one must be installed?

deivid-rodriguez:
> Sorry, I think there might have been a misunderstanding as I wasn't clear on the last part of my comment as by "one of those dependencies is always included" I was thinking of sass-rails and webpacker, not sprockets and sassc as you mentioned.

Aaaah, I see :+1:.

> What you covered I know and I think it makes sense. That flexibility will benefit all parties. I don't see how this could impact much in a negative way since the default Gemfile for Rails still includes sass-rails v6 as a minimum which means sassc-rails is included anyway as the former is now a lightweight wrapper around the latter. So if you are going the default route for CSS that's all there, but if not you can easily remove it and just use webpacker with this change for ActiveAdmin and not be pulling in an unnecessary dependency. Win-win. 👍🏻

Yes, I think this should be backwards compatible in most of the cases since as you point out rails generators add one of these by default to the generated `Gemfile`. If for some reason an application was relying on activeadmin to provide this dependency, then I guess this could break, that's why it seems to me that 3.0 is a good opportunity to do this.

> Is there any reason or way to handle the case where neither are installed and to have an explicit error that one must be?

This could be a good idea but I can't spot a place where it would make sense for us to check this, it feels more like something Rails itself could/should do?

javierjulio:
Yeah neither can I. I'm new to handling that kind of approach but figured I'd mention it. Either way I think as long as the changelog states clearly that it is now required that a user set either a dependency on `sass-rails` or `webpacker` in their Gemfile as ActiveAdmin no longer specifies either, that'll be good.

deivid-rodriguez:
Ok! I'll merge this!

PR #6585, opened 2020-12-03
Bundler 2.2 locks the specific platform in the lockfile

deivid-rodriguez:
I forgot to commit this part when I upgraded to the latest bundler prerelease.


javierjulio (approved):
Thanks!

PR #6606, opened 2020-12-11
Use latest rubygems & bundler in CI

deivid-rodriguez:
We released rubygems 3.2.0 and bundler 2.2.0.

Just checking everything is good since we're having a few issues with the final
versions.


deivid-rodriguez:
We got one flaky (shit! :disappointed:), but I don't see how it could be related to these updates. So I restarted CI and it's all good now.

javierjulio (approved):
Sounds good, thanks! Don't worry, I'm sure we'll find them eventually!

Issue #2514, opened 2013-09-28
The AA generator should detect and deal with require_tree


seanlinsley:
This is part of #2513

If AA is installed inside of an app that uses `require_tree .`, the generator should pause and prompt the user to choose an option:
1. Install AA assets in `vendor/assets` instead of `app/assets`
2. Remove the `require_tree` directive from `application.css`
3. Remove the `application.css` file entirely (only appropriate for root-level installs)

ccallebs:
@seanlinsley Is there anything left actionable on this issue?

javierjulio:
Closing out as its gone stale. I'm afraid we are overthinking this. At most what we should do is show a warning on install if we detect that require_tree is in either file that it has to be removed for the admin styles to not be included in other parts of their app.

CelsoDeSa:
## Solution for Rails 5.2.2

In **application.css**

Add:
`*= stub "active_admin"`

Bellow:
`*= require_self`

PR #6676, opened 2021-01-08
Make the AA installer interactive - Issue #2513

### LINKED ISSUE #2513: Interactive AA installer
Since Rails generators have the ability to pause and prompt the user for input, we should be using it to smooth over common configuration issues that people run into. This ticket will be used to keep track of and discuss new features. Here's the list so far:

- [ ] choosing the class name for the user & whether or not to use an existing class
- [ ] creating an install without authentication
- [ ] display warning if `require_tree` detected in application.css or application.js
- [x] ~~default user credentials configuration -> #3457~~ replaced by #3988

LeGorge:
- The installer asks the user for input in whether or not he wants Devise and the name of the user class;
 - It also checks for the 'require_tree .' common problem and prints a warning'
 - Creates an option for no interactivity, which is used in our tests
 - Updates the documentation

closes #2513

I had to change the "hook_for" implementation that was calling the devise generator, so the "interactive part" had more freedom, and for keeping compatibility I created the option --skip-users

I changed all variables "name" to "@name" as now the installer is changing that value on the interaction with the user, and the updated value needs to be coherent everywhere.

For the 'require_tree .' problem, I checked discussions [here](https://github.com/activeadmin/activeadmin/issues/2514) and [here](https://github.com/activeadmin/activeadmin/pull/3159) and ultimately added a warning message for the user with information about the problem and the way of solving it, discussed in those threads.

I was not sure how to implement tests for the interactivity of the installer, and as the current code is not doing tests on the options(only on resulted files), I decided to leave as is, but perhaps that would make for another issue.

I would like to add that this is my first PR ever and I had to learn a lot, so any reviews or comments will be well appreciated.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,7 @@ group :test do
   gem "rails-i18n" # Provides default i18n for many languages
   gem "rspec-rails"
   gem "sqlite3", "~> 1.4", platform: :mri
+  gem "ammeter"
 end
 
 group :release do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -84,6 +84,10 @@ GEM
       zeitwerk (~> 2.3)
     addressable (2.7.0)
       public_suffix (>= 2.0.2, < 5.0)
+    ammeter (1.1.5)
+      activesupport (>= 3.0)
+      railties (>= 3.0)
+      rspec-rails (>= 2.2)
     arbre (1.4.0)
       activesupport (>= 3.0.0, < 6.2)
       ruby2_keywords (>= 0.0.2, < 1.0)
@@ -460,6 +464,7 @@ PLATFORMS
 DEPENDENCIES
   activeadmin!
   activerecord-jdbcsqlite3-adapter (~> 61.0)
+  ammeter
   cancancan
   capybara (~> 3.14)
   chandler (= 0.9.0)
--- a/docs/0-installation.md
+++ b/docs/0-installation.md
@@ -21,7 +21,21 @@ that can be injected into your existing Ruby on Rails application.
 
 ## Setting up Active Admin
 
-After installing the gem, you need to run the generator. Here are your options:
+After installing the gem, you need to run the generator. It will check with you
+things like whether you want to use Devise or not and the name of the User class.
+
+  ```sh
+  rails g active_admin:install
+  ```
+
+If you don't want this interaction, run it with `--no_interaction` and we will use
+default values and create an `AdminUser` class to use with Devise
+
+  ```sh
+  rails g active_admin:install --no_interaction
+  ```
+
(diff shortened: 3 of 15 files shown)


LeGorge:
About i18n: I did not put too much thought on this because the errors in the AA instalation are being returned without it anyways. So no translations for the installer just yet. But let me know if this needs to be corrected here or if it will be turned into an issue of its own.

deivid-rodriguez:
Hi @LeGorge, thanks for this, it looks like a very nice addition.

Some comments that come to mind:

* We have several default changes scheduled for AA 3.0 (hopefully we'll release it eventually). I guess it makes sense to make the installer still non-interactive by default for now and make the switch to default interactivity on 3.0?

* The `require_tree` warnings seem like a different issue, right? I know the issue referenced included that as a task, but to me it seems like something completely separate. Can you move that to a separate PR?

* Regarding tests, I guess we can look at how railties tests mock UI for testing their own options?

LeGorge:
Hello @deivid-rodriguez, thanks...
 - So, first point: I didn't really understand if I have to do something here or if this was more of a piece of information.

 - Second point: I've moved the require_tree warning bit to a different [PR](https://github.com/activeadmin/activeadmin/pull/6847)

 - Third point: Thank you for pointing me to the right direction, I had a great time looking into rails and how they test their generators, as well as how RSpec does it.

Anyway, the biggest problem I faced with the tests was that, as one would imagine, in order to test an install generator we would have to create sample apps and call active_admin:install on them, and our entire test suite is already on top of a customized sample app, so, in the end, I couldn't come up with a new "rake test_application" as I don't quite grasp everything about it yet, but I was able to create the samples I needed from inside the customized test app anyways.

I made sure the generator samples were created inside tmp/test_apps folder, but I understand all this is a bit sub-optimal.

Anyway, I'll be here if more changes are needed.

LeGorge:
Hi @deivid-rodriguez, it seems like something is wrong on your end, the master branch is giving errors. The first time we execute bin/rake, it creates the sample test app and everything goes smoothly, but from the second time on, it skips the creation of the app and a lot of tests fail.

The same happens with my branch. And the CI seems to be following that path.

What are your thoughts on this?

LeGorge:
Hi @deivid-rodriguez 
I had a little bit of trouble with the gemfiles for different versions, but it's ok now.
I think this -0.1% in total-coverage is legit. The install generator test was testing if some files were generated in the sample test app, now it's testing on the results of its own "active_admin:install" run, so there's no point in testing the test-code(creation of said sample test app)

javierjulio:
Not sure if it's worth the overhead of supporting and testing an interactive installer with the options we support. I'd be happy with an improvements to the docs if it helps and if we want to revisit this, it can be reopened as well. Thank you.

PR #6826, opened 2021-03-11
Add webrick to gemfiles

deivid-rodriguez:
Since it's a separate gem on ruby 3.


javierjulio (approved):
I didn't realize we were using it. I see that we set it for Capybara. Thanks!

PR #6847, opened 2021-03-20
Adds warning for users that are using "require_tree ."

LeGorge:
Adds warning for users that are using "require_tree ." in their js and css config files

 - This change is being split from PR #6676

### DIFF
--- a/lib/generators/active_admin/install/install_generator.rb
+++ b/lib/generators/active_admin/install/install_generator.rb
@@ -38,6 +38,15 @@ module ActiveAdmin
       end
 
       def create_assets
+        js_file = File.join(destination_root, "app", "assets", "javascript", "application.js")
+        css_file = File.join(destination_root, "app", "assets", "stylesheets", "application.css")
+        if File.foreach(css_file).grep(/\*= require_tree \./).any? || File.foreach(js_file).grep(/\/\/= require_tree \./).any?
+          say_status(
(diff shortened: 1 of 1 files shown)


lib/generators/active_admin/install/install_generator.rb
```diff
@@ -38,6 +38,15 @@ def setup_routes
       end
 
       def create_assets
+        js_file = File.join(destination_root, "app", "assets", "javascript", "application.js")
+        css_file = File.join(destination_root, "app", "assets", "stylesheets", "application.css")
+        if File.foreach(css_file).grep(/\*= require_tree \./).any? || File.foreach(js_file).grep(/\/\/= require_tree \./).any?
+          say_status(
+            :warning, "The code #{set_color("require_tree .", :red)} in either application.css or application.js files
+              makes Active Admin styles or scripts to be included in other parts of your app.
+              If that is not desired you can stub active_admin at the end of those files.")
```
deivid-rodriguez, line 47: This looks good to me! With the stubbing I guess you're referring to https://github.com/activeadmin/activeadmin/issues/2514#issuecomment-462473973. If that's the case, I think we could be more explicit recommending the exact solution? Also, an alternative solution would be to completely remove the `require_tree .` in favour of explicit requires, correct?

javierjulio:
If this is still a helpful warning to show I will accept it if updated. Let me know and I'll reopen. Thank you.

PR #6872, opened 2021-03-30
Clarify docs on user setup

javawizard:
Note that passing a custom user class to `rails g active_admin:install` can also be used to generate a user class with a customized name. This would have saved me a bit of time trying to figure out how to convince Active Admin to generate a model named `User` instead of `AdminUser` for a new project without pivoting to using Devise's generators instead.


deivid-rodriguez (approved):
Looks perfect @javawizard, thanks!

javierjulio (approved):
Thanks!

PR #6873, opened 2021-03-30
Change Korean i18n that misleading

1000ship:
Replace `%{title} 선택됨` to `선택한 항목 %{title}`

In English, "Delete selected" makes sense what it works.
But in Korean, "삭제 선택됨" is so weird. ( Delete = 삭제, selected = 선택됨 )

> 선택됨(selected) can't be 'noun' in Korean. it can be only 'adjective', thus it must be use "selected stuff", "selected item", "selected things" ... etc

"삭제 선택됨" means, I selected something that name is "delete"
"선택한 항목 삭제" means, exactly I will remove what I selected
('선택한 항목' 삭제 = Delete 'selected item')

So if you don't change these wording, many Koreans can't get understand
Lots of Koreans would click that button, thinking 'what's this button do?'

### DIFF
--- a/config/locales/ko.yml
+++ b/config/locales/ko.yml
@@ -69,7 +69,7 @@ ko:
         one: "성공적으로 1개 %{model}을/를 삭제하였습니다"
         other: "성공적으로 %{count}개의 %{plural_model}을/를 삭제하였습니다"
       selection_toggle_explanation: "(선택 항목 바꾸기)"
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (approved):
Thanks for the fix and the explanation! :heart:

javierjulio (approved):
Thanks!

PR #6874, opened 2021-03-30
Workaround flaky CI

deivid-rodriguez:
The [latest Github Actions environment update](https://github.com/actions/virtual-environments/releases/tag/ubuntu20%2F20210318.0) made our CI flaky. I don't know the specific change that caused it but seems definitely tied to the upgrade. Even if there's no public report, I think it's due to the new image installing chromium from apt/snap sources which seems problematic (it should be installed from source). This suspicion is based on the work being done in this regard since the release: https://github.com/actions/virtual-environments/pull/3029.

I checked that downgrading to ubuntu-18.04 fixes the issue by rerunning things a few times there.

This PR also includes a few other CI changes that I made while investigating this.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,10 +35,10 @@ jobs:
           bundler: none
 
       - name: Install a specific rubygems version
-        run: gem update --system 3.2.14
+        run: gem update --system 3.2.15
 
       - name: Install a specific bundler version
-        run: gem install bundler -v 2.2.14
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
@javierjulio I'm merging this to unblock other PRs. Feel free to leave any comments you may have and I'll address them on a separate PR :+1:.

javierjulio:
No comments, all looks good, thanks! 🙌🏻

PR #6884, opened 2021-04-01
Scope new record instantiation by authorization scope

ngouy:
Lets have a basic structure :
`asset belongs to company`
`company have_many admin_users`
`an admin_user can manage (see/create/update) assets of the company`

this last assertion is be described/implemented into the `AssetPolicy::Scope#resolve` (using pundit)

When we have policy scopes (such as pundit), we are not leveraging them for instantiating (building) new resources (in `#new` and `#create` controller methods)

I face 2 issues because of this:
- issue with my `:create` policy raises me an error when hitting the assets/new URL, as my `create` policy is checking that the asset#company matches the admin_user#company. But company is blank (because we are not leveraging the policy scope).
- I have to overwrite the controller "create" method just to add the scope that I already set (in this case, adding the company_id to the asset).

### DIFF
--- a/spec/unit/resource_controller/data_access_spec.rb
+++ b/spec/unit/resource_controller/data_access_spec.rb
@@ -225,5 +225,16 @@ RSpec.describe ActiveAdmin::ResourceController::DataAccess do
     it "should assign nested attributes once" do
       expect(subject.posts.size).to eq(1)
     end
+
+    context "given authorization scope" do
+      let(:authorization) { controller.send(:active_admin_authorization) }
+
+      it "should apply authorization scope" do
+        expect(authorization).to receive(:scope_collection) do |collection|
+          collection.where(age: "42")
+        end
+        expect(subject.age).to eq(42)
+      end
+    end
   end
 end
--- a/lib/active_admin/resource_controller/data_access.rb
+++ b/lib/active_admin/resource_controller/data_access.rb
@@ -128,7 +128,7 @@ module ActiveAdmin
       #
       # @return [ActiveRecord::Base] An un-saved active record base object
       def build_new_resource
-        scoped_collection.send(
+        apply_authorization_scope(scoped_collection).send(
           method_for_build,
           *resource_params.map { |params| params.slice(active_admin_config.resource_class.inheritance_column) }
         )


deivid-rodriguez (approved):
Nice catch! Not sure if it would also make sense to apply other filtering here, like keeping the current scope or filters, but this one definitely make sense to me :+1:.

spec/unit/resource_controller/data_access_spec.rb
```diff
@@ -225,5 +225,15 @@
     it "should assign nested attributes once" do
       expect(subject.posts.size).to eq(1)
     end
+
+    context "given authorization scope" do
+      let(:authorization) { controller.send(:active_admin_authorization) }
+
+      before { allow(authorization).to receive(:scope_collection) { |collection| collection.where(age: "42") } }
```
javierjulio, line 232: What about using `expect` instead to ensure the method is called so if it's not then the test case fails? By using `allow` it means the method call is optional.
```suggestion
      before { expect(authorization).to receive(:scope_collection) { |collection| collection.where(age: "42") } }
```

ngouy: I have mixed feelings about using `expect` in `before` blocks.
Before block are for context, we should not "test" anything in them.
If we really need to be sure it is called from the scope collection I feel like it would require a proper dedicated test.

I guess that's the difference in approaches between "integration" (it's not really the case here but you get the point) test, rather than pure unit tests.

Here if we have a scope where your age is 42, then your output is a user with age 42. Little chance it came from everything else than the authorization scope. I tend to focus on what's in what's out vs what is called with which arguments.

It's a bit like line 214
```
let(:controller) do
  rc = Admin::UsersController.new
  allow(rc).to receive(:params) do
    params
  end
  rc
end
```

Open to talk about it

javierjulio: I would have preferred that in the test case than a `before` block but no matter. To clarify, I reviewed the added lines only, as I'm not familiar with this file. In my opinion, using `allow` is a red flag because it can cause false positives. I'd rather a test fail loudly if a condition isn't meant which is a requirement for this test case. If that approach is used elsewhere and that's preferred though that's fine. I just point this out of caution.

Could you please rebase and push? Since this was opened we had updated our CI config, sorry.

deivid-rodriguez: I don't have a strong opinion either, specially if using `allow` is already the most common style in this kind of tests. I also sympathize with the idea of making tests less dependent on the implementation, so that if you refactor the implementation the test still passes.

However I agree with @javierjulio that `expect` in this case is better. If it's not necessary to call this method here, and the call ends up being removed in the future to some refactor, keeping the `allow` will allow the test to still pass, but will most likely result in dead test code, since the line will be most likely left in there.

But no big deal, I'm happy with both versions.

ngouy: Then I guess the best of both worlds would be to remove it from the before block and put it in the testing block

Will update and rebase when I found some time thanks for your input guys ✌️

deivid-rodriguez (approved):

javierjulio (approved):
Thanks! ❤️

agrobbin:
Unfortunately, when attempting to ugprade to AA 2.10, I ran into an issue to do with this change. It seems this breaks when working with a ["singleton controller"](https://github.com/activeadmin/inherited_resources#singletons), which is a feature of Inherited Resources. Since [`scoped_collection`](https://github.com/activeadmin/activeadmin/blob/c9695082a328d4f438ee75d7799c6cb54c3907e4/lib/active_admin/resource_controller/data_access.rb#L65-L76) calls out to [`end_of_association_chain`](https://github.com/activeadmin/inherited_resources/blob/f5ceafb10f0ba371e97f9164bc4cff1b887d310b/lib/inherited_resources/base_helpers.rb#L170-L186), it's possible for that to return a single object, rather than a collection-like thing. When that happens, we end up passing that instance in to `apply_authorization_scope`, which is unexpected and will not work properly with the authorization adapter. In the case of Pundit, it will attempt to invoke `Scope#resolve` with that instance, which will not be expecting it as an argument.

For reference, we have a resource that looks like this:

```ruby
ActiveAdmin.register Foo do
  menu false

  belongs_to :bar

  # ...

  controller do
    defaults singleton: true
  end

  # ...
end
```

Before this change, Active Admin + Inherited Resources would, without issue, call `Bar#build_foo` when going to the `/new` action. Now, unfortunately, we hit an exception because our Pundit policy scope is receiving the `Bar` instance as its argument. We are able to work around it by defining our own `build_new_resource`, but I think Active Admin might want to explicitly account for singleton controllers here.

deivid-rodriguez:
Thanks for the feedback! Can you open a new ticket to track this, ideally with repro steps?

agrobbin:
@deivid-rodriguez absolutely! I've opened #7339 with a reproducible test case.

tf:
I realize it has been while, but wanted to note that this change is a sort of breaking change for the CanCan adapter. See #8067.

PR #6889, opened 2021-04-06
Don't include patch level in check names

deivid-rodriguez:
This is to avoid that we need to change our branch configuration every time we change the rubies we test against.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   lint:
-    name: lint (${{ matrix.ruby }})
+    name: lint (${{ matrix.ruby.name }})
     runs-on: ${{ matrix.os }}
 
     timeout-minutes: 15
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
 
       matrix:
-        ruby: [2.7.2]
+        ruby: [{ name: 2.7, value: 2.7.2 }]
         os: [ubuntu-20.04]
 
     steps:
@@ -36,7 +36,7 @@ jobs:
 
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: ${{ matrix.ruby.value }}
           bundler: none
 
       - name: Install a specific rubygems version
@@ -52,7 +52,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ env.BUNDLE_PATH }}
-          key: gems-${{ matrix.os }}-${{ matrix.ruby }}-rails_61-${{ hashFiles('current_gemfile.lock') }}
+          key: gems-${{ matrix.os }}-${{ matrix.ruby.name }}-rails_61-${{ hashFiles('current_gemfile.lock') }}
 
       - name: Install dependencies
         run: bundle install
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
 
       matrix:
-        ruby: [2.7.2]
+        ruby: [{ name: 2.7, value: 2.7.2 }]
```
javierjulio, line 23: The action takes care of using the latest patch version available. Do we still prefer to lock this by patch level?

.github/workflows/ci.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: ${{ matrix.ruby.value }}
           bundler: none
```
javierjulio, line 40: From what I understand, we don't need to install bundler manually and can either specify the version here so this would become:
```suggestion
          bundler: "2.2.15"
```
but also since we have `Gemfile.lock` files, the action by default will use what is specified in the `BUNDLED WITH` section as the bundler version so the bundler key here could be omitted entirely. https://github.com/ruby/setup-ruby#bundler

.github/workflows/ci.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ env.BUNDLE_PATH }}
-          key: gems-${{ matrix.os }}-${{ matrix.ruby }}-rails_61-${{ hashFiles('current_gemfile.lock') }}
+          key: gems-${{ matrix.os }}-${{ matrix.ruby.name }}-rails_61-${{ hashFiles('current_gemfile.lock') }}
```
javierjulio, line 55: It's recommended that we don't cache the gems manually but instead allow the action to do so by specifying `bundler-cache: true` in the config. https://github.com/ruby/setup-ruby#caching-bundle-install-manually If we did that, then that would mean we could remove the following sections:
- Restore cached dependencies
- Cat current Gemfile to a specific file so caching works
- Install dependencies

javierjulio (approved):
Thanks! I'll leave this approved for you to merge. I have comments but they are out of scope for this and shouldn't hold it up.

Now that I'm more familiar with the ruby/setup-ruby action, we do more work that the action actually takes care of. I don't know though if we have a reason or preference for it the way it is now. I just wanted to point some things out since it would cut some cruft from this file.

deivid-rodriguez:
Yes, the comments are all correct!

I do prefer to lock ruby by patch level, specially since we have suffered from regressions in ruby patch level versions not long ago, but other than than I agree with simplifying these things!

I'll merge this and then simplify the workflows in a separate PR :+1:.

.github/workflows/ci.yaml
deivid-rodriguez: I think at some point I implemented these tricks because `ruby/setup-ruby` couldn't properly handle separate Gemfiles, but I'm pretty sure it handles them just fine now, we'll see :)

.github/workflows/ci.yaml
javierjulio: We will maintain this as we prefer to lock by patch level.

.github/workflows/ci.yaml
javierjulio: That makes sense, thanks! 🙌🏻

javierjulio:
That sounds great to me, thanks David! ❤️

PR #6890, opened 2021-04-07
The `env` helper method is dead and not working

paracycle:
Based on [this bug report](https://github.com/Shopify/tapioca/issues/280) in Tapioca (which generates interface files for dynamically generated methods for various DSLs), I was able to figure out that this `env` helper method is dead-code in ActiveAdmin. Moreover, if the `env` helper method is ever called, it will fail since there is no actual method to proxy to anymore.

For more details, see my analysis here: https://github.com/Shopify/tapioca/issues/280#issuecomment-815070624

### DIFF
--- a/lib/active_admin/base_controller.rb
+++ b/lib/active_admin/base_controller.rb
@@ -6,7 +6,6 @@ module ActiveAdmin
   # It implements ActiveAdmin controllers core features.
   class BaseController < ::InheritedResources::Base
     helper ::ActiveAdmin::ViewHelpers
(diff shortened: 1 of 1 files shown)


javierjulio (approved):
Dug through the issue and the change looks good. Thanks again for reporting @paracycle! 

If anything where this was used, the `unsupported_browser` component, is something I'd like to get rid of.

deivid-rodriguez (approved):
Looks good to me too!

PR #6900, opened 2021-04-09
Let `ruby/setup-ruby` handle installing and caching

deivid-rodriguez:
Try out @javierjulio's suggestions in #6889.


deivid-rodriguez:
I need to verify that caching works, and the current CI failure is actually a bug in bundler. Other than that this seems fine.

deivid-rodriguez:
Bundler issue will be fixed by https://github.com/rubygems/rubygems/pull/4523.

javierjulio (approved):
Awesome, thanks!

PR #6906, opened 2021-04-14
Make `permit_params` and `belongs_to` order independent

deivid-rodriguez:
Currently `belongs_to` needs to be defined before `permit_params` or otherwise belongs to params will fail to be allowed and forms on belongs to resources will fail to render with errors like:

```
found unpermitted parameter: :user_id (ActionController::UnpermittedParameters)
./lib/active_admin/resource_dsl.rb:75:in `block (2 levels) in permit_params'
./lib/active_admin/resource_controller/data_access.rb:133:in `build_new_resource'
./lib/active_admin/resource_controller/data_access.rb:116:in `build_resource'
```

I feel this mistake is quite common, given it was happening on our own demo application. I think we can fix it by lazily loading params without causing any issues.

### DIFF
--- a/features/belongs_to.feature
+++ b/features/belongs_to.feature
@@ -103,6 +103,23 @@ Feature: Belongs To
     And I should see the attribute "Body" with "This is the body"
     And I should see the attribute "Author" with "Jane Doe"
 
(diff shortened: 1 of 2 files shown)


javierjulio (approved):
This makes sense, thanks!

PR #6909, opened 2021-04-15
 feat: Customize the title in batch action.

lanzhiheng:
Sometime i would like to custom the title of batch_action.


deivid-rodriguez:
I don't have a strong opinion but I guess it's fine. @javierjulio?

javierjulio:
@deivid-rodriguez I don't see why the i18n option wouldn't suffice for this as I think that would be better but then I tried thinking of what would be similar to this elsewhere in AA and that would probably be filters as a label can be given to override the i18n default. So I think this would be ok. The one thing I'd bring up is considering using `label:` instead of `title:` for the option name as we use the former mostly and with menu-like components.

deivid-rodriguez:
Yeah, I had the same initial thought ("just use i18n"), but then I thought having to use i18n on monolingual apps to configure these simple things is a bit annoying an somewhat a "feature abuse" so this thing seems simple and could come handy sometimes. Agreed on using `label` to name things consistently :+1:.

deivid-rodriguez:
Although forms seems to already have a `title` option that behaves the same way, so we're already inconsistent, so I guess we should choose the name that makes most sense for batch actions.

PR #6911, opened 2021-04-21
Optimize count query for `pagination_total: false` option

deivid-rodriguez:
This is a rebased version of https://github.com/activeadmin/activeadmin/pull/5389.

Some of the changes proposed there have already been introduced, but there was some extra stuff that still seems useful. I guess this is lacking a spec, but seems ok to me other than that.


mgrunberg:
I updated the PR.
The changes are the following:

- Reimplement `reorder` using `except` because `ActiveAdmin::ResourcesControler::Decorator::Wrapper` already handles it. My understanding is that the result is the same.
- Add a spec to test the behaviour: check that we produces a count with limit but without order by.
- Call `.reset` on the existing test before testing decorated scenario. This way we are sure that the input is not loaded (a side-effect of multiple calls to `paginated_collection` in the same spec.

javierjulio (approved):
Thanks! If I understand right, this is an optimization for when using the `pagination_total: false` option? I've updated the PR title to reflect that for the release notes. Feel free to change if you think it can be improved as that's my only concern here. I think it would help to make any pagination query improvements we can so thank you for seeing this through.

Issue #6914, opened 2021-04-23
I18n::InvalidPluralizationData translation data {:one=>"inregistrare", :other=>"inregistrari"} can not be used with :count => 2. key 'few' is missing.


lubosch:
### Did you find a bug?

Pagination does not work in romanian lang.

```
I18n::InvalidPluralizationData
translation data {:one=>"inregistrare", :other=>"inregistrari"} can not be used with :count => 2. key 'few' is missing.
```

### Expected behavior

It should work in romanian lang

### Actual behavior

it throws error 500

### How to reproduce

set RO lang and use paginated_collection.

rails-i18n/lib/rails-i18n/common_pluralizastions/romanian.rb defined 'few' requirement for RO lang

### How to solve

Add 'few' key to config/locales/ro.yml  -> 'active_admin.pagination.entry.few' should exist

deivid-rodriguez:
Pull request appreciated!

lubosch:
but I don't speak romanian :D google translator is enough?

deivid-rodriguez:
I guess a potentially poor translation is better than a runtime error, yeah!

PR #6915, opened 2021-04-23
Add few key to RO pagination.entry

lubosch:
should fix https://github.com/activeadmin/activeadmin/issues/6914


javierjulio (approved):
Noted that this will fix a runtime error so fine as is.

When translating the original text in Google Translate it comes out as "registration" in English. For the en.yml locale we use "entry" and "entries" though so this text might be incorrect.

lubosch:
it can mean different things, look at this from a random dictionary site:

image

deivid-rodriguez:
Thank you!

PR #6922, opened 2021-04-28
Include print styles in main stylesheet

### LINKED ISSUE #6908: Issue with Webpacker 6 (beta) and print.css
When configuring activeadmin to use webpacker, and using webpacker 6 an issue arise with activeadmin/print.css

Maybe renaming it to activeadmin_print.css would avoid having a directory structure and simplify this ?

removing this stylesheet fixes the issue.

deivid-rodriguez:
No need to do this separately which makes an extra http request, plus this has the benefit of delaying print styles until needed. I’ve been doing this for a long time now and as its been well supported to include print styles using a media block. The HTML5 Boilerplate has stressed this as a good default approach for many years now. https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L205-L209

Extracted from #3862.

Closes #6908.

### DIFF
--- a/app/assets/stylesheets/active_admin/_base.scss
+++ b/app/assets/stylesheets/active_admin/_base.scss
@@ -4,38 +4,45 @@
 
 // Partials
 @import "./typography";
-@import "./header";
-@import "./forms";
-@import "./components/comments";
-@import "./components/flash_messages";
-@import "./components/date_picker";
-@import "./components/tables";
-@import "./components/batch_actions";
-@import "./components/modal_dialog";
-@import "./components/blank_slates";
-@import "./components/breadcrumbs";
-@import "./components/dropdown_menu";
-@import "./components/buttons";
-@import "./components/grid";
-@import "./components/links";
-@import "./components/pagination";
-@import "./components/panels";
-@import "./components/columns";
-@import "./components/scopes";
-@import "./components/status_tags";
-@import "./components/table_tools";
-@import "./components/index_list";
-@import "./components/unsupported_browser";
-@import "./components/tabs";
-@import "./pages/logged_out";
-@import "./structure/footer";
-@import "./structure/main_structure";
-@import "./structure/title_bar";
 
-body {
-  @include sans-family;
-  line-height: 1.5;
-  font-size: 72%;
-  background: $body-background-color;
-  color: $text-color;
+@media screen {
+  @import "./header";
+  @import "./forms";
+  @import "./components/comments";
+  @import "./components/flash_messages";
+  @import "./components/date_picker";
+  @import "./components/tables";
+  @import "./components/batch_actions";
+  @import "./components/modal_dialog";
+  @import "./components/blank_slates";
+  @import "./components/breadcrumbs";
+  @import "./components/dropdown_menu";
+  @import "./components/buttons";
+  @import "./components/grid";
+  @import "./components/links";
+  @import "./components/pagination";
+  @import "./components/panels";
+  @import "./components/columns";
+  @import "./components/scopes";
+  @import "./components/status_tags";
+  @import "./components/table_tools";
+  @import "./components/index_list";
+  @import "./components/unsupported_browser";
+  @import "./components/tabs";
+  @import "./pages/logged_out";
+  @import "./structure/footer";
+  @import "./structure/main_structure";
+  @import "./structure/title_bar";
+
+  body {
+    @include sans-family;
+    line-height: 1.5;
+    font-size: 72%;
+    background: $body-background-color;
+    color: $text-color;
+  }
+}
+
+@media print {
+  @import "./print";
 }
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -164,8 +164,7 @@ module ActiveAdmin
     private
 
     def register_default_assets
-      register_stylesheet "active_admin.css", media: "screen"
-      register_stylesheet "active_admin/print.css", media: "print"
+      register_stylesheet "active_admin.css"
       register_javascript "active_admin.js"
     end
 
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/active_admin/_base.scss
```diff
@@ -4,38 +4,45 @@
...
+  @import "./structure/title_bar";
+
+  body {
+    @include sans-family;
+    line-height: 1.5;
+    font-size: 72%;
+    background: $body-background-color;
+    color: $text-color;
+  }
+}
```
javierjulio, line 44: The admin CSS is already loaded as a screen media stylesheet so this isn't necessary. It would be best to remove the screen media wrapper part.

```suggestion
@import "./header";
@import "./forms";
@import "./components/comments";
@import "./components/flash_messages";
@import "./components/date_picker";
@import "./components/tables";
@import "./components/batch_actions";
@import "./components/modal_dialog";
@import "./components/blank_slates";
@import "./components/breadcrumbs";
@import "./components/dropdown_menu";
@import "./components/buttons";
@import "./components/grid";
@import "./components/links";
@import "./components/pagination";
@import "./components/panels";
@import "./components/columns";
@import "./components/scopes";
@import "./components/status_tags";
@import "./components/table_tools";
@import "./components/index_list";
@import "./components/unsupported_browser";
@import "./components/tabs";
@import "./pages/logged_out";
@import "./structure/footer";
@import "./structure/main_structure";
@import "./structure/title_bar";

body {
  @include sans-family;
  line-height: 1.5;
  font-size: 72%;
  background: $body-background-color;
  color: $text-color;
}
```

app/assets/stylesheets/active_admin/_base.scss
```diff
@@ -4,38 +4,45 @@
...
+
+@media print {
```
javierjulio, line 46: I had this line defined in the `_print.scss` partial instead. I'm not sure but I figured that is required for it to work because I think Sass only allows `@import "...";` at the outermost level.

javierjulio (commented):
I think this will be worthwhile, thank you!

app/assets/stylesheets/active_admin/_base.scss
deivid-rodriguez: Actually, this is an unintentional mistake on my part, I'm loading the (now only) stylesheet with `media="screen"`, but the stylesheet should be loaded for all media types, right?

If I understood your idea correctly, these blocks are correct because otherwise all the styles will be applied in the "print view" which is unintented (at least not how it works currently).

lib/active_admin/application.rb
```diff
@@ -164,8 +164,7 @@ def controllers_for_filters
     private
 
     def register_default_assets
-      register_stylesheet "active_admin.css", media: "screen"
-      register_stylesheet "active_admin/print.css", media: "print"
+      register_stylesheet "active_admin.css"
```
deivid-rodriguez, line 167: Here's the mistake I mentioned at https://github.com/activeadmin/activeadmin/pull/6922#discussion_r628010825.

This should be:

```suggestion
      register_stylesheet "active_admin.css", media: "all"
```

deivid-rodriguez (commented):
Thanks for having a look @javierjulio.

I replied to your comments, the PR as it was didn't work correctly. I'll push an update with a working version so you can re-review.

app/assets/stylesheets/active_admin/_base.scss
deivid-rodriguez: Yes, I moved it ouside, because since I was wrapping the rest of the styles under `@media screen`, moving it outside brought a nice simmetry in my opinion.

Not sure about Sass an not outermost imports, it seems to work just fine.

lib/active_admin/application.rb
deivid-rodriguez: Fix pushed!

deivid-rodriguez:
This should be working now, but it seems that I need to fix some specs. Anyways, the idea of the outermost `@media screen` and `@media print` blocks is to make the styles mutually exclusive, which is how it was working up until now. I only left the `normalize` and `typograhpy` imports outside of the blocks because they were the only ones loaded by both stylesheets.

deivid-rodriguez:
Alright, I spotted yet another error. I had changed the load order for the print media, so that a couple of SASS variables were being set _after_ the typography import was using them, so they were not getting applied. I think the current version should be 100% equivalent to what we have now, and only provide the improvement of loading a single stylesheet.

app/assets/stylesheets/active_admin/_base.scss
javierjulio: Yes that makes sense about the separate screen/print blocks because in #3862 the print styles were updated so many of the elements that were originally displayed in print mode weren't anymore. Thanks!

app/assets/stylesheets/active_admin/_base.scss
javierjulio: Ok great, thank you!

javierjulio:
Yes that sounds right and makes sense. The two I had changed/removed were `normalize` and `typography` partials when working on the print portion. Thank you! I will approve and leave for you to merge when best. 🙌🏻

javierjulio (approved):
Thank you! ❤️

baarkerlounger:
@deivid-rodriguez would it be possible to do another release to include this now that it's merged?

deivid-rodriguez:
Sorry for the delay, I'll work on a release 👍.

Adeynack:
@deivid-rodriguez Was it released since? I can't find references to issue `6908` in any release note. I am upgrading an app from Rails 6 to Rails 7 and I get `Sprockets::FileNotFound: couldn't find file 'active_admin/print.css'`. Trying to understand the issue and came upon this PR. I am using the latest published Active Admin GEM (no version limit in my Gemfile).

Issue #6883, opened 2021-04-01
Filters values aren't using authorization scopes


ngouy:
### Did you find a bug?

filter values doesn't use policy scopes.

I understand there is a manual way to scope values of a filter, but as a collection of any resources indexes are by default scoped by the policy adapter (when there is one), so should be filter values, it can lead to a serious data breach.

I am even surprised to found nothing on the internet about that

Another possibility is that I am missing something?

### Expected behavior

On the user index page, In the filter "company", I should only see my own company

### Actual behavior

I see all companies

### How to reproduce
```rb
class Company < ApplicationModel
  has_many :users
  has_many :admin_users
end

class AdminUser < ApplicationModel
  # my actual active_admin / devise user, the one that log into the portal
end

class User < ApplicaitonModel
  belongs_to :company
end

ActiveAdmin.register(User) {} # on this page, I have a company filter, I can try to filter by any company (I see all existing companies)
```

deivid-rodriguez:
Are you up for working on a fix @ngouy?

ngouy:
mhm 🤔 

Have a lot of work ATM (including in my week end 😢 ) so I don't have any time for this
I would be glad to work on this but it may be in 1 month

deivid-rodriguez:
That'd be great, you help would be super appreciated if you end up finding some time!

ngouy:
@deivid-rodriguez quick question from a kinda noob active admin open sourcer

What would be the best way to have access to `active_admin_config` or `active_admin_namespace`or the current controller to have access to the `authorization_adapter` from the ActiveAdmin::Inputs::Filters::SelectInput? 

Once this part is done the rest is formality

deivid-rodriguez:
Hello!

I'm not fully sure, would have to dig in deeper. I think at this level you only have access to the underlying activerecord object you're creating the form for, so I guess you'd need to explicitly pass the correctly scoped collection to the form manually.

So your later idea would seem like the right way to go to me.

ngouy:
Thank you for your guidance @deivid-rodriguez

Will take a look at it

ngouy:
@deivid-rodriguez I've sent the first shot
There are no tests yet, and var namings etc... are far from definitive, but on my current project it seems to work fine

What do you think about this approach? I'm waiting for even a slight input before "committing" myself into righting the final version and tests

deivid-rodriguez:
[...]

PR #6923, opened 2021-04-29
#6883 Force scope on default filters collection

ngouy:
When we use default filters, or filter of an existing relationship without providing a collection, we enforce the collection to be scope
It is useful when you are using authorization scope to define what resources your current user can access, even in filter values.


ngouy:
fix #6923

ngouy:
don't really have the bandwidth and not using AA anymore. Feel free to use the code as you wich but I close this

deivid-rodriguez:
So sorry for not attending this earlier :(. Will try to move this forward at #7511, thanks for your contribution 💪.

PR #6944, opened 2021-05-07
Remove bundler binstub

deivid-rodriguez:
All the logic for the selection of the version of bundler is now built-in to rubygems, so this is unnecessary.


javierjulio (approved):
This is great, thanks!

PR #6952, opened 2021-05-07
Remove unnecessary conditions

deivid-rodriguez:
Mergify respects branch protection settings by default.

Not having to maintain this list reduces the chance of errors, like the one I made that caused #6951 to not be automatically merged (I included two checks that don't exist).


javierjulio (approved):
Thanks!

deivid-rodriguez:
Worked perfectly this time, let's see how it goes!

PR #6956, opened 2021-05-12
feat: support alias_attribute as valid keys for filter

brunohkbx:
## Summary

Given a model with an alias_attribute like so:
`alias_attribute :title_alias, :title`

Then it will be possible to use the alias as a valid filter:
`filter :title_alias, as: :select`


javierjulio:
Does this work with other filter types? What if using a Ransack based filter? This just needs a rebase and push to run CI. Thank you.

PR #7000, opened 2021-06-03
Support Turbo loading

lanzhiheng:
I think it is a good idea to support turbo loading.


nitsujri:
@lanzhiheng manually pulled all these changes in. One area that's still broken seems to be form validation errors. For some reason they do redirects back to the same page and Turbo is throwing the [`Form responses must redirect to another location`](https://github.com/hotwired/turbo-rails/issues/12#issuecomment-754629885). 

Not sure if you've seen this yourself, but we ended up disabling turbo completely for now.

I'm not a contributor at all, so just some feedback.

javierjulio:
We've had another PR in #7451 for Turbo support. If anyone still interested in that, likely using these as a starting point will help.

javierjulio:
Part of the changes here are in #7002

PR #7002, opened 2021-06-04
Update config.register_javascript with options support

lanzhiheng:
I have replace the turbolink with turbo in my project. so I think it will be better to generate `<script>` tag like below

``` html
<script src='assets/admin.js' data-turbo-track='reload'></script>
```

However, the method `register_javascript` don't support `options` like `  config.register_stylesheet`

I think it will be a good idea to make it customizable

``` ruby
config.register_stylesheet 'active_admin.css', media: 'screen', 'data-turbo-track': 'reload'
config.register_javascript 'active_admin.js', 'data-turbo-track': 'reload'
```

### DIFF
--- a/app/views/layouts/active_admin_logged_out.html.erb
+++ b/app/views/layouts/active_admin_logged_out.html.erb
@@ -12,11 +12,11 @@
       <%= stylesheet_link_tag style, **options %>
     <% end %>
   <% end %>
-  <% ActiveAdmin.application.javascripts.each do |path| %>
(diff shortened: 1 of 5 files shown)


nitsujri:
Would love to get this one loaded as hotwire becomes more readily used.

deivid-rodriguez (approved):
Looks good to me, can you rebase this to make sure CI is ok?

javierjulio (approved):
Thanks!

PR #7026, opened 2021-06-25
Use type=number for filters

c960657:
Use `<input type=number>` for filter inputs.


javierjulio:
Thank you. This is worth considering for mobile support. I'm working on the next major version so I will keep this in mind then.

javierjulio:
To follow up here, I'm not sure why this was even needed since the filter already inherits from Formtastic's NumberInput which uses an HTML input type=number control. Perhaps the issue was just on the styling side? If so that is already addressed on the default branch. Without these changes, mobile was already showing the numbers at the top of the keyboard since we were always using a type=number input.

PR #7085, opened 2021-07-30
Displays a action_item only if the user is authorized

### LINKED ISSUE #7084: Should not display any span element for an action item that the user is not able to use
### Did you find a bug?

ActiveAdmin renders empty `span` elements for actions that the user is not authorized to use. That is a problem when you need to customize the css style of those elements.

### Expected behavior

Render no `span` element inside the `action_items` `div` element

```
<div class="action_items">
</div>
```

Or even no `div.action_items` at all

### Actual behavior

It renders empty span for each action that te user is not allowed to use

```
<div class="action_items">
   <span class="action_item"></span>
</div>
```

### How to reproduce

* Take an app with 2 users with roles, 1 user with `admin` who can CRUD a resource `Post` and another with a role `reader`, who can only read thoses posts.
* Connect with the user `reader`
* Go to the `Post` resource index
* Inspect the dom within `.title_bar .titlebar_right .action_items` 

https://www.loom.com/share/e72c9af9d26344baa0fcc933084cedbb

shanser:
fix #7084


Fivell (approved):
looks reasonable

shanser:
Thanks @Fivell
Any idea when this PR could be merged ? (We are counting on it for https://github.com/betagouv/eva-serveur)
Thanks ! 🙏

shanser:
👋 @Fivell Do you have any idea when this PR could be merged ?

Thanks

shanser:
👋  @Fivell do you have any news for me ? :)

Fivell (approved):

varyonic (commented):
(edited) I think the original lines could be simply transposed, so:
```ruby
def add_default_destroy_action_item
  if controller.action_methods.include?("destroy") && authorized?(ActiveAdmin::Auth::DESTROY, resource)
    add_action_item :destroy, only: :show do
      localizer = ActiveAdmin::Localizers.resource(active_admin_config)
      link_to localizer.t(:delete_model), resource_path(resource), method: :delete,
                                                                       data: { confirm: localizer.t(:delete_confirmation) }
    end
  end 
end
```

Fivell:
hello @varyonic, do you mean this?

```ruby
def add_default_show_action_item
  if controller.action_methods.include?("destroy") 
    add_action_item :destroy, only: :show, if: -> { authorized?(ActiveAdmin::Auth::DESTROY, resource) } do
      localizer = ActiveAdmin::Localizers.resource(active_admin_config)
      link_to localizer.t(:delete_model), resource_path(resource), method: :delete,
                                                                       data: { confirm: localizer.t(:delete_confirmation) }
    end
  end 
end   
  ```

Fivell:
@shanser hello, second approval required to make contribution merged, however maintainers usually do not want to accept such contributions without rspec part

shanser:
I understand @Fivell. This part is not tested at all, that's why I didn't add any spec.
You would argue that it could be the right time to do it but to be honest I don't really know where to start because the setup would be really complicated.

varyonic:
I clarified my original suggestion above.  A test in `action_items_spec` would be nice to have, but the only similar test I see where `authorized?` is stubbed false is in `auto_link_spec.rb`  I'm not against merging with no test, however.

javierjulio:
I'd accept this if we can get tests for each. The suggestions are a good place to start. Thanks!

javierjulio:
The original issue is resolved on the default branch. An element wrapper is no longer rendered per action, the action item contents are rendered directly into the container. I'd still like to do this change since logically it makes sense regardless of the original issue. Adding tests may be tricky but should be doable. I would like to accept this with updated tests.

javierjulio:
This will be incorporated into the default branch soon. Thank you.

PR #7087, opened 2021-08-01
Adds hide_attribute to Resource DSL for hiding attribute from UI.

garo:
The hide_attribute is useful if the default index and show interfaces
would be perfect without any customisation, except that one of more
attributes (database columns) should just be completely hidden from the UI.

The hide_attribute works like the global Application filter_attributes
setting which allows hiding certain attributes from every resource.

### DIFF
--- a/lib/active_admin/resource.rb
+++ b/lib/active_admin/resource.rb
@@ -67,6 +67,9 @@ module ActiveAdmin
     # nil to not decorate.
     attr_accessor :decorator_class_name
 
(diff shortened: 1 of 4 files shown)


javierjulio:
Personally, I'm not a fan in general with this and the defaults we use. I wouldn't accept this as I think ideally all pages should be declared by the AA user rather than dynamically generated. I wouldn't want the overhead of having to maintain this as well and the various combinations or side effects introduced. If other maintainers want though, it can be reopened.

PR #7127, opened 2021-08-27
Show ransackable_scopes filters in search results 

### LINKED ISSUE #6887: Custom Filter Empty Summary
Hi,
I'm using activeadmin with custom filters:

`
filter :pratica_filter, label: "Pratica Contiene", as: :select, collection: TipoPratica.all.sort_by{|t| t.descrizione.capitalize }.map{|t| [t.descrizione.capitalize, t.id] }
...
def self.pratica_filter(value)
		# My query here...
end

def self.ransackable_scopes(_auth_object = nil)
		%i[pratica_filter]
end
`
It works but the problem is that I see an empty summary like this (even if the filter is active):

Screenshot from 2021-04-06 12-04-36

Please note that "Nessuno" in English means "Nobody, no one".

What I'm doing wrong? It seems that activeadmin doesn't understand the name of the filter and its value in the summary.
Any help will be highly appreciated.
Thank you,
Marco

### LINKED ISSUE #7042: No current filters when use filters in sidebar
### Desc

The issue seems to be similar to the [#4342](https://github.com/activeadmin/activeadmin/issues/4342) but a  bug appears for ActiveAdmin 2.9.0.

image

1. Add filters filter (Example):

```
ActiveAdmin.register Admin::MyResource, as: 'Resource Page' do
  filter :field_eq, label: 'Resource field'
  filter :ransackable_field_eq, label 'Ransackable scope field'
end

###################

module Admin
  class MyResource

    class << self

      def ransackable_field_eq(value)
        merge('condition with value')
      end

      def ransackable_scopes(_auth_object = nil)
        %i[ransackable_field_eq]
      end

    end

  end
end

```

### Actual behavior

When I use both filters on a page I see only the first one in the Current Filters (Search Status section).

### Expected behavior

Both filters are shown in the Current Filters

### Why does it happen

In the `ActiveAdmin::Filters::ActiveSidebar`

I can see the line:
```
active_filters = ActiveAdmin::Filters::Active.new(active_admin_config, assigns[:search])

active_filters.filters.each do |filter|
  # render filter
end
```
where `assigns[:search]` is an instance of `Ransack::Search`.

The first search criteria is stored in the Ransack::Search instance -> @base -> conditions,
But the second one is in the Ransack::Search instance -> @scope_args

Looks like @scope_args do not appear in the active filters so they are not rendered on a page.
Could you please fix it o provide a workaround?

vlad-psh:
In AA, we can filter our data using `ransackable_scopes` https://github.com/activerecord-hackery/ransack#using-scopesclass-methods
Data filtering works just fine, but the "Search Results" sidebar shows "Current filters: None"

This PR adds proper indication of those selected scopes. For example:

Screen Shot 2021-08-27 at 21 33 13

This PR

* Fixes #7042
* Fixes #6887 (duplicating)

### DIFF
--- a/lib/active_admin/filters/active_sidebar.rb
+++ b/lib/active_admin/filters/active_sidebar.rb
@@ -11,32 +11,47 @@ module ActiveAdmin
 
       def block
         -> do
-          active_filters = ActiveAdmin::Filters::Active.new(active_admin_config, assigns[:search])
-          span do
-            if current_scope
-              h4 I18n.t("active_admin.search_status.current_scope"), style: "display: inline"
-              b scope_name(current_scope), class: "current_scope_name", style: "display: inline"
-            end
-            div style: "margin-top: 10px" do
-              h4 I18n.t("active_admin.search_status.current_filters"), style: "margin-bottom: 10px"
-              ul do
-                if active_filters.filters.blank?
-                  li I18n.t("active_admin.search_status.no_current_filters")
-                else
-                  active_filters.filters.each do |filter|
-                    li filter.html_options do
-                      span do
-                        text_node filter.label
-                      end
-                      b do
-                        text_node to_sentence(filter.values.map { |v| pretty_format(v) })
-                      end
-                    end
-                  end
-                end
+          def scope_block(current_scope)
+            return unless current_scope
+
+            h4 I18n.t("active_admin.search_status.current_scope"), style: "display: inline"
+            b scope_name(current_scope), class: "current_scope_name"
+          end
+
+          def filters_list(active_filters, active_scopes)
+            h4 I18n.t("active_admin.search_status.current_filters")
+            ul do
+              if active_filters.filters.blank? && active_scopes.blank?
+                li I18n.t("active_admin.search_status.no_current_filters")
+              else
+                active_filters.filters.each { |filter| filter_item(filter) }
+                active_scopes.each { |name, value| scope_item(name, value) }
               end
             end
           end
+
+          def filter_item(filter)
+            li filter.html_options do
+              span filter.label
+              b to_sentence(filter.values.map { |v| pretty_format(v) })
+            end
+          end
+
+          def scope_item(name, value)
+            filter = active_admin_config.filters[name.to_sym]
+            label = filter ? filter[:label] : name.titleize
+
+            li class: "current_filter_#{name}" do
+              span "#{label} #{Ransack::Translate.predicate('eq')}"
+              b value
+            end
+          end
+
+          active_filters = ActiveAdmin::Filters::Active.new(active_admin_config, assigns[:search])
+          active_scopes = assigns[:search].instance_variable_get("@scope_args")
+
+          scope_block(current_scope)
+          filters_list(active_filters, active_scopes)
         end
       end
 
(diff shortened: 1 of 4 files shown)


penghouho:
I verified that the change can solve my issue of ransackable_scopes is not considered in search result, subscribed this PR.

lib/active_admin/filters/active_sidebar.rb
```diff
@@ -11,32 +11,47 @@ def initialize
...
+          def scope_block(current_scope)
+            return unless current_scope
+
+            h4 I18n.t("active_admin.search_status.current_scope"), style: "display: inline"
+            b scope_name(current_scope), class: "current_scope_name"
+          end
+
+          def filters_list(active_filters, active_scopes)
+            h4 I18n.t("active_admin.search_status.current_filters")
+            ul do
```
Fivell, line 23: @vlad-psh  👍 , what do you think can methods be defined out of the block ?

vlad-psh: @Fivell Thanks for your comment.

Initially I wanted to make changes as minimal as possible, so I just edited our existing (pretty long) `block` method (just splitted internal lambda with loops into smaller methods for readability; max nesting level were reduced from 13 to just 8)

But I agree, it would be even better to extract those methods outside lambda. So I created new commit adding new view component `active_filters_sidebar_content` to render content for active filters.

Please check, what do you think?

Fivell: @vlad-psh  looks good thanks, lets wait for CI

Insercomhai: To get the process it would be even better to extract those methods outside lambda. So I created new commit adding new view component active_filters_sidebar_

Insercomhai (approved):

Fivell:
@vlad-psh  thanks! @deivid-rodriguez  seems it is ready to merge

Fivell (approved):

pilgunboris:
Hello, is there some estimation of when this fix might be released?

lib/active_admin/filters/active_sidebar.rb
```diff
@@ -10,34 +10,7 @@ def initialize
...
       def block
-        -> do
-          active_filters = ActiveAdmin::Filters::Active.new(active_admin_config, assigns[:search])
-          span do
-            if current_scope
-              h4 I18n.t("active_admin.search_status.current_scope"), style: "display: inline"
-              b scope_name(current_scope), class: "current_scope_name", style: "display: inline"
-            end
-            div style: "margin-top: 10px" do
-              h4 I18n.t("active_admin.search_status.current_filters"), style: "margin-bottom: 10px"
```
deivid-rodriguez, line 21: I see that you removed `style:` parameters and the wrapper div. I see how that inline CSS felt hacky but the previous spacing felt a bit more visually pleasant at least to me. Should we restore it and consider cleaning that up separately?

deivid-rodriguez (commented):
Just a small comment, looks good. A bit of messing with ransack internals, but seem to works fine so let's do it.

lib/active_admin/filters/active_sidebar.rb
vlad-psh: Ok, sure )
Fixed and rebased.

deivid-rodriguez: Thanks!

Issue #6787, opened 2021-03-02
id_column should not link to show page if policy forbids show?


adibsaad:
### Did you find a bug?

I'm using AA with the Pundit adapter. I have a resource and in its policy class, I define `show?` as `false` and `index?` as `true`. I'm able to go to the index page for the resource (which is correct), and when I see the rows of each record for that resource, I noticed the `View` button in the actions column is not visible, which is also correct, since `show?` is `false`. However, the `id_column` still links to the Show page.

### Expected behavior

The ID column should not link to the show page, since the policy does not allow it. It should only show the ID of the record without a link to anywhere.

deivid-rodriguez:
Agreed, a PR would be appreciated :+1:.

rahulknojha:
@deivid-rodriguez I'm working on it.

jmschp:
Hi,

I believe there is more to it than just the link on the `id` column.

Should the user also be block to access the show page, even if we manually  insert the url in the browser?

javierjulio:
We have a PR ready for this and just needs an update with tests. Thank you.

PR #7136, opened 2021-09-15
Fix: id_column link to show page if policy forbids

rahulknojha:
This PR is raised to fix => https://github.com/activeadmin/activeadmin/issues/6787


javierjulio:
This is just for not including a link to the show page but if you were to access the show page using that link and the policy class forbids it, you do get redirected with a flash message right? If so I think this isn't much of a worry. I would still accept it though since it's good, just needs tests. Thank you.

PR #7144, opened 2021-09-24
Fix pundit namespace detection

vlad-psh:
Given a following config:

```
```

ActiveAdmin will search policies with `Admin::` namespace.
But if model already contains the configured namespace name (anywhere in the model name), AA won't tell Pundit to search policy with configured namespace.

For example, if we have model called `ShopAdmin`, we'll look up for `ShopAdminPolicy` instead of `Admin::ShopAdminPolicy`

Replacing condition in https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/pundit_adapter.rb#L61 to something like this, fixes this bug:

```diff
- if default_policy_namespace && !object.class.to_s.include?(default_policy_namespace.to_s.camelize)
+ if default_policy_namespace && !object.class.to_s.match?(/^#{default_policy_namespace.to_s.camelize}::/)
```

### DIFF
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -31,13 +31,11 @@ module ActiveAdmin
     end
 
     def retrieve_policy(subject)
-      case subject
-      when nil then Pundit.policy!(user, namespace(resource))
-      when Class then Pundit.policy!(user, namespace(subject.new))
-      else Pundit.policy!(user, namespace(subject))
-      end
+      Pundit.policy!(user, namespace(policy_target(subject)))
     rescue Pundit::NotDefinedError => e
-      if default_policy_class
+      if (policy = compat_policy(subject))
+        policy
+      elsif default_policy_class
         default_policy(user, subject)
       else
         raise e
@@ -57,8 +55,41 @@ module ActiveAdmin
 
     private
 
+    def policy_target(subject)
+      case subject
+      when nil then resource
+      when Class then subject.new
+      else subject
+      end
+    end
+
+    # This method is needed to fallback to our previous policy searching logic.
+    # I.e.: when class name contains `default_policy_namespace` (eg: ShopAdmin)
+    # we should try to search it without namespace. This is because that's
+    # the only thing that worked in this case before we fixed our buggy namespace
+    # detection, so people are probably relying on it.
+    # This fallback might be removed in future versions of ActiveAdmin, so
+    # pundit_adapter search will work consistently with provided namespaces
+    def compat_policy(subject)
+      target = policy_target(subject)
+
+      return unless default_policy_namespace &&
+        target.class.to_s.include?(default_policy_module) &&
+        (policy = Pundit.policy(user, target))
+
+      policy_name = policy.class.to_s
+
+      Deprecation.warn "You have `pundit_policy_namespace` configured as `#{default_policy_namespace}`, " \
+        "but ActiveAdmin was unable to find policy #{default_policy_module}::#{policy_name}. " \
+        "#{policy_name} will be used instead. " \
+        "This behavior will be removed in future versions of ActiveAdmin. " \
+        "To fix this warning, move your #{policy_name} policy to the #{default_policy_module} namespace"
+
+      policy
+    end
+
     def namespace(object)
-      if default_policy_namespace && !object.class.to_s.include?(default_policy_namespace.to_s.camelize)
+      if default_policy_namespace && !object.class.to_s.match?(/^#{default_policy_module}::/)
         [default_policy_namespace.to_sym, object]
       else
         object
@@ -75,6 +106,10 @@ module ActiveAdmin
 
     def default_policy_namespace
       ActiveAdmin.application.pundit_policy_namespace
+    end
+
+    def default_policy_module
+      default_policy_namespace.to_s.camelize
     end
 
   end
--- a/spec/unit/pundit_adapter_spec.rb
+++ b/spec/unit/pundit_adapter_spec.rb
@@ -91,6 +91,34 @@ RSpec.describe ActiveAdmin::PunditAdapter do
       auth.authorized?(:index)
     end
 
+    context "when model name contains policy namespace name" do
+      include_context "capture stderr"
+
+      before do
+        allow(ActiveAdmin.application).to receive(:pundit_policy_namespace).and_return :pub
+        namespace.register(Publisher)
+        ActiveSupport::Deprecation.behavior = :stderr
+      end
+
+      after do
+        ActiveSupport::Deprecation.behavior = :raise
+      end
+
+      it "looks for a namespaced policy" do
+        expect(Pundit).to receive(:policy!).with(anything, [:pub, Publisher]).and_return(DefaultPolicy.new(double, double))
+        auth.authorized?(:read, Publisher)
+      end
+
+      it "fallbacks to the policy without namespace" do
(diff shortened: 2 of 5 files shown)


deivid-rodriguez:
Makes sense to me, can you add a spec?

vlad-psh:
> Makes sense to me, can you add a spec?

@deivid-rodriguez  Thanks for your comment.

Sure, I'll add specs for this change, but if possible, I'd like to hear your thoughts regarding this PR first.

The thing is, although this PR fixes policies lookup to be working as expected, it probably also will break existing policies for some users.
I believe there are users who has all AA policies in `app/policies/admin/*`, but policy for `ShopAdmin` (for example) in `app/policies/shop_admin_policy.rb` instead of namespaced `app/policies/admin/shop_admin_policy.rb` just because it worked that way before.

What do you think? How can we made this transition smooth?

deivid-rodriguez:
I see, nice catch!

So, I guess we should keep both conditions, the old one and the new one. If the old one is met, but the new one is not met, then we should keep current behavior and print a warning, so that users can accomodate their policy to the proper convention. Something like

```ruby
def namespace(object)
  if default_policy_namespace && !object.class.to_s.include?(default_policy_namespace.to_s.camelize)
    if object.class.to_s.match?(/^#{default_policy_namespace.to_s.camelize}::/)
      [default_policy_namespace.to_sym, object]
    else
      warn "<explain the issue and how to fix it>"
      object
    end
  else
    object
  end
end
```

Makes sense?

deivid-rodriguez:
Ping @vlad-psh, still interested?

vlad-psh:
@deivid-rodriguez Thanks for your suggestion!

I found out that actual logic would be a little bit more complicated

We need first search namespaced policy (eg: `Admin::ShopAdminPolicy`), if this policy weren't found, we can try to search for a policy without namespace (eg: `ShopAdminPolicy`) but only if `default_namespace` is configured and model name contains namespace in the name.

We'll show a warning only if policy were found at second try.

Otherwise, we fallback to `default_policy` (as we always did)

Please, take a look, what do you think?

If all good, I'll write the specs for it

vlad-psh:
Rebased onto current `master` to see if tests are still passing

lib/active_admin/pundit_adapter.rb
```diff
@@ -57,8 +55,47 @@ def format_action(action, subject)
...
+      target = policy_target(subject)
+
+      if target.class.to_s.include?(default_policy_namespace.to_s.camelize)
+        policy = Pundit.policy(user, target)
+
+        if policy
+          policy_name = policy.class.to_s
+          ns_name = default_policy_namespace.to_s.camelize
+
+          warn "WARN: You have `pundit_policy_namespace` configured, " \
```
deivid-rodriguez, line 85: We have a module for this

```suggestion
          Deprecation.warn "You have `pundit_policy_namespace` configured, " \
```

Also probably worth mentioning the actual value configured?

```suggestion
          Deprecation.warn "You have `pundit_policy_namespace` configured as #{default_policy_namespace}, " \
```

spec/unit/pundit_adapter_spec.rb
```diff
@@ -91,6 +91,34 @@ def resolve
...
+      before do
+        allow(ActiveAdmin.application).to receive(:pundit_policy_namespace).and_return :pub
+        namespace.register(Publisher)
+      end
+
+      around do |example|
+        original_stderr = $stderr
+        $stderr = StringIO.new
+        example.run
+        $stderr = original_stderr
```
deivid-rodriguez, line 104: This trick is also used at `spec/unit/action_builder_spec.rb`. Can we extract it to a helper, like `capturing_stderr(&example)` or something?

deivid-rodriguez (commented):
It looks really good, I only left some minor comments.

Also I noticed that our test app has this issue since we have an `AdminUser` policy at the top level. Can you move the policy to fix the warnings being introduced?

vlad-psh:
@deivid-rodriguez Thank you for review 👍
Fixed everything except following two things:

> Also I noticed that our test app has this issue since we have an AdminUser policy at the top level. Can you move the policy to fix the warnings being introduced?

As I can see, we don't have configured `config.pundit_policy_namespace` in our test app. Can you point me where you saw new warnings? Those probably are just some special cases and needed to be fixed individually 🤔

```diff
- warn "WARN: You have `pundit_policy_namespace` configured, " \
+ Deprecation.warn "You have `pundit_policy_namespace` configured, " \
```

I tried this, but `Deprecation.warn` will throw an exception. I thought this shouldn't be a fatal error, but just a simple annoying warning to the console 🤔

deivid-rodriguez:
Yes, Pundit is not enabled by default (perhaps it should be?) but can be enabled by setting `authorization_adapter` to `ActiveAdmin::PunditAdapter` and `pundit_default_policy = "ApplicationPolicy"` in activeadmin configuration's of the generated test application.

That configuration should live at the `tmp/development_apps/rails_70/config/initializers/active_admin.rb` file generated by running `bin/rake local s`. If you make those changes and restart the application, visiting `http://localhost:3000/admin/admin_users` should trigger the warning (because the `admin_users_policy` is not scoped to `admin`).

> I tried this, but Deprecation.warn will throw an exception. I thought this shouldn't be a fatal error, but just a simple annoying warning in the console 🤔

Interesting... I tried the above and didn't raise in the development application. I think this depends on the warning configuration of each specific application, you probably have `config.activesupport.deprecation` set to `:raise`? See https://guides.rubyonrails.org/configuring.html#config-active-support-deprecation

vlad-psh:
> Interesting... I tried the above and didn't raise in the development application. I think this depends on the warning configuration of each specific application, you probably have config.activesupport.deprecation set to :raise? See https://guides.rubyonrails.org/configuring.html#config-active-support-deprecation

@deivid-rodriguez Thanks for directing me, I found the reason. It's in https://github.com/activeadmin/activeadmin/blob/master/spec/rails_helper.rb#L29
This option configured explicitly to `:raise` for rspec.

Should I change rspec example to catch an error instead of capturing stderr?
I'm not sure because this will be a little bit different behavior comparing to the scenario with default configuration (which is `:stderr`)

> Yes, Pundit is not enabled by default (perhaps it should be?) but can be enabled by setting authorization_adapter to ActiveAdmin::PunditAdapter and pundit_default_policy = "ApplicationPolicy" in activeadmin configuration's of the generated test application.

I think this is fine as it is now. We don't have to configure `pundit_default_policy` in our test app.
Or if we do, we should move all the policies (not just a single `AdminUserPolicy`) into our configured namespace.
I don't think it is necessary for test app. What do you think?

deivid-rodriguez:
Regarding deprecations, the motivation for having `:raise` in our test suite is to make sure we get all deprecations fixed right when deprecations are first introduced, so that we can easily add support for future Rails versions later. But I understand how this caused confusion here since it deviates from the standard configuration for test environments. So, not fully sure what's best, what do you think?

Regarding "fixing" deprecations in the test app, you're totally right, it's ok as it is, nevermind.

vlad-psh:
@deivid-rodriguez I decided to just add `ActiveSupport::Deprecation.behavior = :stderr` for these few tests (and to restore it back to `:raise` in the `after` block). Should be a good compromise for everybody 🤓

deivid-rodriguez:
Sounds good! I'll have a final look soon and get this shipped with the next release. Thanks so much!

deivid-rodriguez:
I just added a very minor style tweak (keep the variable to capture the exception named as `e` instead of `error`, for consistency with the other similar code in this file, and to keep the fix more focused) and rebased this.

Thanks so much for the outstanding work! 💪

deivid-rodriguez (approved):

Issue #7196, opened 2021-11-22
Support Rails 7


thanhtoan1196:
Can't install activeadmin using rails 7.0.0.alpha2, please help

Logs
```
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies....
Bundler could not find compatible versions for gem "actionpack":
  In snapshot (Gemfile.lock):
    actionpack (= 7.0.0.alpha2)

  In Gemfile:
    activeadmin was resolved to 0.4.0, which depends on
      meta_search (>= 0.9.2) was resolved to 1.1.0.pre2, which depends on
        actionpack (~> 3.1.0.alpha)

    rails (~> 7.0.0.alpha2) was resolved to 7.0.0.alpha2, which depends on
      actionpack (= 7.0.0.alpha2)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.

Bundler could not find compatible versions for gem "rails":
  In snapshot (Gemfile.lock):
    rails (= 7.0.0.alpha2)

  In Gemfile:
    activeadmin was resolved to 1.0.0.pre3, which depends on
      rails (< 5.0, >= 3.2)

    turbo-rails (>= 0.7.11) was resolved to 0.8.3, which depends on
      rails (>= 6.0.0)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.

Bundler could not find compatible versions for gem "railties":
  In snapshot (Gemfile.lock):
    railties (= 7.0.0.alpha2)

  In Gemfile:
    activeadmin was resolved to 1.0.0.pre5, which depends on
      railties (< 5.1, >= 3.2)

    rails (~> 7.0.0.alpha2) was resolved to 7.0.0.alpha2, which depends on
      railties (= 7.0.0.alpha2)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

Fivell:
related https://github.com/activeadmin/activeadmin/issues/7175

tagliala:
I'm trying to update activeadmin to use rails 7 (https://github.com/tagliala/activeadmin/pull/1), but there are some blocking dependencies:

- [inherited_resources](https://github.com/activeadmin/inherited_resources) activeadmin/inherited_resources#707
- [ransack](https://github.com/activerecord-hackery/ransack) activerecord-hackery/ransack#1266 
[...]

PR #7232, opened 2021-12-18
[Rails 7] Get rid of db-query-matchers gem

alejandroperea:
I've seen in this issue https://github.com/activeadmin/activeadmin/issues/7196 that `db-query-matchers` gem is incompatible with Rails 7 and looks no longer maintained.

I've introduced a custom Rspec matcher that subscribes to `sql.active_record` event and checks if the passed query matches (or not) with the executed ones.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 require "simplecov" if ENV["COVERAGE"] == "true"
 
+require_relative "support/matchers/perform_database_query_matcher"
+
 RSpec.configure do |config|
   config.disable_monkey_patching!
   config.filter_run focus: true
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,6 @@ group :test do
   gem "cuprite"
   gem "capybara"
   gem "webrick"
-  gem "db-query-matchers"
 
   gem "simplecov", require: false # Test coverage generator. Go to /coverage/ after running tests
   gem "cucumber-rails", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -158,9 +158,6 @@ GEM
       activerecord (>= 5.a)
       database_cleaner-core (~> 2.0.0)
     database_cleaner-core (2.0.1)
-    db-query-matchers (0.10.0)
-      activesupport (>= 4.0, < 7)
-      rspec (~> 3.0)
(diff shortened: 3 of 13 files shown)


varyonic:
Interesting.  This dependency was introduced by #5811  The latest publisher of `db-query-matchers` is still active using Ruby on GitHub so I think too soon to say abandoned.

rafaelfranca:
Is this dependency worth though? It is only used in two tests, which can be easily rewritten to check the same thing.

javierjulio:
If we can do the same thing through ActiveRecord notification subscription, I'd be ok with that and would approve. Ideally, we'd still want the database query check for these test cases.

rafaelfranca:
> If we can do the same thing through ActiveRecord notification subscription

That is how db-query-matchers is implemented.

I agree we should keep the SQL check.

@alejandroperea do you want to improve the test to use the `ActiveSupport::Notification.subscribe('sql.active_record')`?

alejandroperea:
@rafaelfranca Sure! I'll take a look this weekend 👍

alejandroperea:
@rafaelfranca Done! It is the first time I use `ActiveSupport::Notifications` (super cool, btw), I hope I did it properly 🙏 

I've added an Rspec matcher for doing this, just in case we want to use it in another place in the future.

javierjulio (approved):
Thanks!

spec/spec_helper.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 require "simplecov" if ENV["COVERAGE"] == "true"
 
+Dir[File.expand_path("#{File.dirname(__FILE__)}/support/matchers/*.rb")].each { |file| require file }
```
deivid-rodriguez, line 4: My only piece of feedback/nitpick would be that `require_relative "support/matchers/perform_database_query_matcher"` feel simpler and more explicit to me. I also liked the fact that this was only required lazily when needed and I think it might've been helpful to understand that this gem was only needed for two specs.

deivid-rodriguez (approved):
Just a very small feedback that you're free to address but I'm good with the current version too. Thanks so much!

alejandroperea:
I agree @deivid-rodriguez ! Change applied 👍 

Thanks for the review! :)

deivid-rodriguez:
Thanks so much!!

PR #7234, opened 2021-12-21
[Rails 7] Bump Railties

baarkerlounger:
Bump railties for Rails 7


tagliala:
Hi, there is an ongoing issue with some comments at #7196. Before upgrading ActiveAdmin itself, there are some dependencies to fix

Here it is an ongoing PR: tagliala/activeadmin#1

PR #7236, opened 2021-12-21
Drop support for Ruby 2.5

alejandroperea:
Ruby 2.5 arrived to EOL almost 9 months ago. Could be the moment to drop the support.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -83,19 +83,6 @@ jobs:
 
         deps: [rails_52, rails_60, rails_61, rails_61_turbolinks, rails_61_webpacker]
 
-        include:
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Hi, I'll slowly get back to working on ActiveAdmin. I think it's fine to drop support for Ruby 2.5, yeah.

deivid-rodriguez (approved):
This seems fine to me. I'll also propose dropping Rails 5.2 and Rails 6.0 independently.

deivid-rodriguez:
It's cumbersome to support both Ruby 2.5 and Ruby 3.1 due to https://github.com/mikel/mail/pull/1439, so that's another reason to drop support.

javierjulio (approved):
Thanks!

PR #7235, opened 2021-12-21
Add Rails 7 Support

tagliala:
# ⚠️  Do not rely on this branch for deployment ⚠️

Hi, this is a work in progress and I'm opening this PR as a reference to #7196 

I'm trying to update activeadmin to use Rails 7, but there are some blocking dependencies:

### ✅  No known issues
- [x] Incompatible with Server Timing middleware: https://github.com/activeadmin/activeadmin/issues/7196#issuecomment-1002878544 #7262
- [x] Autoloading issues when `before_action` option is enabled:  https://github.com/activeadmin/activeadmin/issues/7196#issuecomment-1004789728 https://github.com/activeadmin/activeadmin/pull/7235#issuecomment-1030803856 #7340
### ✅ Dependencies checklist

- [x] [inherited_resources](https://github.com/activeadmin/inherited_resources) activeadmin/inherited_resources#707
- [x] [ransack](https://github.com/activerecord-hackery/ransack) activerecord-hackery/ransack#1266 
- [x] [cucumber-rails](https://github.com/cucumber/cucumber-rails) cucumber/cucumber-rails#523 cucumber/cucumber-rails/pull/522 - #7366 #7369 
- [x] [rails-i18n](https://github.com/svenfuchs/rails-i18n) svenfuchs/rails-i18n#957 svenfuchs/rails-i18n#958
- [x] [db-query-matchers](https://github.com/civiccc/db-query-matchers) civiccc/db-query-matchers#42 - downgrades to an old version without issues #7232
- [x] [arbre](https://github.com/activeadmin/arbre) activeadmin/arbre#314 - downgrades to an old version **with** issues - **ISSUES FIXED, WAITING FOR RELEASE**

### ✅ Removed dependencies
- [x] [activerecord-jdbc-adapter](https://github.com/jruby/activerecord-jdbc-adapter) jruby/activerecord-jdbc-adapter#1103

At the moment, there are no stable JRuby versions supporting MRI 2.7, the minimum required version to run Rails 7.0. There was too much effort in trying to build against jruby-head, so JRuby dependency has been temporarily removed

#### 1. Changes to Gemfile

Add the following to your Gemfile:

```rb
# NOTE: DO NOT RELY ON THIS BRANCH FOR DEPLOYMENT!
gem 'activeadmin', github: 'tagliala/activeadmin', branch: 'feature/railties-7' # FIXME: revert to stable
```

### Tests

- [x] I cannot replicate the zeitwerk failure I see on CI on my local development machine (specs are green)
[...]

PR #7275, opened 2022-01-20
Test with Ruby 3.0 and Ruby 3.1

deivid-rodriguez:
Try to slowly get back to helping out here.

I want to get Ruby 3.0 and Ruby 3.1 officially tested.

Some changes are also proposed separately at #7235 so I added @tagliala as a co-author.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,7 @@ jobs:
           bundler-cache: true
 
       - name: Install a specific rubygems version
-        run: gem update --system 3.2.17
+        run: gem update --system 3.3.5
 
       - name: Generate docs
         run: bin/rake docs:build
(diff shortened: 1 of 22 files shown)


deivid-rodriguez:
👍 I'll switch to strings.

Not sure about Rails support, I guess not, I will make sure and whatever not supported.

deivid-rodriguez:
This one should be ready from my side! I want to drop support for Rails 5.2 and Rails 6.0, but I guess that can be done/discussed separately.

Also I'm slimming down our CI matrix a bit by running webpacker and turbo links checks only against the latest supported Ruby, so we need to reconfigure our default branch settings to remove some required checks now.

javierjulio (approved):
All sounds good to me on the Rails support side. Thanks David! ❤️

PR #7289, opened 2022-01-21
Small follow up to ruby 2.5 support drop

deivid-rodriguez:
We were testing against a Ruby 2.5-compatible JRuby, so we need to bump that too.


Fivell (approved):
@deivid-rodriguez  thanks !

PR #7291, opened 2022-01-21
Remove unused lets

alejandroperea:
Very small PR for cleaning some unused lets. I was expecting to find more, as this gem is big, but very few found.


deivid-rodriguez (approved):

javierjulio (approved):
Thanks!

PR #7292, opened 2022-01-22
Remove unused spec templates pundit policies folder

alejandroperea:
The folder `spec/support/templates/policies/` seems useless, I've run the tests locally and it is passing, let's see in the CI.

In addition, `module NoActiveModel` has been removed from the specs as it is unused since: b7b7401466daf463c696deaefba43aca7e5ef7f6


alejandroperea:
Nevermind, it is used. Closing it.

PR #7293, opened 2022-01-23
Drop rails 5.2 support

deivid-rodriguez:
It's hard for us to support a big range of Rails versions, and Rails 5.2 is essentially not maintained anymore except for severe security issues. So I propose that we move on.


deivid-rodriguez:
This one should be ready. Codeclimate coverage checks didn't run not sure why, and branch settings need to be edited to reflect the new required checks.

deivid-rodriguez:
Alright, as soon as I changed the required checks, the code coverage checks went through 👍.

Fivell (approved):
👍

deivid-rodriguez:
Thanks!

PR #7302, opened 2022-01-30
Bump i18n version to 1.9.1 as current version has been yanked

alejandroperea:
The current version of `i18n` (1.9.0) gem is no longer in Rubygems, looks like has been yanked. A new patch version has been released reverting the code that was not working properly in the previous version.

Screenshot 2022-01-30 at 09 55 58

So, this PR fixes the following error:

Screenshot 2022-01-30 at 09 46 33

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -215,7 +215,7 @@ GEM
       actionpack (>= 5.2)
       activesupport (>= 5.2)
     highline (2.0.3)
-    i18n (1.9.0)
+    i18n (1.9.1)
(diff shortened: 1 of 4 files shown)


javierjulio (commented):
Thanks! We just need to update the rails_60 gemfile too and this should be good to go.

alejandroperea:
@javierjulio You're right! I missed that one when pushing the commit, now the change is committed in Rails 6.0 also 👍

javierjulio (approved):
Thanks!

PR #7304, opened 2022-01-31
Remove unused module NoActiveModel in specs

alejandroperea:
`NoActiveModel` module is dead code and can be safely removed from the specs

Dead since: b7b7401466daf463c696deaefba43aca7e5ef7f6


deivid-rodriguez (approved):
❤️

javierjulio (approved):
Awesome, thank you!

PR #7307, opened 2022-02-04
Add test for helper method with kwargs call in form fails on Ruby 3

javierjulio:
When a helper method accepts keyword arguments and **they are given**, required or not, this causes a failure in Ruby 3.

> wrong number of arguments (given 2, expected 1) (ActionView::Template::Error)
> ./tmp/test_apps/rails_61/app/helpers/time_helper.rb:3:in `format_time'

Note if the keyword arguments are all optional and not given, then no error occurs. I've added a second test case that indicates that it passes while the former case is failing but should pass.

This is due to an Arbre issue https://github.com/activeadmin/arbre/issues/312 but I was not able to figure out how to create a failing test in Arbre's test suite. Any attempts didn't recreate a failed test but it's easy to replicate on ActiveAdmin's end.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -519,4 +519,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   2.3.5
+   2.3.6
--- a/features/new_page.feature
+++ b/features/new_page.feature
@@ -56,6 +56,36 @@ Feature: New Page
     And I should see the attribute "Title" with "Hello World"
     And I should see the attribute "Body" with "This is the body"
 
+  Scenario: A form where calling a helper method with given kwargs is successful
+    Given a configuration of:
+    """
+      ActiveAdmin.register Post do
+        form do |f|
+          f.inputs "Publishing" do
+            f.input :published_date, input_html: { "data-time" => format_time(Time.current, format: :short) }
+          end
+          f.actions
+        end
+      end
+    """
+    And I follow "New Post"
+    Then I should see a fieldset titled "Publishing"
+
+  Scenario: A form where calling a helper method with no kwargs is successful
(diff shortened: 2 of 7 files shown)


deivid-rodriguez:
Arbre fix over at https://github.com/activeadmin/arbre/pull/318 seems to be working! 🎉

javierjulio:
Awesome thanks David! ❤️

Gemfile
```diff
@@ -64,3 +64,4 @@ group :docs do
 end
 
 gemspec path: "."
+gem "arbre", github: "activeadmin/arbre", branch: "fix-keyword-delegation-in-form-component"
```
Fivell, line 67: @javierjulio  thanks, are we going to merge it or first we will update arbre ?

Fivell (approved):

Gemfile
deivid-rodriguez: I think we can merge and release arbre, and then update this?

Fivell: @deivid-rodriguez  mee too

javierjulio: Thanks guys! I will clean up the test cases I added before merging. If we release the Arbre gem before then, feel free to update here or I can do that.

javierjulio: Ok, all done updating tests so this is ready.

deivid-rodriguez:
@javierjulio I released arbre, can you update this to test against the newly released version?

javierjulio:
Yes I will do that now, thanks!

deivid-rodriguez:
@javierjulio Mergify updated arbre automatically in separated pull requests, so I rebased this to only add the spec. Thanks so much for this!

PR #7338, opened 2022-02-16
Allow tables/panels to scroll horizontally when they get too wide.

gregschmit:
I've been using Active Admin and the only issue I've had is that while I don't necessarily need a "fully" responsive design so the app works on phones, sometimes when the index table is really wide or on tablets, the entire page is pushed off the right side of the page. I made some improvements to the CSS to just allow the tables to start to scroll horizontally when they hit that point, and would like to contribute these changes back.

To be clear, these are really the only rules I'm applying; though I put them where they seemed to make sense in context:

```css
#wrapper {
  table-layout: fixed;
  .panel_contents { overflow-x: auto; }
  .index_as_table { overflow-x: auto; }
  .attributes_table { overflow-x: auto; }
}
```

I noticed that there weren't many comments in the CSS files, but let me know if these directives should be commented to explain why they are there.

I can also revert the changes and put these grouped together somewhere or into a `_responsive.scss` if you would prefer, though that seemed to go against the design of the current CSS rules.

### DIFF
--- a/app/assets/stylesheets/active_admin/components/_tables.scss
+++ b/app/assets/stylesheets/active_admin/components/_tables.scss
@@ -63,23 +63,30 @@ table.index_table {
 
 // --------- Tables inside Panels
 
-.panel_contents table {
-  margin-top: 5px;
-  th {
-    padding-top: 10px;
-    background: none;
-    color: $primary-color;
-    @include no-shadow;
-    @include text-shadow;
-    text-transform: uppercase;
-    border-bottom: 1px solid #ccc;
+.panel_contents {
+  overflow-x: auto;
+
+  table {
+    margin-top: 5px;
+    th {
+      padding-top: 10px;
+      background: none;
+      color: $primary-color;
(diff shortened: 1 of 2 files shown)


javierjulio:
Thanks for the submission! I don't think this is something we should include. It's simple enough for others to include as needed if they want it but as a default I don't think it should be there.

gregschmit:
That sucks; it really makes the default UI significantly better IMO. The default UI is really crappy if tables are wider than the window. Am I not even able to convince you with a demo?

gregschmit:
It's also really jarring because the window fits the show and edit views almost always, but then snaps off the screen when you go to the index view.

javierjulio:
I've tried this before when I first started #3862 years ago so its not new to me. No issue with that as one can recreate on our demo app. There are major issues with UI, very dated CSS, etc. that in general needs a massive revamp which is why I don't think this should go in as a default. It's simple enough for someone to add in as they need it and should be opt in.

PR #7340, opened 2022-02-17
Remove deprecation warning using controller filters inside initializer

### LINKED ISSUE #5772: Zeitwerk autoloading deprecation on Rails 6.0.0.pre1
### Expected behavior

Upon upgrading a Rails 5.2.3 app to Rails 6.0.0.pre1 and regenerating the ActiveAdmin initializer, I expect to see no deprecation warnings.

### Actual behavior

When running `rspec` (v3.8) on Rails 6.0.0.pre1 and Ruby 2.5.5 I get the following deprecation warning from [Zeitwerk](https://github.com/fxn/zeitwerk) the new default autoloader in Rails 6:

```
DEPRECATION WARNING: Initialization autoloaded the constants ApplicationHelper, ActiveAdmin::ViewsHelper, [...snip lots of other classes...], InheritedResources::Base, Base, DeviseController, Devise::SessionsController, Devise::PasswordsController, Devise::UnlocksController, Devise::RegistrationsController, and Devise::ConfirmationsController.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload ApplicationHelper, for example,
the expected changes won't be reflected in that stale Module object.

These autoloaded constants have been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
```
When I comment out the ActiveAdmin initializer, there is no deprecation warning. 

### How to reproduce

Run `rspec` (v3.8), or anything that will output deprecations, on Rails 6.0.0.pre1 and Ruby 2.5.5.

mgrunberg:
Defining Action filters (`before_action`, `skip_before_action`, etc) using `ActiveAdmin.before_action` on the initializer file produces the following warning.

```
DEPRECATION WARNING: Initialization autoloaded the constants ApplicationHelper, TimeHelper, DeviseHelper, ApplicationController, InheritedResources::Base, DeviseController, Devise::SessionsController, Devise::PasswordsController, Devise::UnlocksController, Devise::RegistrationsController, and Devise::ConfirmationsController.

Being able to do this is deprecated. Autoloading during initialization is going to be an error condition in future versions of Rails.
```

NOTE: On Rails 7.0 it produces the following error.

```
lib/active_admin/base_controller/authorization.rb:3:in `<module:ActiveAdmin>': uninitialized constant InheritedResources::Base (NameError)
Did you mean?  Base64
```

ActiveAdmin module forwards the filter call to all controllers. That means that we are using classes meant to be autoloaded inside the initializer. That's the source of the warning and it's not allowed starging Rails 7.x (see the second red box of https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#config-autoload-paths).

This PR defers the forward until controllers are loaded by using ActiveSupport.on_load hook.

Closes #5772.

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -150,8 +150,8 @@ module ActiveAdmin
     AbstractController::Callbacks::ClassMethods.public_instance_methods.
       select { |m| m.match(/_action\z/) }.each do |name|
       define_method name do |*args, &block|
-        controllers_for_filters.each do |controller|
-          controller.public_send name, *args, &block
+        ActiveSupport.on_load(:active_admin_controller) do
+          public_send name, *args, &block
         end
       end
     end
--- a/lib/active_admin/base_controller.rb
+++ b/lib/active_admin/base_controller.rb
@@ -77,5 +77,6 @@ module ActiveAdmin
       { controller: controller, action: action }
     end
 
+    ActiveSupport.run_load_hooks(:active_admin_controller, self)
   end
 end
--- a/lib/active_admin/devise.rb
+++ b/lib/active_admin/devise.rb
@@ -53,22 +53,32 @@ module ActiveAdmin
 
     class SessionsController < ::Devise::SessionsController
       include ::ActiveAdmin::Devise::Controller
+
+      ActiveSupport.run_load_hooks(:active_admin_controller, self)
     end
 
     class PasswordsController < ::Devise::PasswordsController
       include ::ActiveAdmin::Devise::Controller
+
+      ActiveSupport.run_load_hooks(:active_admin_controller, self)
     end
 
     class UnlocksController < ::Devise::UnlocksController
       include ::ActiveAdmin::Devise::Controller
+
+      ActiveSupport.run_load_hooks(:active_admin_controller, self)
     end
 
     class RegistrationsController < ::Devise::RegistrationsController
       include ::ActiveAdmin::Devise::Controller
+
+      ActiveSupport.run_load_hooks(:active_admin_controller, self)
     end
 
     class ConfirmationsController < ::Devise::ConfirmationsController
       include ::ActiveAdmin::Devise::Controller
+
+      ActiveSupport.run_load_hooks(:active_admin_controller, self)
     end
 
     def self.controllers_for_filters


mgrunberg:
I can't find a good way to write a test for this. The issue is at boot time, calling `ActiveAdmin.before_action` on the initializer. It's totally fine to do it inside a test.

The initializers run very early on test setup, even on feature tests. I tried to write a feature test that restart/reload the server but couldn't make it.

I test this PR manually on each rails version by adding: `config.before_action -> (controller) { Rails.logger.info "HEY, before action #{controller.class.name}##{action_name}" }` to `config/initializers/active_admin.rb`

tagliala:
Hi,

I can see that all Rails 6.1 builds are failing. Errors seem related to this change:

```undefined method `prepend_before_action' for #<Class:0x000055f32e0131b0>```

https://github.com/activeadmin/activeadmin/runs/5233438250?check_suite_focus=true#step:9:87

**Comment by deivid-rodriguez** [comment]
I'm guessing this change is correct (passes realworld cucumber features), but the unit tests are broken by it. If that's the culprit, we need to adapt unit tests.

**Comment by mgrunberg** [comment]
Hi, I believe the failures are related to test execution order. I'm going to check to confirm and see how to fix it.

**Comment by mgrunberg** [comment]
Ended up using the alternative to prevent problems when `spec/unit/device_spec.rb` runs before `spec/unit/application_spec.rb`

File: lib/active_admin/application.rb [hunk 1]
```diff
@@ -150,8 +150,8 @@ def routes(rails_router)
     AbstractController::Callbacks::ClassMethods.public_instance_methods.
       select { |m| m.match(/_action\z/) }.each do |name|
       define_method name do |*args, &block|
-        controllers_for_filters.each do |controller|
-          controller.public_send name, *args, &block
+        ActiveSupport.on_load(:active_admin_controller) do
+          public_send name, *args, &block
```
javierjulio, line 154: Just curious since this previously was looping for each controller, why is that no longer necessary?

javierjulio (commented):
@mgrunberg this is not an area I really know anything about so hoping the others can help. Thank you!

lib/active_admin/application.rb
javierjulio: Thanks! I think it makes sense now. From there though I didn't understand why the separate `run_load_hooks` calls using `..._base` or `..._devise` when they have the same base object argument?

deivid-rodriguez: Yes, I have the same question as @javierjulio. Only running hooks we registered ourselves (the ones tagged as `:active_admin_controller`) seems to fix the deprecation for me. Why do we need to run the others?

mgrunberg: Sorry for the confusion. `..._base` and `..._devise` are not required to fix the issue. 
I was inspired by Rails and they provide hooks for controller, API controllers and base controller. I thought it was a good idea to leave a door open to define hooks for only base controller or devise controllers exclusively (without an if inside `:active_admin_controller`.
I just change my mind and remove them. They are a source of confusion and aren't needed right now.

deivid-rodriguez (approved):
This looks good to me! Can you rebase and squash to a single commit?

deivid-rodriguez:
Awesome thanks so much!

deivid-rodriguez:
This actually closes https://github.com/activeadmin/activeadmin/issues/5772, right?

mgrunberg:
> This actually closes #5772, right?

Yes, that correct

javierjulio (approved):
Thanks again! ❤️

lib/active_admin/application.rb
javierjulio: No worries at all. Thank you so much for clarifying @mgrunberg!

PR #7341, opened 2022-02-17
Fix form SCSS variables no longer being defined in the outermost scope, so no longer being accessible

gigorok:
- Make SCSS forms variables to be accessible in the global SCSS scope
- Allow to override $filter-field-seperator-width SCSS variable

### DIFF
--- a/app/assets/stylesheets/active_admin/mixins/_variables.scss
+++ b/app/assets/stylesheets/active_admin/mixins/_variables.scss
@@ -45,7 +45,7 @@ $blank-slate-border: 1px dashed #DADADA !default;
 $sidebar-inner-content-width: $sidebar-width - ($section-padding * 2);
 
 // Filter Forms
-$filter-field-seperator-width: 12px;
+$filter-field-seperator-width: 12px !default;
 $side-by-side-filter-input-width: ($sidebar-inner-content-width * 0.5) - ($text-input-horizontal-padding * 2) - $filter-field-seperator-width;
 $side-by-side-filter-select-width: ($sidebar-inner-content-width * 0.5) - $filter-field-seperator-width;
 $date-range-filter-input-width: ($sidebar-inner-content-width * 0.5) - ($filter-field-seperator-width * 0.5);
--- a/app/assets/stylesheets/active_admin/mixins/_variables.scss
+++ b/app/assets/stylesheets/active_admin/mixins/_variables.scss
@@ -40,3 +40,12 @@ $text-input-horizontal-padding: 10px !default;
 $text-input-total-padding: $text-input-horizontal-padding * 2 + $border-width * 2;
 
 $blank-slate-border: 1px dashed #DADADA !default;
+
+// Sidebar Forms
+$sidebar-inner-content-width: $sidebar-width - ($section-padding * 2);
+
+// Filter Forms
+$filter-field-seperator-width: 12px;
+$side-by-side-filter-input-width: ($sidebar-inner-content-width * 0.5) - ($text-input-horizontal-padding * 2) - $filter-field-seperator-width;
+$side-by-side-filter-select-width: ($sidebar-inner-content-width * 0.5) - $filter-field-seperator-width;
+$date-range-filter-input-width: ($sidebar-inner-content-width * 0.5) - ($filter-field-seperator-width * 0.5);
--- a/app/assets/stylesheets/active_admin/_forms.scss
+++ b/app/assets/stylesheets/active_admin/_forms.scss
@@ -250,8 +250,6 @@ form {
 
 // -------------------------------------- Sidebar Forms
 
-$sidebar-inner-content-width: $sidebar-width - ($section-padding * 2);
-
 .sidebar_section {
 
   label {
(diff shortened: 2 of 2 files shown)


Fivell (approved):
@gigorok  thanks

Fivell:
@javierjulio , hello, what do you think about this?

javierjulio (commented):
They don't need to be moved, just marked with a `!default` at the end which the form variables don't have. The gradient variables do have it already though. Are you sure you can't override those gradient variables currently?

deivid-rodriguez (commented):
I had a look at this PR and I think it's correct.

I believe the issue is that since https://github.com/activeadmin/activeadmin/pull/6922, these variables are only defined under a `@media screen` scope, so they no longer work when accessed from active_admin_datepicker outside of any scopes: 

https://github.com/activeadmin-plugins/active_admin_datetimepicker/blob/9a7993b9c566926d57273c16fe0bd9308e46b7af/app/assets/stylesheets/active_admin_datetimepicker.scss#L1-L15

If the variables are moved to `_variables.scss`, they are again defined outside of any `@media` scope as they used to (because that file gets imported from the top level), so the issue gets fixed. This explains why also the alternative fix of wrapping rules in `active_admin_datepicker` under a `@media` scope also works.

app/assets/stylesheets/active_admin/mixins/_gradients.scss
```diff
@@ -1,6 +1,3 @@
-$secondary-gradient-start: #efefef !default;
-$secondary-gradient-stop: #dfe1e2 !default;
-
```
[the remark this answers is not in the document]
gigorok: thanks

gigorok:
@deivid-rodriguez yep. you are right. I'm going to fix `active_admin_datetimepicker`. unfortunately wrapping rules in `active_admin_datepicker` under a @media scope is not working (https://github.com/activeadmin-plugins/active_admin_datetimepicker/pull/73)

deivid-rodriguez:
Alright, so if I understand correctly, this is actually fixing two issues:

* One is the variables no longer being accessible outside of `@media` scopes, which is fixed by moving them to `_variables.scss`. That fixes the error you're getting in active_admin_datepicker `Error: Undefined variable: "$filter-field-seperator-width`.

* Another fix is, even with the above patch, the overridden value not taking effect due to missing `!default`.

Correct?

gigorok:
> Alright, so if I understand correctly, this is actually fixing two issues:
> 
> * One is the variables no longer being accessible outside of `@media` scopes, which is fixed by moving them to `_variables.scss`. That fixes the error you're getting in active_admin_datepicker `Error: Undefined variable: "$filter-field-seperator-width`.
> * Another issue is that, even with the above patch, the overridden value is not taking effect due to missing `!default`.
> 
> Correct?

Correct. After merging of this PR `$filter-field-seperator-width` SCSS variable can be overridden & can be accessible in the global SCSS scope

deivid-rodriguez:
@javierjulio We have confirmed that you were right that gradient variables were fine, and that `!default` was missing from form variables, but we did find that they do need to be moved since #6922 as explained above. This PR should be now ready from my side other than the commit splitting which is nice but not strictly necessary and I can do it myself if needed anyways.

app/assets/stylesheets/active_admin/mixins/_variables.scss
```diff
@@ -40,3 +40,12 @@ $text-input-horizontal-padding: 10px !default;
...
 $blank-slate-border: 1px dashed #DADADA !default;
+
+// Sidebar Forms
+$sidebar-inner-content-width: $sidebar-width - ($section-padding * 2);
+
+// Filter Forms
+$filter-field-seperator-width: 12px !default;
+$side-by-side-filter-input-width: ($sidebar-inner-content-width * 0.5) - ($text-input-horizontal-padding * 2) - $filter-field-seperator-width;
+$side-by-side-filter-select-width: ($sidebar-inner-content-width * 0.5) - $filter-field-seperator-width;
+$date-range-filter-input-width: ($sidebar-inner-content-width * 0.5) - ($filter-field-seperator-width * 0.5);
```
javierjulio, line 51: The `$filter-field-seperator-width` variable is now marked `!default` here whereas in its original location it wasn't. If we are moving all the form based variables (note: include `$text-input-total-padding` as well) I think we should just mark these all with `!default` otherwise they can't be overridden.

javierjulio: Hmm no, I guess not since we can always do so later if needed. We already have another calculated variable here so it should be fine. Thanks!

javierjulio (approved):

app/assets/stylesheets/active_admin/mixins/_variables.scss
deivid-rodriguez: I think the current changeset respects the previous criteria of only marking variables set to a literal as `!default`, not variables derived from other variables. Not sure if that's right or not, but this change keeps that consistent.

gigorok:
> Alright, this looks good to me. Only nitpick would be if you would mind to split both changes into two separate commits, and explain each of them on its own commit message.

Done. Please, check it up.

deivid-rodriguez (approved):
Great!

As far as I understand, the addition of `!default` is not necessary for `active_admin_datetimepicker` because:

* Variables are not overridden, just used.
* Even if they were overridden, they are overridden _after_ the default values are set, so `!default` would have no effect.

That said, the change keeps consistency with all the other variables in `_variables.scss`, so it seems good!

deivid-rodriguez:
Thanks @gigorok and @javierjulio! ❤️

PR #7344, opened 2022-02-19
Be able to run `rake cucumber` twice on local environment

mgrunberg:
Once I run `rake cucumber` I can run it anymore without failures, the test app is broken.

The problem is that `features/development_reloading.feature` creates a `admin/posts.rb` file but is not tagged as `@changes-filesystem`. The created file is not deleted after the test.

This PR marks address that by rollbacking filesystem after the scenario


Fivell (approved):

javierjulio (approved):
Great find! I had this fail for me recently and wondered what may be wrong. Thank you! ❤️

PR #7347, opened 2022-02-21
Properly check whether columns are counter cache in order to display them or not

### LINKED ISSUE #7346: A show page does not display a field ends with a "_count" substring
I have a model with some column ends with a "_count" substring. This column doesn't appear on a show page. It should appear as It is a regular column just like any other.

I've figured out that this is because this column filtered out as a counter cache column by the following code in ```ActiveAdmin::Resource::Attributes#counter_cache_col?```

The code simply checks that the column has a name that ends with "_count". On the one hand, this leads to the situation where a regular columns are filtered out. On the other hand, this code will not filter a counter cache column with a custom name that doesn't match "*_count".

plashchynski:
Check the column is actually a counter cache column. Fixes #7346

### DIFF
--- a/lib/active_admin/resource/attributes.rb
+++ b/lib/active_admin/resource/attributes.rb
@@ -37,7 +37,10 @@ module ActiveAdmin
       end
 
       def counter_cache_col?(c)
-        c.name.end_with?("_count")
+        counter_cache_columns = resource_class.reflect_on_all_associations(:has_many)
+                                              .select(&:has_cached_counter?)
+                                              .map(&:counter_cache_column)
+        counter_cache_columns.include?(c.name)
       end
 
       def filtered_col?(c)
--- a/features/show/default_content.feature
+++ b/features/show/default_content.feature
@@ -42,3 +42,19 @@ Feature: Show - Default Content
     And I should see the attribute "Body" with "Empty"
     And I should see the attribute "Created At" with a nicely formatted datetime
     And I should not see the attribute "Author"
+
+  @changes-db-schema
+  Scenario: Columns with "counter cache"-like names
+    Given a new "tagging_count" counter column is added to "posts"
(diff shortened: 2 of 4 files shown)


lib/active_admin/resource/attributes.rb
```diff
@@ -37,7 +37,10 @@ def sti_col?(c)
       end
 
       def counter_cache_col?(c)
-        c.name.end_with?("_count")
+        counter_cache_columns = resource_class.reflect_on_all_associations(:has_many)
+                                              .select(&:has_cached_counter?)
```
Fivell, line 41: @plashchynski thanks for contribution 👍 , I suggest to add new column  `some_count` to Post model in specs, and check it is rendered in features/index/format_as_csv.feature in order to merge it

Fivell (changes requested):

lib/active_admin/resource/attributes.rb
deivid-rodriguez: Are you still interested @plashchynski?

javierjulio:
I think I would have rather wanted an attribute to an existing model but I’m fine with either approach. Can you please rebase @plashchynski and we’ll merge this? Thank you.

lib/active_admin/resource/attributes.rb
```diff
@@ -37,7 +37,10 @@ def sti_col?(c)
...
+                                              .map(&:counter_cache_column)
+        counter_cache_columns.include?(c.name)
```
javierjulio, line 43: Seeing where this method is called as ultimately part of a loop, should we memoize the value?

javierjulio (commented):
Since we have parallel tests, I don't want to do the "@changes-db-schema" rollback approach so we can just update an existing model with the cache counter.

PR #7349, opened 2022-02-25
Apply `box-sizing: border-box` globally

deivid-rodriguez:
This is extracted from #3862. I think it shouldn't change the default appearance but should make it easier to customize it.

I slightly changed the original commit to propose the best practice recommended here: https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/.

### DIFF
--- a/app/assets/stylesheets/active_admin/_base.scss
+++ b/app/assets/stylesheets/active_admin/_base.scss
@@ -33,6 +33,12 @@
   @import "./structure/main_structure";
   @import "./structure/title_bar";
 
+  *,
+  *:before,
+  *:after {
+    box-sizing: border-box;
+  }
+
   body {
     @include sans-family;
     line-height: 1.5;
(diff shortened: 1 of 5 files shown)


javierjulio (approved):
Awesome thanks! ❤️

deivid-rodriguez:
@javierjulio Do you think it makes sense to add a new changelog section for this kind of changes like `type: appearance`, `type: design` or something like that, or better to ship them as `type: enhancement`?

javierjulio:
I think using "enhancement" works great. If we do the former, then I'd think we'd need more sub groups specific to appearance or design. Thanks!

johnmaxwell:
image

We've upgraded to ActiveAdmin 2.10.1 which includes this change. It appears to have affected the formtastic hints for checkboxes. They now start on the same line as the checkbox/label and then wrap, instead of just being directly beneath the checkbox. Anyone else or just us?

PR #7351, opened 2022-02-25
Get ready for 2.10.1 release

deivid-rodriguez:
I'd like to start doing frequent releases and try to ship constant improvements quickly. I'd like to aim for one release every two weeks, let's see how it goes.


javierjulio (approved):
Thanks!

PR #7358, opened 2022-02-26
Drop support for Ruby 2.6

alejandroperea:
Ruby 2.6 is going to be unsupported soon, one month left for arriving to its EOL.

Maybe could be the moment to drop the support also for ActiveAdmin.


alejandroperea:
Looks like JRuby supports 2.6.x.....

Not ready for Ruby 2.7: https://github.com/jruby/jruby/issues/6464

PR #7366, opened 2022-03-04
Update cucumber-rails dependency

tagliala:
`cucumber-rails` 2.5 with Rails 7 support has not been released yet
at the time of this commit

This change will allow #7235 to be merged


deivid-rodriguez (approved):
Awesome, thanks!! ❤️

javierjulio (approved):
Thanks!

PR #7384, opened 2022-03-09
Add turbolinks support to has many js

### LINKED ISSUE #7383: hasMany.js does not support turbolinks
### Did you find a bug?

While most other javascript features of ActiveAdmin are set up to be initialized on `page:load` and `turbolinks:load`, the hasMany feature is initialized at run time.

### Expected behavior

Has many forms should still work after turbolinks transitions.

### Actual behavior

The has many initializer is run on page load but not after turbolinks transitions.

### How to reproduce

1. Set up a form with `has_many`
2. Set up turbolinks
3. In the ActiveAdmin UI, create an instance of the resource with has many
4. Load the index page and transition to the edit page with turbolinks enabled
5. Notice that the form on the edit page does not have has many features

### Extra Notes

Hi, I'm working on a PR to fix this now. If anyone wants more detail on this issue, I'm happy to provide it, but I think that the fix will be strait-forward and I just wanted to provide enough context here to support the PR.

Also, it would be nice to support turbo transitions, but that's a job for another PR.

amiel:
This closes #7383.

## Notes

* No regression tests

  There are very little javascript tests, and none around the has many feature. I'm not really sure how to get started but would be happy to look in to it as a followup.

* ~~Formatting changes~~

  ~~My editor automatically made some formatting changes based on the eslint configuration of the project. Would you like me to split these out or remove them?~~

* Turbo vs turbolinks

  I would also like to add support for turbo, and not just turbolinks, but that is also a different project.

### DIFF
--- a/features/has_many.feature
+++ b/features/has_many.feature
@@ -52,3 +52,27 @@ Feature: Has Many
 
     When I click "Add New Post"
     Then I should see a link to "Hide"
+
+  Scenario: Sortable is initialized when transitioning to edit with existing data
+    Given a configuration of:
+    """
+      ActiveAdmin.register User do
+        form do |f|
+          f.inputs do
+            f.has_many :posts, sortable: :position do |ff|
+              ff.input :title
+              ff.input :body
+            end
+          end
+          f.actions
+        end
+      end
+      ActiveAdmin.register Post
+    """
+    When I go to the last author's show page
+    And I follow "Edit User"
+    Then I should see a link to "Add New Post"
+    And there should be 1 "input" tag within ".ui-sortable"
+
+    When I click "Add New Post"
+    Then I should see a link to "Remove"
--- a/app/assets/javascripts/active_admin/base.js
+++ b/app/assets/javascripts/active_admin/base.js
@@ -392,7 +392,6 @@
     $(document).on("change", '.has_many_container[data-sortable] :input[name$="[_destroy]"]', function() {
       recompute_positions($(this).closest(".has_many"));
     });
-    init_sortable();
     $(document).on("has_many_add:after", ".has_many_container", init_sortable);
   });
   var init_sortable = function init_sortable() {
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
Yes, please, no formatting changes! A way to get this tested would be awesome, I'll investigate too. Other than that looks good!

amiel:
@deivid-rodriguez I've updated this to remove the formatting changes. I'll look in to tests as another PR if that's ok with you.

deivid-rodriguez:
I prefer to add tests together with this feature, or at least try :)

features/has_many.feature
```diff
@@ -26,6 +26,7 @@ Feature: Has Many
     When I go to the last author's show page
     And I follow "Edit User"
     Then I should see a link to "Add New Post"
+    And there should be 1 "input" tag within ".ui-sortable"
```
amiel, line 29: I know this is a little awkward, but unfortunately I couldn't find any other observable change for when has many is initialized. The sortable option adds some classes that allowed me to make an assertion that fails before the changes in this PR and passes with the changes.

deivid-rodriguez:
@amiel This looks good to me! Only thing I would change is to completely split the feature test to a separate scenario, so that we can test the `has_many` feature with sortable attribute separately. What do you think?

features/has_many.feature
deivid-rodriguez: It looks good enough to me! Easier that simulating a real drag and drop action.

amiel: That makes sense. Will do

amiel: Ok, how do you feel about this?

deivid-rodriguez (approved):
It looks perfect to me now, thanks! 🎉 

Only modification I would make is to squash both commits, but that can be done when merging!

amiel:
Squashed and rebased to be current with `master`.

deivid-rodriguez:
Awesome, I'll let @javierjulio double check this if he is available!

Are you planning to also add turbo support, by the way?

javierjulio (approved):

Issue #3910, opened 2015-04-25
Custom form in partial doesn't work


adisos:
```
# admin/post.rb
ActiveAdmin.register Post do
  form partial: 'form'
end
```

```
# app/views/admin/posts/_form.html.arb
insert_tag active_admin_form_for resource do |f|
  inputs :title, :body
  actions
end
```
Error:
```undefined method `posts_path' for #<#<Class:0x007f9980312588>:0x007f9981f96358>```

**Comment by noijd** [comment]
I have a similar problem and I fix it establishing the controller action.

In my case I have a `member action` like this:

```ruby
  member_action :publish, method: [:get, :post] do
    @page_title = "Publish #{resource.title}"

    if request.get?
      render :publish
    else
      # ...
    end
  end
```

So this would render the view on `views/admin/posts/publish.html.arb`, which contains this code:

```ruby
active_admin_form_for assigns[:post], url: { action: 'publish' } do
  semantic_errors
  inputs
  actions
end
```

reicheltd:
+1 same problem using rails 4.2.3 and activeadmin 1.0.0.pre1

jtomaszewski:
Use `active_admin_form_for [:admin, resource] do |f|` instead of `insert_tag active_admin_form_for resource do |f|`.

We should update this in the documentation.

timoschilling:
I’m going to close this since you haven’t responded. Please feel free to update this issue with additional information so it can be reconsidered.

davidstosik:
After trying to update (from a really old version!), I had the same issue:
`undefined method `posts_path' for #<#<Class:0x007f9980312588>:0x007f9981f96358>`

@jtomaszewski's suggestion helped me fix it: although ActiveAdmin's documentation recommends
`active_admin_form_for resource do |f|`, it only works when I use `active_admin_form_for [:admin, resource] do |f|`.

AlexVentura:
I had a problem described here https://github.com/activeadmin/activeadmin/issues/5410 I could not found what was causing the wrong URL params in my resource. So, I decided to override each and every `action_item`, `member_action`, `form_action`, etc.

To solve the form concept I did the following and it really works pretty fine:

```ruby
# app/admin/organizations.rb
ActiveAdmin.register Post do
  form partial: 'form'
end
```

```ruby
# app/views/admin/organizations/_form.html.arb
active_admin_form_for [:admin, resource] do |f|
  tabs do
    tab 'General Configuration' do
[...]

PR #7394, opened 2022-03-13
Remove `insert_tag` from Form-Partial docs

TonyArra:
This PR contains only documentation changes. It references #3910, which was closed due to inactivity.

This removes the use of `insert_tag` in the documentation for form-partials. 

As discussed in #3910, using `insert_tag active_admin_form_for resource` in a form-partial raises the following error (where `post` is the name of the resource):

```ruby
undefined method 'posts_path' for #<#Class:0x007f9980312588:0x007f9981f96358>
```

The commonly accepted solution in #3910 is the usage of  `active_admin_form_for[:admin, resource]` instead. This is the solution that worked for me as well. I'm uncertain if this requires any additional explanations regarding the namespace of the resource.

### DIFF
--- a/docs/5-forms.md
+++ b/docs/5-forms.md
@@ -56,7 +56,7 @@ Which looks for something like this:
 
 ```ruby
 # app/views/admin/posts/_form.html.arb
-insert_tag active_admin_form_for resource do |f|
+active_admin_form_for [:admin, resource] do |f|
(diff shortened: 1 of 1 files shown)


javierjulio (approved):

deivid-rodriguez (approved):
Thanks!

PR #7437, opened 2022-04-08
run update_resource inside a transaction to avoid autosaving relationships through assign_attributes when the record is invalid

### LINKED ISSUE #5430: Associations are persisted before validation of main record

deivid-rodriguez:
A copy of #6202, to not let it be forgotten.

Fixes https://github.com/activeadmin/activeadmin/issues/5430

When submitting a form with invalid attributes, some associations are autosaved via assign_attributes before calling save_resource. As a consequence, when updating a resource, the user sees a validation error and is not aware that the changes to the associations got persisted.

This PR runs the code inside the update_resource method in an ActiveRecord::Transaction so that everything is rolled back when the model is not valid.


javierjulio (approved):
Thanks David! Not sure if you wanted to get this in as is or with tests.

matgaw:
@deivid-rodriguez @javierjulio is this planned to be merged at some point? it solves an important issue

thibaudgg:
I also just encountered that issue, this fix is important and should be merged.

ggstroligo:
Is this PR be merged or there is another fix for this issue already on the codebase?

deivid-rodriguez:
Hello, sorry for the radio silence here. I'll try to find some time to re-evaluate this in the following weeks!

deivid-rodriguez:
> Is this PR going to be merged, or is there another fix for this issue already in the codebase?

Are you experiencing this problem? I personally don't know if it has been fixed independently, but if you're experiencing the issue, I think you're very well equipped to verify it!

mgrunberg:
> > Is this PR going to be merged, or is there another fix for this issue already in the codebase?
> 
> Are you experiencing this problem? I personally don't know if it has been fixed independently, but if you're experiencing the issue, I think you're very well equipped to verify it!

@deivid-rodriguez I manage to replicate the original issue in v4 branch adding some models to the template app. This confirms the errors hasn't been fixed.

I want to write a feature test about the change. I don't have time now but should have it in the following days.

mgrunberg:
@deivid-rodriguez I found time to write the spec.

I rebased the PR. The first commit shows the failure and the second one, a cherry-pick of the original code, fixes it.

About the spec. It involved a new model so there are a lot of new files just for this. It may be the wrong call but I considered it cleaner than trying to find a way to mimic the issue scenario with existing models (for example: using `Post` and `Tag`). 

Now we have `Company` with a `has_and_belongs_to`  `Store`. The feature test changes the `Company` to a state where validations fail (`name` can't be blank) and the test expects `stores` changes to be rollbacked.

javierjulio:
@mgrunberg thank you. Did you consider or try doing a controller or unit test instead for this? Cucumber adds significant overhead that makes it hard to follow. Then again I don't know how easy it is to do in a controller or unit test, it would just be ideal if we could I think.

mgrunberg:
> @mgrunberg thank you. Did you consider or try doing a controller or unit test instead for this? Cucumber adds significant overhead that makes it hard to follow. Then again I don't know how easy it is to do in a controller or unit test, it would just be ideal if we could I think.

@javierjulio I didn't consider a controller test. No strong argument here. I simply went for a feature test. It should be possible to add a spec in `spec/unit/resource_controller_spec.rb` (at least I can't think a blocker) but I won't have time to do this for a few days. If that's not a problem I'm ok on give it a try to controller specs.

Just one question, are you fine with the introduction of the new modals?
Because even if I replace the feature spec with a controller one I still need the new model, migrations, etc. The amount of new files won't decrease.

javierjulio:
@mgrunberg no worries, take your time. Thank you. Yes the models, migrations, etc. are ok I just prefer to avoid Cucumber if we can. Although feel free to leave the feature spec as is since it's already written. I just think a unit or controller test would be easier to understand and review if we can. Thank you again.

mgrunberg:
@javierjulio I wrote the controller spec. Let me know what you think

spec/unit/resource_controller_spec.rb
```diff
@@ -124,6 +124,24 @@ def call_after_destroy(obj); end
...
+  describe "resource update" do
+    let(:controller) { Admin::CompaniesController.new }
+
+    around do |example|
+      with_resources_during(example) do
+        ActiveAdmin.register Company
+      end
+    end
+
+    it "should not update habtm associations when validation fails" do
```
javierjulio, line 137:
```suggestion
    it "should not update habtm associations when the resource validation fails" do
```

javierjulio (approved):
Thanks!

Fs00:
@javierjulio I see this hasn't been backported to 3.x, would you accept a PR that does it? We're hitting the bug fixed by this PR in a production app running AA 3.

javierjulio:
@Fs00 yes. Typically these changes have been backported so I'm not sure if we skipped it for some reason so let's confirm first. @mgrunberg any concern with this change being backported to v3?

mgrunberg:
@javierjulio no problems at all on backporting. I guess it got misplaced. Let me backport this

Fs00:
Many thanks! Looking forward to it 👀🙏

mgrunberg:
@Fs00, @javierjulio backport released as v3.5.2

PR #7451, opened 2022-04-20
Begin turbo support

amiel:
In #7384, I mentioned wanting to also add support for turbo.

I just spent the last 2 days trying to get cucumber tests using turbo in order to have a good failing test that shows that the turbo support is working.

I tried every combination of (rails 6.1, rails 7) X (sprockets, webpacker), as well as just rails 7 with importmaps, and I in each case I ran in to an issue I couldn't get past.

Anyway, I thought I'd get this PR opened to at least get conversation started. I'm not sure if anyone can help get tests working...

### DIFF
--- a/app/assets/javascripts/active_admin/base.js
+++ b/app/assets/javascripts/active_admin/base.js
@@ -125,7 +125,7 @@
       });
     }
   };
-  $(document).ready(onDOMReady$2).on("page:load turbolinks:load", onDOMReady$2);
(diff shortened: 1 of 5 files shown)


javierjulio:
I had to disable the webpacker CI build as it's been failing for awhile and not sure what's wrong. If someone wants to take that on I'm happy to accept a fix and also these improvements for Turbo support. Thanks.

javierjulio:
Closing in favor of #7000 since this also requires options support on registering JS which is also separately in #7002.

PR #7455, opened 2022-04-22
Use squash merge strategy for mergify

deivid-rodriguez:
This matches how we usually merge other PRs: merge strategy in general, except for single commit PRs, where we use squash. Dependabot PRs are single commit, so we should use squash for them too.


javierjulio (approved):
Yes, thanks!

PR #7476, opened 2022-05-03
Drop rails 6.0 support

deivid-rodriguez:
It only receives severe security fixes, like 5.2: https://guides.rubyonrails.org/maintenance_policy.html.

Time to drop support!


deivid-rodriguez:
(Took the chance to update a bunch of other stuff in CI).

javierjulio (approved):
Thanks!

rylanb:
Are you still supporting Rails 6.1? Just wanted to check on that as I wasn't sure

deivid-rodriguez:
Yes.

In general, whatever combination of stable releases of ActiveAdmin and Rails that Bundler allows to be installed together can be considered supported.

rylanb:
Awesome thanks so much! I just wanted to make sure there. 

Thanks for all your hard work and the quick reply!

PR #7479, opened 2022-05-06
Fix pundit performance

### LINKED ISSUE #7477: Slow view performance after upgrading from 2.11.2 to 2.12.0
### Did you find a bug?

All ActiveAdmin pages are much slower after performing this upgrade (e.g. from 1s to 9s per page load)

### Expected behavior

ActiveAdmin page performance should not change significantly

### Actual behavior

All ActiveAdmin pages are much slower after performing this upgrade (e.g. from 1s to 9s per page load)

### How to reproduce

Upgrade from 2.11.2 to 2.12.0.

Based on our instrumentation, the slowdown appears to be in the view layer. What would be most helpful for us to provide to track down this issue?

deivid-rodriguez:
Our fallback behavior introduced at #7144 to keep backwards compatibility when looking for compatible Pundit policies introduced severe performance degradation.

This PR fixes the issue by introducing a cache of Pundit policies. It also introduces other minor improvements but the performance culprit should hopefully be fixed by 63e3f5f850a364faf58979637a7b70431630442c.

Fixes #7477.

### DIFF
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -89,7 +89,7 @@ module ActiveAdmin
     end
 
     def namespace(object)
-      if default_policy_namespace && !object.class.to_s.match?(/^#{default_policy_module}::/)
+      if default_policy_namespace && !object.class.to_s.start_with?("#{default_policy_module}::")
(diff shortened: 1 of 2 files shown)


Fivell (approved):
👍

PR #7487, opened 2022-05-10
Fix passing expected hash payload argument

### LINKED ISSUE #7342: Possible inconsistency with ActiveSupport::Notifications' interface breaks when using Sentry
### Did you find a bug?

Trying to install the latest release (2.10.0) in an existing Rails 6 app that uses Sentry for monitoring, I get the following error when running `rake db:setup`:

```
NoMethodError: undefined method `[]=' for #<ActiveAdmin::Application:0x0000563b7de34130 @namespaces=#<ActiveAdmin::Namespace::Store:0x0000563b7de2bf08 @namespaces={}>, @settings=#<Class:0x0000563b7de2bcd8>, @namespace_settings=#<Class:0x0000563b7de2bb48>, @stylesheets={"active_admin.css"=>{:media=>"all"}}, @javascripts=#<Set: {"active_admin.js"}>>
/usr/local/bundle/ruby/3.0.0/bundler/gems/activeadmin-c9695082a328/lib/active_admin/application.rb:37:in `method_missing'
/usr/local/bundle/ruby/3.0.0/gems/sentry-rails-4.6.1/lib/sentry/rails/tracing.rb:40:in `instrument'
/usr/local/bundle/ruby/3.0.0/gems/activesupport-6.1.4.6/lib/active_support/notifications.rb:203:in `instrument'
/usr/local/bundle/ruby/3.0.0/bundler/gems/activeadmin-c9695082a328/lib/active_admin/application.rb:115:in `load!'
/usr/local/bundle/ruby/3.0.0/bundler/gems/activeadmin-c9695082a328/lib/active_admin/application.rb:142:in `routes'
/usr/local/bundle/ruby/3.0.0/bundler/gems/activeadmin-c9695082a328/lib/active_admin.rb:76:in `routes'
/srv/oysterhr/config/routes.rb:7:in `block in <main>'
# ... rest of backtrace omitted
```

I tracked the error on the sentry-ruby gem, and it happens when handling ActiveSupport::Notifications events, in particular the `ActiveAdmin::Application::BeforeLoadEvent`.

I noticed AA is calling `ActiveSupport::Notifications.instrument` passing an ActiveAdmin::Application instance as payload when instrumenting this event:

```
ActiveSupport::Notifications.instrument ActiveAdmin::Application::BeforeLoadEvent, ActiveAdmin.application
```

This is the exact case that causes Sentry to crash. Sentry expects the payload argument to be a hash-like object and tries adding a new key to an `ActiveAdmin::Application` instance, which fails with the reported error.

The [documentation of ActiveSupport::Notifications](https://www.rdoc.info/docs/rails/6.0.2.1/ActiveSupport/Notifications) for rails 6 (and 7 as well):
> Payloads consist of a hash whose contents are arbitrary and generally depend on the event.

So I believe this is an issue with ActiveAdmin, not sentry. I've managed to fix the issue locally by wrapping the payload as a hash and adapting the subscribers:

```rb
ActiveSupport::Notifications.instrument ActiveAdmin::Application::BeforeLoadEvent, { active_admin_application: ActiveAdmin.application }
```

And on the subscribers, extracting that key from the hash:
```rb
# activeadmin/lib/active_admin.rb:109
def wrap_block_for_active_support_notifications block
  proc { |_name, _start, _finish, _id, payload| block.call payload[:active_admin_application] }
end
```

Am I missing anything here? Shall I open a PR? This might be a breaking change if other applications depend on the previous format.

### Expected behavior

Active Admin should use the ActiveSupport::Notifications interface consistently with the documented behavior, and not cause any other gems/applications that expect a hash on the payload argument to crash.

### Actual behavior

The rails app crashes on boot when a subscriber of the events instrumented by AA tries to treat AtiveAdmin::Application as a hash-like object.

### How to reproduce

Note: could not reproduce this with the bug report template. Let me know if you need me to set up a demo app (if my assumptions about the fact we're violating the AS::Notifications interface are correct, we might want to fix it anyway).

- Add the following gems to your gemfile:
```rb
gem "activeadmin"
gem "sentry-ruby"
gem "sentry-rails"
```

In an initializer, initialize sentry so it listens for instrumented events:
```rb
  Sentry.init do |config|
    config.dsn = 'https://12345678910456789aeee6@o640421.ingest.sentry.io/123456789' # a sentry dsn config (only needs to be a valid URL, not a valid sentry credential)

    config.traces_sample_rate = 1 # Not sure this is necessary
  end
```

ispyropoulos:
Passing non hash objects as payload arguments to an event instrumenter breaks subscribers implemented by other libraries like Sentry.

https://api.rubyonrails.org/v6.1.5.1/classes/ActiveSupport/Notifications.html
https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html

Fixes #7342

### DIFF
--- a/features/step_definitions/configuration_steps.rb
+++ b/features/step_definitions/configuration_steps.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 module ActiveAdminReloading
   def load_aa_config(config_content)
-    ActiveSupport::Notifications.instrument ActiveAdmin::Application::BeforeLoadEvent, ActiveAdmin.application
+    ActiveSupport::Notifications.instrument ActiveAdmin::Application::BeforeLoadEvent, { active_admin_application: ActiveAdmin.application }
     eval(config_content)
-    ActiveSupport::Notifications.instrument ActiveAdmin::Application::AfterLoadEvent, ActiveAdmin.application
+    ActiveSupport::Notifications.instrument ActiveAdmin::Application::AfterLoadEvent, { active_admin_application: ActiveAdmin.application }
     Rails.application.reload_routes!
     ActiveAdmin.application.namespaces.each &:reset_menu!
   end
--- a/lib/active_admin.rb
+++ b/lib/active_admin.rb
@@ -113,7 +113,7 @@ module ActiveAdmin
     private
 
     def wrap_block_for_active_support_notifications block
-      proc { |_name, _start, _finish, _id, payload| block.call payload }
+      proc { |_name, _start, _finish, _id, payload| block.call payload[:active_admin_application] }
     end
 
   end
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -76,7 +76,7 @@ module ActiveAdmin
 
       namespace = namespaces[name.to_sym] ||= begin
         namespace = Namespace.new(self, name)
-        ActiveSupport::Notifications.instrument ActiveAdmin::Namespace::RegisterEvent, namespace
+        ActiveSupport::Notifications.instrument ActiveAdmin::Namespace::RegisterEvent, { active_admin_namespace: namespace }
         namespace
       end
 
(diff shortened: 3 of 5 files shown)


deivid-rodriguez (approved):
Thanks!

javierjulio (approved):
Thanks!

PR #7488, opened 2022-05-12
Honor load paths order when loading admin files

tf:
[PR #6124](https://github.com/activeadmin/activeadmin/pull/6124) started sorting files when loading to ensure `ActiveAdmin.routes` has a consistent order. This order - based on admin file names - took precedence over the order of load paths.

When admins are loaded from different locations, e.g. different gems and the main application, it can be helpful to control load
order. That way the main application can override settings of admins defined in gems.

The previous behavior also made load order depend on the file system location of gems and the application since expanded paths are used when sorting. For example, in development mode, admins from a gem in `/home/user/.rvm/...` might be loaded before admins from the app in `/home/user/my_app/app/admin`, while being loaded after app admins in production when the gem lives in `/home/app/my_app/vendor/...`.

We now instead sort files of each load path individually to achieve consistent ordering while still preserving the overall order as
prescribed by the `application.load_paths` array.

### DIFF
--- a/lib/active_admin/application.rb
+++ b/lib/active_admin/application.rb
@@ -126,7 +126,7 @@ module ActiveAdmin
 
     # Returns ALL the files to be loaded
     def files
-      load_paths.flatten.compact.uniq.flat_map { |path| Dir["#{path}/**/*.rb"] }.sort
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Makes sense to me, thanks!

javierjulio (approved):
Thanks!

PR #7489, opened 2022-05-14
Collection Helper should not make redundant COUNT queries.

### LINKED ISSUE #6492: Disable COUNT(*) for filtered collections
### Did you find a bug?

I'm using pagination_total false on the index collection, which works fine and as expected. However, I'm noticing in the SQL debug logs that when I filter a collection (by user_id), it is still doing a `SELECT COUNT(*) FROM (SELECT 1 AS one FROM "large_table" WHERE "large_table"."user_id" = 123 ORDER BY "large_table"."id" desc LIMIT $1 OFFSET $2) subquery_for_count`.

Is there a way to disable this for filters? I have an extremely large table and it is causing my app to timeout.

I'm just using the default filter method on my activeadmin model

### Expected behavior

Filtering large tables should not run an expensive `COUNT(*)` or `ORDER BY` (even if it is using limit and offset)

### Actual behavior

Request times out

### How to reproduce

Have an extremely large table and filter on an indexed column like user_id. Expect it to be relatively performant and not run things like COUNT(*) unnecessarily.

allisonphillips:
I have been looking into why `pagination_total: false` is not preventing the execution of multiple `subquery_for_count` queries. I found the issue appeared rooted in the `collection_size` method not validating whether a collection is already loaded. General context on my changes:

- This method is called several times when building an index page
- Because the current method definition uses `count`, this triggers several redundant count subqueries (`subquery_for_count`) that cause latency and even timeouts when loading index pages
- I also modified the guards at the beginning of the method because `Array#length` should be more performant than `Array#count` within this context

Resolves  #6492

## Testing/QA
Because this is a refactor and not new behavior I did not add new specs. When looking into this, I did come across some bugs in the paginated collection specs that should probably be fixed at some point though; for example, the [examples pertaining to pagination_total](https://github.com/activeadmin/activeadmin/blob/master/spec/unit/views/components/paginated_collection_spec.rb#L231-L248) will always result in a false pass because they're testing for the absence of queries that will never occur, and it's going uncaught because the author did not test the opposite assertion when the option is false. 

I've QA'd this as an override in a staging environment that uses a sanitized production DB snapshot by loading index pages for tables that I know to have millions of records. The staging environment has fewer computing resources, but after pushing this change it is rendering large index pages _faster_ than the production environment &mdash; it's even rendering `pagination_total: false` pages that are _timing out_ in production (due to the COUNT subqueries). 

In a development environment, I can also see that this prevents the redundant COUNT subqueries that drive many of us to attempt using the `pagination_total` option; here is my sanitized server log:

### Before
```ruby
Rendering redacted/app/views/active_admin/resource/index.html.arb
  ...
   (41.4ms)  SELECT COUNT(*) FROM (SELECT  1 AS one FROM `redacted_models ` LIMIT 20 OFFSET 0) subquery_for_count
   (2.6ms)  SELECT COUNT(*) FROM (SELECT  1 AS one FROM `redacted_models ` LIMIT 20 OFFSET 0) subquery_for_count
   (4.9ms)  SELECT COUNT(*) FROM (SELECT  1 AS one FROM `redacted_models ` ORDER BY `redacted_models `.`id` desc LIMIT 1 OFFSET 20) subquery_for_count
   (0.4ms)  SELECT COUNT(*) FROM (SELECT  1 AS one FROM `redacted_models ` LIMIT 20 OFFSET 0) subquery_for_count
  RedactedModel Load (0.4ms)  SELECT  `redacted_models`.* FROM `redacted_models ` ORDER BY `redacted_models `.`id` desc LIMIT 20 OFFSET 0

  Rendered redacted/app/views/active_admin/resource/index.html.arb (233.7ms)
```

### After
```ruby
Rendering redacted/app/views/active_admin/resource/index.html.arb
  ...
  RedactedModel Load (1.6ms)  SELECT  `redacted_models`.* FROM `redacted_models` ORDER BY `redacted_models`.`id` desc LIMIT 20 OFFSET 0
   (0.6ms)  SELECT COUNT(*) FROM (SELECT  1 AS one FROM `redacted_models` ORDER BY `redacted_models`.`id` desc LIMIT 1 OFFSET 20) subquery_for_count
  Rendered redacted/app/views/active_admin/resource/index.html.arb (2815.6ms)
```


javierjulio (commented):
Thanks for tackling this! Would you consider adding some specs for the cases you've mentioned? It would help to have some more coverage around this as we'd appreciate any help and input on improving that.

Nkadze:
Guys any plans to merge this PR?

deivid-rodriguez:
Can you help us add a spec?

Nkadze:
Just checked this fork on my project and I still get `subquery_for_count` when used with `scoped_collection` including joins and merges

deivid-rodriguez:
So what you're saying is that this does not actually work?

Nkadze:
Right now, in my case
if you use 
```
controller do
    def scoped_collection
      super.joins(:parent).includes(:parent).merge(Parent.some_simple_scope).order('parent.column DESC')
    end
  end
```
something like this, then you use filter on one of the columns, it still generates heavy `subquery_for_count`s

hwo411:
There is also a problem when using pagination_total: false with https://github.com/activeadmin/activeadmin/blob/3f2c8d2b7636c2e6d009093d1d9ec70e4ed722b0/lib/active_admin/views/components/paginated_collection.rb#L106 this part when custom select and order is used.

We monkey-patched it with `collection.model.from` to perform count on subquery. Not sure if it's good for general use-case, so not providing and MR, but JFYI, if someone is more experienced with the codebase who can fix it reads it.

paulmthiebauth:
Bump! It'd be great to get this in. The overhead of the `count` calls adds so much to the overall page costs when using `filters` with millions and millions of records

PR #7511, opened 2022-05-24
Force scope on default filters collection

### LINKED ISSUE #6883: Filters values aren't using authorization scopes

deivid-rodriguez:
In-repo clone of https://github.com/activeadmin/activeadmin/pull/6923.

> When we use default filters, or filter of an existing relationship without providing a collection, we enforce the collection to be scope
> It is useful when you are using authorization scope to define what resources your current user can access, even in filter values.

Fixes #6883.

### DIFF
--- a/lib/active_admin/filters/resource_extension.rb
+++ b/lib/active_admin/filters/resource_extension.rb
@@ -75,6 +75,7 @@ module ActiveAdmin
       def add_filter(attribute, options = {})
         raise Disabled unless filters_enabled?
 
+        options = filter_default_collection(attribute.to_sym).merge(options)
         (@filters ||= {})[attribute.to_sym] = options
       end
 
@@ -93,7 +94,7 @@ module ActiveAdmin
 
         if filters.empty? || preserve_default_filters?
           default_filters.each do |f|
-            filters[f] ||= {}
+            filters[f] ||= filter_default_collection(f)
           end
         end
 
@@ -102,6 +103,16 @@ module ActiveAdmin
         end
 
         filters
+      end
+
+      def filter_default_collection(filter)
+        association = resource_class.reflect_on_all_associations.find { |association| association.name == filter }
+        return {} unless association
+
+        scoped_collection = proc do
+          active_admin_authorization.scope_collection(association.klass.all)
+        end
+        { collection: scoped_collection }
       end
 
       # @return [Array] The array of default filters for this resource


javierjulio:
The tests are failing. Likely need some updates since this was last looked at.

mgrunberg:
@javierjulio This is now a breaking change. `scoped_collection` is not raising `PunditError` anymore.

mgrunberg:
@javierjulio took me some time to build a proper spec. Now is ready for review.
For the record, tests were failing because the code assumes that we can always get a scoped collection. But the template app define a default scope nor a policy for Tagging class (which is fine). That makes `scoped_collection` to raise an error.
I didn't want to investigate why raising an error is the expected result. In the context of building filters, I find it good enough to fallback to "all" scope if this happens.

lib/active_admin/error.rb
```diff
@@ -19,6 +19,16 @@ def message
     end
   end
 
+  class ScopeNotDefined < StandardError
```
javierjulio, line 22: I'd like to include "Policy" in the error class name so it's clear since when I think "scopes" it's something else. Perhaps "PolicyScopeNotDefined" as the class name?

javierjulio (commented):
What is the benefit of adding this? If I understand right it only affects default filters, so if you set filters explicitly which is best, this would not apply. Or I guess it would since there is an update to `add_filter` method. I'm not convinced on this change.

mgrunberg:
> What is the benefit of adding this? If I understand right it only affects default filters, so if you set filters explicitly which is best, this would not apply. Or I guess it would since there is an update to `add_filter` method. I'm not convinced on this change.

@javierjulio  The change affects filters and default filters. I added more test scenarios to demonstrate the usage. I understand the previous commit was misleading.
If you use an implicit filter collection, we make sure it is accessible by the user. That's the main idea.

lib/active_admin/error.rb
mgrunberg: I understand your concern with the name. I'm open to use another one. I just want to avoid `Policy` because it's a  `Pundit` strong word. I want to model something more abstract, the `ActiveRecord::RecordNotFoundError` for `scope_collection` method. I don't want to tailor the error for a specific AuthorizationAdapter implementation.

I renamed the class to `ScopeAuthorizationError`. Is it better?

Issue #7516, opened 2022-05-25
Show all batch action names in the batch action selector's title


jadilet:
It would be good to show all batch action names in the batch action selector's title, so a user doesn’t need to click the batch action selector to show the full list of batch actions.
<img width="1385" alt="Feature" src="https://user-images.githubusercontent.com/6593519/170226136-c792be41-b944-4813-a824-5797fdfa44ba.png">

hubertjakubiak:
It could be useful for some admin users and should be configurable. We had monkey patch to achieve this behaviour as client requested such feature.

javierjulio:
I'm not sure this makes any sense to me. You want to append a delimited list of the drop down options to the drop down button label so they don't have to click the drop down to see all the options? I don't think this is a good idea nor something we'd want to do and support with a button that could break the layout if the text were too large.

PR #7532, opened 2022-06-01
Implement override batch action selector label

jadilet:
https://github.com/activeadmin/activeadmin/issues/7516


javierjulio:
I don't think this is a good idea to display all drop down options in the drop drown button itself. The discussion can continue in the issue. Thank you.

PR #7556, opened 2022-06-15
Add more documentation about PORO decorator requirements

sanfrecce-osaka:
Some methods need to be implemented in the following cases when using PORO decorator, but documentation did not mention it.

- use any actions with param(e.g. `show`, `edit`, `destroy` )
  - If the condition is not satisfied, there will be a problem so that decorator doesn't respond to param(by default, param is id)
- includes comments on resources
  - If the condition is not satisfied, there will be a problem with `ActiveAdmin::ResourceController::Decorators.undecorate` .

So I add explanation of the above cases.

### DIFF
--- a/docs/11-decorators.md
+++ b/docs/11-decorators.md
@@ -53,8 +53,8 @@ class PostDecorator
 end
 ```
 
-If you include comments on resources, your decorator class must respond to
-`model` returning a resource, and `decorated?` returning true.
+If a given resource uses ActiveAdmin's Comments feature, then that resource's decorator class must respond to
+`model` where it returns the model instance and `decorated?` returns `true`.
 
 ```ruby
 # app/decorators/post_decorator.rb
--- a/docs/11-decorators.md
+++ b/docs/11-decorators.md
@@ -53,6 +53,48 @@ class PostDecorator
 end
 ```
 
+If you include comments on resources, your decorator class must respond to
+`model` returning a resource, and `decorated?` returning true.
+
+```ruby
+# app/decorators/post_decorator.rb
+class PostDecorator
+  attr_reader :post
+  delegate_missing_to :post
+
+  def initialize(post)
+    @post = post
+  end
+
+  def decorated?
+    true
+  end
+
+  def model
+    post
+  end
+end
+```
+
+If you use any actions with param(e.g. show, edit, destroy), your decorator
+class must respond to `to_param`.
+
+```ruby
+# app/decorators/post_decorator.rb
+class PostDecorator
+  attr_reader :post
+  delegate_missing_to :post
+
+  def initialize(post)
+    @post = post
+  end
+
+  def to_param
+    post.id
+  end
+end
+```
+
 ## Forms
 
 By default, ActiveAdmin does *not* decorate the resource used to render forms.
--- a/docs/11-decorators.md
+++ b/docs/11-decorators.md
@@ -77,20 +77,17 @@ end
 ```
 
 If you use any actions with param(e.g. show, edit, destroy), your decorator
-class must respond to `to_param`.
+class must explicitly delegate `to_param` to the decorated model.
 
 ```ruby
 # app/decorators/post_decorator.rb
 class PostDecorator
   attr_reader :post
   delegate_missing_to :post
+  delegate :to_param, to: :post
 
   def initialize(post)
     @post = post
-  end
-
-  def to_param
-    post.id
   end
 end
 ```


sanfrecce-osaka:
appendix

## On using any actions with param(e.g. `show`, `edit`, `destroy` )

If the following decorator is passed to `decorate_with`,

```ruby
class PostDecorator
  attr_reader :post
  delegate_missing_to :post

  def initialize(post)
    @post = post
  end
end
```

raise following error on accessing `Admin::PostsController#show`.

```
ActiveRecord::RecordNotFound in Admin::PostsController#show

Couldn't find Post with 'id'=#< PostDecorator:0x0000000117067288>
```

Because `Object#to_param` returns value of `Object#to_s`.

cf. https://github.com/rails/rails/blob/bbbc861f717689a0a28f031fe2176d0f8a6b07c7/activesupport/lib/active_support/core_ext/object/to_query.rb#L7-L9

In this case, if `PostDecorator` implements `#to_param` returning `resource_id`, it works as expected.

## On including comments on resources

If the following decorator is passed to `decorate_with`,

```ruby
class PostDecorator
  attr_reader :post
  delegate_missing_to :post

  def initialize(post)
    @post = post
  end

  def to_param
    post.id
  end
end
```

raise following error on accessing `Admin::PostsController#show`.

```
NoMethodError in Admin::Posts#show

Showing /Users/xxx/sample_app/vendor/bundle/ruby/3.1.0/gems/activeadmin-2.13.1/app/views/active_admin/resource/show.html.arb where line #3 raised:

undefined method `base_class' for PostDecorator:Class

      ResourceController::Decorators.undecorate(resource).class.base_class.name.to_s
                                                               ^^^^^^^^^^^
```

Because `ResourceController::Decorators.undecorate` returns `PostDecorator` and it doesn't respond to `PostDecorator#decorated?`.

https://github.com/activeadmin/activeadmin/blob/b45b1fb05af9a7f6c5e2be94f61cf4a5f60ff3bb/lib/active_admin/resource_controller/decorators.rb#L20-L26

In this case, if `PostDecorator` implements `#model` returning resource and `decorated?` returning true, it works as expected.

deivid-rodriguez (approved):
Looks good to me!

docs/11-decorators.md
```diff
@@ -53,6 +53,48 @@ class PostDecorator
 end
 ```
 
+If you include comments on resources, your decorator class must respond to
+`model` returning a resource, and `decorated?` returning true.
```
javierjulio, line 57: Are you referring to "comments" as an example association for the example below or ActiveAdmin comments module?

docs/11-decorators.md
```diff
@@ -53,6 +53,48 @@ class PostDecorator
...
+  end
+
+  def model
+    post
+  end
+end
+```
+
+If you use any actions with param(e.g. show, edit, destroy), your decorator
+class must respond to `to_param`.
```
javierjulio, line 80: Are you saying that for any action with a default id parameter, the decorator requires this? Or its only required if you choose a different parameter other then the default id?

javierjulio (commented):
All good with this change, thanks! I just want to clarify each and then update the descriptions.

docs/11-decorators.md
sanfrecce-osaka: I'm referring to  "comments" as an ActiveAdmin comments module.

sanfrecce-osaka: For example, if you have disabled comments as follows,

```ruby
ActiveAdmin.register Post do
  config.comments = false
end
```

your decorator class doesn't require `model` and `decorated?`.

docs/11-decorators.md
sanfrecce-osaka: I intended the former.

sanfrecce-osaka: For example, if you choose a different parameter as follows,

```ruby
ActiveAdmin.register Post do
  decorate_with PostDecorator
  
  controller do
    def find_resource
      scoped_collection.where(name: params[:id]).first!
    end
  end
end

# app/models/post.rb
class Post < ApplicationRecord
  def to_param
    name
  end
end
```

your decorator class require `to_param` and returning name.

```ruby
class PostDecorator
  attr_reader :post
  delegate_missing_to :post

  def initialize(post)
    @post = post
  end

  def to_param
    post.name
  end
end
```

docs/11-decorators.md
javierjulio: Ha, I thought it was the former so glad I ask. Thanks. I'd suggest this to clarify.

```suggestion
If a given resource uses ActiveAdmin's Comments feature, then that resource's decorator class must respond to
`model` where it returns the resource and `decorated?` returns true.
```

docs/11-decorators.md
javierjulio: Sorry, I'm still confused by this one since you said its the former but showing where the parameter is customized. Are you saying that if `to_param` is overridden in the model, then the decorator should also respond to that method? Or actually just delegate to it no?

docs/11-decorators.md
sanfrecce-osaka: Thank you very much for your suggestion 🤩

docs/11-decorators.md
sanfrecce-osaka: Sorry for the lack of clarity 🙏 
PORO decorator always includes Object class in ancestor classes, so it can respond to `to_param`.
This means that if you expect the PORO decorator's `to_param` to act as decorated model's `to_param`, we need to explicitly delegate `to_param` to the decorated model.
From the above, I realized that my proposed code could be a bit more concise and clear.

```diff
  # app/decorators/post_decorator.rb
  class PostDecorator
    attr_reader :post
    delegate_missing_to :post
+   delegate :to_param, to: :post

    def initialize(post)
      @post = post
    end
- 
-   def to_param
-     post.id
-   end
  end
```

deivid-rodriguez: The additional clarification makes sense to me, can you include it in this patch?

sanfrecce-osaka: Thank you for your feedback, I update this path!

cf. https://github.com/activeadmin/activeadmin/pull/7556/commits/d7c8a38acd563fc4e14aa9307f7b371d6d7febe9

javierjulio (approved):
Thanks!

PR #7577, opened 2022-07-03
chore: Set permissions for GitHub actions

neilnaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -6,6 +6,9 @@ on:
   schedule:
     - cron: '1 1 * * *'
 
+permissions:
+  contents: read
+
 jobs:
   test:
     name: test (${{ matrix.ruby.name }}, ${{ matrix.deps }})


deivid-rodriguez:
Hello! I was reading about this and I have some questions.

* We don't use the `pull_request_target` workflow trigger, so we should be safe already, right? Is this in case at some point we change this workflow to be triggered on that event?

* Why didn't you change the other workflow in the same way?

javierjulio:
My understanding is this primarily around `pull_request_target` but if not, we can reopen. Thanks!

PR #7660, opened 2022-10-01
Replace gsub with tr on csv_filename

RicardoTrindade:
Minor performance improvement on streaming.rb#csv_filename.
As per https://github.com/fastruby/fast-ruby/blob/main/code/string/gsub-vs-tr.rb calling `tr` is usually faster than using `gsub`.

### DIFF
--- a/lib/active_admin/resource_controller/streaming.rb
+++ b/lib/active_admin/resource_controller/streaming.rb
@@ -31,7 +31,7 @@ module ActiveAdmin
       end
 
       def csv_filename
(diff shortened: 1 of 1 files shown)


javierjulio:
If accepting this, since we have Rubocop well integrated, I've used rubocop-performance elsewhere. I think it would be worthwhile to just include that and enable the rule for this. That way we get other instances and can start using rubocop-performance too. If that interests you for taking on, I would accept a PR for that along with this change. Thanks!

PR #7666, opened 2022-10-03
Add 'has_nested' class to menu only if at least one child is visible

icnagy:
`has_nested` class is applied to any menu item that has children without checking if those should be displayed or not. This results an empty dropdown in case where a menu item has no visible children.

image


javierjulio:
Can you please remove the `yarn.lock` update, then rebase and push to run updated CI? Thank you.

javierjulio:
Thank you. Closing this out since this will be changing significantly in the next major version.

Issue #7702, opened 2022-10-26
Using multi-attribute Ransack filter not properly creating/pulling label for active filters sidebar content display


corlissc:
When utilizing Ransack's DSL to create an index filter for multiple attributes, `ActiveAdmin::Views::ActiveFiltersSidebarContent` doesn't properly create or pull the label to display within the active filters sidebar. Instead it pulls the first attribute specified to do a lookup within the `ActiveFilter#resource.filters` hash. This leads to misleading active filter data within the UI.

For example with the following Rails model:
```
class Address < ActiveRecord::Base
  #Contains the following attributes:
  # id
  # street_line_1
  # street_line_2
  # city
  # state
  # zip
  # county
end
```
And a corresponding ActiveAdmin object with a filter searching across city and county:
```
ActiveAdmin.register Address do
  filter :city_or_county, as: :string, label: 'City or county'
end
```

### Expected behavior
When the city_or_county filter is utilized, I expect the active filters sidebar content to include an item with the label configured for the city_or_county filter (i.e. City or county) by utilizing the entirety of the filter (each attribute, any association, etc...) in order to build or properly lookup the correct filter metadata

### Actual behavior
The active filters sidebar content includes an item with only the name of the first attribute accounted for "City". Furthermore, if there was a custom label for the "city" filter, it would pull that filter's label.

### How to reproduce
./app/models/address.rb
```
class Address < ActiveRecord::Base
  #Contains the following attributes:
  # id
  # street_line_1
  # street_line_2
  # city
  # state
  # zip
  # county
end
```

./app/admin/addresses.rb
```
ActiveAdmin.register Address do
  filter :city_or_county, as: :string, label: 'City or county'
end
```

Start a rails server and utilize the filter... If the server started at http://localhost:3000, hitting the following URL would exercise the filter and show the output http://localhost:3000/admin/addresses?q[city_or_county_contains]=ma&commit=Filter&order=id_desc

ray-curran:
This looks similar to https://github.com/activeadmin/activeadmin/issues/5885. [The fix](resource.filters[condition.key.to_sym]) added the ability for the custom label when the condition has a predicate. So if you only need the single filter, including the predicate might help in the short term:

./app/admin/addresses.rb
```
ActiveAdmin.register Address do
  filter :city_or_county_contains, as: :string, label: 'City or county'
end
```

[...]

PR #7717, opened 2022-11-09
Prioritize filters with predicates in active filter labels (#7702)

ray-curran:
When there are filters for a single field and a multi-field ransack search with predicate, the multi-field label on the active filter should use the custom label from the filter input box.

Reproduction steps:

./app/admin/posts.rb
```
ActiveAdmin.register Post do
  filter :title_or_body_contains, as: :string, label: 'Title or Body'
  filter :title, as: :string

  ...
end
```

Navigate to the posts path for a user like
`localhost:3000/admin/users/4/posts`

Complete search in both fields, note that both active filters use the label "Title contains". The `title_or_body_contains` should use the label from the filter "Title or Body contains".


corlissc (approved):
@ray-curran looks good. This will resolve the issue that I submitted.

PR #7807, opened 2023-03-03
Use path_for helper to include url_options in breadcrumb links

drcapulet:
This isn't exactly the same behavior you'd get with a named path helper since it doesn't require the `params` key, but it is very close.


javierjulio:
What's the real objective/purpose here? I haven't had to do anything really with breadcrumb links so don't see what the need is here for.

drcapulet:
@javierjulio We need to propagate a query parameter to all links - something you can do by defining `url_options` in the controller to override [the default](https://github.com/rails/rails/blob/25f8d57100546bad54fbcf2487a34f459e8d2989/actionpack/lib/action_dispatch/routing/url_for.rb#L111-L116).

Issue #7290, opened 2022-01-21
Intermittent spec failure when testing with webpacker


deivid-rodriguez:
### Did you find a bug?

An intermittent spec failure.

### Expected behavior

Specs pass consistently.

### Actual behavior

```
Failures:

  1) ActiveAdmin::ResourceController::PolymorphicRoutes polymorphic_url with multiple belongs_to (not fully supported yet) within the posts page works with a category as parent
     Failure/Error: javascript_tag = active_admin_namespace.use_webpacker ? javascript_pack_tag(path) : javascript_include_tag(path)

     ActionView::Template::Error:
       767: unexpected token at ''
     # ./lib/active_admin/views/pages/base.rb:42:in `block (2 levels) in build_active_admin_head'
     # ./lib/active_admin/views/pages/base.rb:41:in `block in build_active_admin_head'
     # ./lib/active_admin/views/pages/base.rb:27:in `build_active_admin_head'
     # ./lib/active_admin/views/pages/base.rb:9:in `build'
     # ./app/views/active_admin/resource/index.html.arb:3:in `block in __home_runner_work_activeadmin_activeadmin_app_views_active_admin_resource_index_html_arb__2517167330308779720_47007235259800'
     # ./app/views/active_admin/resource/index.html.arb:1:in `new'
     # ./app/views/active_admin/resource/index.html.arb:1:in `__home_runner_work_activeadmin_activeadmin_app_views_active_admin_resource_index_html_arb__2517167330308779720_47007235259800'
     # ./lib/active_admin/resource_controller/streaming.rb:13:in `index'
     # ./spec/unit/resource_controller/polymorphic_routes_spec.rb:19:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # JSON::ParserError:
     #   767: unexpected token at ''
     #   ./lib/active_admin/views/pages/base.rb:42:in `block (2 levels) in build_active_admin_head'

Finished in 33.13 seconds (files took 4.71 seconds to load)
674 examples, 1 failure
```

### How to reproduce

Hard, for now.

tagliala:
Is it possible that this issue is caused by the following line?

https://github.com/activeadmin/activeadmin/blob/1b6dd5c2c5722f896ca36c3fa447b4db63038446/tasks/test_application.rb#L81

In any case, it should be fixed, the proper path is "gemfiles/rails_61_webpacker/Gemfile"

deivid-rodriguez:
[...]

PR #7851, opened 2023-03-12
Use Node 16 in webpacker

tagliala:
There is an issue with the Node 18 version installed on the CI that prevents webpacker to run correctly

```
Run tmp/test_apps/rails_61_webpacker/bin/webpack
node:internal/crypto/hash:71
  this[kHandle] = new _Hash(algorithm, xofLen);
```

Ref: activeadmin/activeadmin#7290

---

I think that the failure is not related. It should be because secrets are not available in forks and in PRs originating from a fork?

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,9 +27,9 @@ jobs:
         deps: [rails_61, rails_70]
 
         include:
-          # - ruby: 3.1
-          #   os: ubuntu-latest
-          #   deps: rails_61_webpacker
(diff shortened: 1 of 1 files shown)


javierjulio (approved):
Thank you so much! ❤️ I'm aware of the code coverage failure but not sure what's causing it. May just end up trying to switch to Codecov if all else fails. In the meantime, I'll leave this approved so I can merged once resolved.

PR #7852, opened 2023-03-12
CI Improvements

tagliala:
- Test against JRuby 9.4
- Update bug report template to Rails 7

### DIFF
--- a/tasks/bug_report_template.rb
+++ b/tasks/bug_report_template.rb
@@ -12,12 +12,23 @@ gemfile(true) do
   end
 
   # Change Rails version if necessary.
-  gem "rails", "6.1.4"
+  gem "rails", "~> 7.0.0"
 
-  gem "sprockets", "3.7.2"
-  gem "sassc-rails", "2.1.2"
-  gem "sqlite3", "1.4.2", platform: :mri
-  gem "activerecord-jdbcsqlite3-adapter", "61.0", platform: :jruby
+  gem "sprockets", "~> 3.7"
+  gem "sassc-rails"
+  gem "sqlite3", platform: :mri
+  gem "activerecord-jdbcsqlite3-adapter", platform: :jruby
+
+  # Fixes an issue on CI with default gems when using inline bundle with default
+  # gems that are already activated
+  # Ref: rubygems/rubygems#6386
+  if ENV["CI"]
+    require "net/protocol"
+    require "timeout"
+
+    gem "net-protocol", Net::Protocol::VERSION
+    gem "timeout", Timeout::VERSION
+  end
 end
 
 require "active_record"
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -18,8 +18,8 @@ jobs:
 
       matrix:
         os: [ubuntu-latest]
-        ruby: [jruby-9.3]
-        deps: ["rails_61"]
+        ruby: [jruby-9.4]
+        deps: ["rails_70"]
 
     env:
       JRUBY_OPTS: -J-Xmx1024m --dev
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
 
 env:
-  JRUBY_VERSION: jruby-9.3.4.0
+  JRUBY_VERSION: jruby-9.4
   RUBY_VERSION: 3.1
 
 jobs:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,7 +5,7 @@ on:
 
 env:
   JRUBY_VERSION: jruby-9.4
-  RUBY_VERSION: 3.1
+  RUBY_VERSION: 3.2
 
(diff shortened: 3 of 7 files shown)


tasks/bug_report_template.rb
```diff
@@ -12,12 +12,20 @@
...
-  gem "activerecord-jdbcsqlite3-adapter", "61.0", platform: :jruby
+  gem "activerecord-jdbcsqlite3-adapter", "70.0", platform: :jruby
+
+  # Prevents an issue with inline bundle when default gems have been
+  # already activated. May not be needed in newer versions of Ruby / bundle
+  # Ref: rubygems/rubygems#6386
+  if ENV["CI"]
+    gem "net-protocol", "0.1.1", platform: :jruby
+    gem "timeout", "0.3.0", platform: :jruby
+  end
```
javierjulio, line 28: Thank you! 🙌🏻 I could not figure this out for the life of me. The script was moved to the lint.yml workflow for now but it's commented out. Please uncomment it and let's see if its running here now.

.github/workflows/ci.yaml
```diff
@@ -113,7 +113,7 @@ jobs:
           path: coverage
 
       - name: Format, sum & upload results to Code Climate
-        uses: paambaati/codeclimate-action@v2.7.4
+        uses: paambaati/codeclimate-action@v3.2.0
```
javierjulio, line 116: Not sure why this is still failing but thank you for updating! Let's see if we can figure this out, otherwise I'll see about switching to Codecov which I have more experience with.

tagliala: I think this may be related to secrets and PR from forks, it is not the update itself. It is failing also with the very same version

javierjulio: Yes, you're right. Sorry about that. I read your message in the middle of the migration and realized then that makes sense. Either way though, I've been committed to migrating to Codecov which is now ready.

javierjulio: Ok the Codecov migration is merged so we can drop the this file change. Thank you again for your patience and help! Always appreciated.

tasks/bug_report_template.rb
javierjulio: Still failing. 😔 Not sure what it could be but appreciate you looking into it. If you are able to figure it out, would love to know what the issue is. I can accept the other changes in the meantime. Whether you prefer to resolve and do all at one, or separate the changes, is fine with me. Thank you.

.github/workflows/ci.yaml
tagliala: @javierjulio I will rebase #7851 so it can be merged.

As per this one, there are other issues to fix and I can't figure out the correct approach. I will try to suggest one. I think it will be acceptable since this only affects the CI

tasks/bug_report_template.rb
tagliala: >  would love to know what the issue is. 

Now it is MRI Ruby that has activated still different versions. I think this may be fixed in future releases of the offending dependencies. I will do some more tests in my fork and then provide a solution

.github/workflows/ci.yaml
javierjulio: Ok that sounds good to me. I appreciate your help. Thank you!

javierjulio (approved):
You're amazing. I've been trying to fix this one for awhile but didn't want to get rid of it. Thank you! ❤️ I really appreciate it.

tasks/bug_report_template.rb
```diff
@@ -12,12 +12,23 @@
...
+  # Fixes an issue on CI with default gems when using inline bundle with default
+  # gems that are already activated
+  # Ref: rubygems/rubygems#6386
+  if ENV["CI"]
+    require "net/protocol"
+    require "timeout"
+
+    gem "net-protocol", Net::Protocol::VERSION
+    gem "timeout", Timeout::VERSION
+  end
```
tagliala, line 31: a little bit hacky, I've tried several things, like

- `bundle exec ruby`
- Move this check to a separate workflow and do not bundle before

It appears that at the moment that ruby is running, timeout gem with an older version is already activated, so the inline bundle check fails

javierjulio: No issue with this at all. It didn't make much sense to me why it was happening. Just glad it's working now. Thanks again!

tasks/bug_report_template.rb
```diff
@@ -12,12 +12,23 @@
...
   # Change Rails version if necessary.
-  gem "rails", "6.1.4"
-
-  gem "sprockets", "3.7.2"
-  gem "sassc-rails", "2.1.2"
-  gem "sqlite3", "1.4.2", platform: :mri
-  gem "activerecord-jdbcsqlite3-adapter", "61.0", platform: :jruby
+  gem "rails", "~> 7.0.0"
+
+  gem "sprockets", "~> 3.7"
```
tagliala, line 17: `3.7` is still needed because 4.0 will ask for a manifest file, and I do not know how to avoid that

PR #7858, opened 2023-03-12
Do not specify net-smtp explicitly

tagliala:
mail gem has been fixed and this is not needed anymore


gemfiles/rails_61_webpacker/Gemfile.lock
```diff
@@ -255,6 +256,9 @@ GEM
       net-protocol
     nio4r (2.5.8)
     nio4r (2.5.8-java)
+    nokogiri (1.14.2)
+      mini_portile2 (~> 2.8.0)
+      racc (~> 1.4)
```
tagliala, line 261: those were missing from this lockfile only, I've left them in this commit

javierjulio (approved):
Great catch! I forgot about this. Thank you.

PR #7920, opened 2023-04-10
Fix typo in Vietnamese locale for filter text

tvziet:
Fixed a typo in the Vietnamese locale: "fortune" instead of "filter".


javierjulio:
@tvziet can you please share the translation was and the corrected form is now?

PR #7926, opened 2023-04-13
Update pagination wording when pagination_total is false

drcapulet:
Started out trying to fix `Displaying posts <b>1 - 0</b>` for empty collections with `pagination_total: false`, but then realized I could make a few other improvements without triggering a `total_count` call.


javierjulio:
Sorry, I'm not able to replicate this for an empty collection or for a collection with a single item, on v4. This PR I think was introduced before that work began where I've made several pagination changes. Since you mentioned making other improvements after opening it would be best to isolate those so please consider opening a new PR for just that at least. Thank you.

PR #7934, opened 2023-04-24
Enable Pundit authorization with namespaced decorators

### LINKED ISSUE #7933: Pundit authorization fails when using Admin namespaced decorator
### Did you find a bug?

I have a working Pundit setup, but I'm now trying to put my pundit policies under the `Admin` namespace (`config.pundit_policy_namespace = :admin).

I'm decorating some objects with a decorator class that is also in the `Admin` namespace, and this seems to cause authorization to fail.

### Expected behavior

When calling ActiveAdminPunditAdapter#authorized? the return value should be based on the policy for the wrapped subject.

### Actual behavior

ActiveAdminPunditAdapter#authorized? looks for a policy for the wrapper class, not finding it, and thus ends up raising an error (or basing the return value on the default policy, if defined).

### How to reproduce

I've set up a template below that tests 3 cases:

1. Subject decorated with a decorator in the Admin namespace (fails – pundit can't find policy)
2. Subject decorated with a decorator in the top level namespace (succeeds)
3. Subject not decorated (succeeds)

```ruby
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  if ENV["ACTIVE_ADMIN_PATH"]
    gem "activeadmin", path: ENV["ACTIVE_ADMIN_PATH"], require: false
  else
    gem "activeadmin", github: "activeadmin/activeadmin", require: false
  end

  gem "rails", "~> 7.0.0"

  gem "sprockets", "~> 3.7"
  gem "sassc-rails"
  gem "sqlite3", platform: :mri
  gem "activerecord-jdbcsqlite3-adapter", platform: :jruby

  gem "pundit"

  # Fixes an issue on CI with default gems when using inline bundle with default
  # gems that are already activated
  # Ref: rubygems/rubygems#6386
  if ENV["CI"]
    require "net/protocol"
    require "timeout"

    gem "net-protocol", Net::Protocol::VERSION
    gem "timeout", Timeout::VERSION
  end
end

require "active_record"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

  end

    t.string :full_name
  end
end

require "action_controller/railtie"
require "action_view/railtie"
require "active_admin"

  config.root = __dir__
  secrets.secret_token = "secret_token"
  secrets.secret_key_base = "secret_key_base"

  config.eager_load = false

  config.hosts = "www.example.com"
end

end

end

module Admin
  class UserPresenter
    attr_reader :user
    delegate_missing_to :user
    delegate :to_param, to: :user

    def initialize(user)
      @user = user
    end
  end
end

class UserPresenter
  attr_reader :user
  delegate_missing_to :user
  delegate :to_param, to: :user

  def initialize(user)
    @user = user
  end
end

module Admin
  class ApplicationPolicy
    attr_reader :user, :record

    def initialize(user, record)
      @user = user
      @record = record
    end

    def index?
      false
    end

    def show?
      false
    end

    def create?
      false
    end

    def new?
      create?
    end

    def update?
      false
    end

    def edit?
      update?
    end

    def destroy?
      false
    end

    class Scope
      def initialize(user, scope)
        @user = user
        @scope = scope
      end

      def resolve
        raise NotImplementedError, "You must define #resolve in #{self.class}"
      end

      private

      attr_reader :user, :scope
    end
  end
end

module Admin
  class UserPolicy < ApplicationPolicy
    def index?
      true
    end
  end
end


ActiveAdmin.setup do |config|

#  config.pundit_default_policy = 'Admin::ApplicationPolicy'
end

Rails.application.initialize!

  ActiveAdmin.routes(self)
end

require "minitest/autorun"
require "rack/test"
require "rails/test_help"

  def test_authorization_with_decorator_in_admin_namespace
    user = User.create!
    subject = Admin::UserPresenter.new(user)
    adapter = ActiveAdmin::PunditAdapter.new(subject, nil)

    assert adapter.authorized?(:index, subject)
  end

  def test_authorization_with_decorator_without_namespace
    user = User.create!
    subject = NonNamespacedUserPresenter.new(user)
    adapter = ActiveAdmin::PunditAdapter.new(subject, nil)

    assert adapter.authorized?(:index, subject)
  end

  def test_authorization_with_plain_subject
    user = User.create!
    subject = user
    adapter = ActiveAdmin::PunditAdapter.new(subject, nil)

    assert adapter.authorized?(:index, subject)
  end


  private

  def app
    Rails.application
  end
end
```

rogerkk:
I guess the testing could do with some love, and perhaps we should add tests for both namespaced and non-namespaced decorators. Am I on the right track here?

# What

When retrieving auth policies and the subject is wrapped in a namespaced decorator, Pundit is not able to find the policy. My original issue with full  description and code to reproduce is in issue #7933.

# How

This fix makes use of `ResourceController::Decorators.undecorate` to undecorate the target before asking pundit to fetch the policy.

It does this in `PunditAdaper#policy_target`, so as to have the fix affect `PunditAdapter#retrieve_policy` which in turn is used by  `PunditAdapter#authorized`.

Unless I'm missing something the remaining public methods are not affected by the issue at hand.

---

Fixes #7933

### DIFF
--- a/lib/active_admin/pundit_adapter.rb
+++ b/lib/active_admin/pundit_adapter.rb
@@ -56,7 +56,7 @@ module ActiveAdmin
       case subject
       when nil then resource
       when Class then subject.new
-      else subject
+      else undecorate(subject)
       end
     end
 
@@ -118,6 +118,9 @@ module ActiveAdmin
       @policies ||= {}
     end
 
+    def undecorate(subject)
+      ResourceController::Decorators.undecorate(subject)
+    end
   end
 
 end
--- a/spec/unit/pundit_adapter_spec.rb
+++ b/spec/unit/pundit_adapter_spec.rb
@@ -130,6 +130,38 @@ RSpec.describe ActiveAdmin::PunditAdapter do
       end
     end
 
+    context "when decorator name contains policy namespace name" do
+      before do
+        allow(ActiveAdmin.application).to receive(:pundit_policy_namespace).and_return :foobar
+      end
+
+      class Admin::PostDecorator
+        attr_reader :object
+        delegate_missing_to :object
+
+        def initialize(object)
+          @object = object
+        end
+
+        def decorated?
+          true
+        end
+
+        def model
+          object
+        end
+      end
+
+      it "asks Pundit for a policy for the undecorated object" do
+        policy = DefaultPolicy.new(double, double)
+        allow(policy).to receive(:show?).and_return(true)
+
+        expect(Pundit).to receive(:policy).with(anything, [:foobar, an_instance_of(Post)]).and_return(policy)
+
+        auth.authorized?(:read, Admin::PostDecorator.new(Post.new))
+      end
+    end
+
     context "when Pundit is unable to find policy scope" do
       let(:collection) { double("collection", to_sym: :collection) }
       subject(:scope) { auth.scope_collection(collection, :read) }


lukeasrodgers:
fwiw @rogerkk this would fix a similar issue I'm having with the `CanCanAdapter`, have been working around it by implementing custom `authorized?` logic that uses `subject.try(:decorated?)` but I think using `undecorate` is better

rogerkk:
@lukeasrodgers Ah, thanks for the verification!

If you want to have a stab at making codecov happy, I'll be happy to share the glory ;) If not I'll see if I can set off some time do it and see if it's possible to get the attention of a maintainer.

rogerkk:
Changing state of this PR from a draft, in the hopes of attracting maintainer attention. :sweat_smile: 

Is there any interest in getting this into master? If so then I can put a little effort into improving the tests.

PR #7935, opened 2023-04-27
Sanitize HTML in page title (remove HTML tags)

leoc:
This change removes the HTML from the title before it's added to the document `<head>`.

It allows to use links as return values from `#name` methods of decorators without causing weird HTML tags in the browsers or title tab bar.

### DIFF
--- a/lib/active_admin/views/pages/base.rb
+++ b/lib/active_admin/views/pages/base.rb
@@ -25,7 +25,7 @@ module ActiveAdmin
 
         def build_active_admin_head
           within head do
-            html_title [title, helpers.active_admin_namespace.site_title(self)].compact.join(" | ")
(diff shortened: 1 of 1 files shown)


javierjulio:
Can you please provide a code example(s) of the issue you ran into and your expected output as the title tag? It's very likely we'd want to do this anyway since it could contain user provided data and we should protect against, even if considered more the responsibility of the ActiveAdmin user. If you can add tests here, I'll accept the change. Thank you.

javierjulio:
@leoc I'm not able to update your PR or rebase it. I've already incorporated this change into a feature branch where this along with all other html head/assets configs being removed and replaced by standard Rails partials. By default the title tag rendered will now sanitize and remove all tags as you did here but can be overridden through a simple partial if needed. Thank you!

PR #7993, opened 2023-06-27
Improve Swedish translations

carlottostromstedt:
I saw translations being improved when looking at the last changelog so i thought i could help by improving the swedish translations.

There were some translations missing (compared to en.yml) and many translations were wrong. Should be fixed now :+1:


javierjulio (approved):
Thanks!

PR #8007, opened 2023-07-10
Allow `config.paginate` to be a block

allomov:
Hello, everyone!

I propose this change to allow `config.paginate` option to be set as a block that will return a value based on resource configuration.

In my case, I have a Sessions resource with an index that can be displayed as a table (default view) and as a calendar. With the calendar view, the pagination is replaced by the start/end time filter.

Using this change I can initialize resources in the following way:
```ruby
ActiveAdmin.register Session, as: "All Sessions" do
  config.paginate = proc { params[:as] != "calendar" }
  index(as: :calendar)
end
```
which will allow `ActiveAdmin::Views::IndexAsCalendar` will handle filtering instead of pagination.

Thank you for your attention! Please, tell me what you think about it.

### DIFF
--- a/lib/active_admin/resource/pagination.rb
+++ b/lib/active_admin/resource/pagination.rb
@@ -20,8 +20,12 @@ module ActiveAdmin
         @max_per_page = namespace.max_per_page
       end
 
-      def paginate
-        @paginate.is_a?(Proc) ? instance_eval(&@paginate) : @paginate
+      def paginate(controller: nil)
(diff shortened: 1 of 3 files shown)


allomov:
I am not sure why CI is failing, because the same command for [test (3.2, rails_70)](https://github.com/activeadmin/activeadmin/actions/runs/5511547683/jobs/10049692497?pr=8007#logs) and other configurations work fine, but fail on `test (3.2, rails_61)`

PR #8012, opened 2023-07-11
Add additional metadata to activeadmin.gemspec

Aesthetikx:
This commits adds additional metadata to `activeadmin.gemspec` based on the [rubygems specification](https://guides.rubygems.org/specification-reference/#metadata). This should add a few additional links to the sidebar on rubygems.org, namely Changelog, and also improve discoverability with other tools using the rubygems ecosystem.

### DIFF
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -18,7 +18,16 @@ Gem::Specification.new do |s|
 
   s.extra_rdoc_files = %w[CHANGELOG.md CODE_OF_CONDUCT.md CONTRIBUTING.md README.md]
 
-  s.metadata = { "rubygems_mfa_required" => "true" }
+  s.metadata = {
+    "bug_tracker_uri" => "https://github.com/activeadmin/activeadmin/issues",
+    "changelog_uri" => "https://github.com/activeadmin/activeadmin/blob/master/CHANGELOG.md",
+    "documentation_uri" => "https://activeadmin.info",
+    "homepage_uri" => "https://activeadmin.info",
+    "mailing_list_uri" => "https://groups.google.com/group/activeadmin",
+    "rubygems_mfa_required" => "true",
+    "source_code_uri" => "https://github.com/activeadmin/activeadmin",
+    "wiki_uri" => "https://github.com/activeadmin/activeadmin/wiki"
+  }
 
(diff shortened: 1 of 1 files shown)


activeadmin.gemspec
```diff
@@ -18,7 +18,16 @@ Gem::Specification.new do |s|
...
-  s.metadata = { "rubygems_mfa_required" => "true" }
+  s.metadata = {
+    "bug_tracker_uri" => "https://github.com/activeadmin/activeadmin/issues",
+    "changelog_uri" => "https://github.com/activeadmin/activeadmin/blob/master/CHANGELOG.md",
+    "documentation_uri" => "https://activeadmin.info/documentation.html",
+    "homepage_uri" => "https://activeadmin.info",
+    "mailing_list_uri" => "https://groups.google.com/group/activeadmin",
+    "rubygems_mfa_required" => "true",
+    "source_code_uri" => "https://github.com/activeadmin/activeadmin",
+    "wiki_uri" => "https://github.com/activeadmin/activeadmin/wiki"
```
javierjulio, line 29: I'd rather only declare what RubyGems doesn't pick up since the rest it's already figuring out correctly, as you alluded to. Thanks.

```suggestion
    "changelog_uri" => "https://github.com/activeadmin/activeadmin/blob/master/CHANGELOG.md",
    "rubygems_mfa_required" => "true"
```

Aesthetikx: Right on ✌️

tagliala: Hi,

I would suggest to reconsider and revert to the full definition

I do not know how rubygems figures out links, but it appears that links are outdated and non-secure

image

I guess that defining them in the gemspec will fix with the secure versions

javierjulio: Thanks. That's a good point @tagliala. I'm ok with the original submission then but with the caveat that the `documentation_uri` point to the base domain instead, same as homepage. @Aesthetikx sorry about that. Would you mind redoing the changes?

Aesthetikx: Right on 2.0 ✌️

javierjulio: @Aesthetikx thanks!

javierjulio (approved):
Thanks!

PR #8016, opened 2023-07-14
Create SECURITY.md

Joebonesthefig:
. don't forget to get the phone number for the badman


javierjulio:
Thank you. We've had security submissions so others are finding that already without this. Further, I don't know what content would go here initially. We'll revisit when we can. Thanks again.

PR #8021, opened 2023-07-22
Improve SimpleCov configuration

tagliala:
- Track all files in `/lib` for accurate coverage
- Use `test_frameworks` profile to exclude test folders
- Add `Libraries` and `Tasks` groups for HTML export

Note: `/tasks` files not yet covered

---

Other notes:

- You can see at https://app.codecov.io/gh/activeadmin/activeadmin that `features` is included and `tasks` only shows a single file
- I've deliberately ignored tasks files to avoid decreasing coverage too much and because I understand that tasks are tested in a way that coverage can't be checked
- In order to see the effects of `track_files`, you can run
   ```
   COVERAGE=true bin/rspec spec/unit/view_factory_spec.rb --seed 1
   ```
   on this branch (`2149 / 4785 LOC (44.91%)`) and against master (`2149 / 3954 LOC (54.35%)`)

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -1,8 +1,15 @@
 # frozen_string_literal: true
+
 SimpleCov.start do
-  add_filter %r{^/spec/}
+  load_profile "test_frameworks"
+
   add_filter "tmp/development_apps/"
   add_filter "tmp/test_apps/"
+
+  add_group 'Libraries', 'lib/'
+  add_group 'Tasks', 'tasks/'
+
+  track_files "lib/**/*.rb" # TODO: track "tasks/**/*.rb"
 end
 
 if ENV["COVERAGE"] == "true"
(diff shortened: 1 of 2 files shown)


.simplecov
```diff
@@ -1,8 +1,15 @@
 # frozen_string_literal: true
+
 SimpleCov.start do
-  add_filter %r{^/spec/}
+  load_profile "test_frameworks"
```
javierjulio, line 4: It seems this does something with features directory as well. My main concern there is I want to make sure its actually tracking the step files.

tagliala: > My main concern there is I want to make sure its actually tracking the step files.

Confirmed. this excludes test, specs, features

Ref: https://github.com/simplecov-ruby/simplecov/blob/main/lib/simplecov/profiles/test_frameworks.rb

I understand this use case, I will suggest a different approach

PR #8023, opened 2023-07-23
Drop support for Ruby 2.6

### LINKED ISSUE #8022: Ruby 2.6 is not supported anymore
### Did you find a bug?

ActiveAdmin is supposed to work with Ruby 2.6, as per gemspec

https://github.com/activeadmin/activeadmin/blob/8ed827b1f2ca4c319c4353efa8de38a1854b62bd/activeadmin.gemspec#L23

### Expected behavior

Works

### Actual behavior

Does not work, because Ransack 4 requires Ruby 2.7

Also, this going unnoticed because CI is not testing against all supported ruby versions

### How to reproduce

Attempt to `bundle` with 2.6

```
BUNDLE_GEMFILE=gemfiles/rails_61/Gemfile bundle install
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Could not find compatible versions

Because every version of activeadmin depends on ransack >= 4.0, < 5
  and ransack >= 3.1.0 depends on Ruby >= 2.7,
  every version of activeadmin requires Ruby >= 2.7.
So, because Gemfile depends on activeadmin >= 0
  and current Ruby version is = 2.6.10,
  version solving has failed.
```

tagliala:
Ransack 4 requires Ruby 2.7, so ActiveAdmin 3.x series does not support
Ruby 2.6 already and this commit fixes the misleading gemspec version

It has been decided not to run CI tests against 2.7 because it is in EOL

Close #8022

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,7 +11,7 @@ require:
 
 AllCops:
   DisabledByDefault: true
-  TargetRubyVersion: 2.6
+  TargetRubyVersion: 2.7
 
   Exclude:
     - tmp/development_apps/**/*
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
 
   s.metadata = { "rubygems_mfa_required" => "true" }
 
-  s.required_ruby_version = ">= 2.6"
+  s.required_ruby_version = ">= 2.7"
 
   s.add_dependency "arbre", "~> 1.2", ">= 1.2.1"
   s.add_dependency "formtastic", ">= 3.1", "< 5.0"


.github/workflows/ci.yaml
```diff
@@ -22,6 +22,7 @@ jobs:
           - 3.2
           - 3.1
           - '3.0'
+          - 2.7
```
javierjulio, line 25: No need to test on Ruby 2.7 since it's already EOL. I removed that intentionally. We can allow users to use Ruby 2.7 but no interest in testing that version in CI.

```suggestion
```

.rubocop.yml
```diff
@@ -11,7 +11,7 @@ require:
 
 AllCops:
   DisabledByDefault: true
-  TargetRubyVersion: 2.6
+  TargetRubyVersion: 2.7
```
javierjulio, line 14:
```suggestion
  TargetRubyVersion: 3.0
```

javierjulio (commented):
Thanks for doing this!

.rubocop.yml
tagliala: 3.0 may suggest changes that are not compatible with 2.7, actually breaking 2.7 compatibility

javierjulio: Thanks! Then we can leave the Rubocop config as it was. I don't care about having Ruby 2.7 in CI since its EOL. It's best-case scenario for others still on it that want to use it whether it works for them. It's intentional to not run Ruby 2.7 in CI.

tagliala: Done and clarified in the commit message why the CI will not run against 2.7

javierjulio: Thanks!

javierjulio (approved):
Thanks!

PR #8024, opened 2023-07-23
Introduce RuboCop Performance

tagliala:
Commits are separated by scope:

- Add gem
- Autofix safe, non-production offenses
- Autofix unsafe, non-production offenses
- Autofix safe, production offenses
- Autofix unsafe, production offenses
- Manually fix remaining production offenses

Notes:
- All changed lines of code are covered by specs (required #8034)
- `Performance/EndWith` changes should be safe because minimum required ruby version is 2.7, and 2.7 also supports `Symbol#end_with?`
- `Performance/InefficientHashSearch` fixes should be safe, because both `@menus` and `@children` are hashes
- `Performance/StringInclude` fixes should be safe, because `@column` is the result of a RegExp match
- `changelog:resync` and `local` rake tasks have been tested manually

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ inherit_mode:
 
 require:
   - rubocop-packaging
+  - rubocop-performance
   - rubocop-rails
   - rubocop-rspec
 
@@ -98,6 +99,162 @@ Packaging/RequireHardcodingLib:
   Enabled: true
 
 Packaging/RequireRelativeHardcodingLib:
+  Enabled: true
+
+Performance:
+  Enabled: true
+
+Performance/AncestorsInclude:
+  Enabled: false
+
+Performance/ArraySemiInfiniteRangeSlice:
+  Enabled: false
+
+Performance/BigDecimalWithNumericArgument:
+  Enabled: true
+
+Performance/BindCall:
+  Enabled: true
+
+Performance/BlockGivenWithExplicitBlock:
+  Enabled: true
+
+Performance/Caller:
+  Enabled: true
+
+Performance/CaseWhenSplat:
+  Enabled: true
+
+Performance/Casecmp:
+  Enabled: false
+
+Performance/ChainArrayAllocation:
+  Enabled: false
+
+Performance/CollectionLiteralInLoop:
+  Enabled: true
+
+Performance/CompareWithBlock:
+  Enabled: true
+
+Performance/ConcurrentMonotonicTime:
+  Enabled: true
+
+Performance/ConstantRegexp:
+  Enabled: true
+
+Performance/Count:
+  Enabled: true
+
+Performance/DeletePrefix:
+  Enabled: true
+
+Performance/DeleteSuffix:
+  Enabled: true
+
+Performance/Detect:
+  Enabled: true
+
+Performance/DoubleStartEndWith:
+  Enabled: true
+  IncludeActiveSupportAliases: true
+
+Performance/EndWith:
+  Enabled: true
+
+Performance/FixedSize:
+  Enabled: true
+
+Performance/FlatMap:
+  Enabled: true
+  EnabledForFlattenWithoutParams: false
+
+Performance/InefficientHashSearch:
+  Enabled: true
+
+Performance/IoReadlines:
+  Enabled: true
+
+Performance/MapCompact:
+  Enabled: false
+
+Performance/MapMethodChain:
+  Enabled: false
+
+Performance/MethodObjectAsBlock:
+  Enabled: true
+
+Performance/OpenStruct:
+  Enabled: true
+
+Performance/RangeInclude:
+  Enabled: true
+
+Performance/RedundantBlockCall:
+  Enabled: false
+
+Performance/RedundantEqualityComparisonBlock:
+  Enabled: false
+
+Performance/RedundantMatch:
+  Enabled: true
+
+Performance/RedundantMerge:
+  Enabled: true
+  MaxKeyValuePairs: 2
+
+Performance/RedundantSortBlock:
+  Enabled: true
+
+Performance/RedundantSplitRegexpArgument:
+  Enabled: true
+
+Performance/RedundantStringChars:
+  Enabled: true
+
+Performance/RegexpMatch:
+  Enabled: true
+
+Performance/ReverseEach:
+  Enabled: true
+
+Performance/ReverseFirst:
+  Enabled: true
+
+Performance/SelectMap:
+  Enabled: false
+
+Performance/Size:
+  Enabled: true
+
+Performance/SortReverse:
+  Enabled: true
+
+Performance/Squeeze:
+  Enabled: true
+
+Performance/StartWith:
+  Enabled: true
+
+Performance/StringIdentifierArgument:
+  Enabled: true
+
+Performance/StringInclude:
+  Enabled: true
+
+Performance/StringReplacement:
+  Enabled: true
+
+Performance/Sum:
+  Enabled: false
+
+Performance/TimesMap:
+  Enabled: true
+
+Performance/UnfreezeString:
+  Enabled: true
+
+Performance/UriDefaultParser:
   Enabled: true
 
 Rails/FilePath:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -133,6 +133,8 @@ Performance/ChainArrayAllocation:
 
 Performance/CollectionLiteralInLoop:
   Enabled: true
+  Exclude:
+    - spec/**/*
 
 Performance/CompareWithBlock:
   Enabled: true
@@ -186,6 +188,9 @@ Performance/MethodObjectAsBlock:
 
 Performance/OpenStruct:
   Enabled: true
+  Exclude:
+    - features/**/*
+    - spec/**/*
 
 Performance/RangeInclude:
   Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -188,9 +188,6 @@ Performance/MethodObjectAsBlock:
 
 Performance/OpenStruct:
   Enabled: true
-  Exclude:
-    - features/**/*
-    - spec/**/*
 
 Performance/RangeInclude:
   Enabled: true
--- a/lib/active_admin/resource_controller/streaming.rb
+++ b/lib/active_admin/resource_controller/streaming.rb
@@ -24,7 +24,7 @@ module ActiveAdmin
         headers["Last-Modified"] = Time.current.httpdate
 
         if ActiveAdmin.application.disable_streaming_in.include? Rails.env
-          self.response_body = block[String.new]
+          self.response_body = block[String.new] # rubocop:disable Performance/UnfreezeString to preserve encoding
         else
           self.response_body = Enumerator.new &block
         end
--- a/lib/active_admin/resource_controller/streaming.rb
+++ b/lib/active_admin/resource_controller/streaming.rb
@@ -31,7 +31,7 @@ module ActiveAdmin
       end
 
       def csv_filename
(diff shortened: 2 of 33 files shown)


lib/active_admin/resource_controller/streaming.rb
```diff
@@ -24,14 +24,14 @@ def stream_resource(&block)
         headers["Last-Modified"] = Time.current.httpdate
 
         if ActiveAdmin.application.disable_streaming_in.include? Rails.env
-          self.response_body = block[String.new]
+          self.response_body = block[(+"")]
```
javierjulio, line 27: This one I wasn't familiar with. This comes from the `Performance/UnfreezeString` cop but we don't have it enabled in our config. If you are confident with this change, then I'd be ok with it, otherwise I would suggest we disable that cop for now.

spec/unit/views/components/paginated_collection_spec.rb
```diff
@@ -53,7 +53,7 @@ def paginated_collection(*args)
 
     context "when specifying :param_name option" do
       let(:collection) do
-        posts = 10.times.map { Post.new }
+        posts = Array.new(10) { Post.new }
```
javierjulio, line 56: This is also coming from a cop that we don't have explicitly enabled. I've used it and since it's just for a spec too, I don't mind at all having it on. I'll comment separately about the cops config.

.rubocop.yml
```diff
@@ -100,6 +101,68 @@ Packaging/RequireHardcodingLib:
...
+  Enabled: true
+
+Performance/StringIdentifierArgument:
+  Enabled: true
+
+Performance/StringInclude:
+  Enabled: true
+
+Performance/Sum:
+  Enabled: true
```
javierjulio, line 164: Please change the enabled and option values here as needed. I just copied a config from elsewhere that has the full list of cops so you don't have to write out each one manually.

```suggestion
Performance/AncestorsInclude:
  Enabled: false

Performance/ArraySemiInfiniteRangeSlice:
  Enabled: false

Performance/BigDecimalWithNumericArgument:
  Enabled: true

Performance/BindCall:
  Enabled: true

Performance/BlockGivenWithExplicitBlock:
  Enabled: true

Performance/Caller:
  Enabled: true

Performance/CaseWhenSplat:
  Enabled: true

Performance/Casecmp:
  Enabled: false

Performance/ChainArrayAllocation:
  Enabled: false

Performance/CollectionLiteralInLoop:
  Enabled: true

Performance/CompareWithBlock:
  Enabled: true

Performance/ConcurrentMonotonicTime:
  Enabled: true

Performance/ConstantRegexp:
  Enabled: true

Performance/Count:
  Enabled: true

Performance/DeletePrefix:
  Enabled: true

Performance/DeleteSuffix:
  Enabled: true

Performance/Detect:
  Enabled: true

Performance/DoubleStartEndWith:
  Enabled: true
  IncludeActiveSupportAliases: true

Performance/EndWith:
  Enabled: true

Performance/FixedSize:
  Enabled: true

Performance/FlatMap:
  Enabled: true
  EnabledForFlattenWithoutParams: false

Performance/InefficientHashSearch:
  Enabled: true

Performance/IoReadlines:
  Enabled: true

Performance/MapCompact:
  Enabled: false

Performance/MapMethodChain:
  Enabled: false

Performance/MethodObjectAsBlock:
  Enabled: true

Performance/OpenStruct:
  Enabled: true

Performance/RangeInclude:
  Enabled: true

Performance/RedundantBlockCall:
  Enabled: false

Performance/RedundantEqualityComparisonBlock:
  Enabled: false

Performance/RedundantMatch:
  Enabled: true

Performance/RedundantMerge:
  Enabled: true
  MaxKeyValuePairs: 2

Performance/RedundantSortBlock:
  Enabled: true

Performance/RedundantSplitRegexpArgument:
  Enabled: true

Performance/RedundantStringChars:
  Enabled: true

Performance/RegexpMatch:
  Enabled: true

Performance/ReverseEach:
  Enabled: true

Performance/ReverseFirst:
  Enabled: true

Performance/SelectMap:
  Enabled: false

Performance/Size:
  Enabled: true

Performance/SortReverse:
  Enabled: true

Performance/Squeeze:
  Enabled: true

Performance/StartWith:
  Enabled: true

Performance/StringIdentifierArgument:
  Enabled: true

Performance/StringInclude:
  Enabled: true

Performance/StringReplacement:
  Enabled: true

Performance/Sum:
  Enabled: false

Performance/TimesMap:
  Enabled: true

Performance/UnfreezeString:
  Enabled: true

Performance/UriDefaultParser:
  Enabled: true
```

javierjulio (commented):
@tagliala thank you again for this. I've implemented rubocop-performance at work so I'm familiar with it. It's been a helpful tool. The main concern I have is with what was auto corrected vs what is reported in future updates.

With our config, there are some changes here that shouldn't have occurred since the cops are disabled by default. I believe if you **auto correct by rule** manually that will override our config settings. Reviewing your commits I believe it was auto correct by rule.

Can you please declare each rubocop-performance cop in the config and whether it should be enabled or not? This is important as we don't want to auto correct something here that isn't being reported as a warning in future updates. I've included a copy of all the cops config so you don't have to write out each one. Please change the values for each as needed.

lib/active_admin/resource_controller/streaming.rb
tagliala: >  If you are confident with this change, then I'd be ok with it,

Using this approach from a while, I would say that I'm quite confident of `+''` to generate unfrozen strings for the use cases I work with

Hovewer, according to https://github.com/fastruby/fast-ruby#stringdup-vs-string-code, `String.new` generates a `ASCII-8BIT` string, while `+''` generates an UTF-8 string. If this can be an issue in this context, we can disable it

tagliala:
Failure on updating code coverage does not seem to depend by these changes

.rubocop.yml
```diff
@@ -100,6 +101,168 @@ Packaging/RequireHardcodingLib:
...
+  Enabled: false
+
+Performance/MethodObjectAsBlock:
+  Enabled: true
+
+Performance/OpenStruct:
+  Enabled: true
+  Exclude:
+    - features/**/*
+    - spec/**/*
```
javierjulio, line 194: I didn't realize we use OpenStruct. We have two instances. I'd prefer that we remove them since we shouldn't be using them at all. I'll leave it up to you if you'd rather do that here or through a separate PR. It's just for two specs. 

For this PR though, at a minimum we should remove the `features` directory since there are no instances there and if there were any, we'd want to enforce this.

```suggestion
Performance/OpenStruct:
  Enabled: true
  Exclude:
    - spec/**/*
```

.rubocop.yml
```diff
@@ -100,6 +101,168 @@ Packaging/RequireHardcodingLib:
...
+  Enabled: false
+
+Performance/ChainArrayAllocation:
+  Enabled: false
+
+Performance/CollectionLiteralInLoop:
+  Enabled: true
+  Exclude:
+    - features/**/*
+    - spec/**/*
```
javierjulio, line 138: No need to exclude these since there are no violations currently so we can enforce for now.

```suggestion
```

lib/active_admin/resource_controller/streaming.rb
```diff
@@ -24,14 +24,14 @@ def stream_resource(&block)
...
-          self.response_body = block[String.new]
+          self.response_body = block[String.new] # rubocop:disable Performance/UnfreezeString to preserve encoding
         else
           self.response_body = Enumerator.new &block
         end
       end
 
       def csv_filename
-        "#{resource_collection_name.to_s.gsub('_', '-')}-#{Time.zone.now.to_date.to_formatted_s(:default)}.csv"
+        "#{resource_collection_name.to_s.tr('_', '-')}-#{Time.zone.now.to_date.to_formatted_s(:default)}.csv"
```
javierjulio, line 34: The merge conflict is on this line since `to_formatted_s(:default)` was changed to `to_s` in #8119.

javierjulio (approved):
Thanks @tagliala! Just two minor changes. This is ready to merge but has a single merge conflict.

.rubocop.yml
tagliala: There are four violations:

```
spec/unit/resource_controller/polymorphic_routes_spec.rb:42:9: C: Performance/CollectionLiteralInLoop: Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
        %w(posts user_posts).each do |current_page|
        ^^^^^^^^^^^^^^^^^^^^
spec/unit/resource_controller/polymorphic_routes_spec.rb:87:9: C: Performance/CollectionLiteralInLoop: Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
        %w(posts category_posts user_posts).each do |current_page|
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/unit/views/components/attributes_table_spec.rb:78:11: C: Performance/CollectionLiteralInLoop: Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
          [ ...
          ^
spec/unit/views/components/attributes_table_spec.rb:196:48: C: Performance/CollectionLiteralInLoop: Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
            expect(col.class_names).to include(["even", "odd"][index % 2])
```

.rubocop.yml
tagliala: Will remove those here

spec/unit/views/pages/form_spec.rb
```diff
@@ -16,7 +16,7 @@
     end
 
     let(:arbre_context) do
-      OpenStruct.new(params: params, helpers: helpers, assigns: {})
+      instance_double(Arbre::Context, helpers: helpers, assigns: {})
```
tagliala, line 19: @javierjulio does this work?

.rubocop.yml
javierjulio: Great thanks!

spec/unit/views/pages/form_spec.rb
javierjulio: Yes, I think so since tests run fine. Thanks!

.rubocop.yml
```diff
@@ -100,6 +101,164 @@ Packaging/RequireHardcodingLib:
 Packaging/RequireRelativeHardcodingLib:
   Enabled: true
 
+Performance:
+  Enabled: true
```
tagliala, line 105: should I remove this? This will implicitly enable all cops under `Performance`

javierjulio: No. It won't because we have `DisabledByDefault: true` set. This means that cops are now opt-in vs opt-out. On future Rubocop and sibling gem releases, new cops will be disabled and a log message will appear when running rubocop. Easier to see this in action. Here is sample output (partial) from running the rubocop command in CI from another repo with a rubocop-rails Dependabot PR.

```
...
  The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.
  
  Please also note that you can opt-in to new cops by default by adding this to your config:
    AllCops:
      NewCops: enable
  Rails/DangerousColumnNames: # new in 2.21
    Enabled: true
  Rails/RedundantActiveRecordAllMethod: # new in 2.21
    Enabled: true
  Rails/UnusedRenderContent: # new in 2.21
    Enabled: true
  For more information: https://docs.rubocop.org/rubocop/versioning.html
...
```

javierjulio:
@tagliala this is ready for you to merge. Thank you again for contributing this! ❤️

.rubocop.yml
tagliala: Ok, so I will do the same in capybara

tagliala:
@javierjulio welcome. Should I use squash & merge for this PR?

.rubocop.yml
javierjulio: Great thanks!

javierjulio:
@tagliala yes

PR #8025, opened 2023-07-23
Introduce RuboCop Capybara

tagliala:


features/step_definitions/tab_steps.rb
```diff
@@ -5,14 +5,15 @@
...
 end
 
 Then("I should see tab content {string}") do |string|
-  expect(page).to have_css(".tabs .tabs-content", text: string, visible: true)
+  expect(page).to have_css(".tabs .tabs-content", text: string, visible: :visible)
 end
 
 Then("I should not see tab content {string}") do |string|
+  # FIXME: change to `:all` or `:hidden` according to the expected behavior
   expect(page).to have_css(".tabs .tabs-content", text: string, visible: false)
```
tagliala, line 18: Quoting from https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/VisibilityMatcher

> `visible: false` does not find just invisible elements, but both visible and invisible elements. For expressiveness and clarity, use one of the # symbol values, `:all`, `:hidden` or `:visible`

@javierjulio if I use `:hidden` here, this spec fails. I can't tell if the expected behavior is `:hidden` or `:all`

```
(::) failed steps (::)

expected to find non-visible css ".tabs .tabs-content" with text "tab 2" but there were no matches. Also found "tab 1\ntab 2", which matched the selector but not all filters.  (RSpec::Expectations::ExpectationNotMetError)
./features/step_definitions/tab_steps.rb:17:in `"I should not see tab content {string}"'
features/show/tabs.feature:34:in `I should not see tab content "tab 2"'

Failing Scenarios:
cucumber features/show/tabs.feature:6 # Scenario: Set a method to be called on the resource as the title
```

javierjulio: Yes, this likely due to the JS update which now uses Flowbite instead of jQuery UI tabs on master. You may have created this PR before that and not updated afterwards since I see its behind. Sorry about that. I'm not sure what would be applicable here. I don't recall what happened with jQuery UI tabs, but now each of the tab content areas are defined but hidden using a CSS class. If you rebase this branch, run `bin/rake local server`, login and go to http://localhost:3000/admin/kitchensink you'll see a tabs demo to get an idea what's going. That should help identify what needs to change for the test if anything. Thanks!

tagliala: I do not remember if Cucumber + Capybara in this situation treats `class="hidden"` elements as hidden ones

However, this should fix the spec

```diff
 Then("I should see tab content {string}") do |string|
-  expect(page).to have_css(".tabs .tabs-content", text: string, visible: true)
+  expect(page).to have_css(".tabs .tabs-content :not(.hidden)", text: string)
 end
 
 Then("I should not see tab content {string}") do |string|
-  expect(page).to have_css(".tabs .tabs-content", text: string, visible: false)
+  expect(page).to have_css(".tabs .tabs-content .hidden", text: string)
 end
```

javierjulio: Ok great thanks!

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -202,17 +202,17 @@ def filter(name, options = {})
       let(:body) { filter :published_date, input_html: { 'autocomplete': "off" } }
 
       it "should generate provided input html for both ends of date range" do
-        expect(body).to have_selector("input.datepicker[name='q[published_date_gteq]'][autocomplete=off]")
-        expect(body).to have_selector("input.datepicker[name='q[published_date_lteq]'][autocomplete=off]")
+        expect(body).to have_css("input.datepicker[name='q[published_date_gteq]'][autocomplete=off]")
+        expect(body).to have_css("input.datepicker[name='q[published_date_lteq]'][autocomplete=off]")
```
javierjulio, line 206: There are a few of these in this file (at least) but I saw that it's using `have_css` vs `have_field`. Shouldn't this be using the latter with the class as a keyword argument?

spec/unit/form_builder_spec.rb
```diff
@@ -98,20 +98,20 @@ def build_form(options = {}, form_object = Post.new, &block)
...
     it "should only generate the form once" do
-      expect(body).to have_selector("form", count: 1)
+      expect(body).to have_css("form", count: 1)
     end
 
     it "should generate actions" do
-      expect(body).to have_selector("input[type=submit][value='Submit Me']")
-      expect(body).to have_selector("input[type=submit][value='Another Button']")
+      expect(body).to have_css("input[type=submit][value='Submit Me']")
+      expect(body).to have_css("input[type=submit][value='Another Button']")
```
javierjulio, line 114: Shouldn't this use the `have_button` matcher?

spec/unit/form_builder_spec.rb
```diff
@@ -98,20 +98,20 @@ def build_form(options = {}, form_object = Post.new, &block)
     end
 
     it "should generate a text input" do
-      expect(body).to have_selector("input[type=text][name='post[title]']")
+      expect(body).to have_field("post[title]", type: "text")
     end
 
     it "should generate a textarea" do
-      expect(body).to have_selector("textarea[name='post[body]']")
+      expect(body).to have_css("textarea[name='post[body]']")
```
javierjulio, line 105: Shouldn't this also be another `have_field` matcher? From the docs it says it should work with textarea's.

javierjulio (commented):
@tagliala thanks again for setting this up! I only left a few comments because there are many instances of each so it would apply across the board.

tagliala:
Maybe RuboCop Capybara does not flag some offenses:

```
# Offenses: Capybara/RSpec/HaveSelector, Capybara/SpecificMatcher
have_selector("input[name='title']")

# No offenses detected - Interpolation
have_selector("input[name='#{title}']")
```

```
# Offense: Capybara/SpecificMatcher
have_css("input[type=submit]")

# No offenses detected (multiple attributes)
have_css("input[type=submit][value='x']")
```

```
# Offense: Capybara/SpecificMatcher
have_css("input[name='post_author_attributes__first_name_']")

# No offenses: name with `[]`
have_css("input[name='post[author_attributes][first_name]']")
```

spec/unit/filters/filter_form_builder_spec.rb
tagliala: I was not able to find a way to test for `autocomplete` with `have_field` matcher

spec/unit/form_builder_spec.rb
tagliala: The matcher works, I've did the change in a separate commit, because `have_button` does not allow to specify the type, in case of revert

.rubocop.yml
```diff
@@ -27,6 +28,33 @@ AllCops:
 
   StyleGuideCopsOnly: false
 
+Capybara:
+  Enabled: true
+
+Capybara/ClickLinkOrButtonStyle:
+  Enabled: true
+
```
javierjulio, line 36: We should have all `rubocop-capybara` cops listed here. Several are missing.

Can you also list these in alphabetical order, please?

spec/unit/form_builder_spec.rb
javierjulio: Does the `have_field` matcher not work here and in general for any other tests using textarea? The docs state that it will.

.rubocop.yml
tagliala: is there a command line to list them all?

javierjulio: I don't know. When I set up Rubocop for a work project, I just copied the rules from the docs using some JS in a browser console. When I checked the `rubocop-capybara` docs, it was only a handful of rules missing. I haven't set up `rubocop-capybara` before, otherwise I would have shared the full list of cops like for `rubocop-performance`.

tagliala: I've used all the cops listed at https://docs.rubocop.org/rubocop-capybara/cops.html

spec/unit/form_builder_spec.rb
tagliala: I think it is because of `post[body]`. I've reported an issue at https://github.com/rubocop/rubocop-capybara/issues/78

javierjulio: Ok great thanks!

.rubocop.yml
javierjulio: Awesome thanks! These all look good.

javierjulio (approved):
@tagliala this is also ready for you to merge too. Thank you again! 🙌🏻

PR #8026, opened 2023-07-24
Use https where possible

tagliala:
- Use https where possible
- Separate commit: Remove links to the Demo hosted on Heroku (it does not work)


tagliala:
Eslint is failing for other reasons. Ref: #8027

javierjulio (approved):
Thanks!

PR #8028, opened 2023-07-24
Fix new eslint offenses

tagliala:


app/javascript/active_admin/features/batch_actions.js
```diff
@@ -45,7 +46,7 @@ const toggleDropdown = function(condition) {
   }
 }
 
-const toggleAllChange = function(event) {
+const toggleAllChange = function(_event) {
```
tagliala, line 49: Not sure if these changes are ok or if these function are missing `event.preventDefault` maybe someone with a deeper knowledge can check faster than me

javierjulio:
@tagliala thanks! I pushed the update directly on master because I preferred to disable the no-unused-vars rule since for the simple case of function arguments, a code editor is already well suited for highlighting unused variables. I'm not a fan of renaming arguments before or after the fact. Since we also have existing cases, I think it's fine to leave those as is.

PR #8031, opened 2023-07-26
Fix CSV responses when index overriding with the call alias method

megos:
CSV responses are implemented by overriding the index as below.

<https://github.com/activeadmin/activeadmin/blob/2692838b7520a6b78478beeda9e0adb8179e2f39/lib/active_admin/resource_controller/streaming.rb#L12-L17>

If a user overrides the index using the alias method instead of the super, this implementation is ignored and CSV responses are not working.

```rb
controller do
  def index
    index! do |format| # Cause "No renderer defined for format: csv"
      ...
    end
  end
end
```

I add an alias method to make this implementation work anytime.

### DIFF
--- a/app/controllers/active_admin/resource_controller/streaming.rb
+++ b/app/controllers/active_admin/resource_controller/streaming.rb
@@ -15,6 +15,7 @@ module ActiveAdmin
           yield(format) if block_given?
         end
       end
+      alias :index! :index
 
       protected
 
--- a/spec/unit/authorization/index_overriding_spec.rb
+++ b/spec/unit/authorization/index_overriding_spec.rb
@@ -5,19 +5,44 @@ RSpec.describe "Index overriding", type: :controller do
   before do
     load_resources { ActiveAdmin.register Post }
     @controller = Admin::PostsController.new
+  end
 
-    @controller.instance_eval do
-      def index
-        super do
-          render body: "Rendered from passed block"
-          return
+  context "with the call super" do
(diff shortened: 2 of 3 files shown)


mgrunberg:
@javierjulio this LGTM. What do you think?

javierjulio:
@mgrunberg not quite. It's why I held off but left this here as I'd like to see it resolved. I'm just not sure what the solution should be. I encountered this too the year before and resolved by calling `super` instead of `index!` inside the `index` action I had overridden. The following is the case I ran into:

```diff
controller do
  def index
    params[:some_flag] = "blah"
-   index!
+   super
  end
end
```

This is puzzling because I'm not sure why we are getting a new action method if InheritedResources would be defining this alias anyway since the Base applies the Actions module. Can you check if the controller has protected methods for `show!`, `create!`, etc.? I realized while responding, that this PR is setting a public alias but InheritedResources declares them as protected. Perhaps that would suffice here as I do recall one concern I had was the unexpected assertion change in the resource_controller_spec.rb file.

I'd like to see tests that have a custom index action that does something, e.g. set a param, before calling `index!` where a request is made for both `html` and `csv` to confirm it works as expected.

mgrunberg:
@javierjulio thanks for the feedback. I need some time to think about your comment. I'll come back to this in a few days. I don't discard even copy this PR as an "in repo".

Jcpayvoice23 (approved):

PR #8034, opened 2023-07-28
Remove multiline ternary operator

tagliala:
Replace multiline ternary operator with a standard `if...else` for readability and better code coverage

### master

Fake Coverage 100%

image

### this branch

Real Coverage 96.43%

image

### DIFF
--- a/lib/active_admin/order_clause.rb
+++ b/lib/active_admin/order_clause.rb
@@ -29,8 +29,11 @@ module ActiveAdmin
     end
 
     def table_column
-      (@column =~ /\./) ? @column :
+      if (@column =~ /\./)
+        @column
+      else
         [table, active_admin_config.resource_quoted_column_name(@column)].compact.join(".")
+      end
     end
 
     def sql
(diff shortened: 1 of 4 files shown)


lib/active_admin/order_clause.rb
```diff
@@ -29,8 +29,11 @@ def table
     end
 
     def table_column
-      (@column =~ /\./) ? @column :
+      if (@column =~ /\./)
+        @column
```
javierjulio, line 33: This class was new to me. This is also a worthwhile improvement so thank you. It would be best to resolve the code coverage patch here though before merging. Should be simple enough to do.

I tried out the spec and figured out that the following sample will trigger this branch in the code. Feel free to use this as a starting point for a basic test here for coverage.

```ruby
  describe "" do
    let(:clause) { "posts.id_asc" }
    subject { super().table_column }
    it { is_expected.to eq("posts.id") }
  end
```

javierjulio (approved):
Thanks!

Issue #6378, opened 2020-07-28
Filter Support for Citext Datatypes


dbenton9:
TL;DR: Add `:citext` as supported field detection for filters

I was recently trying to determine why my Active Admin panel was not showing the filter I had set for a field on my database:
```
filter :keyword
```
We discovered a workaround by changing the previous line to:
```
filter :keyword, as: :string
```
It was frustrating that before we fixed the issue Active Admin just appeared to be ignoring the requested filter. I began to dig around the Active Admin code base to discover this https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/filters/forms.rb.

In the `forms.rb` file, default filter types are determined for a variety of different data types, but not for citext. Our field `keyword` is a citext. Thus why it wasn't supported by default. 

Rails began [supporting citext](https://github.com/rails/rails/pull/12523) about 5 years ago.

I'd love to see Active Admin support citexts with a default filter type of `:string` like the datatypes `string` and `text` currently receive. I'm new to contributing, but I read I should get the feature request approved first before I start coding. So this is my feature request.

javierjulio:
Sure, we would be open to a PR that adds `citext` support for filters.

PR #8089, opened 2023-10-03
Add support for citext column types

Fire-Dragon-DoL:
https://github.com/activeadmin/activeadmin/issues/6378

Citext is now supported in filters using the same string filtering input.

For discussions, I bring the following proposals:
- [default_input_type](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/filters/forms.rb#L31-L40) in the filter form builder should raise an exception on the "case else" portion of the code, otherwise the inputs with invalid types are swallowed and hidden. Quite hard to notice on a deployment
- It doesn't seem like creating a `CitextInput` class is "found" by formtastic. What if we added an option to customize the "default input type" for filters?


Fire-Dragon-DoL:
I just realized that citext is not a thing in sqlite, so the tests cannot pass. I'll think of a solution

Fire-Dragon-DoL:
I couldn't find any equivalent.
What's the solution here, given the 1-liner:

Should I introduce postgres to the test suite?

lib/active_admin/filters/forms.rb
```diff
@@ -31,7 +31,7 @@ def default_input_type(method, options = {})
           case column.type
           when :date, :datetime
             :date_range
-          when :string, :text
+          when :string, :text, :citext
```
javierjulio, line 34: I'd be ok with just this since it's meant to behave no different a string.

I don't want or have the time to deal with introducing Postgres to the test suite.

Fire-Dragon-DoL: Ok, I will test one thing this weekend, I suspect I could define a custom citext type (it doesn't exist since there is no postgres). If that works, I should be able to test with sqlite, even if behind the scenes is just varchar

javierjulio: @Fire-Dragon-DoL are you able to patch this change and test it in your setup? I use Postgres but haven't used citext before. I'm ok with accepting just this change and no tests. It should work fine as is with Postgres since its text based like the other two.

Fire-Dragon-DoL: I'll update the pr to remove tests tomorrow morning. Sorry for the long wait

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -155,6 +155,26 @@ def filter(name, options = {})
     end
   end
 
+  describe "citext attribute" do
```
Fire-Dragon-DoL, line 158: Ok I tried hard creating a tests and it is possible, but everything is a messy hack unless postgres is introduced.

lib/active_admin/filters/forms.rb
Fire-Dragon-DoL: If more postgres specialization do arise, I should probably work on a way to introduce postgres to a subset of the test suite

javierjulio (approved):
Thanks!

PR #8091, opened 2023-10-06
Rails 7.1 Support

caseyprovost:
This is a WIP. Looks like assets need some fixing. We are working towards polishing this PR, but happy to have any assistance along the way :)


javierjulio:
Thank you. I'm closing in favor of #8102 since other changes will be needed to support Rails 7.1 and no need to install InheritedResources or Arbre from git now.

PR #8102, opened 2023-10-11
Support Rails 7.1

mgrunberg:
Related to https://github.com/activeadmin/activeadmin/issues/8090.

I made the necessary changes to run test against Rails 7.1

Some dependencies are in process of adding support to Rails 7.1:
- [x] Devise: https://github.com/heartcombo/devise/pull/5639
- [x] Ransack: https://github.com/activerecord-hackery/ransack/pull/1439
- [x] Formtastic: https://github.com/formtastic/formtastic/pull/1371
- [ ] Cucumber-Rails: https://github.com/cucumber/cucumber-rails/pull/575 - ⏳ **Pending release**

ActiveAdmin is set to raise an exception on any deprecation message.  Because of that, I had to update dependencies to non-released code. This doesn't seem right. 
There is also the possibility to log deprecations (not raise errors), let contributors to make PR to remove messages when they arise. Similar argument than the one used to remove upper bound dependencies limits from gemspec (https://github.com/activeadmin/activeadmin/pull/8098).
I'm open to suggestions.

Ransack is a special case because it's not just a deprecation warning, they use code that Rails 7.1 removed.

Finally, there are two changes that can be extracted from here into separate PR:
- [use to_s on dates](https://github.com/activeadmin/activeadmin/commit/e09d20beac1809c7a914bfa70ef1c66b0cee749f)
- [remove use of ActiveSupport::Deprecation singleton](https://github.com/activeadmin/activeadmin/commit/838b91bed63934254a1aea0ab660628c657a7a15)

PR #8105, opened 2023-10-12
Replace to_formatted_s with to_s to convert date to string

mgrunberg:
Solves the following deprecation warning in Rails 7.1:
```
DEPRECATION WARNING: to_default_s is deprecated and will be removed from Rails 7.2 (use to_s instead)
```

Check deprecation
[here](https://github.com/rails/rails/blob/v7.1.1/activesupport/lib/active_support/core_ext/date/conversions.rb#L59)

It shouldn't be a problem in [Rails 6.1](https://github.com/rails/rails/blob/v6.1.7.6/activesupport/lib/active_support/core_ext/date/conversions.rb#L59 ) or [Rails 7.0](https://github.com/rails/rails/blob/v7.0.8/activesupport/lib/active_support/core_ext/date/conversions.rb#L59 )

### Background

This commit was extracted from #8102. 

Rails 7.1 is technically supported in 3-0-stable because #8098 got merged. We can't run test against Rails 7.1 yet (#8102) but makes sense to me to merge the required changes while waiting.

### DIFF
--- a/lib/active_admin/resource_controller/streaming.rb
+++ b/lib/active_admin/resource_controller/streaming.rb
@@ -31,7 +31,7 @@ module ActiveAdmin
       end
 
       def csv_filename
-        "#{resource_collection_name.to_s.gsub('_', '-')}-#{Time.zone.now.to_date.to_formatted_s(:default)}.csv"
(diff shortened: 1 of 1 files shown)


tagliala (approved):

javierjulio (approved):
Thanks! Yes this is safe since it's just converting a date to string. The default output would be the same.

PR #8106, opened 2023-10-12
Remove deprecated usage of ActiveSupport::Deprecation singleton

mgrunberg:
Check more here https://edgeguides.rubyonrails.org/7_1_release_notes.html#add-rails-application-deprecators

### Background

This commit was extracted from #8102. 

Rails 7.1 is technically supported in 3-0-stable because #8098 got merged. We can't run test against Rails 7.1 yet (#8102) but makes sense to me to merge the required changes while waiting.


lib/active_admin.rb
```diff
@@ -61,6 +60,10 @@ def application
       @application ||= ::ActiveAdmin::Application.new
     end
 
+    def deprecator
+      @deprecator ||= ActiveSupport::Deprecation.new("4.0", "active-admin")
```
tagliala, line 64: should this use 3.0?

```suggestion
      @deprecator ||= ActiveSupport::Deprecation.new("3.0", "active-admin")
```

javierjulio: From what I've seen on how this is used/defined, I believe the version specified here is meant to default to the next major version so the original value is correct.

mgrunberg: First argument is `deprecation_horizon`, according to [rails comments](https://github.com/rails/rails/blob/v7.1.1/activesupport/lib/active_support/deprecation.rb#L61): `The version number in which the deprecated behavior will be removed`. 
It could be 3.2 (assuming the next release is 3.1). I chose 4.0 because according to https://github.com/activeadmin/activeadmin/pull/8099 the default branch would be v4. I thought it wouldn't harm to use next major.

tagliala: thanks for the clarification

javierjulio (approved):
Thanks!

PR #8117, opened 2023-10-22
Provide detail in DB statement timeout error for filters

Vicentecorrea:
Hello everyone! 👋

When a database timeout error occurs querying the values for a sidebar filter, the error message does not report that it is an error in the filters nor does it indicate the specific filter with the error, so I propose this change to give more information to the developer about the error so he can work on it.

Context: using a Rails application in which I use Active Admin, I accessed the view of a resource with many records and after a few seconds I got the error `PG::QueryCanceled: ERROR: canceling statement due to statement timeout`, but the error was not querying records for the table, it was querying values for a sidebar `SelectInput` filter.

### DIFF
--- a/lib/active_admin/inputs/filters/select_input.rb
+++ b/lib/active_admin/inputs/filters/select_input.rb
@@ -43,6 +43,8 @@ module ActiveAdmin
           else
             super
           end
+        rescue ActiveRecord::QueryCanceled => error
+          raise ActiveRecord::QueryCanceled.new "#{error.message.strip} while querying the values for the ActiveAdmin :#{method} filter"
         end
 
         def pluck_column
--- a/spec/unit/filters/filter_form_builder_spec.rb
+++ b/spec/unit/filters/filter_form_builder_spec.rb
@@ -173,6 +173,21 @@ RSpec.describe ActiveAdmin::Filters::ViewHelper do
         }
         body
       end
+
+      context "and a statement timeout error occurs" do
+        let(:body) { filter :title, as: :select, collection: ["foo"] }
+        let(:input_super_class) { Formtastic::Inputs::Base::Collections }
+        let(:db_timeout_exception) { ActiveRecord::QueryCanceled.new("ERROR: canceling statement due to statement timeout") }
+        let(:expected_exception_message) { "ERROR: canceling statement due to statement timeout while querying the values for the ActiveAdmin :title filter" }
+
+        before do
+          expect_any_instance_of(input_super_class).to receive(:collection).and_raise(db_timeout_exception)
+        end
+
+        it "should raise a database timeout error with a message indicating which filter was the cause" do
+          expect { body }.to raise_error(ActiveRecord::QueryCanceled, expected_exception_message)
+        end
+      end
     end
   end
 


Vicentecorrea:
@tagliala I am trying to code a spec (on the `spec/unit/filters/filter_form_builder_spec.rb` file within the `"when loading collection from DB"`) but in order to test the call to the `collection` method of the `ActiveAdmin::Inputs::Filters::SelectInput` class, I need to instantiate that class, and that has been difficult, since it needs several complex arguments, like an instance of `ActiveAdmin::Filters::FormBuilder` (which in turn for the `template` parameter needs an instance of `ActionView`), a `Ransack` object, etc.

Is there any way to facilitate the construction of this spec?

Vicentecorrea:
@tagliala good idea. Spec ready, thank you!

lib/active_admin/inputs/filters/select_input.rb
```diff
@@ -43,6 +43,8 @@ def collection
           else
             super
           end
+        rescue ActiveRecord::QueryCanceled => e
+          raise ActiveRecord::QueryCanceled.new "#{e.message.strip} while querying the values for the :#{method} filter"
```
javierjulio, line 47: We should include "ActiveAdmin" in the error message for clarity.

```suggestion
        rescue ActiveRecord::QueryCanceled => error
          raise ActiveRecord::QueryCanceled.new "#{error.message.strip} while querying the values for the ActiveAdmin :#{method} filter"
```

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -173,6 +173,21 @@ def filter(name, options = {})
         }
         body
       end
+
+      context "and a timeout error occurs" do
+        let(:body) { filter :title, as: :select, collection: ["foo"] }
+        let(:input_super_class) { Formtastic::Inputs::Base::Collections }
+        let(:db_timeout_exception) { ActiveRecord::QueryCanceled.new("ERROR: canceling statement due to statement timeout") }
+        let(:expected_exception_message) { "ERROR: canceling statement due to statement timeout while querying the values for the :title filter" }
```
javierjulio, line 181:
```suggestion
        let(:expected_exception_message) { "ERROR: canceling statement due to statement timeout while querying the values for the ActiveAdmin :title filter" }
```

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -173,6 +173,21 @@ def filter(name, options = {})
...
+
+        before do
+          allow_any_instance_of(input_super_class).to receive(:collection).and_raise(db_timeout_exception)
```
javierjulio, line 184: This is a hard requirement for the test to pass so it must use `expect*` instead of `allow*` as the latter can introduce false positives.

```suggestion
          expect_any_instance_of(input_super_class).to receive(:collection).and_raise(db_timeout_exception)
```

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -173,6 +173,21 @@ def filter(name, options = {})
...
+        end
+
+        it "should raise a DB timeout error with a message indicating which filter has failed" do
```
javierjulio, line 187:
```suggestion
        it "should raise a database timeout error with a message indicating which filter was the cause" do
```

spec/unit/filters/filter_form_builder_spec.rb
```diff
@@ -173,6 +173,21 @@ def filter(name, options = {})
         }
         body
       end
+
+      context "and a timeout error occurs" do
```
javierjulio, line 177:
```suggestion
      context "and a statement timeout error occurs" do
```

javierjulio (approved):
Thanks!

PR #5994, opened 2019-12-30
Extend menu to allow for nested submenus.

taralbass:
We have so many Active Admin resources that our menus were becoming overwhelming. I added the ability to created nested submenus by monkey-patching Active Admin. There is no issue for this feature, but I decided to make a pull request, as it seemed that this feature might be useful based on assorted old conversations that I came across. My implementation does not require (or support) that the parent is a resource. It just extends how parent menus work today to allow multiple levels of parent, and handles the associated styling.

active_admin_nested_menu_demo

PR #8132, opened 2023-11-03
Make sure menu creation does not modify menu options

### LINKED ISSUE #8078: Nested menu items are rendered top-level
### Did you find a bug?

We are still encountering the behavior described in this issue: https://github.com/activeadmin/activeadmin/issues/5712

### Expected behavior

The menus should be rendered as expected, with nested menu items being nested under the appropriate parent menu.

### Actual behavior

At times the nested menus are rendered on the top-level instead, as described in the linked issue (#5712). When the bug appears, the menus are randomly rendered correctly/incorrectly on page load. Restarting the servers solved the issue for us (temporarily, I'm assuming it can reappear at any time).

### How to reproduce

This is not easy to reproduce, in fact, in the same code base, this is the first time we have encountered the issue in 5-6 years in production.

### Info

ActiveAdmin version: 2.13.1
Rails version: 6.1.7.6
Ruby version: 3.1.4

dssjoblom:
This is an attempt to fix/work around #8078 and related issues.

If you look at the mentioned and related tickets, the problem is that sometimes nested menu items are rendered on the top-level, without a parent. I'm not 100% sure this PR fixes the problem, as it is difficult to reproduce, but this is a place in the code where the parent property of the menu items is modified unnecessarily.

fixes #8078

### DIFF
--- a/lib/active_admin/menu.rb
+++ b/lib/active_admin/menu.rb
@@ -48,6 +48,7 @@ module ActiveAdmin
       #   menu.add parent: 'Dashboard', label: 'My Child Dashboard'
       #
       def add(options)
+        options = options.dup # Make sure parameter is not modified
         parent_chain = Array.wrap(options.delete(:parent))
 
         item = if parent = parent_chain.shift


javierjulio:
Thank you @dssjoblom. I can understand how you think this is possibly related to the issue at hand. I do believe it's worth trying. I don't want to merge a change though without some idea that it's an improvement. For the original issue, how often do you think you would encounter it? For example, at least once a week? We would need to know some timeline and for how long you've run this patch in production for. So with that example, if you were running it for two weeks now and haven't encountered it again, that can suggest it's been resolved with the patch. Have you been running this in production as a monkey patch and/or a fork and how long for?

dssjoblom:
@javierjulio we aren't running it in production yet. Our release cycle is about once every 2 weeks, so I can include it in the next release (next week or the week after that). This issue doesn't happen very often (twice since September 15th), so it might be difficult to prove its absence even with a production test. In any case, I will let you know how it goes :slightly_smiling_face:

jdurand:
We were having the same problem when the initial request to an ActiveAdmin route upon app load was an XHR request. This modification resolves the issue, ensuring that the menu now loads as intended irrespective of the first request type.

dssjoblom:
@javierjulio as an update, we have been running this patch in production for almost a month now without encountering the menu rendering issue again, so I'm guessing this PR fixes it.

javierjulio:
@dssjoblom that's great news, thank you. I'll rebase and merge this shortly.

javierjulio (approved):
Thanks!

javierjulio:
Worth noting that if this really does fix it, the bug may have been introduced by #5994 which I have already planned to revert for the next major release.

jdurand:
@javierjulio you're going to revert #5994 because submenus were reimplemented in a different way or are you doing away with submenus entirely?

javierjulio:
@jdurand submenus are supported, with one level. That just introduced further nested levels which won't be supported in the next release. I don't have plans to reimplement that in a different way but I would consider improvements there. I'm working on extracting the menu rendering to a partial so it can be customized.

PR #8138, opened 2023-11-17
Prefer `match?` over `=~` to avoid MatchData

tagliala:
When the specific `MatchData` object is not needed, using `regexp.match?` is preferred over `=~` as it avoids the creation of a MatchData object, resulting in improved performance and reduced memory usage.

Example:

```rb
PARTS = ["admin", "users", "2b2f0fc2-9a0d-41b8-b39d-aa21963aaee4"]

%i[ips memory].each do |benchmark|
  Benchmark.send(benchmark) do |x|
    x.report('=~') { PARTS.each { |part| part =~ ID_FORMAT_REGEXP } }
    x.report('match?') { PARTS.each { |part| ID_FORMAT_REGEXP.match?(part) } }
    x.compare!
  end
end
```

```
Comparison (IPS):
              match?:   943080.8 i/s
                  =~:   621955.5 i/s - 1.52x  (± 0.00) slower

Comparison (Memory):
              match?:          0 allocated
                  =~:        168 allocated - Infx more
```

### DIFF
--- a/lib/active_admin/view_helpers/breadcrumb_helper.rb
+++ b/lib/active_admin/view_helpers/breadcrumb_helper.rb
@@ -2,6 +2,7 @@
 module ActiveAdmin
   module ViewHelpers
     module BreadcrumbHelper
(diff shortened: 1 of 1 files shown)


javierjulio (approved):
Thanks! Odd that the Performance/RegexpMatch rule didn't pick this up.

PR #8143, opened 2023-11-29
Send auth_object option to ransack

### LINKED ISSUE #8108: ransack never receives an auth object
### Did you find a bug?

`ransack` method never receives an "auth object" to be used in model methods.
If some filters are shown/hidden using `if:` based on authorization, this raises an exception now because `ransackable_attributes` might not be authorizing certain attributes to be searched for certain users.

### Expected behavior

https://github.com/activeadmin/activeadmin/blob/ecf730d8d51088690ea4fa5b6177d2e588b0b261/lib/active_admin/resource_controller/data_access.rb#L218

This line should pass a second argument that represents the auhorizing user. In case of devise, I expect this to be either a user, or a "cancancan" ability (I'm not sure which is the most correct), or something else if a custom auth plugin is used.
Maybe it could pass `auth_object: active_admin_authorization`, which is effectively what's used for authorization in activeadmin.

### Actual behavior

What actually happens?
The auth_object in `def self.ransackable_attributes(auth_object = nil)` is always nil

### How to reproduce

I will omit this since the source code is very explicit on the topic

### Ideas

Thinking about this, it opens an incredible opportunity where the auth object could be a wrapper around the authorization adapter **and the controller** (or the active admin config). In this way, it could be possible to provide a default implementation of ransackable attributes that's based on the action filters (I'm not sure if the same is possible for associations)

mgrunberg:
fixes #8108

The issue suggest to use `active_admin_authorization` as auth_object. It make sense to me.

I'm not implementing the idea of a default implementation of ransackable attributes. That doesn't seems easy to me. I would appreciate a PR or a battle tested example for that.


CHANGELOG.md
javierjulio, line 1: No need to change this file manually as the release notes are auto generated. I'm planning to remove the local tool and we'll use the generated changelog from GitHub Releases here in the future. For now you can just remove the changelog file changes here.

Choose from one of the PR labels that are prefixed with "type: ..." and apply it here. Those labels are used by the Ruby script when generating the changelog. We'll use the same in the future with GitHub releases.

mgrunberg: Thanks! Done

javierjulio: Thanks!

javierjulio (approved):
Thanks!

baldarn:
Sorry, is this going to be released anytime soon?

I saw that was merged, but still in 3.2.1 is not released

javierjulio:
@baldarn it is already released in v4. This change was seen as a breaking change so we won't release on v3.

qdegraeve:
@javierjulio sorry to reopen the discussion but v4 is not released as a stable version yet and is a huge change.
I don't really see why this is considered as a breaking change as the change goes from `nil` to a very specific `auth_object` that can only be activeadmin related.

Would it be possible to reconsider releasing it on v3 ?

baldarn:
> @javierjulio sorry to reopen the discussion but v4 is not released as a stable version yet and is a huge change. I don't really see why this is considered as a breaking change as the change goes from `nil` to a very specific `auth_object` that can only be activeadmin related.
> 
> Would it be possible to reconsider releasing it on v3 ?

we added a simple patch in our app to get this feature.

just monkey patch 

`lib/active_admin/resource_controller/data_access.rb`

and you are good to go!

javierjulio:
@qdegraeve those that worked on the feature view it as a breaking change and I trust that so no sorry it will not be released in v3. I don't see much concern either since it's a one line change in a two line method that hasn't been modified since it's original implementation and Metasearch-to-Ransack migration over a decade ago. As @baldarn mentioned it is simple and safe to just monkey patch if you need the behavior now on v3. Thank you.

qdegraeve:
Thanks for your reply, I'll add the patch as well

PR #8158, opened 2023-12-06
Targeting minimum Ruby version to 3.0

alejandroperea:
Some time ago Ruby 2.7 support was dropped, removing it from the CI but some references remained in the code.

I think we can target the minimum Ruby version to 3.0.

Let me know if you think something different!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,7 +13,7 @@ require:
 
 AllCops:
   DisabledByDefault: true
-  TargetRubyVersion: 2.7
+  TargetRubyVersion: 3.0
 
   Exclude:
     - tmp/development_apps/**/*
--- a/activeadmin.gemspec
+++ b/activeadmin.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
     "wiki_uri" => "https://github.com/activeadmin/activeadmin/wiki"
   }
 
-  s.required_ruby_version = ">= 2.7"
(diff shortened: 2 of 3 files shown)


mgrunberg:
Thanks for the PR!

Ruby 2.7 reached EOL about 9 months ago. And you are right that we are not testing against Ruby 2.7 anymore. But given the fact that Rails 7.1 still supports Ruby 2.7, I am more inclined to keep support for now.

Would you like to add Ruby 2.7 back to test matrix instead? I'm ok just testing Ruby 2.7 against latest Rails.

tagliala:
Attempted to test against 2.7, but the bundled nokogiri does not support 2.7, so we need to change gemfile.lock

I guess it is to much of effort

Ref: https://github.com/activeadmin/activeadmin/actions/runs/7499835745/job/20417354050#step:5:38

javierjulio:
@alejandroperea if you still want to complete this change, it just needs a rebase. The single merge conflict is because that file was deleted so you can drop that change and leave the other two. Thanks.

tagliala:
But 2.7 will not install because of Nokogiri (1.16 requires 3.0+, it should be changed in gemfiles)

javierjulio:
@tagliala yes, that's right.

mgrunberg:
closed in favor of https://github.com/activeadmin/activeadmin/pull/8259

PR #8194, opened 2023-12-30
Migrate docs to VitePress

javierjulio:
This migrates the custom docs site to now be built using VitePress which works with existing Markdown files. It helpful features built in like dark mode, mobile support, etc. Some Markdown files will need formatting updates to render correctly (e.g. wrong heading level).


mgrunberg:
@javierjulio have you considered keeping the content of current docs (v3)? 
This has two pros:
1) Users will be able to access old versions doc. Something good until they manage to upgrade to latest version.
2) We can merge this PR and keep evolving v4 docs on other PRs. Now we have to merge stuff to master and rebase this PR to make changes.

Just to be clear, I'm talking about the content, not the styles/UI (we should use VitePress now on for any version doc).

I can work on this idea if it makes sense.

javierjulio:
@mgrunberg yes but only with a repository link to the docs folder on the 3-0-stable branch. Vitepress doesn't support documentation versioning. We could probably move the v3 docs into a subfolder and then use the multiple sidebars feature which is path based alongside a nav option to select the current or v3 docs.

javierjulio:
@mgrunberg I extracted the general updates that were version independent #8424 and merged that in. Let me know if you are still interested in trying something out for docs versioning or to first share an approach to see if it's something we'll consider. Thanks!

mgrunberg:
> @mgrunberg I extracted the general updates that were version independent #8424 and merged that in. Let me know if you are still interested in trying something out for docs versioning or to first share an approach to see if it's something we'll consider. Thanks!

I'm still interested in this. I'm still figuring out how to do this but here is what I have in mind.

> use the multiple sidebars feature which is path based alongside a nav option to select the current or v3 docs.

I want to achieve something like that.

What I have in mind is:
1) Keep v3 doc in 3-0-stable branch
2) Fetch docs during doc-build process

Why fetching during build? I don't want to carry over old doc files in master. If old docs need an update (backport of a feature?), it make more sense to do the PR in that version branch.
I understand that we still gonna need to trigger a workflow in master to deploy the docs. It's a downside but nothing too bad in my opinion.

I haven't thought a lot on how to fetch. Could be a `node fetch`, a git export, a git submodule. There are ways for sure.

Following this idea, I'm wondering if the first step should be migrate current doc (v3) to vite-press. I don't know if there is a blocker for that.

Anyway, that's on my head. What do you think.

Oh, before I forget. I won't be able to work on this until Friday.

javierjulio:
I think that may be too much or too problematic since the v3 docs have to map to the expected pages which would change in the new setup. I don't mind duplicating the v3 docs on master branch since we need to build a single site and deploy it. I believe it would be easier to include the v3 docs in the new Vitepress setup. We'd have to do some formatting updates (e.g. fix headings) in some cases anyway. You are welcome to try either approach but figure just easiest to work with duplicate v3 docs in master. No rush on this. Thanks for expressing an interest in tackling it!

javierjulio:
@mgrunberg I've updated this PR so it just migrates the existing docs content to VitePress. The v4 docs changes I'll have in a new branch. We can figure out there how to incorporate both v3 and v4 docs in the VitePress site. At least this way we can migrate the docs to VitePress and hopefully others will want to make documentation updates. Let me know what you think.

.github/workflows/docs-deployment.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: yarn
+      - uses: actions/configure-pages@v4
+      - run: yarn install
+      - name: Build docs site
+        run: |
+          yarn docs:build
+          touch docs/.vitepress/dist/.nojekyll
```
mgrunberg, line 36: Seems that `.nojekyll` this is not needed anymore. 
It got removed from [github pages deploy docs](https://vitepress.dev/guide/deploy#github-pages) in march by https://github.com/vuejs/vitepress/pull/3630

```suggestion
        run: yarn docs:build
```

.github/workflows/docs-deployment.yml
```diff
@@ -0,0 +1,50 @@
+name: Docs Deployment
+
+on:
+  push:
+    branches:
+      - master
```
mgrunberg, line 6: It's better to deploy automatically when we cut a new release. If we need to update the docs without releasing a version we can trigger the workflow manually.

I want to prevent an out-synced doc if we merge a PR today and release it in a week.

```suggestion
  release:
    types: [published]
```

tasks/docs.rake
mgrunberg, line 1: If we are dropping `Yard` task, why don't we also delete https://github.com/activeadmin/activeadmin/blob/master/Gemfile#L53-L56?

package.json
```diff
@@ -29,14 +29,18 @@
     "eslint": "^9.8.0",
     "gherkin-lint": "^4.2.2",
     "rollup": "^4.19.2",
-    "tailwindcss": "^3.3.5"
+    "tailwindcss": "^3.3.5",
+    "vitepress": "^1.3.1"
```
mgrunberg, line 33: `vitepress` 1.4.0 is out. Does it make sense to bump before merge it the PR?

mgrunberg:
> @mgrunberg I've updated this PR so it just migrates the existing docs content to VitePress. The v4 docs changes I'll have in a new branch. We can figure out there how to incorporate both v3 and v4 docs in the VitePress site. At least this way we can migrate the docs to VitePress and hopefully others will want to make documentation updates. Let me know what you think.

@javierjulio The new docs look amazing! I left a few comments but the PR looks really good

.github/workflows/docs-deployment.yml
javierjulio: Great catch, thanks! I've made this change.

.github/workflows/docs-deployment.yml
javierjulio: This has been updated.

tasks/docs.rake
javierjulio: I've removed the docs group from the Gemfile.

package.json
javierjulio: The installed version is 1.4.0 if you check the lock file but for some reason the package.json doesn't change when updating. I'm not sure what command does it. I updated package.json and reran `yarn install` manually.

PR #8195, opened 2023-12-30
Update nl.yml for v4

caifara:
Adapts V4 i18n changes for Dutch translations. 

Also: chose "reactie" instead of "commentaar" (as both translations were in use).

### DIFF
--- a/config/locales/nl.yml
+++ b/config/locales/nl.yml
@@ -1,4 +1,8 @@
 nl:
+  views:
+    pagination:
+      previous: "Vorige"
+      next: "Volgende"
   active_admin:
     dashboard: Dashboard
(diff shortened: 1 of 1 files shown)


javierjulio (commented):
@caifara thank you. 👏🏻 This is missing a single v4 default change. Would you mind translating the `status_tag.unset` from "No" to "Unknown" in the Dutch equivalent? The default was changed to separate `false` from `nil`, e.g. in the case of booleans.

caifara:
@javierjulio done, thank you for all your effort!

javierjulio (approved):
@caifara thank you for your support and contribution! I really appreciate it. ❤️

PR #8209, opened 2024-01-05
Format daterange filter input values according to i18n default format

eikes:
When my locale is `de`, I expect my date format to be `%d.%m.%Y`.  The current behavior forces a different format. This PR makes sure, that the default format is used. Just like the formtastic date input does.


eikes:
Or might I even suggest to split the method and change it like this?
```ruby
        def input_html_options_for(input_name, placeholder)
          { placeholder: placeholder,
            value: input_value_for(input_name) }.merge(input_html_options)
        end

        def input_value_for(input_name)
          @object.public_send(input_name).to_s.to_date.to_s
        rescue
          ""
        end
```

lib/active_admin/inputs/filters/date_range_input.rb
```diff
@@ -39,7 +39,7 @@ def input_html_options_for(input_name, placeholder)
                             nil
                           end
           { placeholder: placeholder,
-            value: current_value ? current_value.strftime("%Y-%m-%d") : "" }.merge(input_html_options)
+            value: current_value ? current_value.to_s : "" }.merge(input_html_options)
```
tagliala, line 42: If this is correct, then it should be the same as

```suggestion
            value: current_value.to_s }.merge(input_html_options)
```

javierjulio (commented):
@eikes thank you. It would help to have a feature test for this. We have a locale step ("set my locale to") so you should be able to test this with a Capybara assertion that the input value is now set to expected format for the locale.

features/index/filters.feature
```diff
@@ -35,6 +35,20 @@ Feature: Index Filtering
...
+    """
+      ActiveAdmin.register Post
+    """
+    When I set my locale to "de"
+    And I am on the index page for posts
+    And I fill in "q[published_date_gteq]" with "20.10.1999"
+    And I press "Filtern"
+    Then the "q[published_date_gteq]" field should contain "20.10.1999"
+    When I set my locale to "en"
+    And I press "Filtern"
```
javierjulio, line 49:
```suggestion
    And I press "Filter"
```

eikes: No, it's still a German page at that point. I am only doing that, because the I18n locale is not reset otherwise...

eikes:
@javierjulio I created a test and the gherkin-linter is happy now as well.

features/index/filters.feature
eikes: (the  `When I set my locale to "en"` does not reload the page)

javierjulio (commented):
@eikes does this work as intended for you locally with the generated app? Granted, I haven't loaded a different locale before but tried with this and I'm not seeing the change, unless I'm misunderstanding it. Although, seems that [changing the language on the page will not suffice to see it in the browser](https://stackoverflow.com/a/2969367) which would explain why the date input appears the same for me. Just want to be sure it's working as intended for you. If so, then we'll get this in. Thanks!

features/index/filters.feature
javierjulio: Yes, that makes sense. At the time I thought this may have been one possible issue so wanted to call it out just in case. Thank you.

javierjulio:
@eikes can you describe a bit more why you want to make this change and for which version? Note that the default branch of ActiveAdmin is for v4 and we've removed jQuery UI and its date picker. Any date inputs will now be the native HTML date input. From what I can tell, [the date input parsed value is always in the same format of `yyyy-mm-dd` but the display will be based on the browser's locale](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#value) so if anything we wouldn't want to make this change and leave the previous format as it was. Are we missing something?

javierjulio:
@eikes from what we can tell, what we do currently is meant to be the way so I'm going to close this out. Thanks. Feel free to continue any discussion.

eikes:
I wasn‘t aware of the switch to the native datepicker. The PR is not needed in this case.

javierjulio:
@eikes that makes sense. No worries. We figured that was the case. Thank you for the follow up and confirmation.

PR #8213, opened 2024-01-07
Test against Ruby 3.3

tagliala:


mgrunberg:
@tagliala LGTM.

I'm wondering if we must narrow the matrix to reduce the amount of combinations. For example, since Ruby 3.0 is about to reach EOL we could just test against Rails 7.1. 

What do you think?

tagliala:
> For example, since Ruby 3.0 is about to reach EOL we could just test against Rails 7.1.

I will do the same for 2.7 and address #8158 too

mgrunberg (approved):

tagliala:
The following should be changed:

image

PR #8218, opened 2024-01-07
Move pagination translations to active_admin scope

### LINKED ISSUE #8215: v4 beta 3: Kaminari translation override may affect the host application
Not sure that this should be part of activeadmin's `locale.yml` as it is

Kaminari may be used also in the "host" application, so this will override translations of the host front-end

https://github.com/activeadmin/activeadmin/blob/4318a7a5ca1845e08437e8c0d0321ea659b6aa08/config/locales/en.yml#L2-L5

Introduced in 9999e80cadaf16a8d1d0a6e092229107979f5760

tagliala:
Active Admin uses the Kaminari gem for pagination, which may also be used by the host application. Moving the translations from `views.pagination` to `active_admin.pagination` prevents potential conflicts.

Fix #8215

### DIFF
--- a/app/views/active_admin/kaminari/_next_page.html.erb
+++ b/app/views/active_admin/kaminari/_next_page.html.erb
@@ -8,7 +8,7 @@
 -%>
 <% unless current_page.last? %>
   <%= link_to url, rel: 'next', remote: remote, class: "flex items-center justify-center px-2.5 py-3 h-8 leading-tight text-gray-500 dark:text-gray-400 hover:bg-gray-100 hover:text-gray-700 dark:hover:bg-gray-800 dark:hover:text-white rounded no-underline" do %>
-    <span class="sr-only"><%= t('views.pagination.next').html_safe %></span>
+    <span class="sr-only"><%= t('active_admin.pagination.next') %></span>
     <svg class="w-2.5 h-2.5 rtl:rotate-180" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 6 10">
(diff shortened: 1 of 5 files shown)


app/views/active_admin/kaminari/_next_page.html.erb
```diff
@@ -8,7 +8,7 @@
 -%>
 <% unless current_page.last? %>
   <%= link_to url, rel: 'next', remote: remote, class: "flex items-center justify-center px-2.5 py-3 h-8 leading-tight text-gray-500 dark:text-gray-400 hover:bg-gray-100 hover:text-gray-700 dark:hover:bg-gray-800 dark:hover:text-white rounded no-underline" do %>
-    <span class="sr-only"><%= t('views.pagination.next').html_safe %></span>
+    <span class="sr-only"><%= t('active_admin.pagination.next').html_safe %></span>
```
tagliala, line 11: not sure about this, since the default template is using this translation in a screen reader only span, wouldn't be better to remove `html_safe`?

javierjulio: Yes, I think so. I left this as is because it's what Kaminari used but then it had to as it used HTML entities for arrows which we don't want. We just need the plain text. I'll leave it up to you to make that change if you want. This is all set to merge either way so thank you!

javierjulio: I just rebased so the tests should now pass. Sorry about that. I was adding the changelog from 3.2.0 into default branch and forgot we run tests on the changelog but no more.

javierjulio (approved):
Thanks! Feel free to remove the `.html_safe` if you want. I can release beta4 once you merge.

javierjulio (approved):
Thanks! ❤️

PR #8228, opened 2024-01-11
Make importmap-rails optional

mgrunberg:
Currently, the engine requires the `importmap-rails` gem during initialization. Is not possible to install the gem without it.

I want to be able to test other alternatives to bundle the javascript. They won't be supported by default, like importmaps-rails.

### DIFF
--- a/lib/active_admin.rb
+++ b/lib/active_admin.rb
@@ -9,6 +9,12 @@ require "formtastic_i18n"
 require "inherited_resources"
 require "arbre"
 
+begin
+  require "importmap-rails"
+rescue LoadError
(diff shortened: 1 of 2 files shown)


jeremybdk:
It would be great to be able to use this new version with shakapacker + existing tailwind setup or any other setup without depending on import map thanks

javierjulio:
@jeremybdk yes, that's the intention of the v4. It was just an oversight to not guard in the engine file. So don't worry. We'll get to this soon and have it be in a beta4 release. Thanks.

javierjulio (approved):
Thanks!

PR #8231, opened 2024-01-12
Cleanup tailwind template

mgrunberg:
Small tweaks tailwind template.

I believe there is no need to scan our plugin.js. It's not javascript that we use in our pages.

Also, we don't have views ending with html / rb extensions.

I understand if these changes are rejected


javierjulio:
Since we define these in the components layer, we scan the file so Tailwind picks them up otherwise they aren’t applied. We aren’t scanning the lib directory as I replaced that with this. We could change that all in the future but this is required currently. Sorry.

mgrunberg:
> Since we define these in the components layer, we scan the file so Tailwind picks them up otherwise they aren’t applied. We aren’t scanning the lib directory as I replaced that with this. We could change that all in the future but this is required currently. Sorry.

Oh, interesting. Excelent, I'm closing the PR then

PR #8232, opened 2024-01-12
Exclude primary key from generated resource

mgrunberg:
Resource generator produces an index definition with two "id" columns: one by calling `column :id` and other calling `id_column`

```
  # Add or remove columns to toggle their visiblity in the index action
  index do
    selectable_column
    id_column
    column :id
    actions
  end
```

This PR makes sure we generate just one.


javierjulio (approved):
Great catch, thanks!

PR #8234, opened 2024-01-12
Make NPM package compatible with tools like vite_rails

mgrunberg:
### Context

This PR is related to https://github.com/activeadmin/activeadmin/pull/8228. 
It changes the content of our npm package to be compatible with tools like `vite_rails`.

_Note:_ Our source code is still written for a host app using `importmap-rails`. This PR doesn't change that. Our tests and development apps work like that and there is no plan to change it.

### Description

Our source code is written with importmaps in mind. All imports are absolute and based on `active_admin` alias/map.

We don't distribute any importmap definition within our npm package. That makes it impossible to use our package with bundlers/tools like vite, webpack, rollup, etc out of the box.

To integrate with `vite_rails` I want to define the following file

```javascript
// app/javascript/entrypoints/active_admin.js
import '@activeadmin/activeadmin'
```

Our main file is `app/javascript/active_admin.js`, when `vite` parses the file it fails to resolve imports like `import "active_admin/features/dark_mode_toggle"`. Even more, since we are vendoring external dependencies `vite` can't process `import "flowbite"`.

This PR does two things:
1) add external dependencies to package.json
2) add a build step to preprocess our code and generate a dist folder where alias is replaced with relative imports.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,10 +1,10 @@
 import path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { URL, fileURLToPath } from 'node:url';
 import { readFileSync } from 'node:fs';
 import alias from '@rollup/plugin-alias';
 
 const packageJson = JSON.parse(
-	readFileSync(new URL('./package.json', import.meta.url))
+  readFileSync(new URL('./package.json', import.meta.url))
 );
 
 const __dirname = fileURLToPath(new URL('.', import.meta.url));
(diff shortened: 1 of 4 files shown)


rollup.config.js
```diff
@@ -0,0 +1,40 @@
...
+				entries: [
+					{
+						find: 'active_admin',
+						replacement: path.join(assetsDir, 'active_admin'),
+					},
+				]
+			})
+		]
+	}
+];
```
tagliala, line 40: indentation here seems to use tabs instead of spaces

javierjulio: Yes should be spaces, sorry. Of course this is always the first thing I'd bring up too. 🙈

mgrunberg: Fixed

javierjulio (approved):
Thanks! Interesting, so all this does is just set relative paths but the JS output and structure, is all the same. Would the same import and setup work with other JS bundlers, say like esbuild in jsbundling-rails?

mgrunberg:
> Thanks! Interesting, so all this does is just set relative paths but the JS output and structure, is all the same. Would the same import and setup work with other JS bundlers, say like esbuild in jsbundling-rails?

Correct, this should work with esbuild and other bundlers.

PR #8235, opened 2024-01-12
Improve Italian translation

tagliala:
- Add new strings
- Use plural in `pagination.empty` key, since `model` is plural there. Also slightly change the sentence so that it could work with all grammatical genders


javierjulio (approved):
Thanks! When you're ready to merge, just apply the "type enhancement" label. That way this will show up under enhancements in the next release notes.

PR #8237, opened 2024-01-15
Sync vendor and package json

mgrunberg:
This PR is a WIP. I'm not gonna be able to continue working on it for a while. I'm leaving as a way to show what I was doing.

My goal is to write something general, independent of the dependencies we have today. 

I found an issue with this approach. npm registry is hierarchical while importmaps is not. Let's take `flowbite` dependency as an example. When you pin it using importmaps it adds `@popperjs/core` to the map.

Flowbite has one file in it's package that includes popperjs in it. It's not picked by default. I guess it will be hard to code a universal way of finding this files, if they exists.

I don't want to end with an importmap of the importmap (a file saying which file to download for each importmap exception).

On idea to explore is use importamp.rb comments to decide what to do with the dependency.

```
# Default import
pin "@rails/ujs", to: "@rails--ujs.js" # @7.1.2

# Custom import 
pin "flowbite", preload: true # downloaded from https://cdnjs.cloudflare.com/ajax/libs/flowbite/2.2.1/flowbite.min.js
```

Default imports can be handled with the code that is right now in the PR.

For custom import we can rebuild the URL using package.json version and download the file. If the file is gone, raising an error force us to rewrite the comment with proper URLs.

The idea is not simple. Maybe we can use another approach, like always generate the importmap from package.json. In that case I'm not sure how to force/check that importmap.rb is in sync with package.json.

### DIFF
--- a/bin/prep-release
+++ b/bin/prep-release
@@ -29,5 +29,6 @@ end
 bump_version_file(version)
 system "bin/bundle"
 bump_npm_package(version)
+system "yarn install --frozen-lockfile"
 system "bin/rake", "dependencies:vendor"
 system "yarn build"
--- a/bin/prep-release
+++ b/bin/prep-release
@@ -29,4 +29,5 @@ end
 bump_version_file(version)
 system "bin/bundle"
 bump_npm_package(version)
+system "bin/rake", "dependencies:vendor"
 system "yarn build"
--- a/tasks/dependencies.rake
+++ b/tasks/dependencies.rake
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+namespace :dependencies do
+  desc "Copy package.json dependencies into vendor/javascript"
+  task :vendor do
+    # Update dependencies
+    system "yarn install --frozen-lockfile"
+
+    node_modules = File.expand_path("../node_modules", __dir__)
+    vendor = File.expand_path("../vendor/javascript", __dir__)
+
+    # Copy flowbite to vendor
+    FileUtils.cp(
+      File.join(node_modules, 'flowbite', 'dist', 'flowbite.min.js'),
+      File.join(vendor, 'flowbite.js')
+    )
+
+    # Delete sourcemaps refs
+    Dir.glob(File.join(vendor, '**', '*.js')).each do |file|
+      content = File.read(file)
+
+      File.write(file, content.gsub(/\/\/# sourceMappingURL=\S+/, ''))
+    end
+  end
+end
--- a/tasks/dependencies.rake
+++ b/tasks/dependencies.rake
@@ -3,9 +3,6 @@
 namespace :dependencies do
   desc "Copy package.json dependencies into vendor/javascript"
   task :vendor do
-    # Update dependencies
-    system "yarn install --frozen-lockfile"
-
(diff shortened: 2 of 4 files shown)


mgrunberg:
As we spoke in slack, [Rails deprecated rails_ujs in Rails 7.1](https://github.com/rails/rails/pull/50535) so it doesn't make sense to do an automatic task with it. I'm using rake to copy node_modules dependencies into vendor. Why rake and note a node script? Because I started the PR with rake.

javierjulio:
@mgrunberg thanks! All this looks good, I'm just having trouble running it locally. I have this branch checked out and tried running both `bin/rake dependencies:vendor` and `rake dependencies:vendor` but both fail. I don't know why.

```
> bin/rake dependencies:vendor
rake aborted!
Don't know how to build task 'dependencies:vendor' (See the list of available tasks with `rake --tasks`)

(See full trace by running task with --trace)
```

mgrunberg:
@javierjulio please try again. I forgot to import the task in Rake file. Honestly, I don't know how this worked for me without that :thinking:

tasks/dependencies.rake
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+namespace :dependencies do
+  desc "Copy package.json dependencies into vendor/javascript"
+  task :vendor do
+    # Update dependencies
+    system "yarn install --frozen-lockfile"
+
```
javierjulio, line 8: We are running this twice now in the test app setup which we should avoid. I think we need this in the `bin/prep-release` script instead which would be consistent with the `bin/bundle` line we run there.

```suggestion
```

bin/prep-release
```diff
@@ -29,4 +29,5 @@ end
 bump_version_file(version)
 system "bin/bundle"
 bump_npm_package(version)
+system "bin/rake", "dependencies:vendor"
```
javierjulio, line 32:
```suggestion
system "yarn install --frozen-lockfile"
system "bin/rake", "dependencies:vendor"
```

javierjulio:
@mgrunberg thanks! That makes sense. I couldn't recall that yesterday. It's working now.

tasks/test_application.rb
```diff
@@ -18,6 +18,7 @@ def soft_generate
       end
       Bundler.with_original_env do
         Kernel.system("yarn install") # so tailwindcss/plugin is available for test app
+        Kernel.system("rake dependencies:vendor") # so flowbite is available for test app
```
javierjulio, line 21: This would be available regardless (different reason from the line comment above), we just want to ensure the file is updated from the installed version in our npm package dependencies.

```suggestion
        Kernel.system("rake dependencies:vendor") # ensure flowbite is updated for test app
```

tasks/dependencies.rake
mgrunberg: done

bin/prep-release
mgrunberg: done

tasks/test_application.rb
mgrunberg: Thanks! Done

javierjulio (approved):
Thanks!

PR #8255, opened 2024-02-06
Allow to use custom presenters for download links

k0va1:
Sometime we want to be able to change `download_links` configuration for different actions. For example:
```ruby
index download_links: [:json] do
...
end

show do
...
end
```
Now `download_links` config from `index` works for both index and show actions. My PR allows to specify `download_links` separately


mgrunberg:
@k0va1 thanks for the PR. Would you add some test? Maybe [here](https://github.com/activeadmin/activeadmin/blob/master/features/index/formats.feature) is a good place.

Also, would you make the PR against the default branch? We can backport the enhancement later

PR #8260, opened 2024-02-15
Run rubocop against ruby 3.3 by default

mgrunberg:
Now that we test against ruby 3.3 (#8213) it make sense to run Rubocop against Ruby 3.3 by default


javierjulio (approved):
Thanks!

PR #8275, opened 2024-03-03
Improve RuboCop exclusions configuration

tagliala:
- Sort list
- Exclude all `vendor/bundle` in `gemfiles` subfolder
- Exclude `node_modules`
- Exclude `tmp`, since it is in `.gitignore`

This will improve performance when `node_modules` has content

### `master` (third run)

```
real	0m5.448s
user	0m5.614s
sys	0m1.096s
```

### This branch (third run)

```
real	0m2.910s
user	0m3.176s
sys	0m0.920s
```


.rubocop.yml
```diff
@@ -16,9 +16,9 @@ AllCops:
   TargetRubyVersion: 3.0
 
   Exclude:
-    - tmp/development_apps/**/*
-    - tmp/test_apps/**/*
-    - gemfiles/rails_61/vendor/bundle/**/*
+    - gemfiles/**/vendor/bundle/**/*
```
javierjulio, line 19: We should also include the `.git` directory, since we override the `Exclude` config we have to redeclare it as the defaults aren't merged with our custom setting. https://github.com/rubocop/rubocop/issues/9325

```suggestion
    - .git/**/*
    - .github/**/*
    - bin/**/*
    - gemfiles/**/vendor/bundle/**/*
```

javierjulio (approved):
Thanks!

.rubocop.yml
tagliala: Thanks, good catch.

I've force pushed with some changes (just adding the defaults + customs, so `vendor/` instead of `vendor/bundle`

The performance is also improved

javierjulio: Ok great, thank you. I've had to use the same config at work previously after uncovering that issue. All the changes look good and nice to see the performance improvement.

tagliala:
Failure looks unrelated

```
error - 2024-03-04 15:52:11,018 -- Commit creating failed: {"error": "Server Error (500)"}
Error: Codecov: Failed to properly create commit: The process '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1
```

Will wait and retry later

javierjulio:
@tagliala yes, it happens from time to time. A short wait and then retrying should resolve it. Thank you.

PR #8278, opened 2024-03-13
Recommend using target="_blank" instead of target="blank" (incorrect)

ndbroadbent:
Hello,

This is just a small change to some documentation. I noticed that there were some incorrect usages of `target: :blank`, which has also found it's way into some StackOverflow and ChatGPT answers.

You can find more information about this here: https://stackoverflow.com/questions/35703005/what-is-the-difference-between-target-blank-and-target-blank

> blank targets an existing frame or window called "blank". A new window is created only if "blank" doesn't already exist.
> _blank is a reserved name which targets a new, unnamed window.

So I just thought it might be worth opening a PR to show users the correct way to do it. Thanks!

### DIFF
--- a/docs/1-general-configuration.md
+++ b/docs/1-general-configuration.md
@@ -206,7 +206,7 @@ ActiveAdmin.setup do |config|
   config.namespace :admin do |admin|
     admin.build_menu :utility_navigation do |menu|
       menu.add label: "ActiveAdmin.info", url: "https://www.activeadmin.info",
-                                          html_options: { target: :blank }
+                                          html_options: { target: "_blank" }
       admin.add_current_user_to_menu  menu
       admin.add_logout_button_to_menu menu
     end
--- a/docs/2-resource-customization.md
+++ b/docs/2-resource-customization.md
@@ -300,7 +300,7 @@ config.namespace :admin do |admin|
     menu.add label: "Sites" do |sites|
       sites.add label: "Google",
                 url: "https://google.com",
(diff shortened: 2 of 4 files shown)


docs/1-general-configuration.md
```diff
@@ -206,7 +206,7 @@ ActiveAdmin.setup do |config|
   config.namespace :admin do |admin|
     admin.build_menu :utility_navigation do |menu|
       menu.add label: "ActiveAdmin.info", url: "https://www.activeadmin.info",
-                                          html_options: { target: :blank }
+                                          html_options: { target: :_blank }
```
tagliala, line 209: @javierjulio isn't it better to have a string now that this symbol is changing?

ndbroadbent: Hi @tagliala, I've changed it to use strings instead of symbols

tagliala (approved):

docs/1-general-configuration.md
javierjulio: No, it won't matter because it will still covert that to a string. A symbol with a leading underscore is valid. I would have left the format as is but doesn't matter if its a string or not.

tagliala:
> Usage of quotes does not raise a RuboCop issue but it is not uniform with other quotes in this context

@ndbroadbent can you use double quotes instead of single quote?

ndbroadbent:
@tagliala I've rebased and updated the PR, thanks!

tagliala (approved):

PR #8288, opened 2024-03-29
Use primary color with ActiveAdmin 4 instead of blue, indigo

cprodhomme:
It's will be more simple custom the interface


javierjulio:
@cprodhomme thank you. Sorry, I will not review or consider an approach like this until Tailwind v4 which I was hoping would have been released already (originally Q1) but should be out by summer instead. I learned that Tailwind v4 is changing configuration so I'd rather wait until we know what that looks like before implementing further changes in our Tailwind config file as it would be a breaking change. Ideally, we'd have something like this, although not the first commit as the color choice there was deliberate. Thanks!

PR #8297, opened 2024-04-13
Update zh-TW locale with new v4 keys

dlackty:
As titled, thanks for all the efforts and v4 looks so good!


javierjulio (approved):
Thanks!

PR #8309, opened 2024-04-30
Fix typo in spec context

tagliala:
Found with `codespell`

The typo was there before #2601, but according to #2601 the intent is to test `/posts/1` and not `comments` anymore


javierjulio (approved):
Thanks!

PR #8310, opened 2024-04-30
Backport Suppress ruby 3.3 warning

mgrunberg:
Backport #8303 

--------

```
/myapp/vendor/bundle/gems/zeitwerk-2.6.13/lib/zeitwerk/kernel.rb:34: warning: csv was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add csv to your Gemfile or gemspec. Also contact author of activeadmin-3.2.0 to add csv into its gemspec.
```


javierjulio (approved):
Thanks!

PR #8311, opened 2024-04-30
Backport Recommend using target="_blank" instead of target="blank" (incorrect)

mgrunberg:
Backport #8278

--------

Use target="_blank" instead of target="blank" (incorrect)


javierjulio (approved):
Thanks!

PR #8318, opened 2024-05-01
Disallow sqlite3 2.x

tagliala:
Ref: rails/rails#51636

---

Not quite sure of this fix, especially on the dependabot side

### DIFF
--- a/gemfiles/rails_70/Gemfile.lock
+++ b/gemfiles/rails_70/Gemfile.lock
@@ -238,6 +238,7 @@ GEM
     matrix (0.4.2)
     method_source (1.0.0)
     mini_mime (1.1.5)
+    mini_portile2 (2.8.6)
     minitest (5.22.3)
     multi_test (1.1.0)
     net-imap (0.4.10)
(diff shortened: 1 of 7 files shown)


javierjulio:
@tagliala this would be right as I had to [lock the version to v1 for the bug report template scheduled workflow](https://github.com/activeadmin/activeadmin/commit/125688645f761f6ea4df5fa2d0868caa7318d108) for that test to pass. For the Dependabot config file change, wouldn't we want to duplicate that for the other bundler entries?

javierjulio:
Ok great, thank you!

gemfiles/rails_70/Gemfile.lock
```diff
@@ -371,6 +377,7 @@ GEM
 PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
+  ruby
```
tagliala, line 380: this is being added by my local configuration. I have a global `force_ruby_platform` setting

tagliala:
Supersedes #8315, #8316, #8320

javierjulio (approved):
Thanks!

PR #8325, opened 2024-05-05
Update de locale with new v4 keys

til:
Hi, new ActiveAdmin fan here 👋🏾 ! This PR adds the 4 missing keys for v4 to the `de` locale, and I also took the liberty to clean it up and remove the outdated `de-CH` variant, in separate commits.

PR #8326, opened 2024-05-05
Remove de-CH locale

til:
A lot of keys were missing, and it did not contain any CH specific translations.

(previously part of #8325)


javierjulio (approved):
Thanks!

PR #8329, opened 2024-05-09
Update vendored jquery-ui from 1.12.1 to 1.13.3

tagliala:
JavaScript files have been manually updated because:
- `jquery-ui-rails` is unmaintained and it is bundling 1.13.0
- An active fork of `jquery-ui-rails` is bundling 1.13.2, but there is
  a more recent version (1.13.3)
 
`escape-selector` has been removed because jQuery UI is using
jQuery's built-in escape selector. This selector is available from
version 3.0, which is required by Active Admin, so it should not be
a problem

Ref #8327

Ref:
- https://bugs.jqueryui.com/ticket/14991/
- jquery/jquery-ui#1957

---

I've manually tested

- [x] `$.ui.version` Version in console
- [x] Datepicker at `http://localhost:3000/admin/categories?order=name_asc`
- [x] Tabs switching at `http://localhost:3000/admin/kitchensink`
- [x] dropdown menu at `http://localhost:3000/admin/users`
- [x] hover menu in the navigation bar at `http://localhost:3000/admin`
- [x] sorting at `http://localhost:3000/admin/comments`
- [x] clear filters at `http://localhost:3000/admin/comments?order=id_desc`
- [x] multiple table checkboxes at `http://localhost:3000/admin/tags` 
- [x] multiple checkboxes at `http://localhost:3000/admin/users`


package.json
```diff
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "jquery": "^3.4.1",
-    "jquery-ui": "^1.12.1",
+    "jquery-ui": "^1.13.2",
```
javierjulio, line 47: The lock file shows that 1.13.3 was installed, not 1.13.2, so just for consistency.

```suggestion
    "jquery-ui": "^1.13.3",
```

javierjulio (approved):
Thanks!

package.json
tagliala: Fixed, and reworded the commit message

javierjulio (approved):
Thanks!

PR #8338, opened 2024-05-15
Add accessible names for navigation buttons

### LINKED ISSUE #8328: v4: add accessible names for buttons
### Expected behavior

No warnings

### Actual behavior

image

### How to reproduce

1. Go to https://activeadmin-demo.onrender.com/admin
2. Login

Run Lighthouse

tagliala:
- Remove reference to `drawer-navigation-label`, which is not defined
- Change `aria-controls` to `main-menu`
- Add `aria-label` to buttons without text

Close https://github.com/activeadmin/activeadmin/issues/8328

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -224,4 +224,4 @@ This release includes several locale changes. Please [reivew the en.yml locale](
 - The `comments.title_content` text has been updated with an "All " prefix.
 - The `comments.delete_confirmation` text has been fixed to use singular form.
 - Inconsistent use of login/sign-in related terms so text now uses "Sign in", Sign out", and "Sign up" throughout.
-- The `toggle_dark_mode`, `toggle_main_navigation_menu`, `toggle_section`, and `toggle_user_menu` keys have been added
+- The `toggle_dark_mode`, `toggle_main_navigation_menu`, `toggle_section`, and `toggle_user_menu` keys have been added.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -224,3 +224,4 @@ This release includes several locale changes. Please [reivew the en.yml locale](
 - The `comments.title_content` text has been updated with an "All " prefix.
 - The `comments.delete_confirmation` text has been fixed to use singular form.
 - Inconsistent use of login/sign-in related terms so text now uses "Sign in", Sign out", and "Sign up" throughout.
+- The `toggle_dark_mode`, `toggle_main_navigation_menu`, `toggle_section`, and `toggle_user_menu` keys have been added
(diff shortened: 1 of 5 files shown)


javierjulio (commented):
I'm fine with either. Is the `aria-label` the recommended approach? If so, then we should use that.

If we label these buttons, wouldn't we need to label others like the main navigation?

The upgrading guide should also be updated to include the new locale keys.

tagliala:
> I'm fine with either. Is the `aria-label` the recommended approach? If so, then we should use that.

That's a good question. I'm not sure

1. Bootstrap uses `aria-label` for its main navigation button
2. Tailwind uses `aria-labelledby` with references to a visually hidden `label` near the button
3. One of the first google results is https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/ (which is 4 year old)

> If we label these buttons, wouldn't we need to label others like the main navigation?

That's correct. Since that button is not visible in large desktop, it was not spotted by lighthouse

> The upgrading guide should also be updated to include the new locale keys.

Will do that

tagliala:
I'm still not sure on how to proceed here

javierjulio:
What characters need to be escaped? I don't see anything in the changes here that is out of the ordinary.

tagliala:
> What characters need to be escaped? I don't see anything in the changes here that is out of the ordinary.

Yes, this approach is fine, but moving to `aria-label` (which I've done in my local environment) adds interpolation in element attributes. I've force pushed

---

image

How would you call this button? As a non-native I'm having troubles. It is also not a dropdown and "toggle" does not fit: "Exand/Collapse group"?

javierjulio:
I've seen both sets of terms used. Vitepress uses `aria-label="toggle section"` for that same menu action, so we could use that there as well.

319684023:
> * Remove reference to `drawer-navigation-label`, which is not defined
> * Change `aria-controls` to `main-menu`
> * Add `aria-label` to buttons without text
> 
> Close #8328

brunoprietog:
Hi, blind developer here. Don't worry about using `aria-label` or `aria-labelledby`, both gibe the same results. Simply, `aria-labelledby` allows you to reuse text from another element, there are some cases where it's useful, but I don't think this is one of them.

UPGRADING.md
```diff
@@ -224,3 +224,4 @@ This release includes several locale changes. Please [reivew the en.yml locale](
 - The `comments.title_content` text has been updated with an "All " prefix.
 - The `comments.delete_confirmation` text has been fixed to use singular form.
 - Inconsistent use of login/sign-in related terms so text now uses "Sign in", Sign out", and "Sign up" throughout.
+- The `toggle_dark_mode`, `toggle_main_navigation_menu`, `toggle_section`, and `toggle_user_menu` keys have been added
```
javierjulio, line 227:
```suggestion
- The `toggle_dark_mode`, `toggle_main_navigation_menu`, `toggle_section`, and `toggle_user_menu` keys have been added.
```

javierjulio (approved):
Thanks!

javierjulio:
@tagliala this is ready to merge. I wasn't sure if you were holding off though for any reason? I'd like to cut a release with this change. Thanks again!

tagliala:
> this is ready to merge

No reasons, will merge, thanks

PR #8344, opened 2024-05-28
Make lockfile platforms easier to maintain

deivid-rodriguez:
This should work equally well and has less churn when maintainers upgrade their OS or switch platforms.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -489,4 +489,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   2.5.9
+   2.5.11
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -447,10 +447,8 @@ GEM
 
 PLATFORMS
   aarch64-linux
-  arm64-darwin-21
-  arm64-darwin-22
-  x86_64-darwin-22
-  x86_64-darwin-23
+  arm64-darwin
+  x86_64-darwin
   x86_64-linux
 
 DEPENDENCIES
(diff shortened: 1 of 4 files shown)


javierjulio (commented):
Do you know where the `*-21` and other platform values come from (how they get added/removed) or if Bundler has docs that cover these?

Do you want to make the same changes to the lock files in the `/gemfiles` directory too?

deivid-rodriguez:
Yeah, I should've added a bit more context!

So these platform values ultimately come from Ruby. For example,

```
$ ruby -e 'puts RUBY_PLATFORM'
arm64-darwin21
```

When Bundler is run, it uses the value of the local platform, and finds the optimal resolution for the local platform. For example, if you're on `x86_64-linux` and say, nokogiri, has a `x86_64-linux` version, it will choose that and include that in the lockfile. I will also include the local platform in the `PLATFORMS` lockfile section if it's not already there.

The specific number in this case is supposed to track the specific macOS version. However, in practice, gem authors don't release macOS version specific variants of their gems, since the platforms are the same in practice. So, there's no `arm64-darwin-21` specific version of nokogiri, just a `arm64-darwin`.

Recent versions of Bundler will "optimize" this, and as long as no variants to a specific macOS version exist, it will lock only arch + OS, but not the specific macOS version.

So, these should not get added in the future again as long as we use a recent enough version of Bundler.

I'll add the same changes to the other gemfiles, good idea!

tagliala:
> So, these should not get added in the future again as long as we use a recent enough version of Bundler.

Ah great, just tested because my dev environment was adding platform and it does not happen on this branch

javierjulio (approved):
Thanks!

deivid-rodriguez:
Thanks both!

PR #8345, opened 2024-05-28
Use `yarn` consistently everywhere

deivid-rodriguez:
Just that, I saw that we normally use just `yarn` so I think we can get rid of the binstub.


javierjulio (approved):
Thanks!

Oranuch158:
> > Just that, I saw that we normally use just `yarn` so I think we can get rid of the binstub.
> 
> Oranuch Jaihan callpay0841725297 Bookpay1570900063088

100000bant

PR #8356, opened 2024-06-04
Update Catalan translation

fjorba:
Add new messages and improve a few older ones.


config/locales/ca.yml
```diff
@@ -43,30 +74,46 @@ ca:
...
     devise:
+      username:
+        title: "Usuari"
+      email:
+        title: "Email"
+      subdomain:
+        title: "Subdomini"
+      password:
+        title: "Contrasenya"
+        password_confirmation: "Confirmació de la contrasenya"
```
javierjulio, line 111: Ah I think you meant this. It should also resolve the YAML lint warning.

```suggestion
      password_confirmation: "Confirmació de la contrasenya"
```

fjorba: Yes, you are right, sorry. And I did perform a yamllint locally, but somehow I added lines after and I didn't check again. I'm working on it.

fjorba:
Sorry again, @javierjulio, but this time it seems that the checks don't run automatically.  Now that I'm pretty confident that they should pass...

config/locales/ca.yml
```diff
@@ -14,27 +31,41 @@ ca:
...
-      empty: "No hi ha %{model}"
-      one: "S'està mostrant <b>1</b> %{model}"
-      one_page: "S'estan mostrant <b>tots %{n}</b> %{model}"
-      multiple: "S'estan mostrant %{model} <b>%{from}&nbsp;-&nbsp;%{to}</b> de <b>%{total}</b> en total"
-      multiple_without_total: "S'estan mostrant %{model} <b>%{from}&nbsp;-&nbsp;%{to}</b>"
+      empty: "No s'ha trobat cap %{model}"
+      one: "<b>1</b> %{model} de <b>1</b>"
+      one_page: "%{model}, tots %{n}</b>"
+      multiple: "%{model}, <b>%{from}-%{to}</b> d'un total de <b>%{total}</b>"
+      multiple_without_total: "%{model}, <b>%{from}-%{to}</b>"
```
javierjulio, line 65: These don't follow what we do currently for the `en` locale. Only the `empty` key uses the `model` parameter. You can see a [diff of the pagination keys in the UPGRADING guide](https://github.com/activeadmin/activeadmin/blob/master/UPGRADING.md#localization-updates).

config/locales/ca.yml
```diff
@@ -43,47 +74,69 @@ ca:
...
       unlock:
         title: "Reenvia instruccions per a desbloquejar"
         submit: "Reenvia instruccions per a desbloquejar"
       resend_confirmation_instructions:
         title: "Reenviar instruccions de confirmació"
         submit: "Reenviar instruccions de confirmació"
       links:
-        sign_in: "Registrar"
+        sign_up: "Sign up"
+        sign_in: "Sign in"
```
javierjulio, line 134: These were added but without translation. I used the `title` values from the `sign_up` and `login` keys.

```suggestion
        sign_up: "Doneu-vos d'alta"
        sign_in: "Identifiqueu-vos"
```

javierjulio (commented):
@fjorba no need to apologize, just trying to help. I appreciate you doing this. Thank you.

config/locales/ca.yml
fjorba: Oh yes, sorry again to give you so much work...

javierjulio: It's ok! No worries. Thank you. Just one line here needs to be updated. I'll submit another review.

config/locales/ca.yml
```diff
@@ -14,27 +31,41 @@ ca:
...
-      empty: "No hi ha %{model}"
-      one: "S'està mostrant <b>1</b> %{model}"
-      one_page: "S'estan mostrant <b>tots %{n}</b> %{model}"
-      multiple: "S'estan mostrant %{model} <b>%{from}&nbsp;-&nbsp;%{to}</b> de <b>%{total}</b> en total"
-      multiple_without_total: "S'estan mostrant %{model} <b>%{from}&nbsp;-&nbsp;%{to}</b>"
+      empty: "No s'ha trobat cap %{model}"
+      one: "S'està mostrant <b>1</b> de <b>1</b>"
+      one_page: "S'estan mostrant <b>tots %{n}</b>"
+      multiple: "Se n'estan mostrant <b>%{from}-%{to}</b> d'un total de <b>%{total}</b>"
+      multiple_without_total: "Se n'estan mostrant <b>%{from}-%{to}</b>"
```
javierjulio, line 65: This includes a required change to the `multiple` key since the form should be: "1-10 **of** 100" as an example. But since it overlaps with another question I had, I included another suggested change here.

I noticed the first part of each string is using a slightly different wording but it's all meant to be the same. So in the English version it's just "Showing 1..." as the prefix to keep it succinct. Mind you, I'm using Google Translate here, as while I'm fluent in Spanish, I don't know Catalan but am learning a bit from this so thank you.

Based on the translations I'm seeing in Google, I believe "S'està mostrant" would be the correct or rather intended one for all of these strings. Is there a reason though you choose not to use it?

```suggestion
      one: "S'està mostrant <b>1</b> de <b>1</b>"
      one_page: "S'està mostrant <b>tots %{n}</b>"
      multiple: "S'està mostrant <b>%{from}-%{to}</b> de <b>%{total}</b>"
      multiple_without_total: "S'està mostrant <b>%{from}-%{to}</b>"
```

fjorba: Thank you for your detailed review. The original English wording "Showing xxx" produce strange, not natural, sentences in romance languages, starting with the use or gerunde, not a good choice in any romance language. However, the prevalence of English makes it less strange to newer generations.

I'm translating ActiveAdmin because it is  embedded into Muscat, that I have also translated and done some minor patches. And I have seen the original translation in the context of a real application and it doesn't look natural. So I have reviewed all other romance languages I'm familiared with (French, Italian, Portuguese and Spanish) to look for better ideas, but also this expresssion in other Catalan packages. I've also asked a Catalan philologist, and we've come out with this form. It uses this n, that is called a feeble pronoun, in the sense that it refers to (it stresses what we are referring to) whatever we are counting. Google translator is quite good, but it is not a native Catalan translator.

Summing up: I belive that this is the most natural translation I could find.

javierjulio: Thank you for clarifying. I appreciate it. What about the second suggestion changing "d'un total de" to "de" in the `multiple` key string?

fjorba: This wording is an improvement of seeing a previous form (that I cannot recall now) that was not clear enough in a real Rails application that we are implementing (Muscat).  Now it is explicit about the total amount of elements in the list and what it is showing at this point.

javierjulio (approved):
Thanks!

PR #8362, opened 2024-06-06
include can now include modules with private included method

lomefin:
Using Rails 7, ActionView::Helpers::TagHelper has its method `included` set as `private`.

Trying to make

```
ActiveAdmin.register PaperTrail::Version do

  include ActionView::Helpers::TagHelper

end
```

Resulted in `NoMethodError: private method 'included' called for ActionView::Helpers::TagHelper:Module`

Seems that the method became private, changing the call to `send` overrides this error.

It is important to analyze whether is valid to override the Module's visibility configuration or not.

### DIFF
--- a/lib/active_admin/dsl.rb
+++ b/lib/active_admin/dsl.rb
@@ -53,7 +53,7 @@ module ActiveAdmin
     #
     # @return [Nil]
     def include(mod)
-      mod.included(self)
+      mod.send(:included, self)
     end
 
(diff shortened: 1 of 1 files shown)


lomefin:
So you say that I don't need to `include` TagHelper in ActiveAdmin?

We've had this set as the example for a long while so I thought this was the standard way.

javierjulio:
@lomefin yes, that is correct. Sorry, I don't know.

Issue #7376, opened 2022-03-07
Webpacker is deprecated


jaynetics:
webpacker has been deprecated (or [retired](https://rubyonrails.org/2022/1/21/this-week-in-rails-rails-7-0-1-webpacker-retirement-and-more-10bad602) as they call it) and is no longer included in rails 7 by default: 

> Only security issues will be considered and there will be no version 6 release of Webpacker

> For applications currently using Webpacker, the first recommendation is to switch to jsbundling-rails with Webpack.

as a consequence, i think activeadmin should no longer suggest integrating with webpacker. i'll push a PR shortly to remove the relevant section from the installation instructions.

perhaps the `use_webpacker` config option should start showing a deprecation warning as well?

deivid-rodriguez:
As far as I understand, there's https://github.com/shakacode/shakapacker which works just fine with ActiveAdmin (as per https://github.com/activeadmin/activeadmin/issues/7335) and will continue to be maintained.

So for now I think I would just clarify that even though still supported, it's no longer included with Rails by default.

deivid-rodriguez:
@jaynetics Can you change #7377 as suggested? Feel free to comment on jsbundling-rails/esbuild too, although maybe @tagliala is in a better position to write that improvement.

baarkerlounger:
JSBundling with webpack is another potential upgrade path from Webpacker. Doesn't look like that works out of the box?

tagliala:
Apologies for the silence, I was going to write a long comment and I've also tried to change instructions at least to shakapacker but then I had to stop

@baarkerlounger yes, it works. I was eventually able to remove shakapacker and make activeadmin work with jsbundling-rails (with some little changes because [I've migrated a webpacker workflow](https://github.com/rails/jsbundling-rails/blob/main/docs/switch_from_webpacker.md)).

I also use sass via webpack, but that should not be needed

### `package.json` dependencies
```json
  "dependencies": {
    "@activeadmin/activeadmin": "^2.11.0",
    "@rails/ujs": "7.0.2-3",
    "css-loader": "^6.7.1",
    "mini-css-extract-plugin": "^2.6.0",
    "sass": "^1.49.9",
    "sass-loader": "^12.6.0",
    "webpack": "^5.70.0",
    "webpack-cli": "^4.9.2",
    "webpack-remove-empty-scripts": "^0.7.3"
  },
```

### `webpack.config.js`

```js
const path = require('path')
const webpack = require('webpack')

[...]

PR #8363, opened 2024-06-07
Mention the sass-rails dependency in documentation

elthariel:
If think the dependency on sass-rails was removed because of the newer webpacker option, but in some API apps, you can end up not having the sass compiler at hands, this PR just mention it in the setup documentation

### DIFF
--- a/docs/0-installation.md
+++ b/docs/0-installation.md
@@ -9,6 +9,9 @@ Active Admin is a Ruby Gem.
 ```ruby
 gem 'activeadmin'
 
+# If you're not using webpacker, ensure you have:
+# gem 'sass-rails'
+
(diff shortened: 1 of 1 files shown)


tagliala:
Hello, webpacker support has been dropped in v4

https://github.com/activeadmin/activeadmin/issues/7376#issuecomment-1826867882

Is it sure that this line is required?

I'm not using it without webpacker. this is what I have:

```
gem 'cssbundling-rails'
gem 'importmap-rails'
```

javierjulio:
@elthariel thank you. Yes, that's right. As @tagliala mentioned, we don't use SASS anymore on the default branch as it was replaced entirely with Tailwind.

PR #8369, opened 2024-06-09
Fix strftime month directive

### LINKED ISSUE #8368: Timestamp format bug in rails template does not allow to run specs against 7.2.0.beta2
```
ActiveRecord::InvalidMigrationTimestampError:  (ActiveRecord::InvalidMigrationTimestampError)

Invalid timestamp 20240709180755 for migration file: create_posts.
Timestamp must be in form YYYYMMDDHHMMSS, and less than 20240610160808.
```

https://github.com/activeadmin/activeadmin/blob/feee35c4340a36d055a516b9174d606763aacd7c/spec/support/rails_template.rb#L8

`%M` is minute of the hour

https://ruby-doc.org/stdlib-3.1.0/libdoc/date/rdoc/DateTime.html#method-i-strftime

tagliala:
`%M` is minute of the hour, and Rails 7.2 is guarding against invalid timestamps before executing migrations by default

This commit uses the correct zero-padded month directive `%m`

Fix #8368

Ref: https://ruby-doc.org/stdlib-3.1.0/libdoc/date/rdoc/DateTime.html#method-i-strftime

### DIFF
--- a/spec/support/rails_template.rb
+++ b/spec/support/rails_template.rb
@@ -5,7 +5,7 @@ gem "cssbundling-rails"
 rails_command "css:install:tailwind"
 rails_command "importmap:install"
 
-initial_timestamp = Time.now.strftime("%Y%M%d%H%M%S").to_i
+initial_timestamp = Time.now.strftime("%Y%m%d%H%M%S").to_i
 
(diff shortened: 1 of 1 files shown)


javierjulio (approved):
Thanks!

PR #8370, opened 2024-06-09
Update dependencies

tagliala:
Also:
- Fix new RuboCop offenses
- Use the latest rubygems to allow specs against Ruby 3.0


.github/workflows/ci.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          rubygems: latest
```
tagliala, line 49: I can use a conditional here to only use latest on 3.0

javierjulio: I see the previous CI failure. I want to keep Ruby 3.0 support so this is good. I didn't know about this flag for the ruby/setup-ruby action. Thank you.

javierjulio (approved):
Thanks!

PR #8372, opened 2024-06-09
Use database_cleaner-active_record directly instead

tagliala:
Use `database_cleaner-active_record` directly, as suggested at https://github.com/DatabaseCleaner/database_cleaner


javierjulio (approved):
Thanks!

PR #8375, opened 2024-06-15
Update dependencies and improve CI

tagliala:
- Use strings for Ruby version to uniform with Arbre and Inherited
  Resources and prevent confusion with "3.0"
- Use "rails" instead of "deps" for clarity


javierjulio (approved):
Thanks!

PR #8389, opened 2024-07-01
Allow batch actions without a confirmation dialog

DarkSwoop:
We have batch actions that don't need to be confirmed, and they should work too. Currently, the click event simply fills the hidden field and doesn't submit the form itself.

This change checks if the batch action has a confirm data-attribute, and if not, it simply submits the form.

### DIFF
--- a/app/javascript/active_admin/features/batch_actions.js
+++ b/app/javascript/active_admin/features/batch_actions.js
@@ -1,11 +1,20 @@
 import Rails from '@rails/ujs';
 
+const submitForm = function() {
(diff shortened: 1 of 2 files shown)


javierjulio (commented):
Sounds like a worthwhile change, thanks. Can you please add a feature test for this? Should be able to duplicate the existing JS scenario with some modifications.

DarkSwoop:
I added a test, and it works locally.

javierjulio (approved):
Thanks!

PR #8390, opened 2024-07-01
Add `use_static_assets` feature

jaynetics:
Related to #8246.

This increases the gem size from 173 to 203 KB (as a side note, it might be worth investigating why it is already so big).

Set `use_static_assets` to true to test. I tested it with sprockets, propshaft, and the default pipeline. Of course this would need some automated tests as well. It could also use a mention in the documentation (UPGRADING.md? web docs?), and possibly a generator flag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,6 @@
 /vendor/bundle
 /rails_70
 /dist
+/lib/active_admin/static_assets/assets
 docs/.vitepress/cache
 docs/.vitepress/dist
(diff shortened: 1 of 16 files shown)


jaynetics:
@javierjulio any thoughts on whether it makes sense to hammer this out?

jaynetics:
I've now released a gem to do this, [activeadmin_assets](https://github.com/jaynetics/activeadmin_assets/). Feel free to mention it in the docs as a possible approach (with caveats).

PR #8394, opened 2024-07-09
Support async_count for scopes

jdlubrano:
# Background

In ActiveAdmin applications, a surprising performance bottleneck can be the `COUNT` queries that run to provide counts for each scope made available on an resource's index page. In many cases, performance issues can be addressed by adding missing indices and/or otherwise speeding up the underlying query.

There is another performance boost that could be low-hanging fruit, however. The `COUNT` queries for scopes currently run sequentially.  In Rails 7.1, the framework provides the `async_count` method that allows queries to be run in parallel. For index pages with many scopes (and many COUNTs) running queries in parallel could offer a significant time savings.

This PR will use `async_count` to query the collection size of scopes in versions of Rails where `async_count` is available. Rails apps not running on Rails 7.1 will continue to rely on sequential queries.

# Approach

This PR introduces an `ActiveAdmin::AsyncCount` class that quacks enough like an ActiveRecord Relation to play nicely with the `collection_size` helper.  The `ActiveAdmin::Views::Scopes` class eagerly initializes `AsyncCount`s for each scope that is configured like
```ruby
scope(:my_scope, show_count: :async)
```
which kicks off async queries.  As a result, rendering the scopes component should only be as slow as its slowest `COUNT`.

## Closed Questions

* Should the async query behavior be more opt-in? ✅ (it was made more opt-in in the latest commits)
  * ~Should individual scopes have an `async_count: true` option?~ (I choose `show_count: :async` to leverage the existing `show_count` attribute)
  * ~Should the index options support a `scope_count: :async` option?~ (It seems easy enough to configure many scopes to use `async_count` using the `with_options` Rails/ActiveSupport helper.  Example:
```ruby
with_options(show_count: :async) do
  scope(:one)
  scope(:two)
end
```
  * ~Should consumers be able to configure their own `ScopeCounter` class globally and/or per index?~ Not necessary

# Credits

@tuffylock @aturar

### DIFF
--- a/lib/active_admin/async_count.rb
+++ b/lib/active_admin/async_count.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module ActiveAdmin
+  class AsyncCount
+    def initialize(collection)
+      @collection = collection.except(:select, :order)
+      @promise = @collection.async_count
+    end
+
+    def count
+      value = @promise.value
+      # value.value due to Rails bug https://github.com/rails/rails/issues/50776
+      value.respond_to?(:value) ? value.value : value
+    end
+
+    delegate :except, :group_values, :length, :limit_value, to: :@collection
+  end
+end
--- a/lib/active_admin/async_count.rb
+++ b/lib/active_admin/async_count.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-
 module ActiveAdmin
   class AsyncCount
     def initialize(collection)
--- a/lib/active_admin/async_count.rb
+++ b/lib/active_admin/async_count.rb
@@ -12,6 +12,8 @@ module ActiveAdmin
       value.respond_to?(:value) ? value.value : value
     end
 
+    alias size count
+
     delegate :except, :group_values, :length, :limit_value, to: :@collection
   end
 end
--- a/lib/active_admin/async_count.rb
+++ b/lib/active_admin/async_count.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 module ActiveAdmin
   class AsyncCount
+    class NotSupportedError < RuntimeError; end
+
     def initialize(collection)
+      raise NotSupportedError, "#{collection.inspect} does not support :async_count" unless collection.respond_to?(:async_count)
+
       @collection = collection.except(:select, :order)
       @promise = @collection.async_count
     end
--- a/lib/active_admin/views/components/scopes.rb
+++ b/lib/active_admin/views/components/scopes.rb
@@ -58,9 +58,9 @@ module ActiveAdmin
 
       # Return the count for the scope passed in.
       def get_scope_count(scope)
-        chained = scope_chain(scope, collection_before_scope)
+        chained = @async_counts[scope] || scope_chain(scope, collection_before_scope)
 
-        collection_size(@async_counts[scope] || chained)
+        collection_size(chained)
       end
 
       def group_name(group)
--- a/lib/active_admin/views/components/scopes.rb
+++ b/lib/active_admin/views/components/scopes.rb
@@ -73,14 +73,10 @@ module ActiveAdmin
         call_method_or_exec_proc(scope.display_if_block)
       end
 
-      def async_counts?
-        collection_before_scope.respond_to?(:async_count)
-      end
-
       def prepare_async_counts(scopes, options)
-        @async_counts = if options[:scope_count] && async_counts?
+        @async_counts = if options[:scope_count]
                           scopes
-                            .select(&:show_count)
+                            .select(&:async_count?)
                             .select { |scope| display_scope?(scope) }
                             .index_with { |scope| AsyncCount.new(scope_chain(scope, collection_before_scope)) }
                         else
--- a/lib/active_admin/views/components/scopes.rb
+++ b/lib/active_admin/views/components/scopes.rb
@@ -21,7 +21,7 @@ module ActiveAdmin
         scopes.group_by(&:group).each do |group, group_scopes|
           div class: "index-button-group", role: "group", data: { "group": group_name(group) } do
             group_scopes
-              .select(&method(:display_scope?))
+              .select { |scope| display_scope?(scope) }
               .each { |scope| build_scope(scope, options) }
 
             nil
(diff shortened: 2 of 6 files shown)


lib/active_admin/async_count.rb
```diff
@@ -0,0 +1,17 @@
...
+  class AsyncCount
+    def initialize(collection)
+      @collection = collection.except(:select, :order)
+      @promise = @collection.async_count
+    end
+
+    def count
+      value = @promise.value
+      # value.value due to Rails bug https://github.com/rails/rails/issues/50776
+      value.respond_to?(:value) ? value.value : value
```
jdlubrano, line 12: I'm not entirely sure what a "good" solution is to this.  Rails 7.1.3 has a bug with its query cache such that promises get wrapped by an "extra" promise.  It doesn't seem like ActiveAdmin's responsibility to work around that, but I'm not sure what alternatives there might be.  In any case, this is isolated to this class and is tested, so that seemed like a good starting point.

jdlubrano:
Hi @javierjulio.  I was wondering if there was any interest in this PR from the ActiveAdmin maintainers.  If not, I have no issue closing it.  Sorry about the direct ping, I was looking for a maintainer to nudge.

javierjulio:
@jdlubrano thank you. Just have to be patient for us to find time.

I agree this can be a helpful feature. I haven't tried the new async querying capability in Rails yet so I'll have to rely on others for input. Looking at the scopes DSL in our docs, it seems best that this would be an option per scope and that it should be an opt-in feature.

I'm not sure about your last question though. What need do you see for having the ability to override the class?

jdlubrano:
Thanks so much for the feedback!

I will work on tweaking this to rely on a per-scope, opt-in action. Given the per-scope option, I don't see a need to leave the class open to overrides.

jdlubrano:
I have updated scopes to accept an option like
```ruby
scope(:my_scope, show_count: :async)
```
as a way to opt-in to async queries.  If someone uses `show_count: :async` with a version of Rails/ActiveRecord that does not implement `async_count`, a `NoMethodError` will be raised.  Since this behavior is opt-in, that seems okay to me.

The PR description has also been updated accordingly.

lib/active_admin/views/components/scopes.rb
```diff
@@ -15,11 +16,13 @@ def tag_name
...
+        prepare_async_counts(scopes, options)
+
         scopes.group_by(&:group).each do |group, group_scopes|
           div class: "index-button-group", role: "group", data: { "group": group_name(group) } do
-            group_scopes.each do |scope|
-              build_scope(scope, options) if call_method_or_exec_proc(scope.display_if_block)
-            end
+            group_scopes
+              .select { |scope| display_scope?(scope) }
+              .each { |scope| build_scope(scope, options) }
```
mgrunberg, line 25: I would keep a single iteration. Performance is not the main argument because I belive scopes is a small collection. But by introducing a select before the each iteration it's not so evident that the main change the introduction of the new helper `display_scope?`

```suggestion
            group_scopes.each do |scope|
              build_scope(scope, options) if display_scope?(scope)
            end
```

spec/unit/views/components/scopes_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    context "when scopes are configured to query their counts asynchronously" do
+      let(:configured_scopes) do
+        [
+          ActiveAdmin::Scope.new(:all, nil, show_count: :async),
+          ActiveAdmin::Scope.new(:published, nil, show_count: :async) { |posts| posts.where.not(published_date: nil) }
+        ]
+      end
+
+      it "raises an error when ActiveRecord async_count is unavailable", unless: Post.respond_to?(:async_count) do
+        expect { scopes }.to raise_error(NoMethodError, %r{async_count})
```
mgrunberg, line 72: Would be easy to display a nicer error message in this scenario? 
Something like "scope `show_count: :async` is not available in Rails X.Y.Z. This Rails version doesn't support async_count". 
I know this feature is `opt-in` but I'm wondering if we can explain in a better way the cause of the problem. I want to avoid future issues asking why they get NoMethodError.
Of course, if it's easy.

lib/active_admin/async_count.rb
mgrunberg: Implementing the workaround here makes sense. We don't know the Rails version the host app is running.

mgrunberg (commented):
@jdlubrano thanks for your work. I like the PR feature.
I wrote a few comments. I want to know your thoughts about them.

I hope I have more time tomorrow to keep reviewing the PR.

spec/unit/views/components/scopes_spec.rb
jdlubrano: Done in [f4205c3](https://github.com/activeadmin/activeadmin/pull/8394/commits/f4205c30a48a531a627b639bb34d12f8c4870fb1)

lib/active_admin/views/components/scopes.rb
jdlubrano: Done in [7dd372e](https://github.com/activeadmin/activeadmin/pull/8394/commits/7dd372e8beffdf498d181375459dc7689a67e093)

javierjulio (commented):
@mgrunberg I think this is ready and looks good to me. Would you be able to test it out?

mgrunberg (commented):
@javierjulio looks good to me.

I tested it on my machine and it works. I have a simple suggestion before merging.

lib/active_admin/views/components/scopes.rb
```diff
@@ -55,12 +58,31 @@ def current_scope?(scope)
 
       # Return the count for the scope passed in.
       def get_scope_count(scope)
-        collection_size(scope_chain(scope, collection_before_scope))
+        chained = scope_chain(scope, collection_before_scope)
+
+        collection_size(@async_counts[scope] || chained)
```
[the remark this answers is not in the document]
jdlubrano: Done in 8b044d9da1390c17979c40070f1fd4ef440eb2e7

spec/unit/async_count_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      async_count = described_class.new(Post.group(:author_id))
+      expect(async_count.count).to eq(100 => 1, 200 => 1)
+    end
+
+    # See https://github.com/rails/rails/issues/50776
+    it "works around a Rails 7.1.3 bug with wrapped promises" do
+      promise = instance_double(ActiveRecord::Promise, value: Post.count)
+      promise_wrapper = instance_double(ActiveRecord::Promise, value: promise)
+      collection = class_double(Post, async_count: promise_wrapper)
+      allow(collection).to receive(:except).and_return(collection)
```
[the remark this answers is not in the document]
jdlubrano: Done in be9e118a57f1d1b9886b7e3ac9481e8d4bb7d230

mgrunberg (approved):

javierjulio (approved):
Thanks!

PR #8397, opened 2024-07-12
Update dependencies

tagliala:
Needed to test against 7.2

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -92,8 +92,8 @@ GEM
       minitest (>= 5.1)
       mutex_m
       tzinfo (~> 2.0)
-    addressable (2.8.6)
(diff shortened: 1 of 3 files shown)


javierjulio (approved):
@tagliala thanks! I'll merge this now. I wasn't sure if you knew when doing these updates so just to share, we have a `bin/bundle` script that will apply to all gemfiles automatically. So you can run `bin/bundle update` (or any other Bundler command) and it would do what you would expect. Dependabot will only submit these updates monthly (first of the month) but since they are grouped now, if you wanted that to happen sooner, say weekly, that would be ok to change if you prefer. Just FYI.

tagliala:
> So you can run bin/bundle update

Didn't know, thanks

PR #8405, opened 2024-07-25
Remove deprecation warnings when building with dart sass >= 1.77.7

### LINKED ISSUE #8404: Several deprecation warnings when building with dart sass >= 1.77.7
I am using ActiveAdmin 3.2.2 with dart sass 1.77.8.

### Expected behavior

Compiling assets should complete without warnings.

### Actual behavior

The latest version of dart sass emits several warnings about ActiveAdmin's scss files. Here is one example:

DEPRECATION WARNING: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls

```
   ┌──> ../../shared/node_modules/@activeadmin/activeadmin/src/scss/mixins/_gradients.scss
5  │     background-color: $start;
   │     ^^^^^^^^^^^^^^^^^^^^^^^^ declaration
   ╵
   ┌──> ../../shared/node_modules/@activeadmin/activeadmin/src/scss/mixins/_buttons.scss
12 │ ┌   &.disabled {
13 │ │     opacity: 0.5;
14 │ │     cursor: default;
15 │ │   }
   │ └─── nested rule
   ╵
    ../../shared/node_modules/@activeadmin/activeadmin/src/scss/mixins/_gradients.scss 5:3  gradient()
    ../../shared/node_modules/@activeadmin/activeadmin/src/scss/mixins/_buttons.scss 21:3   default-button()
    ../../shared/node_modules/@activeadmin/activeadmin/src/scss/mixins/_buttons.scss 64:3   dark-button()
    ../../shared/node_modules/@activeadmin/activeadmin/src/scss/_forms.scss 223:5           @import
    ../../shared/node_modules/@activeadmin/activeadmin/src/scss/_base.scss 9:11             @import
    app/frontend/entrypoints/active_admin.scss 20:9                                         root stylesheet
```

### How to reproduce

Generate a blank Rails 7.1 app:

```
rails new sass-repro
```

Install dart sass:

```
./bin/bundle add dartsass-rails
./bin/rails dartsass:install
```

Install ActiveAdmin:

```
./bin/bundle add activeadmin
./bin/rails g active_admin:install --skip-users
```

For a simple reproduction, replace the Rails default `application.scss` with ActiveAdmin's:

```
cp app/assets/stylesheets/active_admin.scss app/assets/stylesheets/application.scss
```

Compile assets:

```
./bin/rails assets:precompile
```

Produces:

```
Deprecation Warning: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls
...
```

mgrunberg:
Fixes #8404

The use of declarations after nested rules is currently deprecated in Sass.

We have a bunch of mixins that have nested rules in their definition: basic-button, dark-button, light-button, clearfix.

To be compatible with dartsass, the solution is to move the inclusion of those mixins to the bottom.

In some scenarios, like dropdown_menu_button (dropdown_menu.scss) the intention is to override default button styles. To keep that behaviour I'm using `& {}` which is the recommended way in dartsass to put rules below a nesting.

Read more about the deprecation here https://sass-lang.com/documentation/breaking-changes/mixed-decls/

Sadly libsass does not implement the deprecation. The only way to test this commit is manually (I don't see the point on building a feature test just to check building against a different sass engine. Specially since v4 doesn't use sass).

Steps to test:
````
bin/rake local bundle add dartsass-rails
bin/rake local rails dartsass:install
cp tmp/development_apps/rails_71/app/assets/stylesheets/active_admin.scss tmp/development_apps/rails_71/app/assets/stylesheets/application.scss
bin/rake local rails dartsass:build
````

I tested this won't break anything by compiling CSS using `sassc-rails` and checking rules order. I also did a manual review of development app

NOTE: I tested against Rails 7.1

### DIFF
--- a/app/assets/stylesheets/active_admin/components/_table_tools.scss
+++ b/app/assets/stylesheets/active_admin/components/_table_tools.scss
@@ -1,6 +1,6 @@
 .table_tools {
-  @include clearfix;
   margin-bottom: 16px;
+  @include clearfix;
 }
 
 .table_tools .dropdown_menu {
@@ -9,11 +9,14 @@
 
 a.table_tools_button, .table_tools .dropdown_menu_button {
   @include light-button;
-  @include gradient(#FFFFFF, #F0F0F0);
-  @include border-colors(#d9d9d9, #d0d0d0, #c5c5c5);
-  font-size: 0.9em;
-  padding: 4px 14px 4px;
-  margin: 0;
+
+  & {
+    @include gradient(#FFFFFF, #F0F0F0);
+    @include border-colors(#d9d9d9, #d0d0d0, #c5c5c5);
+    font-size: 0.9em;
+    padding: 4px 14px 4px;
+    margin: 0;
+  }
 
   &:not(.disabled) {
     &:hover {
--- a/app/assets/stylesheets/active_admin/_forms.scss
+++ b/app/assets/stylesheets/active_admin/_forms.scss
@@ -220,9 +220,8 @@ form {
   /* Buttons */
 
   input[type=submit], input[type=button], button {
-    @include dark-button;
     cursor: pointer;
-
+    @include dark-button;
   }
 
   .buttons, .actions {
--- a/app/assets/stylesheets/active_admin/components/_comments.scss
+++ b/app/assets/stylesheets/active_admin/components/_comments.scss
@@ -2,10 +2,10 @@
 .comments {
 
   .active_admin_comment {
-    @include clearfix;
     margin-top: 10px;
     margin-bottom: 20px;
     max-width: 700px;
+    @include clearfix;
 
     .active_admin_comment_meta {
       width: 130px;
--- a/app/assets/stylesheets/active_admin/components/_dropdown_menu.scss
+++ b/app/assets/stylesheets/active_admin/components/_dropdown_menu.scss
@@ -3,9 +3,12 @@
 
   .dropdown_menu_button {
     @include light-button;
-    position: relative;
-    padding-right: 22px !important;
-    cursor: pointer;
+
+    & {
+      position: relative;
+      padding-right: 22px !important;
+      cursor: pointer;
+    }
 
     &:before {
       content: ' ';
@@ -139,11 +142,11 @@
         }
 
         &:last-child {
+          border: none;
           a {
             border-bottom-left-radius: 2px;
             border-bottom-right-radius: 2px;
           }
-          border: none;
         }
       }
     }
--- a/app/assets/stylesheets/active_admin/components/_pagination.scss
+++ b/app/assets/stylesheets/active_admin/components/_pagination.scss
@@ -38,8 +38,11 @@
   margin-left: 4px;
   select {
     @include light-button;
-    @include rounded(0px);
-    padding: 1px 5px;
+
+    & {
+      @include rounded(0px);
+      padding: 1px 5px;
+    }
   }
 }
 
--- a/app/assets/stylesheets/active_admin/components/_tabs.scss
+++ b/app/assets/stylesheets/active_admin/components/_tabs.scss
@@ -28,13 +28,16 @@
 
     a {
       @include light-button;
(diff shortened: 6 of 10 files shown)

### CODE REVIEW

**Review by javierjulio — APPROVED** [review]
@mgrunberg were you able to test this manually with the local generated dev app defaults (sass-rails) as well? My priority is to ensure these style changes don't break there which I don't expect it to. If so, then you are all set to merge and release. Thanks!

**Comment by mgrunberg** [comment]
> @mgrunberg were you able to test this manually with the local generated dev app defaults (sass-rails) as well? My priority is to ensure these style changes don't break there which I don't expect it to. If so, then you are all set to merge and release. Thanks!

@javierjulio perfect. I still have to do that. I'll merge and release once I check this (most likely during Monday).

**Comment by javierjulio** [comment]
@mgrunberg sounds great, thank you!

**Comment by tagliala** [comment]
I'm doing this for a third-party theme

Instead of manually checking the differences, I'm going for css diff before and after: https://github.com/cprodhomme/arctic_admin/issues/123#issuecomment-2254651558

**Comment by tagliala** [comment]
@mgrunberg great, thanks

For people reading here, the fix is not straightforward

If you have "idempotent" / "standalone" / "no side-effects" mixins like `clearfix`, the order doesn't matter

Other mixins like `*-button` may include properties like padding, so their inclusion order should be considered and can't be moved around safely

I've also had to use `& {` for a use case in the third-party theme

**Comment by mgrunberg** [comment]
@tagliala thanks again for the feedback

@javierjulio do you want to review again or should I merge the PR and prepare a release?

File: app/assets/stylesheets/active_admin/components/_table_tools.scss [hunk 1]
```diff
@@ -1,19 +1,22 @@
...
 }
 
 a.table_tools_button, .table_tools .dropdown_menu_button {
-  @include light-button;
   @include gradient(#FFFFFF, #F0F0F0);
   @include border-colors(#d9d9d9, #d0d0d0, #c5c5c5);
-  font-size: 0.9em;
-  padding: 4px 14px 4px;
-  margin: 0;
+  @include light-button;
```
javierjulio, line 13: Wouldn't this change be different visually too since this does use the gradient/border-colors mixing locally to override the values from the button mixin? The same would apply in some other areas since they have the same type of change.

mgrunberg: I see your point. Your logic is correct but I didn't notice changes in my manual check. 
I'll double-check this and come back.

mattbrictson:
@mgrunberg I appreciate all the work you are doing on this. Let me know if I can do anything to help!

app/assets/stylesheets/active_admin/components/_table_tools.scss
mgrunberg: @javierjulio You were right, I was looking at the wrong component. Sorry! this should be the last change.

javierjulio:
@mattbrictson yes definitely. Please test this branch to confirm the CSS updates that required critical order changes ([one example](https://github.com/activeadmin/activeadmin/pull/8405/files#diff-9e37af5677dc2ceb4b84d608610bcc8ba9a2fafb05dfa663bc5df5c3d0eff93f)) in a web browser to ensure everything looks as it was before and of course that any deprecation warnings are now removed too. Thank you!

mattbrictson:
@javierjulio I just tested this branch (as of 6c2ce5ac) and it looks good to me! I see no warnings when I compile assets, and I compared the ActiveAdmin pages against my production environment and did not see any visual regressions. ✅

mgrunberg:
@mattbrictson thanks for the help!

PR #8411, opened 2024-07-30
Better implementation of counter-cache-column check

mgrunberg:
Reimplementation of https://github.com/activeadmin/activeadmin/pull/7347.
Fixes https://github.com/activeadmin/activeadmin/issues/7346.

Thanks @plashchynski for the work!

Original Issue:
```
I have a model with some column ends with a "_count" substring.
This column doesn't appear on a show page.
It should appear as It is a regular column just like any other.
```

### DIFF
--- a/spec/unit/resource/attributes_spec.rb
+++ b/spec/unit/resource/attributes_spec.rb
@@ -32,6 +32,22 @@ module ActiveAdmin
         expect(subject).to_not include :published_date
         ActiveAdmin.application.filter_attributes = keep
       end
+
+      context "when resource has a counter cache" do
+        subject { ActiveAdmin::Resource.new(namespace, Category).resource_attributes }
+
+        it "should not include counter cache column" do
+          expect(subject.keys).not_to include(:posts_count)
+        end
+      end
+
+      context "when resource has a 'counter cache'-like column" do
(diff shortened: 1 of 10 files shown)


spec/unit/resource/attributes_spec.rb
```diff
@@ -32,6 +32,23 @@ module ActiveAdmin
...
+
+          expect(attrs.keys).not_to include(:posts_count)
+        end
+
+        it "should include counter cache like columns" do
+          attrs = ActiveAdmin::Resource.new(namespace, User).resource_attributes
+
+          expect(attrs).to include(sign_in_count: :sign_in_count)
+        end
+      end
```
javierjulio, line 51: This is nitpicky so I'll leave it up to you how best to address. Since it's already using a subject, it could help to have a new context since one of the tests is for a `"column with 'counter-cache'-like name"` context vs a `"counter cache column"` context.

javierjulio (approved):
Thanks!

spec/unit/resource/attributes_spec.rb
mgrunberg: Thanks! Good catch. At the beginning I tried to use just User model but it got hard. I forgot to remove the subject but now I think I'm gonna use it as you suggested (two contexts)

PR #8423, opened 2024-08-04
Add tbody_html and row_html options to TableFor and IndexAsTable

thibaudgg:
This patch adds a `tbody_html` and `row_html` options to the `TableFor` (and
`IndexAsTable`) classes. These options allow you to pass a hash of HTML options
attributes to the `tbody` and `tr` elements.

This is useful when you want to add features to the table that require data
attributes. For example, adding a sortable row via JavaScript.

You can see how I'm currently using this approach in [this app](https://github.com/acp-admin/acp-admin):
- [monkey patching](https://github.com/acp-admin/acp-admin/blob/master/config/initializers/active_admin.rb#L261-L328) (same approach as this PR)
- [using both options on a `table_for` method](https://github.com/acp-admin/acp-admin/blob/master/app/views/active_admin/deliveries/_baskets.html.arb#L10-L13) for using [Stimulus Sortable component](https://www.stimulus-components.com/docs/stimulus-sortable/).

The existing `row_class` option is kept for backwards compatibility, but the
`row_html` will take precedence if both are provided.

### DIFF
--- a/docs/3-index-pages/index-as-table.md
+++ b/docs/3-index-pages/index-as-table.md
@@ -196,13 +196,22 @@ index do
 end
 ```
 
-## Custom row class
+## Custom tbody HTML attributes
 
-In order to add special class to table rows pass the proc object as a `:row_class` option
-of the `index` method.
+In order to add HTML attributes to the tbody use the `:tbody_html` option.
 
 ```ruby
-index row_class: ->elem { 'active' if elem.active? } do
+index tbody_html: { class: "my-class", data: { controller: 'stimulus-controller' } } do
+  # columns
+end
+```
+
+## Custom row HTML attributes
+
+In order to add HTML attributes to table rows, use a proc object in the `:row_html` option.
+
+```ruby
+index row_html: ->elem { { class: ('active' if elem.active?), data: { 'element-id' => elem.id } } } do
   # columns
 end
 ```
--- a/lib/active_admin/views/components/table_for.rb
+++ b/lib/active_admin/views/components/table_for.rb
@@ -16,6 +16,9 @@ module ActiveAdmin
         @resource_class ||= @collection.klass if @collection.respond_to? :klass
 
         @columns = []
+        @tbody_html = options.delete(:tbody_html)
+        @row_html = options.delete(:row_html)
+        # To be deprecated, please use row_html instead.
         @row_class = options.delete(:row_class)
 
         build_table
@@ -91,10 +94,12 @@ module ActiveAdmin
       end
 
       def build_table_body
-        @tbody = tbody do
+        @tbody = tbody **(@tbody_html || {}) do
           # Build enough rows for our collection
           @collection.each do |elem|
-            tr(id: dom_id_for(elem), class: @row_class&.call(elem))
+            html_options = @row_html&.call(elem) || {}
+            html_options.reverse_merge!(class: @row_class&.call(elem))
+            tr(id: dom_id_for(elem), **html_options)
           end
         end
       end
--- a/lib/active_admin/views/index_as_table.rb
+++ b/lib/active_admin/views/index_as_table.rb
@@ -196,17 +196,25 @@ module ActiveAdmin
     # end
     # ```
     #
-    # ## Custom row class
+    # ## Custom tbody HTML attributes
     #
-    # In order to add special class to table rows pass the proc object as a `:row_class` option
-    # of the `index` method.
+    # In order to add HTML attributes to the tbody use the `:tbody_html` option.
     #
     # ```ruby
-    # index row_class: ->elem { 'active' if elem.active? } do
+    # index tbody_html: { class: "my-class", data: { controller: 'stimulus-controller' } } do
     #   # columns
     # end
     # ```
     #
+    # ## Custom row HTML attributes
+    #
+    # In order to add HTML attributes to table rows, use a proc object in the `:row_html` option.
+    #
+    # ```ruby
+    # index row_html: ->elem { { class: ('active' if elem.active?), data: { 'element-id' => elem.id } } } do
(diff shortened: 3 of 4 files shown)


javierjulio (commented):
@thibaudgg thanks for finding the time to submit this. I'm open to suggestions on the new argument name but this has to be generic, not a "data" attributes specific key.

lib/active_admin/views/components/table_for.rb
```diff
@@ -91,10 +93,10 @@ def build_table_header(col)
       end
 
       def build_table_body
-        @tbody = tbody do
+        @tbody = tbody data: @tbody_data do
           # Build enough rows for our collection
           @collection.each do |elem|
-            tr(id: dom_id_for(elem), class: @row_class&.call(elem))
+            tr(id: dom_id_for(elem), class: @row_class&.call(elem), data: @row_data&.call(elem))
```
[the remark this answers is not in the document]
thibaudgg: Good point, I switched to the `tbody_html` options, as I found more used of `*_html` than `*_html_options` in the codebase. 

For the `row_data` it's a bit more complex as we need the proc on the row element. We could add a new `row_html` method with an optional proc arguments, but we would also need to still support the `row_class` for backward compatibility. Which I guess would overwrite a `class` set by `row_html` if both options are used? What would you prefer?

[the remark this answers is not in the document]
javierjulio: If you prefer to keep row_class support, I think it would be fine if it's present to reverse merge that into the new row_html value so the class in row_html would take precedence over row_class. Thank you.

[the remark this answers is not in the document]
thibaudgg: Added `row_html`, but kept `row_class` for backward compatibility. I think we should be good now.

lib/active_admin/views/index_as_table.rb
```diff
@@ -196,17 +196,26 @@ module Views
     # end
     # ```
     #
-    # ## Custom row class
+    # ## Custom tbody HTML attributes
     #
-    # In order to add special class to table rows pass the proc object as a `:row_class` option
-    # of the `index` method.
+    # In order to add special html attributes to table tbody pass a `:tbody_html` option of the `index` method.
```
javierjulio, line 201:
```suggestion
    # In order to add HTML attributes to the tbody use the `:tbody_html` option.
```

lib/active_admin/views/index_as_table.rb
```diff
@@ -196,17 +196,26 @@ module Views
...
-    # index row_class: ->elem { 'active' if elem.active? } do
+    # index tbody_html: { class: "my-class", data: { controller: 'stimulus-controller' } } do
     #   # columns
     # end
     # ```
     #
+    # ## Custom row HTML attributes
+    #
+    # In order to add special html attributes to table rows pass the proc object as a `:row_html` option
+    # of the `index` method.
```
javierjulio, line 212:
```suggestion
    # In order to add HTML attributes to table rows, use a proc object in the `:row_html` option.
```

spec/unit/views/components/table_for_spec.rb
```diff
@@ -294,7 +294,24 @@
       end
     end
 
-    context "when row_class" do
+    context "when tbody_html" do
```
javierjulio, line 297:
```suggestion
    context "with tbody_html option" do
```

spec/unit/views/components/table_for_spec.rb
```diff
@@ -313,6 +330,34 @@
       end
     end
 
+    context "when row_html (take presence over row_class)" do
```
javierjulio, line 333:
```suggestion
    context "with row_html options (takes precedence over deprecated row_class)" do
```

spec/unit/views/components/table_for_spec.rb
```diff
@@ -294,7 +294,24 @@
...
+
+      it "should render data-size attribute within tbody tag" do
+        tbody = table.find_by_tag("tbody").first
+        expect(tbody.attributes).to include(
+          class: "my-class",
+          data: { size: 3 })
+      end
+    end
+
+    context "when row_class (to be deprecated)" do
```
javierjulio, line 314:
```suggestion
    context "with row_class (soft deprecated)" do
```

docs/3-index-pages/index-as-table.md
```diff
@@ -196,13 +196,23 @@ index do
 end
 ```
 
-## Custom row class
+## Custom tbody HTML attributes
 
-In order to add special class to table rows pass the proc object as a `:row_class` option
+In order to add special html attributes to table tbody pass a `:tbody_html` option of the `index` method.
```
javierjulio, line 201:
```suggestion
In order to add HTML attributes to the tbody use the `:tbody_html` option.
```

docs/3-index-pages/index-as-table.md
```diff
@@ -196,13 +196,23 @@ index do
...
+
+```ruby
+index tbody_html: { class: "my-class", data: { controller: 'stimulus-controller' } } do
+  # columns
+end
+```
+
+## Custom row HTML attributes
+
+In order to add special html attributes to table rows pass the proc object as a `:row_html` option
```
javierjulio, line 211:
```suggestion
In order to add HTML attributes to table rows, use a proc object in the `:row_html` option.
```

javierjulio (approved):
@thibaudgg I just have documentation updates, everything else looks great. We'll leave in `row_class` support for now and handle a proper deprecation separately. My main concern is just getting in the options to handle any attribute as you've added here. Thank you again for tackling this.

thibaudgg:
@javierjulio great, thanks for the review. I just updated the PR with your changes. 👍🏻

PR #8425, opened 2024-08-07
Bug Fix: Rails 7.1 Composite Primary Key Display Name Fallback

itay-grudev:
This fixes an issue in Rails 7.1 where because Composite Primary Keys are supported, `primary_key` can return an array.

In this case we iterate the array and produce an output in the form:

```
Model #(key1: value1 / key2: value 2)
```


mgrunberg:
@itay-grudev thanks for the PR.

Can you use master as the base branch? I think this problem also happens in `master` branch. If so the procedure would be fixing `master` and then create a backport PR for `3-0-stable`.

Also, can you add tests?

itay-grudev:
Master is the v4 implementation. The same file doesn't even exist there.

mgrunberg:
> Master is the v4 implementation. The same file doesn't even exist there.

On a quick search, `lib/active_admin/view_helpers/display_helper.rb`  is now `app/helpers/active_admin/display_helper.rb

javierjulio:
@itay-grudev it was moved. Please follow instructions and make the change on the default branch, not v3. You'll have to rebase your branch with master and force push. Make the change in the new file location. Thank you.

javierjulio:
It's worth noting that there are multiple references to `primary_key` in the code base so I'm not sure if this would be the only place to resolve that issue. We would appreciate any investigation into that if you rely on composite primary keys so we address it wholesale. Thank you.

javierjulio:
@itay-grudev you are welcome to recreate this to provide composite primary key support. Please use the default branch and look into the other primary key usage locations since this wasn't the only one. Thank you.

PR #8430, opened 2024-08-15
Add active admin layouts to tailwind config

jpawlyn:
The generated tailwind config includes all the partials that are copied to an app but does not include the layouts. This PR adds the layouts.


lib/generators/active_admin/assets/templates/tailwind.config.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
     './app/admin/**/*.{arb,erb,html,rb}',
     './app/views/active_admin/**/*.{arb,erb,html,rb}',
     './app/views/admin/**/*.{arb,erb,html,rb}',
+    './app/views/layouts/active_admin*.{arb,erb,html,rb}',
```
javierjulio, line 12: The other paths can include more file extensions but the layouts will not and will only be `.html.erb` files.

```suggestion
    './app/views/layouts/active_admin*.{erb,html}',
```

javierjulio (approved):
Thanks!

PR #8432, opened 2024-08-18
Bug Fix: Rails 7.1 Composite Primary Key Display Name Fallback

itay-grudev:
This PR is a rewrite of #8425 applied to the `main` branch.

---

The changes fixe an issue in Rails 7.1 where because Composite Primary Keys are supported, `primary_key` can return an array.

In this case we iterate the array and produce an output of the form:

```
Model #(key1: value1 / key2: value 2)
```

PR #8433, opened 2024-08-18
Bug Fix: Rails 7.1 Query Contraints multiple foreign keys

itay-grudev:
This fixes another Rails 7.1 issue when `query_constraints` are specified. In this case, `foreign_key` returns said query constraints.

See the method implementation here:

```ruby

def foreign_key(infer_from_inverse_of: true)
  @foreign_key ||= if options[:query_constraints]
    options[:query_constraints].map { |fk| fk.to_s.freeze }.freeze
  elsif options[:foreign_key]
    options[:foreign_key].to_s
  else
    derived_fk = derive_foreign_key(infer_from_inverse_of: infer_from_inverse_of)

    if active_record.has_query_constraints?
      derived_fk = derive_fk_query_constraints(derived_fk)
    end

    derived_fk
  end
end
```

This addresses the issue by checking if `foreign_key` is an array and joining it's elements into a string.

### DIFF
--- a/lib/active_admin/resource/attributes.rb
+++ b/lib/active_admin/resource/attributes.rb
@@ -21,7 +21,7 @@ module ActiveAdmin
         @foreign_methods ||= resource_class.reflect_on_all_associations.
           select { |r| r.macro == :belongs_to }.
           reject { |r| r.chain.length > 2 && !r.options[:polymorphic] }.
-          index_by { |r| r.foreign_key.to_sym }
+          index_by { |r| (r.foreign_key.is_a?( Array ) ? r.foreign_key.join( '-' ) : r.foreign_key).to_sym }
       end
 
       def reject_col?(c)
--- a/lib/active_admin/resource/attributes.rb
+++ b/lib/active_admin/resource/attributes.rb
@@ -21,7 +21,7 @@ module ActiveAdmin
         @foreign_methods ||= resource_class.reflect_on_all_associations.
           select { |r| r.macro == :belongs_to }.
           reject { |r| r.chain.length > 2 && !r.options[:polymorphic] }.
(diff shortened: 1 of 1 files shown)


javierjulio:
@itay-grudev thank you. It would be preferable to not link to the rubydoc but the Rails docs instead. Can you please provide more information and a better title as to what this fixes exactly? For example, why are the keys joined with a dash vs another character? How does this affect a resource page? An example, would go a long way. Also, there are multiple instances of `foreign_key` in the codebase so wouldn't others be affected and need an update like this one?

itay-grudev:
@javierjulio The way I understand the code here is that that whatever the foreign key is it needs to be reduced to a string used to build a Hash map - hence the `index_by` method call. I used a dash simply because Rails conventions encourage the use of `_` in database columns and a dash here would help distinguish the two values.

lib/active_admin/resource/attributes.rb
```diff
@@ -21,7 +21,7 @@ def foreign_methods
         @foreign_methods ||= resource_class.reflect_on_all_associations.
           select { |r| r.macro == :belongs_to }.
           reject { |r| r.chain.length > 2 && !r.options[:polymorphic] }.
-          index_by { |r| r.foreign_key.to_sym }
+          index_by { |r| (r.foreign_key.is_a?(Array) ? r.foreign_key.join('-') : r.foreign_key).to_sym }
```
tagliala, line 24: invoking a method here will slow down operations, but the ternary operator hides complexity and codecov is tricked into thinking that this line is being tested.

I would move the conditional to a method and use multiple lines for better code coverage

See warning at https://github.com/activeadmin/activeadmin/pull/8432/files#diff-551e9462ea228e696f89ae7a20c041a4e504d0ab25bad03f2e0c3edae939e344R9

/cc @javierjulio

PR #8371, opened 2024-06-09
Test against Rails 7.2

tagliala:
Rebase after:
- [x] #8370
- [x] #8369
- [x] #8372

Close #8367

PR #8434, opened 2024-08-19
Backport Test against Rails 7.2

mgrunberg:
Backport https://github.com/activeadmin/activeadmin/pull/8371


Gemfile
javierjulio, line 1: Rails 7.2 now supports sqlite3 v2 but it's outside of the changes here so I can't submit a suggestion. In the test groups, let's update `gem "sqlite3", "~> 1.7", platform: :mri` to `gem "sqlite3", platform: :mri` instead.

PR #8439, opened 2024-08-21
Improve form f.inputs attributes rendering

amiel:
This resolves https://github.com/activeadmin/activeadmin/discussions/8436 by using standard rails helpers to build html tags and attributes.

Note that this would be easy to backport to v3 and I'm happy to open another PR if a backported release would be considered.

### DIFF
--- a/lib/active_admin/views/components/active_admin_form.rb
+++ b/lib/active_admin/views/components/active_admin_form.rb
@@ -132,7 +132,7 @@ module ActiveAdmin
         legend = args.shift if args.first.is_a?(::String)
         legend = html_options.delete(:name) if html_options.key?(:name)
         legend_tag = legend ? "<legend class=\"fieldset-title\">#{ERB::Util.html_escape(legend)}</legend>" : ""
-        fieldset_attrs = html_options.map { |k, v| %Q{#{k}="#{v}"} }.join(" ")
+        fieldset_attrs = tag.attributes html_options
         @opening_tag = "<fieldset #{fieldset_attrs}>#{legend_tag}<ol>"
         @closing_tag = "</ol></fieldset>"
         super(*(args << html_options), &block)
--- a/lib/active_admin/views/components/active_admin_form.rb
+++ b/lib/active_admin/views/components/active_admin_form.rb
@@ -131,7 +131,7 @@ module ActiveAdmin
         html_options[:class] ||= "inputs"
         legend = args.shift if args.first.is_a?(::String)
         legend = html_options.delete(:name) if html_options.key?(:name)
-        legend_tag = legend ? "<legend class=\"fieldset-title\">#{ERB::Util.html_escape(legend)}</legend>" : ""
+        legend_tag = legend ? tag.legend(legend, class: "fieldset-title") : ""
         fieldset_attrs = tag_attributes html_options
         @opening_tag = "<fieldset #{fieldset_attrs}>#{legend_tag}<ol>"
         @closing_tag = "</ol></fieldset>"
--- a/lib/active_admin/views/components/active_admin_form.rb
+++ b/lib/active_admin/views/components/active_admin_form.rb
@@ -132,10 +132,22 @@ module ActiveAdmin
         legend = args.shift if args.first.is_a?(::String)
         legend = html_options.delete(:name) if html_options.key?(:name)
         legend_tag = legend ? "<legend class=\"fieldset-title\">#{ERB::Util.html_escape(legend)}</legend>" : ""
-        fieldset_attrs = tag.attributes html_options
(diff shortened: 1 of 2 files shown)


mgrunberg (approved):
@amiel thanks for this! Looks good to me

lib/active_admin/views/components/active_admin_form.rb
```diff
@@ -131,12 +131,22 @@ def build(form_builder, *args, &block)
         html_options[:class] ||= "inputs"
         legend = args.shift if args.first.is_a?(::String)
         legend = html_options.delete(:name) if html_options.key?(:name)
-        legend_tag = legend ? "<legend class=\"fieldset-title\">#{ERB::Util.html_escape(legend)}</legend>" : ""
-        fieldset_attrs = html_options.map { |k, v| %Q{#{k}="#{v}"} }.join(" ")
+        legend_tag = legend ? tag.legend(legend, class: "fieldset-title") : ""
+        fieldset_attrs = tag_attributes html_options
         @opening_tag = "<fieldset #{fieldset_attrs}>#{legend_tag}<ol>"
         @closing_tag = "</ol></fieldset>"
```
amiel, line 137: Note that it's not so trivial to use rails helpers to build the `fieldset` and `ol` since they are fragments.

lib/active_admin/views/components/active_admin_form.rb
```diff
@@ -131,12 +131,22 @@ def build(form_builder, *args, &block)
...
         super(*(args << html_options), &block)
       end
+
+      private
+
+      def tag_attributes(html_options)
+        # When support for Rails 6.x is dropped, use tag.attributes instead.
```
amiel, line 144: Is there a standard way to mark code to be reconsidered after rails upgrades?

lib/active_admin/views/components/active_admin_form.rb
```diff
@@ -131,12 +131,22 @@ def build(form_builder, *args, &block)
...
+        #
+        #     tag.attributes html_options
+        #
+        tag.tag_options(html_options.to_h).to_s.strip.html_safe
+      end
```
javierjulio, line 149: If I understand right this is done to support both Rails 6.1 and 7+ but I'd prefer that we use a condition so if it's Rails 6.1 we use the `tag.tag_options` method, otherwise always use the current `tag.attributes` method instead. I would expect the former to be removed at some point in Rails so that's why I want to make sure we isolate it to only those still running Rails 6.1 which eventually we'll drop support for.

amiel: I couldn't find any examples of checking the rails version in the ActiveAdmin codebase. Would you prefer feature detection or a version check?

For example, something like this?

```ruby
# Rails.version is a String, and both "6.1" <=> "7" and "7.1" <=> "7" work as expected
if Rails.version < "7"
```

or this?

```ruby
if tag.respond_to?(:attributes)
```

amiel: @javierjulio I went with the latter approach, but am happy to revisit.

javierjulio: I would use `if Rails::VERSION::MAJOR <= 6` to isolate the old method and that way the `else` condition is for the current approach. I would be ok with feature detection but favor the version check in this case because otherwise we'd be falling back to the old approach.

amiel: Thanks! Feature detection didn't work anyway 🤦🏼‍♂️

amiel:
Is it possible to re-run CI?

This looks like a transient error:

> [test (3.1, rails_72)](https://github.com/activeadmin/activeadmin/actions/runs/10512423883/job/29125707918)
> GitHub Actions has encountered an internal error when running your job

javierjulio:
@amiel GitHub Actions has been reporting issues for the past hour and was just resolved a few minutes ago. Either way looks like latest build is green so it's ok.

amiel:
Thanks @javierjulio. I think this PR is ready to go.

I appreciate you taking the time to help me through this change. It will clean up a bunch of code at [Bento](https://www.gobento.com/).

Would you consider another PR based on the `3-0-stable` branch? We're not quite ready to upgrade to v4.

javierjulio (approved):
Thanks!

javierjulio:
@amiel you're welcome. I will not do a backport release but @mgrunberg has done those previously. Since he engaged I believe he would consider a release if you open a PR based on the `3-0-stable` branch. I'll let him follow up. Thanks again.

mgrunberg:
@javierjulio, @amiel I have some free time today. I'm going to backport this and release it. @amiel, don't worry about the backport PR.

PR #8442, opened 2024-08-22
Fix batch action with partial and no confirm doesn't submit the form on click

### LINKED ISSUE #8440: Batch action automatically submits form even though a partial is being rendered
It looks like https://github.com/activeadmin/activeadmin/pull/8389 introduced a bug where batch actions rendered with a partial are being submitted immediately.
If you follow the [batch action partial example](https://github.com/activeadmin/activeadmin/blob/master/UPGRADING.md?plain=1#L104) in the upgrade guide, you can see it does not work.

The change [here](https://github.com/activeadmin/activeadmin/pull/8389/files#diff-f36b5577df9210253db0d1c7f87f3544c7ba2e557fc2e26625981f2e81a3eb7fR17) is causing the form to submit immediately because there is no `confirm` attribute.

### Expected behavior

On an index table:
1. Select rows to perform a batch action on
2. Click `Batch Actions` and select an action
3. Partial is rendered
4. User should be able to interact with the form before submitting

### Actual behavior

On an index table:
1. Select rows to perform a batch action on
2. Click `Batch Actions` and select an action
3. Partial is rendered
4. Form is also submitted

### How to reproduce

Create a batch action with a partial:

```
batch_action(
    :mark_published,
    partial: "mark_published_batch_action",
    link_html_options: {
      "data-modal-target": "mark-published-modal",
      "data-modal-show": "mark-published-modal"
    }
  ) do |ids, inputs|
    # ...
end
```

On the index page, select rows, click `Batch Actions` and select `Mark Published`.
See that the partial is rendered, but the form is submitted immediately and the page refreshes before you can interact with the partial.

ryoung:
A batch action rendered with a partial is automatically being submitted when the batch action is clicked as a result of https://github.com/activeadmin/activeadmin/pull/8389. The form is being submitted automatically if there is no `confirm` dataset attribute.

The batch action feature test was incorrectly passing because it was not re-rendering the page unless `inputs["starred"]` was present.

Fixes #8440

### DIFF
--- a/app/javascript/active_admin/features/batch_actions.js
+++ b/app/javascript/active_admin/features/batch_actions.js
@@ -14,7 +14,7 @@ const batchActionClick = function(event) {
     batchAction.value = this.dataset.action
   }
 
-  if (!event.target.dataset.confirm) { submitForm() }
+  if (!event.target.dataset.confirm && !event.target.dataset.modalTarget) { submitForm() }
 }
 
 const batchActionConfirmComplete = function(event) {
--- a/features/index/batch_actions.feature
+++ b/features/index/batch_actions.feature
@@ -204,6 +204,8 @@ Feature: Batch Actions
         batch_action :set_starred, partial: "starred_batch_action_form", link_html_options: { "data-modal-target": "starred-batch-action-modal", "data-modal-show": "starred-batch-action-modal" } do |ids, inputs|
           if inputs["starred"].present?
             redirect_to collection_path, notice: "Successfully flagged 10 posts"
+          else
+            redirect_to collection_path, notice: "Didn't flag any posts"
           end
         end
       end


app/javascript/active_admin/features/batch_actions.js
```diff
@@ -14,7 +14,7 @@ const batchActionClick = function(event) {
     batchAction.value = this.dataset.action
   }
 
-  if (!event.target.dataset.confirm) { submitForm() }
+  if (!event.target.dataset.confirm && !event.target.dataset.modalShow) { submitForm() }
```
javierjulio, line 17: We should rely on this attribute instead which will always be present for a Flowbite modal but the one used initially can be swapped out for another depending on need.

```suggestion
  if (!event.target.dataset.confirm && !event.target.dataset.modalTarget) { submitForm() }
```

javierjulio (approved):
Thanks! I'm not sure how I came up with that logic in the test case since it's clear that it won't do anything outside the condition and isn't right. Just one correction here and we'll get this out in a release this weekend.

app/javascript/active_admin/features/batch_actions.js
ryoung: Got it! I've made the change.

PR #8446, opened 2024-08-23
Backport Improve form f.inputs attributes rendering

mgrunberg:
Backport #8439 and #8448

### DIFF
--- a/lib/active_admin/views/components/active_admin_form.rb
+++ b/lib/active_admin/views/components/active_admin_form.rb
@@ -127,7 +127,7 @@ module ActiveAdmin
         html_options[:class] ||= "inputs"
         legend = args.shift if args.first.is_a?(::String)
         legend = html_options.delete(:name) if html_options.key?(:name)
-        legend_tag = legend ? tag.legend(legend, class: "fieldset-title") : ""
+        legend_tag = legend ? helpers.tag.legend(legend, class: "fieldset-title") : ""
         fieldset_attrs = tag_attributes html_options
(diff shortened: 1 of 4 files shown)


lib/active_admin/views/components/active_admin_form.rb
javierjulio, line 1: Based on your commit and reviewing the CI error, I didn't understand what happened. Do you know what the cause was? Would it be necessary/helpful to do this `helpers` change on the default branch as well?

javierjulio (approved):
Thanks!

mgrunberg:
Note: I'm waiting to merge this because I want to upstream a small fix to default branch first.

lib/active_admin/views/components/active_admin_form.rb
mgrunberg: Correct, I made https://github.com/activeadmin/activeadmin/pull/8448 to use `helpers` in both branches. The problem is not evident and only affects `Tag` resource (edge-case?). 
My theory is that v3 expose it because default form is rendered in the same context as a custom one (ResourceController or something similar where `tag` is an alias of `resource` attr_reader)

javierjulio: It wouldn't be the first time we've had a similar problem. I figured that would be the case, where it's some conflict with a resource. I think it makes sense to add that change back in the default branch too. Thank you again for addressing.

mgrunberg:
@javierjulio, @amiel  I'm going to release a new version with this next week.

amiel:
Thank you so much!

PR #8448, opened 2024-08-23
Be able to customize Tag resource form

mgrunberg:
https://github.com/activeadmin/activeadmin/pull/8439 improved form attributes rendering but introduced an issue if host app has an admin for Tag resource.

The following code (generated by our template app) raise errors

```ruby
# admin/tags.rb
ActiveAdmin.register Tag do
  form do |f|
    f.inputs "Details" do
      f.input :name
    end
    f.actions
  end
end
```

The problem is that `tag` (`tag.attributes`) evaluates to the resource and not the `TagHelper`.

The default form works because it's rendered in a different scope (view context?).

The problem surfaced during backport development: https://github.com/activeadmin/activeadmin/pull/8446.

Let's prefix the call with `helpers`.

### DIFF
--- a/features/edit_page.feature
+++ b/features/edit_page.feature
@@ -6,8 +6,11 @@ Feature: Edit Page
     Given a category named "Music" exists
     And a user named "John Doe" exists
     And a post with the title "Hello World" written by "John Doe" exists
+    And a tag named "Bugs" exists
     And I am logged in
-    And a configuration of:
+
+  Scenario: Default form with no config
+    Given a configuration of:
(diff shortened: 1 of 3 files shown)


amiel:
Apologies and thank you!

javierjulio:
@amiel not your fault and no need to apologize. This has happened before with Arbre so not surprising. @mgrunberg thank you again for addressing in both version tracks.

javierjulio (approved):
Thanks!

PR #8449, opened 2024-08-23
Drop support for Rails 6.1

mgrunberg:
According to [Rails Maintenance Policy](https://guides.rubyonrails.org/maintenance_policy.html), Rails 6.1 is about to loose security update in a month more or less and reach EOL (check [here](https://endoflife.date/rails)).

We can keep support for Rails 6.1 as long as we want.

The PR is to track the required changes to drop support


mgrunberg:
@javierjulio, @tagliala is it too premature to do this now? Considering we are working on a major release.

javierjulio:
@mgrunberg no, I don't think so. We can leave this open until we are ready to merge. The [Rails maintenance page states that 6.1 is supported through October 1, 2024](https://rubyonrails.org/maintenance#security) for security releases only.

tagliala:
Maybe this is a good time to ,erge

Since there are required workflows to pass

image

We can either remove those or bypass the rule for this PR

javierjulio:
When we approve this, we'll remove those required status checks first before merging since they won't apply anymore. Since we are coming up on the date next week, I think it's fine to get ready to merge whenever. Thank you.

tagliala (commented):
Should we add 

```yml
  TargetRailsVersion: 7.0
```

below

https://github.com/activeadmin/activeadmin/blob/3b54bde14bbce666198c73f1aac1e43458bdcc83/.rubocop.yml#L16

?

javierjulio:
@tagliala I didn't realize that was an option in rubocop-rails but makes sense and it is supported as I see mentions in the docs, thank you. Feel free to add it here if you like Matias.

mgrunberg:
I added the TargetRubyVersion. Thank @tagliala for the suggestion.

@javierjulio are we good to merge the PR now?

tagliala (approved):
I would merge this before dropping 3.0 support #8489

I'm going to disable the requirements for 6.1 actions. Done

javierjulio (approved):
Thanks!

PR #8451, opened 2024-08-23
Update bug report template to Rails 7.2

tagliala:
Also set config.eager_load to avoid a warning on stdout:

```
config.eager_load is set to nil. Please update your config/environments/*.rb files accordingly:
```


javierjulio (approved):
Thanks!

PR #8455, opened 2024-08-24
Update @rails/ujs to 7.1.400

tagliala:
It is the same file except for the version number. The new numbering will help with security releases (7.1.4.1 => 7.1.401)

Ref: https://guides.rubyonrails.org/maintenance_policy.html#npm-packages


javierjulio (approved):
Thanks!

PR #8456, opened 2024-08-25
Use consistent text-like inputs selector list for CSS styles

til:
Previously, inputs with type=datetime-local, type=search, type=week and type=month were missing some tailwind styles.

Before:
screenshot-20240825-125818

After:
screenshot-20240825-125752

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -312,7 +312,7 @@ module.exports = plugin(
       '[type=radio]': {
         '@apply w-4 h-4 border-gray-300 focus:ring-2 focus:ring-blue-300 dark:focus:ring-blue-600 dark:focus:bg-blue-600 dark:bg-gray-700 dark:border-gray-600': {}
       },
-      [['[type=date]', '[type=email]', '[type=number]', '[type=password]', '[type=tel]', '[type=text]', '[type=time]', '[type=url]', 'select', 'textarea']]: {
+      [['[type=datetime-local]', '[type=month]', '[type=week]', '[type=search]', '[type=date]', '[type=email]', '[type=number]', '[type=password]', '[type=tel]', '[type=text]', '[type=time]', '[type=url]', 'select', 'textarea']]: {
         '@apply bg-gray-50 border border-gray-300 text-gray-900 rounded-md focus:ring-blue-500 focus:border-blue-500 w-full dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-blue-500 dark:focus:border-blue-500': {}
       },
       'a': {


plugin.js
```diff
@@ -72,6 +72,7 @@ module.exports = plugin(
           'box-shadow': `var(--tw-ring-offset-shadow), var(--tw-ring-shadow), var(--tw-shadow)`,
           'border-color': theme('colors.blue.600', colors.blue[600]),
         },
+        '@apply bg-gray-50 border border-gray-300 text-gray-900 rounded-md focus:ring-blue-500 focus:border-blue-500 w-full dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-blue-500 dark:focus:border-blue-500': {},
```
javierjulio, line 75: I would prefer this stays where it is as this section where you moved the styles too came from Flowbite's default but I intend to remove/consolidate those before final release. I'd prefer the original location be updated with the missing selectors to match.

```suggestion
```

plugin.js
```diff
@@ -312,9 +313,6 @@ module.exports = plugin(
       '[type=radio]': {
         '@apply w-4 h-4 border-gray-300 focus:ring-2 focus:ring-blue-300 dark:focus:ring-blue-600 dark:focus:bg-blue-600 dark:bg-gray-700 dark:border-gray-600': {}
       },
-      [['[type=date]', '[type=email]', '[type=number]', '[type=password]', '[type=tel]', '[type=text]', '[type=time]', '[type=url]', 'select', 'textarea']]: {
-        '@apply bg-gray-50 border border-gray-300 text-gray-900 rounded-md focus:ring-blue-500 focus:border-blue-500 w-full dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-blue-500 dark:focus:border-blue-500': {}
-      },
```
javierjulio, line 317: As mentioned, I want these to remain as is but just updating the selector list with what's missing. At first glance, I think we just need to add datetime-local, month, search and week? Once you update it, I'll double check it's a match. Thank you.

til: Ok, makes sense. Yes, those 4 input types are missing. I noticed that all the `@apply`s were in the lower part of the file and didn't want to move this line up, so my first attempt was to copy the selector from above, like this:

```js
[[
  "[type='text']",
  "[type='email']",
  "[type='url']",
  "[type='password']",
  "[type='number']",
  "[type='date']",
  "[type='datetime-local']",
  "[type='month']",
  "[type='search']",
  "[type='tel']",
  "[type='time']",
  "[type='week']",
   'textarea',
   'select',
]]: {
  '@apply bg-gray-50 border border-gray-300 text-gray-900 rounded-md focus:ring-blue-500 focus:border-blue-500 w-full dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-blue-500 dark:focus:border-blue-500': {}
}
```

However, I then noticed (after some head scratching) that the selector is a key in a JS dictionary and therefore must not occur twice. Keeping the selectors formatted slightly differently, as they are now, will also work – if you are planning to consolidate these styles anyway, this is the better approach. I'll update the PR right away.

til:
Force-pushed to only append the missing types to the selector, leaving all styles in place.

plugin.js
javierjulio: No worries and thank you. The reason for this is we don't want the default Flowbite styles but just implement our own for a consistent default theme. Also to keep a smaller set of styles in place. I just need to get around to finishing that. The update looks good and double checked the element selectors and it's a match. Thanks again.

javierjulio (approved):
Thanks!

plugin.js
til: Cool, thanks for the thorough review and background infos! Yes, a consolidated list of styles would certainly be better – but also quite some work I assume.

PR #8457, opened 2024-08-25
Use attribute_types instead of columns_hash to determine type

til:
Previously, virtual attributes of type boolean with value `nil` would not be rendered as status tag. This also improves boolean detection in rare cases when an `attribute` declaration overrides the type detected in the database.

`attribute_types` existed already in Rails 6.1, so it should be safe to use.

### DIFF
--- a/app/helpers/active_admin/display_helper.rb
+++ b/app/helpers/active_admin/display_helper.rb
@@ -108,8 +108,8 @@ module ActiveAdmin
       when TrueClass, FalseClass
         true
       else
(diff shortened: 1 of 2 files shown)


javierjulio (approved):
Thanks! @tagliala @mgrunberg any concerns with this change? I've used the Attributes API before and makes sense to me. I figured I'd see if either of you know more before merging.

mgrunberg:
@javierjulio no concerns here, LGTM.

PR #8461, opened 2024-08-29
Add favicon to template app to fix Rails 7.2 test suite run

mgrunberg:
Our test suit fails due to
```
No route matches [GET] "/favicon.ico" (ActionController::RoutingError)
```

Our head partial does not define any `icon` tag which makes the browser request `favicon.ico`.

Rails 7.2 [changed the icon of the new generated app](https://github.com/rails/rails/pull/50526) from `favicon.ico` to a PNG file.


javierjulio:
@mgrunberg by including a dummy `public/favicon.ico` file I've confirmed that the tests pass consistently on Rails 7.2 with multiple runs. Once I remove it though, I encounter test failures. Would you mind if I push those changes here? I'd like to keep this focused on just resolving the failed tests for the generated app, nothing user facing.

mgrunberg:
@javierjulio can you recondider? 
I don't like adding a favicon. If we favor something it should be new rails defaults (a PNG).

Trusting on favicon existance forces all new projects to install our views or tweak the favico (if they notice the missing icon).

In this sense, I prefere to render the icon meta if assests exists.

As a non so strong argument, I don't use favicon nor logo.png (new default). I don't want to install the head just for these.

All that been said, despite the extra effort, I'm not against fix the test in one PR and solve user facing problems in another.

javierjulio:
There isn't a user facing problem here though. There is no need to configure a favicon in the layout for them, they are expected to do that if they have one. They can update it as they need to. I don't understand why we'd want to configure something for them. It's not something we need to maintain in the layout file we copy over either. We just need it to fix our test suite where we are using the `rails new` command to generate a dummy app that is causing test failures. If they don't have a favicon, the browser will just log an error so it can be noisy but the site doesn't crash.

mgrunberg:
I understand that you think adding the favicon back is the simplest change to fix the issues and I changed the PR to reflect that.

My point is that our layout uses an old Rails default which increases the chances that users need to make a customization. But site icon per see could require a customization anyway (I'm not using any of the Rails defaults for site icon).

Customizations are not good or wrong. After much thought, I realize I'm inclined to prevent customizations mostly because of the javascript we have on the head. So, I opened another PR (https://github.com/activeadmin/activeadmin/pull/8462) to change the only thing that makes me wonder about head customization. I hope you find both PR ok now.

javierjulio:
@mgrunberg ok great, thank you! Yes, this all we need, the simplest solution for now.

javierjulio (approved):
Thanks!

PR #8462, opened 2024-08-30
Move theme setup script into a partial

mgrunberg:
Keep the logic to intialize the them "private".

Users don't need to know the JS we use when they install views.

Simplify the update process for users if we ever make any change to
that.

### DIFF
--- a/app/views/active_admin/shared/_theme_setup.erb
+++ b/app/views/active_admin/shared/_theme_setup.erb
@@ -1,0 +1,8 @@
+<% # On page load or when changing themes, best to add inline in `head` to avoid FOUC %>
+<%= javascript_tag nonce: true do %>
+  if (localStorage.theme === 'dark' || (!('theme' in localStorage) && window.matchMedia('(prefers-color-scheme: dark)').matches)) {
+    document.documentElement.classList.add('dark')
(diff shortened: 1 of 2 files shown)


app/views/active_admin/shared/_theme_setup.erb
```diff
@@ -0,0 +1,8 @@
+<% # On page load or when changing themes, best to add inline in `head` to avoid FOUC %>
+<%= javascript_tag nonce: true do %>
+  if (localStorage.theme === 'dark' || (!('theme' in localStorage) && window.matchMedia('(prefers-color-scheme: dark)').matches)) {
```
tagliala, line 3: Should `theme` be prefixed by `activeAdmin` or configurable?

I understand that in some circumstances it is better to use the same application theme, but changing the theme in the admin area should not affect the application and vice versa 🤔

/cc @javierjulio

mgrunberg:
@tagliala I don't have an answer for your question but it made me change my mind. I'm not sure anymore that this PR is a good move. Now it seems a premature optimization. I'm closing the PR, you or Javier are open to re-open it (or create a new one. I don't know our policy here :sweat_smile:)

PR #8471, opened 2024-09-04
perform_database_query_matcher should be truthy as long as one query matches

mgrunberg:
Found while working in https://github.com/activeadmin/activeadmin/pull/8470

Before this commit the order or SQL queries matter:

- match, not match

results in a false matching when it should be true

### DIFF
--- a/spec/support/matchers/perform_database_query_matcher.rb
+++ b/spec/support/matchers/perform_database_query_matcher.rb
@@ -7,7 +7,9 @@ RSpec::Matchers.define :perform_database_query do |query|
     @match = nil
 
     callback = lambda do |_name, _started, _finished, _unique_id, payload|
-      @match ||= query_regexp.match?(payload[:sql])
+      if query_regexp.match?(payload[:sql])
+        @match = true
+      end
     end
 
     ActiveSupport::Notifications.subscribed(callback, "sql.active_record", &block)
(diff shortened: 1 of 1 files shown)


spec/support/matchers/perform_database_query_matcher.rb
```diff
@@ -7,7 +7,7 @@
     @match = nil
 
     callback = lambda do |_name, _started, _finished, _unique_id, payload|
-      @match = query_regexp.match?(payload[:sql])
+      @match ||= query_regexp.match?(payload[:sql])
```
javierjulio, line 10: I was confused by this one since `||=` will still evaluate and set the instance variable to `false` in that case. Although it won't if it's first set to `true` where you mentioned ordering matters. This just seems to hide that detail.

Do you think it can help to make this clearer and explicit by setting the flag only when `match?` returns true rather than any return value? Or perhaps a comment can help clarify?

```suggestion
      if query_regexp.match?(payload[:sql])
        @match = true
      end
```

Since at first I didn't understand what this was really solving, I include the following IRB sample for reference that helped me.

```
>> match = nil
=> nil
>> match ||= false
=> false
>> match ||= true
=> true
>> match ||= false
=> true
>> match
=> true
```

mgrunberg: I'm fine with your suggestion. Just don't make me backport it :pray:

PR #8482, opened 2024-09-19
Prefer `require_relative` for internal requires

tagliala:
`require_relative` is preferred over `require` for files within the same project because it uses paths relative to the current file, making code more portable and less dependent on the load path.

This change updates internal requires to use `require_relative` for consistency, performance, and improved portability.

Ref:
- activeadmin/arbre#622
- activeadmin/inherited_resources#939


deivid-rodriguez (approved):

tagliala:
@javierjulio do you want to take a look too?

tagliala:
@deivid-rodriguez apparently this can be considered a breaking change (which is fine in a beta version) because of this side effect: https://github.com/varvet/pundit/pull/831#issuecomment-2367444446

I'm a little bit concerned about `arbre` and `inherithed_resources`, I was assuming that this was fine in a patch release

Any thoughts?

deivid-rodriguez:
I think as long as our release policy permits small enhancements like this in a patch release, then it's fine. I mean, it could make achieving certain things in development a bit harder, but should not be breaking any expected production usage of the gems, so seems fine to me.

PR #8488, opened 2024-09-27
Minor development updates

tagliala:
- Update dependencies
- Improve test app generator
- Test Rails 7.0 (instead of 7.1) against Ruby 3.0

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,7 +32,7 @@ jobs:
         include:
           - ruby: '3.0'
             os: ubuntu-latest
-            rails: rails_71
+            rails: rails_70
     steps:
       - uses: actions/checkout@v4
       - name: Configure bundler (default)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -212,7 +212,7 @@ GEM
       railties (>= 6.0)
       responders (>= 2)
     io-console (0.7.2)
-    irb (1.14.0)
+    irb (1.14.1)
       rdoc (>= 4.0.0)
       reline (>= 0.4.2)
     iso (0.4.0)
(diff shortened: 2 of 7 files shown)


tagliala:
https://github.com/activeadmin/activeadmin/actions/runs/11071311136/job/30762881938#step:5:1278

> sqlite3-2.1.0-x86_64-linux-gnu requires ruby version < 3.4.dev, >= 3.1, which is
> incompatible with the current version, 3.0.7

This is unfortunate

.github/workflows/ci.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
         include:
           - ruby: '3.0'
             os: ubuntu-latest
-            rails: rails_71
+            rails: rails_70
```
tagliala, line 35: If this solution to https://github.com/activeadmin/activeadmin/pull/8488#issuecomment-2379322266 is acceptable, I will change the required actions in settings to allow the branch to be merged without skipping branch rules

javierjulio: I'm fine with dropping Ruby 3.0 from CI now. If that helps here and you are ok with it, let's do that.

tagliala: I would not do that if 3.0 is supported, I'm afraid of unintended breaking changes at patch level, in example by using 3.1 only syntax (like `Array#intersect?`)

This happened several times in the past, a couple of the most recent breaking changes at patch level that I remember are

- sidekiq-scheduler/sidekiq-scheduler#477
- thoughtbot/factory_bot#1614

Please let me know what you think

javierjulio: Well I figure we would bump the minimum required version anyway. I recall we've done that each time we've dropped an old EOL version. You are welcome to open that PR to drop Ruby 3.0 now if you want.

tagliala:
Superseded by #8489

.github/workflows/ci.yaml
tagliala: As a reference, here it is another breaking change introduced in a patch level release because of the shorthand syntax (>= 3.1)

rails/importmap-rails#268

PR #8489, opened 2024-09-30
Drop Ruby 3.0 compatibility

tagliala:
Also improve test app generator flags:
- Add `brakeman`, `ci`, and `rubocop` skips
- Remove `listen`, `spring`, `turbolinks`, and `webpack-install` skips,
  dropped in Rails 7.0
- Sort remaining skips

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,7 +13,7 @@ require:
 
 AllCops:
   DisabledByDefault: true
-  TargetRubyVersion: 3.0
+  TargetRubyVersion: 3.1
   TargetRailsVersion: 7.0
 
   Exclude:
@@ -331,6 +331,7 @@ Style/FrozenStringLiteralComment:
 
 Style/HashSyntax:
   Enabled: true
+  EnforcedShorthandSyntax: never
 
 Style/ParallelAssignment:
   Enabled: true
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -212,7 +212,7 @@ GEM
       railties (>= 6.0)
       responders (>= 2)
     io-console (0.7.2)
-    irb (1.14.0)
+    irb (1.14.1)
       rdoc (>= 4.0.0)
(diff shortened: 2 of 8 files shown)


.rubocop.yml
```diff
@@ -330,6 +330,7 @@ Style/FrozenStringLiteralComment:
 
 Style/HashSyntax:
   Enabled: true
+  EnforcedShorthandSyntax: never
```
tagliala, line 333: After changing the target Ruby version to 3.1, it would be possible to use the hash shorthand syntax

```rb
# bad
{foo: foo, bar: bar}

# good
{foo:, bar:}
```

however, there are 62 offenses across several files and I didn't want to have a lot of changes in this PR

Also, I personally do not like this syntax but I'm not against it

javierjulio: Same here. I think that's best. We'll let that happen naturally over time if others choose as newer versions become the default. Thank you.

tasks/test_application.rb
javierjulio, line 1: This is great. Thank you for keeping it up to date with the latest skip flags.

Gemfile.lock
```diff
@@ -400,10 +400,10 @@ GEM
...
       activesupport (>= 6.1)
       sprockets (>= 3.0.0)
-    sqlite3 (2.0.4-aarch64-linux-gnu)
-    sqlite3 (2.0.4-arm64-darwin)
-    sqlite3 (2.0.4-x86_64-darwin)
-    sqlite3 (2.0.4-x86_64-linux-gnu)
+    sqlite3 (2.1.0-aarch64-linux-gnu)
+    sqlite3 (2.1.0-arm64-darwin)
+    sqlite3 (2.1.0-x86_64-darwin)
+    sqlite3 (2.1.0-x86_64-linux-gnu)
```
tagliala, line 406: require Ruby 3.1

javierjulio (approved):
Thanks!

PR #8497, opened 2024-10-10
Add base-esm.js bundle for esbuild

javierjulio:
This updates the Rollup config to generate an ESM file alongside the existing UMD one. There are some differences here such as using `@rails/ujs` instead of `jquery_ujs` as the former is current and built with ESM. The latter remains in place for the UMD version. To make bundling easier, this [includes a custom build of jquery-ui using these settings](https://jqueryui.com/download/#!version=1.14.0&components=111111011111101010011111010000100000000000000000). The objective is with a new v3 minor release, to provide an additional JS bundle format [that works with jsbundling-rails and esbuild to help resolve known issues](https://github.com/activeadmin/activeadmin/discussions/7947).


package.json
```diff
@@ -43,6 +43,7 @@
     "prepublishOnly": "rm -rf src && cp -R app/javascript/active_admin src && cp -R app/assets/stylesheets/active_admin src/scss"
   },
   "dependencies": {
+    "@rails/ujs": "^7.1.3-4",
```
tagliala, line 46: this can be `7.1.400`

package.json
mgrunberg, line 1: Following [ActionText package.json](https://github.com/rails/rails/blob/v7.2.1/actiontext/package.json#L5), change [`main` and `type`](https://github.com/activeadmin/activeadmin/pull/8497/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R5-R6) with the following:
```
"module": "app/assets/javascripts/active_admin/base-esm.js",
"main": "app/assets/javascripts/active_admin/base.js",
```

mgrunberg (changes requested):

package.json
javierjulio: I've added the module key, thank you. The type=module item has to remain as we are using ESM in our Rollup config.

app/javascript/active_admin/base-esm.js
```diff
@@ -0,0 +1,27 @@
...
+import "./initializers/dropdown-menu"
+import "./initializers/filters"
+import "./initializers/has-many"
+import "./initializers/per-page"
+import "./initializers/table-checkbox-toggler"
+import "./initializers/tabs"
+
+import ModalDialog from "./lib/modal-dialog";
+
+function modal_dialog(message, inputs, callback) {
```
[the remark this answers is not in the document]
javierjulio: I've made this change, thank you.

package.json
javierjulio: I've made this change, thank you. I think they messed up on versioning since installing should have used this one but it didn't. It doesn't seem this version is tagged as the latest in NPM, FYI.

javierjulio:
@tagliala @mgrunberg the last commit contains a different approach for jquery-ui that if it works, I think it's best as we'd avoid having to vendor the jquery-ui source code. Can you please test this in an app to see if that works for you? You'd have to include the JS with a `type: :module` attribute. I had success with the original approach but next week, I'll try it with the new approach that imports each of our jquery-ui dependencies.

tagliala:
@javierjulio did the following changes, bundled & yarned

```diff
-gem 'activeadmin'
+gem 'activeadmin', github: 'activeadmin/activeadmin', branch: '3-0-esm-js-base'

-    "@activeadmin/activeadmin": "^3.2.5",
+    "@activeadmin/activeadmin": "activeadmin/activeadmin#3-0-esm-js-base",
```

What should I do next?

The app is using `webpack`

```
  "scripts": {
    "build": "webpack --config webpack.config.js"
  }
```

`active_admin.js` looks like this:

```
require('jquery')
require('jquery-ui/ui/widgets/datepicker')
require('jquery-ui/ui/widgets/dialog')
require('jquery-ui/ui/widgets/sortable')
require('jquery-ui/ui/widgets/tabs')
require('jquery-ui/ui/widget')
require('jquery-ujs')

require('@activeadmin/activeadmin')
require('arctic_admin')
```

package.json
mgrunberg: The problem with this config is that we claim our package is a module but the entry point (main) is UMD.
This most likely works because bundlers do magic, but I wouldn't say I like the inconsistency.

For v3, and just for v3, I would change "import" with "require" and remove `type=module` from package.json.

mgrunberg:
> @tagliala @mgrunberg the last commit contains a different approach for jquery-ui that if it works, I think it's best as we'd avoid having to vendor the jquery-ui source code. Can you please test this in an app to see if that works for you? You'd have to include the JS with a `type: :module` attribute. I had success with the original approach but next week, I'll try it with the new approach that imports each of our jquery-ui dependencies.

@javierjulio I couldn't test this yet. I'll see if I can this week. Sorry, busy weeks!

PR #8500, opened 2024-10-12
Test against Rails 8.0.0

tagliala:
- [x] #8501
- [x] #8502
- [x] https://github.com/svenfuchs/rails-i18n/pull/1130
- [x] cucumber/cucumber-rails#590


lib/active_admin/router.rb
```diff
@@ -65,10 +65,10 @@ def resource_routes(config)
 
     def page_routes(config)
       page = config.underscored_resource_name
-      router.get "/#{page}" => "#{page}#index"
+      router.get "/#{page}", to: "#{page}#index"
       config.page_actions.each do |action|
         Array.wrap(action.http_verb).each do |verb|
-          build_route(verb, "/#{page}/#{action.name}" => "#{page}##{action.name}")
+          build_route(verb, "/#{page}/#{action.name}", to: "#{page}##{action.name}")
```
javierjulio, line 71: I prefer we do this change in a separate PR to close out the issue since this is not dependent on Rails 8 tests. This route signature is already long supported and thus is backwards compatible.

tagliala: @javierjulio yes, I've committed here just for testing and to make sure that there aren't other deprecations, then I will backport to a separate PR

javierjulio: @tagliala ok great, thank you.

tagliala:
```diff
-    cucumber-rails (3.0.0)
-      capybara (>= 3.11, < 4)
-      cucumber (>= 5, < 10)
-      railties (>= 5.2, < 8)
+    cucumber-rails (1.4.0)
+      capybara (>= 1.1.2)
+      cucumber (>= 1.2.0)
+      nokogiri (>= 1.5.0)
+      rails (>= 3.0.0)
```

tagliala:
Cucumber Rails seems to be the last blocker. There are no unexpected downgrades and CI is green. Is there an action to check for downgrades?

javierjulio:
Ok thank you. I’m not sure if there is.

javierjulio (approved):
Thanks!

PR #8501, opened 2024-10-12
Update dependencies

tagliala:
Pin zeitwerk to `~> 2.6.18` because 2.7.0 only supports Ruby >= 3.2

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,9 @@ group :development, :test do
 
   gem "cssbundling-rails"
   gem "importmap-rails"
+
+  # FIXME: relax this dependency when Ruby 3.1 support will be dropped
+  gem "zeitwerk", "~> 2.6.18"
 end
 
 group :test do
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -18,6 +18,9 @@ group :development, :test do
 
   gem "cssbundling-rails"
   gem "importmap-rails"
+
+  # FIXME: relax this dependency when Ruby 3.1 support will be dropped
+  gem "zeitwerk", "~> 2.6.18"
```
tagliala, line 23: I don't like this but I don't have better ideas.

With this approach, tests agains zeitwerk 2.7 will not be performed, which is not good

javierjulio: That's ok. I'm surprised they dropped support so soon. I'm fine with leaving this as is as I would have done the same, comment as well so it's easy to find. We'll address when it's time. Thanks!

Issue #8498, opened 2024-10-12
Test suite: `ActiveSupport::DeprecationException` with Rails 8


tagliala:
```
ActiveSupport::DeprecationException:
       DEPRECATION WARNING: Drawing a route with a hash key name is deprecated and will be removed in Rails 8.1. Please use the `to:` option with "controller#action" syntax instead. Instead of: match "path" => "controller#action` Please use: match "path", to: "controller#action" (called from page_routes at /Users/geremia/dev/activeadmin/lib/active_admin/router.rb:68)
```

Ref: rails/rails#52422

javierjulio:
Details in the PR https://github.com/activeadmin/activeadmin/pull/8502#issuecomment-2408686158

PR #8502, opened 2024-10-12
Fix routes deprecation warning for Rails 8.x

tagliala:
Close: #8498

Ref: rails/rails#52422

### DIFF
--- a/lib/active_admin/router.rb
+++ b/lib/active_admin/router.rb
@@ -65,10 +65,10 @@ module ActiveAdmin
 
     def page_routes(config)
       page = config.underscored_resource_name
-      router.get "/#{page}" => "#{page}#index"
+      router.get "/#{page}", to: "#{page}#index"
       config.page_actions.each do |action|
         Array.wrap(action.http_verb).each do |verb|
-          build_route(verb, "/#{page}/#{action.name}" => "#{page}##{action.name}")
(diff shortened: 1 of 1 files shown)


tagliala:
Trying to understand how I had this deprecation message on my dev machine, does not seem to affect 8.0.0.beta1

tagliala:
Sorry, I can't replicate the deprecation but this commit should not hurt

javierjulio:
@tagliala the reason is this was actually reverted in https://github.com/rails/rails/commit/5c0b7496ab32c25c80f6d1bdc8b32ec6f75ce1e4 as I noticed from [this rubocop-rails PR comment](https://github.com/rubocop/rubocop-rails/pull/1329#issuecomment-2322661505).

tagliala:
Thanks for noticing. I didn't see the revert PR linked to the PR I was taking a look to.

What should we do?

javierjulio:
@tagliala the reason nothing was linked is that there was no PR. It was a revert commit that was pushed directly to the default branch. Hmm well it doesn't seem worth doing the change because there is no deprecation now. Reviewing other PR's following that, [there was another revert as well](https://github.com/rails/rails/pull/52664). Rails is making changes to routing but seems to be scaled back a bit. I think we should ignore this for now and only make a change once we have deprecations from an official release. Thank you.

PR #8528, opened 2024-11-02
Remove redundant safe navigation operator

tagliala:
Remove redundant safe navigation operator (`&.`) in `result.to_s.strip.present?` check, as `to_s` handles `nil` gracefully by converting it to an empty string.

This makes the code cleaner and more idiomatic without altering behavior.

Ref: https://ruby-doc.org/core-3.1.0/Object.html#method-i-to_s

### DIFF
--- a/app/helpers/active_admin/display_helper.rb
+++ b/app/helpers/active_admin/display_helper.rb
@@ -24,7 +24,7 @@ module ActiveAdmin
     def display_name(resource)
       unless resource.nil?
         result = render_in_context(resource, display_name_method_for(resource))
-        if result.to_s&.strip&.present?
+        if result.to_s.strip.present?
           ERB::Util.html_escape(result)
         else
           ERB::Util.html_escape(render_in_context(resource, DISPLAY_NAME_FALLBACK))


app/helpers/active_admin/display_helper.rb
```diff
@@ -24,7 +24,7 @@ def DISPLAY_NAME_FALLBACK.inspect
     def display_name(resource)
       unless resource.nil?
         result = render_in_context(resource, display_name_method_for(resource))
-        if result.to_s&.strip&.present?
+        if result.to_s.strip.present?
```
tagliala, line 27: unless I'm missing something and in some circumstances `to_s` can be nil (maybe override `to_s`, but that is a user problem)

javierjulio: I made this update in #6342 but I don't recall why I used the safe navigation operator. Likely just out of caution without really knowing that `nil.to_s` will return an empty string so it is redundant as you say. This looks good to me and makes sense to change. Thank you.

javierjulio (approved):
Thanks!

PR #8530, opened 2024-11-02
Use safe navigation for `pundit_default_policy`

tagliala:
Use the safe navigation operator (`&.`) to streamline the
`pundit_default_policy` constantization check, removing the need for a
redundant `&&`.

This change is now possible as the codebase no longer requires Ruby 2.3
support, which did not yet support safe navigation.

The previous implementation was added in commit https://github.com/activeadmin/activeadmin/commit/c93f6f49c6d4a7f35c98466fc1bc1cd0e1c826c5 to ensure
compatibility with older Ruby versions.


javierjulio (approved):
Thanks! I think there is another instance on line 26 that can be changed too.

PR #8533, opened 2024-11-02
Use safe navigation operator in layout helper

tagliala:
Replace a logic `&&` with safe navigation operator now that Ruby < 2.3 support is not required anymore

Ref: #8530

---

NOTE: there is also an (unsafe) cop to detect this use case

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/SafeNavigation

there are a few more occurrences


javierjulio (approved):
Thanks!

PR #8536, opened 2024-11-06
Convert plugin.js to ESM because package.json has type set to module

### LINKED ISSUE #8534: v4: provide plugin.js in both mjs and cjs variants (Node 23)
When trying to use AA v4 with vite_rails the fact that `plugin.js` is written in cjs, but `package.json` has `"type": "module"` causes issues with `require` inside of it

### Expected behavior

Tailwind plugin should work correctly with vite_rails

### Actual behavior

Errors like:

```
17:48:58 vite.1   | Error [ERR_REQUIRE_ESM]: require() of ES Module /redacted/node_modules/.pnpm/@activeadmin+activeadmin@4.0.0-beta13/node_modules/@activeadmin/activeadmin/plugin.js from /redacted/tailwind-active_admin.config.cjs not supported.
17:48:58 vite.1   | plugin.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
17:48:58 vite.1   | Instead either rename plugin.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /redacted/node_modules/.pnpm/@activeadmin+activeadmin@4.0.0-beta13/node_modules/@activeadmin/activeadmin/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

### How to reproduce

Try to require default AA tailwind config from postcss in vite_rails project

4ndv:
This PR makes `plugin.js` use ESM syntax instead of CJS, because `package.json` has `"type": "module"` which causes issues when using bundlers.

I've tested this change against activeadmin demo repo, replacing the file in node_modules, and it works fine with the oldest non-EOL version of node, `18.20.4`.

Looks like autogenerated tailwind config can `require()` ESM version just fine, so probably there is no need to ship both CJS and ESM versions.

Fixes #8534

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -1,6 +1,6 @@
-const plugin = require('tailwindcss/plugin')
-const defaultTheme = require('tailwindcss/defaultTheme');
-const colors = require('tailwindcss/colors');
+import plugin from 'tailwindcss/plugin';
+import defaultTheme from 'tailwindcss/defaultTheme';
+import colors from 'tailwindcss/colors';
 const { spacing, borderWidth, borderRadius } = defaultTheme;
 
 // https://github.com/tailwindlabs/tailwindcss/discussions/9336
@@ -28,7 +28,7 @@ const svgToTinyDataUri = (() => {
   return svgToTinyDataUri;
 })();
 
-module.exports = plugin(
+export default plugin(
   function({ addBase, addComponents, theme }) {
     addBase({
       [[


4ndv:
Well, I've tried to test these changes with Vite and looks like I've *completely* borked my previous attempt, because looks like it works without these changes.

Not sure if this PR should be closed or not, I don't think this migration will do any harm, but the original issue looks like a problem on my side.

Btw [here's my repo](https://github.com/4ndv/activeadmin_vite) with aa4+vite example, because my previous attempt was on work machine, and it's a PITA to separate from the project, so I've started from the clean state

4ndv:
Reading the error message from issue again, I still think that this PR is valid, because the message states very clearly that all of the files inside npm package marked as ESM should be in ESM style and this mix and match can cause some issues.

javierjulio:
@4ndv thank you. We'll keep this open since it makes sense to use ESM here and to ensure compatibility with Vite. I will try to find time soon to test this locally where not using Vite. Let us know what you find out when you have time. Thanks again.

til:
Maybe of interest: after upgrading node from version 22 to 23 in a project with AA 4.0.0.beta13, I saw an error that looks different than yours but (I think) has a related reason:

```
22:45:19 css.1  | (node:17401) ExperimentalWarning: CommonJS module /home/tils/basistool/tailwind-active_admin.config.js is loading ES Module /home/tils/basistool/node_modules/@activeadmin/activeadmin/plugin.js using require().
22:45:19 css.1  | Support for loading ES Module in require() is an experimental feature and might change at any time
22:45:19 css.1  | (Use `node --trace-warnings ...` to show where the warning was created)
22:45:19 css.1  | file:///home/tils/basistool/node_modules/@activeadmin/activeadmin/plugin.js:1
22:45:19 css.1  | const plugin = require('tailwindcss/plugin')
22:45:19 css.1  |                ^
22:45:19 css.1  | 
22:45:19 css.1  | ReferenceError: require is not defined
22:45:19 css.1  |     at file:///home/tils/basistool/node_modules/@activeadmin/activeadmin/plugin.js:1:16
22:45:19 css.1  |     at ModuleJobSync.runSync (node:internal/modules/esm/module_job:367:35)
22:45:19 css.1  |     at ModuleLoader.importSyncForRequire (node:internal/modules/esm/loader:325:47)
22:45:19 css.1  |     at loadESMFromCJS (node:internal/modules/cjs/loader:1396:24)
22:45:19 css.1  |     at Module._compile (node:internal/modules/cjs/loader:1529:5)
22:45:19 css.1  |     at Object..js (node:internal/modules/cjs/loader:1678:16)
22:45:19 css.1  |     at Module.load (node:internal/modules/cjs/loader:1315:32)
22:45:19 css.1  |     at Function._load (node:internal/modules/cjs/loader:1125:12)
22:45:19 css.1  |     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
22:45:19 css.1  |     at wrapModuleLoad (node:internal/modules/cjs/loader:216:24)
22:45:19 css.1  | 
```

Applying this PR fixed it.

4ndv:
> Maybe of interest: after upgrading node from version 20 to 23 in a project with AA 4.0.0.beta13, I saw an error that looks different than yours but (I think) has a related reason:

Just tested running node 23.1.0 with [AA demo repo](https://github.com/activeadmin/demo.activeadmin.info) and can confirm this:

```
▲ demo.activeadmin.info [main] yarn build:css
yarn run v1.22.19
$ tailwindcss -i ./app/assets/stylesheets/active_admin.css -o ./app/assets/builds/active_admin.css --minify -c tailwind-active_admin.config.js

Rebuilding...
(node:497283) ExperimentalWarning: CommonJS module /home/lynx/projects/demo.activeadmin.info/tailwind-active_admin.config.js is loading ES Module /home/lynx/projects/demo.activeadmin.info/node_modules/@activeadmin/activeadmin/plugin.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
file:///home/lynx/projects/demo.activeadmin.info/node_modules/@activeadmin/activeadmin/plugin.js:1
const plugin = require('tailwindcss/plugin')
               ^

ReferenceError: require is not defined
```

I always prefer sticking with LTS versions of node ("even" ones, 20.x, 22.x) so haven't even tested that version, thank you!

Looks like the warning is caused because default `tailwind-active_admin.config.js` is cjs trying to require AA plugin, which node thinks is ESM, and the exception is thrown because that file is actually CJS, and in ESM context there is no global `require`.

tagliala:
Confirmed issue with Node 23

javierjulio (approved):
Thanks!

tagliala:
This deserves an entry in the changelog, because it also fixes compatibility with Node 23

javierjulio:
@tagliala yes, it will be part of the release notes since its generated automatically based on the PR label.

PR #8550, opened 2024-11-19
Fix attributes passed to form has_many not being set on new record form items

Fs00:
This PR fixes a behavior inconsistency that occurs when passing custom attributes to the form builder `has_many` method, like in the following example:
```ruby
f.has_many :posts, data: { controller: "custom" } do |p|
  ...
end
```
which builds the following HTML (I kept only the relevant parts):
```html
<li class="has_many_container posts" data-sortable-start="0">
  <fieldset data-controller="custom" class="inputs has_many_fields">...</fieldset>
   ...
  <fieldset data-controller="custom" class="inputs has_many_fields">...</fieldset>
   ...
</li>
```
However, when clicking the "New Record" link to add a new item to the association, the `<fieldset>` generated for the new item didn't have the user-provided attributes on it (it was just `<fieldset class="inputs has_many_fields">...</fieldset>`).

I've discovered that the HTML used for new items is stored in the `data-html` attribute of the link and generated by a different method that was ignoring all custom attributes except for `class`.
So I've changed the method to make sure that custom attributes are not ignored and added tests for the relevant scenarios.

We'd need this fix on a 3.x patch, should I open the backport PR once this is merged or will you take care of it?

PR #8551, opened 2024-11-21
(Backport) Fix attributes passed to form has_many not being set on new record form items

Fs00:
Backports #8550 for ActiveAdmin 3.x.

The diff is basically the same except for CSS classes that differ between AA 3 and 4.


javierjulio (approved):
Thanks!

Fs00:
Would it be possible to get a new 3.x patch release with this change once #8650 is merged? This change would really be useful to us.
It's already available in 4.0 beta releases since a while, but unfortunately we won't be able to upgrade in the near future.

mgrunberg:
> Would it be possible to get a new 3.x patch release with this change once #8650 is merged? This change would really be useful to us. It's already available in 4.0 beta releases since a while, but unfortunately we won't be able to upgrade in the near future.

@Fs00 A new v3 version is gonna be released in the following days. The plan is to merge that PR and another one, then cut a new release.

Fs00:
This is great news, thank you!

PR #8556, opened 2024-11-27
Backport test against Rails 8.0

mgrunberg:
Backport #8500


tagliala:
@mgrunberg do you want some help in completing this?

mgrunberg:
> @mgrunberg do you want some help in completing this? I can allocate some time

@tagliala thanks for the offer. If you can, please give me a hand. I won't we able to work on this until the end of the month so the help is very welcome.

spec/unit/views/pages/form_spec.rb
```diff
@@ -16,7 +16,7 @@
     end
 
     let(:arbre_context) do
-      OpenStruct.new(params: params, helpers: helpers, assigns: {})
+      Struct.new(:params, :helpers, :assigns).new(params, helpers, {})
```
tagliala, line 19: This is required to allow specs to pass. Apparently, `ostruct` should be explicitly required after some changes to the dependencies but it is not used on `master` anymore and I've changed it to struct for parity

javierjulio: Yes, that's right, thank you for doing this as we don't want to use OpenStruct in general.

javierjulio (approved):
Thanks!

PR #8568, opened 2024-12-06
Use ESM syntax in tailwind template

wrozka:
Recent syntax [change](https://github.com/activeadmin/activeadmin/pull/8536) in plugin.js causes tailwind watch to break. It runs successfully on first run, but then it builds only partially - it doesn't crash, but the output files don't contain all the CSS they should.

I'm not entirely sure what causes it to break only on reloads, but the current version on tailwind config produces this warning:
```
ExperimentalWarning: CommonJS module /Users/wrozka/Projects/orders/tailwind-active_admin.config.js is loading ES Module /Users/wrozka/Projects/orders/node_modules/@activeadmin/activeadmin/plugin.js using require().
```

This PR fixes the broken tailwind watch by migrating tailwind config from CJS to ESM, but it doesn't resolve the warning, it just moves it one level up the stack trace:
```
ExperimentalWarning: CommonJS module /Users/wrozka/Projects/orders/node_modules/tailwindcss/lib/lib/load-config.js is loading ES Module /Users/wrozka/Projects/orders/tailwind-active_admin.config.js using require().
```

I don't think this is the final fix, but at least it fixes the watch problem.

### DIFF
--- a/lib/generators/active_admin/assets/templates/tailwind.config.js
+++ b/lib/generators/active_admin/assets/templates/tailwind.config.js
@@ -1,7 +1,9 @@
-const execSync = require('child_process').execSync;
+import { execSync } from 'child_process';
+import plugin from '@activeadmin/activeadmin/plugin';
+
 const activeAdminPath = execSync('bundle show activeadmin', { encoding: 'utf-8' }).trim();
 
-module.exports = {
+export default {
   content: [
     `${activeAdminPath}/vendor/javascript/flowbite.js`,
     `${activeAdminPath}/plugin.js`,
@@ -14,6 +16,6 @@ module.exports = {
   ],
   darkMode: "selector",
   plugins: [
-    require(`@activeadmin/activeadmin/plugin`)
+    plugin
   ]
 }


lib/generators/active_admin/assets/templates/tailwind.config.js
```diff
@@ -1,7 +1,9 @@
-const execSync = require('child_process').execSync;
+import { execSync } from 'child_process';
+import plugin from '@activeadmin/activeadmin/plugin';
```
javierjulio, line 2: Let's give this an explicit name to make it easier to replace the location where the variable will be used which we need for the generated dev and test apps.

```suggestion
import activeAdminPlugin from '@activeadmin/activeadmin/plugin';
```

lib/generators/active_admin/assets/templates/tailwind.config.js
```diff
@@ -14,6 +16,6 @@ module.exports = {
   ],
   darkMode: "selector",
   plugins: [
-    require(`@activeadmin/activeadmin/plugin`)
+    plugin
```
javierjulio, line 19:
```suggestion
    activeAdminPlugin
```

javierjulio (approved):
Thanks!

matteeyah:
This breaks for Rails installations that don't have a node runtime with `@activeadmin/activeadmin` available in development or during deployment. This is most new Rails installs, since they use `importmap`s by default, instead of the legacy way of including a `package.json` and a full `node` environment.

```
#19 10.94 Error: Cannot find module '@activeadmin/activeadmin/plugin'
#19 10.94 Require stack:
#19 10.94 - /rails/config/tailwind-active_admin.config.js
#19 10.94     at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
#19 10.94     at Function._resolveFilename (pkg/prelude/bootstrap.js:1951:46)
#19 10.94     at Function.resolve (node:internal/modules/cjs/helpers:108:19)
#19 10.94     at Function.resolve (/snapshot/tailwindcss/standalone-cli/patch-require.js:34:38)
#19 10.94     at _resolve (/snapshot/tailwindcss/node_modules/jiti/dist/jiti.js:1:241814)
#19 10.94     at jiti (/snapshot/tailwindcss/node_modules/jiti/dist/jiti.js:1:244531)
#19 10.94     at /rails/config/tailwind-active_admin.config.js:2:15
#19 10.94     at evalModule (/snapshot/tailwindcss/node_modules/jiti/dist/jiti.js:1:247313)
#19 10.94     at jiti (/snapshot/tailwindcss/node_modules/jiti/dist/jiti.js:1:245241)
#19 10.94     at /snapshot/tailwindcss/lib/lib/load-config.js:52:26 {
#19 10.94   code: 'MODULE_NOT_FOUND',
#19 10.94   requireStack: [ '/rails/config/tailwind-active_admin.config.js' ]
#19 10.94 }
```

PR #8574, opened 2024-12-12
Fix input file styles

wrozka:
This PR imports file input styles from [Flowbite](https://flowbite.com/docs/forms/file-input/#file-upload-example) and fixes display [issue](https://github.com/themesberg/flowbite/issues/578) in Safari.

Current look of file inputs:
Screenshot 2024-12-12 at 16 23 00

File inputs after fix:
Screenshot 2024-12-12 at 16 23 22


plugin.js
```diff
@@ -209,13 +209,22 @@ export default plugin(
...
-        'border-width': '0',
-        'border-radius': '0',
-        padding: '0',
-        'font-size': 'unset',
-        'line-height': 'inherit',
+        display: 'block',
+        width: '100%',
+        cursor: 'pointer',
+        overflow: 'hidden',
+        'border-radius': '.5rem',
```
tagliala, line 216: Hi and thanks for this PR.

The border-radius seems off to me by taking a look at the screenshot, and it's confirmed because of `.375rem` in the other buttons

Can you please use the settings from the theme?

I can see 

```
        'border-width': borderWidth['DEFAULT'],
        'border-radius': borderRadius.none,
```

in this file. Don't know exactly how. flowbite works. It looks like it is using `rounded-md` class

wrozka: @tagliala Thanks, done! I've updated the PR to use theme config as much as possible. I reviewed the file again and I've realised I could replace all those styles with just one big `@apply`:

```
[`[type='file']`]: {
  @apply block w-full text-sm text-gray-900 border border-gray-300 rounded-lg cursor-pointer bg-gray-50 dark:text-gray-400 focus:outline-none dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 overflow-hidden;
}
```

Do you have preferred style? I've seen both inlines and applies in this file.

tagliala: Thanks

> Do you have preferred style?

Can't tell exactly, not my expertise, let's wait for @javierjulio 's review, I guess it should be a mix of both but I can't tell exactly what should be in a class and what in that definition

wrozka:
@javierjulio Did you have a chance to take a look at this PR? What do you think?

aalyahya:
Is there anything I can help with to get this PR merged? @tagliala  @javierjulio

tagliala:
I'm sorry, the result looks good to me, but I can't approve this PR

plugin.js
javierjulio: The issue is mostly the `::file-selector-button` styles since it's offset. For example, the block display and overflow hidden are unnecessary. Since it's not addressed here, I'd rather tackle that and keep the styles simpler, matching the base text input instead.

javierjulio:
@wrozka thank you for opening this as I wasn't aware of it and for your patience. I will address this in a new PR (edit: #8809) that finishes the dark mode styles for form controls which were incomplete. The issue here is actually with the `::file-selector-button` styles since it's offset. The block display and overflow hidden are unnecessary as an example.

wrozka:
Thanks, @javierjulio! I have to admit I just blindly copied the styles from Flowbite to fix it 😅

PR #8587, opened 2024-12-27
Test against Ruby 3.4

tagliala:
`bin/bundle update` needs to be executed with Ruby 3.3 for Rails 7.0

- [x] ffi/ffi#1132
- [x] activeadmin/arbre#647
- [x] cucumber/cucumber-ruby#1769


gemfiles/rails_70/Gemfile.lock
tagliala, line 1: I don't know if it is correct to run `bin/bundle update` with Ruby 3.4 for Rails 7.0, because then it will drop the platform-specific sqlite versions, because `sqlite3 (1.7.3)` has not been built against 3.4 (is it still maintained?)

tagliala:
Cucumber Ruby 9.2.1 released, this is ready for merge

javierjulio (approved):
Thanks!

PR #8592, opened 2025-01-03
Background color for opened parent menu section

kevink1103:


javierjulio:
@kevink1103 thank you. The original implementation is meant as a simple base for others to build on instead of baking into the default template which is why we won't accept this change. Sorry about that.

PR #8599, opened 2025-01-04
Fix circular require warning in `belongs_to.rb`

### LINKED ISSUE #8598: Circular reference in `resource/belongs_to.rb`
I don't know if this is a mistake, a left-over, or it is intended to make it this way.

I will address this separately from other warnings

Blame doesn't help

```
lib/active_admin/resource/belongs_to.rb:2: warning: lib/active_admin/resource/belongs_to.rb:2: warning: loading in progress, circular require considered harmful - lib/active_admin/resource.rb
```

Ref:

https://github.com/activeadmin/activeadmin/blob/5f54f9110a82cf6e6a80019ab2f1c64fa5d33d96/lib/active_admin/resource/belongs_to.rb#L2

https://github.com/activeadmin/activeadmin/blob/5f54f9110a82cf6e6a80019ab2f1c64fa5d33d96/lib/active_admin/resource.rb#L15

tagliala:
When running specs with `RUBYOPTS='-w'`, numerous warnings are generated
due to a circular require between:
- `lib/active_admin/resource/belongs_to.rb`
- `lib/active_admin/resource.rb`

The warning message is:

```
lib/active_admin/resource/belongs_to.rb:2: warning: lib/active_admin/resource/belongs_to.rb:2: warning: loading in progress, circular require considered harmful - lib/active_admin/resource.rb
```

Based on the code history, there is no specific reason for the inverse
require from `belongs_to` to `resource`.

This change removes the unnecessary require to eliminate the warnings.

Closes #8598

### DIFF
--- a/lib/active_admin/resource/belongs_to.rb
+++ b/lib/active_admin/resource/belongs_to.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require_relative "../resource"
 
 module ActiveAdmin
   class Resource


lib/active_admin/resource/belongs_to.rb
```diff
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require_relative "../resource"
```
tagliala, line 2: is it possible that this is a breaking change? Like some use cases where `resource/belongs_to` is required individually without other resources?

javierjulio: I don't know. If the warning appears in normal usage of ActiveAdmin v4 and v3 (meaning a real app, not the test/dev app we generate for the project), we should address it in both versions. If we can see it stable in a real app on v3 then I think it's safe.

javierjulio: In commit https://github.com/activeadmin/activeadmin/commit/79c1a980850275c91af1bbe3570062ad8ffa7a25 is where the circular require is introduced but doesn't seem to have any need for it so this removal should be safe. We can backport it as well.

javierjulio (approved):
Thanks!

PR #8600, opened 2025-01-04
Remove `frozen_string_literal` comment from `.arb`

### LINKED ISSUE #8597: Enable some cops to prevent warnings
- [x] Lint/AmbiguousOperator
  ```
  app/controllers/active_admin/resource_controller/data_access.rb:239: warning: `*' interpreted as argument prefix
  ```
- [x] Lint/AmbiguousRegexpLiteral
  ```
  lib/active_admin/resource/naming.rb:8: warning: ambiguous `/`; wrap regexp in parentheses or add a space after `/` operator
  ```
- [x] Lint/UselessAssignment
  ```
  spec/unit/namespace/register_resource_spec.rb:81: warning: assigned but unused variable - user
  ```

tagliala:
The suggestion to add the `# frozen_string_literal: true` comment at
the top of `.arb` files was a false positive for the following reasons:

- It produced a warning.
- It did not have any effect because `.arb` files are evaluated in a
  different context.

To make this work correctly, the comment should be added at the
framework level instead.

This change eliminates the warning when running specs with
`RUBYOPT="-w"`.

Closes #8597

References:
- rails/rails#43725
- rubocop/rubocop#13699

### DIFF
--- a/app/views/active_admin/page/index.html.arb
+++ b/app/views/active_admin/page/index.html.arb
@@ -1,4 +1,3 @@
-# frozen_string_literal: true
 div class: "main-content-container" do
   if page_presenter.block
     instance_exec(&page_presenter.block)
--- a/app/views/active_admin/resource/_form.html.arb
+++ b/app/views/active_admin/resource/_form.html.arb
@@ -1,4 +1,3 @@
-# frozen_string_literal: true
 div class: "main-content-container" do
   if page_presenter.block
     options = {
--- a/app/views/active_admin/resource/_form_default.html.arb
+++ b/app/views/active_admin/resource/_form_default.html.arb
@@ -1,4 +1,3 @@
-# frozen_string_literal: true
 options = {
   url: resource.persisted? ? resource_path(resource) : collection_path,
   as: active_admin_config.param_key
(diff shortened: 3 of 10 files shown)


javierjulio (approved):
Thanks! What's the reason that the `.arb` templates don't need the `# frozen_string_literal: true` comment?

tagliala:
> What's the reason that the .arb templates don't need the # frozen_string_literal: true comment? 

Ah, it does not have effect (tested on arbre) and the warning is legitimate

It has been highlighted here but I did not explicitly : https://github.com/activeadmin/arbre/commit/2bd6d871392641bbdc5fcf2f8955a0e972e7d927

I guess that `.arb` files are treated as string input and parsed somewhere else (like `.erb`) and so magic comments are ignored?

`.erb` is not affected because they are not included: https://github.com/rubocop/rubocop/commit/c7a0f0015e1395824f06371d21f8b893cc64110e

I'm not sure if this is a problem with rubocop itself, or it is left to the user to always exclude `.arb` from that cop

https://github.com/rubocop/rubocop/blob/ccacb30ba92f0b2ec081923875ba8565f56c90e0/config/default.yml#L4039

tagliala:
I'm going to split this so we can merge and fix

```
Lint/AmbiguousRegexpLiteral
Lint/UselessAssignment
```

and leave the decision open for `.arb` templates

tagliala:
rubocop/rubocop#13684

PR #8612, opened 2025-01-23
Update Ruby and Node dependencies

### LINKED ISSUE #8611: Specs started failing on master (tailwindcss@4.0.0)
### Expected behavior

Specs pass

### Actual behavior

```
yarn run v1.22.22
$ tailwindcss -i ./app/assets/stylesheets/application.tailwind.css -o ./app/assets/builds/application.css --minify
/bin/sh: 1: tailwindcss: not found
```

### How to reproduce

https://github.com/activeadmin/activeadmin/actions/runs/12925486188/job/36046546838#step:6:149

(it is not master, but changes are not relevant)

tagliala:
On Rails 7.0, lock concurrent-ruby to 1.3.4 to address rails/rails#54260

Rails 7.0 bundle has been updated with 3.3.x:

```
rvm use 3.3.6 do bash -c "BUNDLE_GEMFILE=gemfiles/rails_70/Gemfile bundle update"
```

Additionally, add a workaround to make sure that the test application
uses Tailwind v3

Close #8611

### DIFF
--- a/tasks/test_application.rb
+++ b/tasks/test_application.rb
@@ -23,6 +23,16 @@ module ActiveAdmin
           Kernel.system("yarn add @activeadmin/activeadmin")
           Kernel.system('npm pkg set scripts.build:css="tailwindcss -i ./app/assets/stylesheets/active_admin.css -o ./app/assets/builds/active_admin.css --minify -c tailwind-active_admin.config.js"')
           Kernel.system("yarn install")
+
+          # Temporary workaround: Downgrade Tailwind CSS to v3.
+          # The `css:install:tailwind` task installs Tailwind CSS v4 by default,
+          # which is suitable for new applications.
+          # Related issues:
+          #  - activeadmin/activeadmin#8611
+          #  - rails/cssbundling-rails#163
+          # TODO: Remove this workaround once Tailwind CSS v4 is supported.
+          Kernel.system('yarn upgrade "tailwindcss@^3.4.17"')
+
           Kernel.system("yarn build:css")
         end
       end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -111,7 +111,7 @@ GEM
       xpath (~> 3.2)
     childprocess (5.1.0)
       logger (~> 1.5)
-    concurrent-ruby (1.3.4)
+    concurrent-ruby (1.3.5)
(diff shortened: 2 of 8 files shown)


tasks/test_application.rb
```diff
@@ -23,6 +23,11 @@ def soft_generate
           Kernel.system("yarn add @activeadmin/activeadmin")
           Kernel.system('npm pkg set scripts.build:css="tailwindcss -i ./app/assets/stylesheets/active_admin.css -o ./app/assets/builds/active_admin.css --minify -c tailwind-active_admin.config.js"')
           Kernel.system("yarn install")
+
+          # Temporary workaround to downgrade Tailwind to v3. Ref: activeadmin/activeadmin#8611
+          # TODO: remove when supporting Tailwind v4
+          Kernel.system('yarn upgrade "tailwindcss@^3.4.17"')
```
tagliala, line 29: workaaround

javierjulio, line 29: I don't recall why this would be necessary if it should be using the lock file to determine the version which is already at v3. Latest failing builds were installing v4 for some reason. Sorry, not sure why this would necessary to fix the issue or why it even occurred.

javierjulio (approved):
Thanks!

tasks/test_application.rb
tagliala: probably there is a better way, I'll investigate

https://github.com/rails/tailwindcss-rails?tab=readme-ov-file#choosing-a-specific-version-of-tailwindcss

tagliala: Ref: https://github.com/rails/cssbundling-rails/issues/163

Relevant code: https://github.com/rails/cssbundling-rails/blob/123569431c0e303c1aca8b275b2d1bc3c68a3a63/lib/install/tailwind/install.rb#L9

I can't find a better workaround to this issue at the moment.

I've tried

```
# Temporary workaround to downgrade Tailwind to v3.
# Ref: activeadmin/activeadmin#8611, cssbundling-rails#163
# TODO: remove when supporting Tailwind v4
gsub_file "package.json", /"tailwindcss": "[^"]*"/, '"tailwindcss": "^3.4.17"'
```

in the rails template, but does not look either

tagliala:
Merging this to have a green build

PR #8613, opened 2025-01-26
Bump Formtastic to 5.0 and remove legacy code

tagliala:
This commit addresses TODO notes in the code related to Formtastic < 4
and sets the minimum required Formtastic version to >= 5.0

Confirmed defaults:
- https://github.com/formtastic/formtastic/blob/473f5cef34de5ffe8812a3b0b07f167a22f8c4b9/lib/formtastic/form_builder.rb#L46
- https://github.com/formtastic/formtastic/blob/473f5cef34de5ffe8812a3b0b07f167a22f8c4b9/lib/formtastic/form_builder.rb#L49

---

Should be also backported to v3

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,8 +4,8 @@ PATH
     activeadmin (4.0.0.beta15)
       arbre (~> 2.0)
       csv
-      formtastic (>= 3.1)
-      formtastic_i18n (>= 0.4)
+      formtastic (>= 5.0)
+      formtastic_i18n (>= 0.7)
       inherited_resources (~> 2.0)
       kaminari (>= 1.2.1)
       railties (>= 7.0)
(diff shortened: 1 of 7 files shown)


javierjulio (approved):
Thanks!

tagliala:
> since the minimum required Formtastic version is now 5.0

Sorry, I was tricked by `Gemfile` and I didn't check `gemspec`.

`master` still supports v3

```
  s.add_dependency "formtastic", ">= 3.1"
  s.add_dependency "formtastic_i18n", ">= 0.4"
```

Are you fine to bump the dependency to >= 5.0?

javierjulio:
@tagliala no worries at all. Right, we'd have to bump here. Yes, I think it would be best to do that bump now on v4. Thanks!

tagliala:
Done, merging

PR #8616, opened 2025-01-29
Fix: Ensure Unique Tab IDs for Non-Latin Titles

aalyahya:
Fix: Ensure Unique Tab IDs for Non-Latin Titles
---

**This PR fixes an issue that occurs when using non-Latin titles for tabs without specifying a custom id** 

## Issue:

When using non-Latin titles for tabs without specifying a custom id, the generated HTML assigns duplicate id values due to `String#parameterize` returning an empty string for non-Latin characters.

## Example:

```ruby
tabs do
  tab "عبدالله" do
    span "tab 1"
  end
  tab "خالد" do
    span "tab 2"
  end
end
```
 
**Generates:**
```html
<nav class="tabs-nav" role="tablist" data-tabs-toggle="#tabs-container-32328">
  <a data-tabs-target="#tabs--32328" role="tab" aria-controls="tabs--32328" href="#" class="text-blue-600 hover:text-blue-600 dark:text-blue-500 dark:hover:text-blue-500 border-blue-600 dark:border-blue-500" aria-selected="true">عبدالله</a>
  <a data-tabs-target="#tabs--32328" role="tab" aria-controls="tabs--32328" href="#" class="dark:border-transparent text-gray-500 hover:text-gray-600 dark:text-gray-400 border-gray-100 hover:border-gray-300 dark:border-gray-700 dark:hover:text-gray-300" aria-selected="false">خالد</a>
</nav>

```

Since both tabs receive the same id, tab navigation breaks.

## Cause

string.parameterize ignores non-Latin characters and returns an empty string:
```ruby
string = "عبدالله"
string.parameterize #=> ""
```

## Fix:

If parameterize returns an empty string, we now generate a random string using `SecureRandom.hex.first(8)` to ensure uniqueness. A new instance variable, `@fragments`, tracks generated IDs to maintain reference consistency.

Changes
	•	Ensured unique tab id generation when parameterize results in an empty string.
	•	Introduced @fragments to store generated IDs.
	•	Added a test case to verify correctness.

### DIFF
--- a/lib/active_admin/views/components/tabs.rb
+++ b/lib/active_admin/views/components/tabs.rb
@@ -15,6 +15,7 @@ module ActiveAdmin
         add_class "tabs"
         @menu = nav(class: "tabs-nav", role: "tablist", "data-tabs-toggle": "#tabs-container-#{object_id}")
         @tabs_content = div(class: "tabs-content", id: "tabs-container-#{object_id}")
+        @fragments = {}
       end
 
       def build_menu_item(title, options, &block)
@@ -33,7 +34,7 @@ module ActiveAdmin
       private
 
       def fragmentize(string)
-        "tabs-#{string.parameterize}-#{object_id}"
+        @fragments[string] ||= "tabs-#{string.parameterize.presence || SecureRandom.hex.first(8)}-#{object_id}"
       end
     end
   end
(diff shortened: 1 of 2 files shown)


javierjulio:
@aalyahya thank you. Sorry, I won't be accepting this change. I've decided for months now that I will remove the tabs component in another v4 beta release as that was my intention. The only component that we don't use directly that will remain is panel. When the tabs component is removed, I will make sure to update the UPGRADING guide to call out that breaking change.

PR #8639, opened 2025-02-26
Support sortable argument in id_column

jaynetics:
This is mostly useful to prevent sorting on tables with uuid PKs.

Sorting by uuid isn't very useful, and it can have problematic performance if the table is very large, even if there is an index.

### DIFF
--- a/spec/unit/views/components/index_table_for_spec.rb
+++ b/spec/unit/views/components/index_table_for_spec.rb
@@ -15,7 +15,8 @@ RSpec.describe ActiveAdmin::Views::IndexAsTable::IndexTableFor do
     let(:assigns) do
       {
         collection: collection,
-        active_admin_config: active_admin_config
+        active_admin_config: active_admin_config,
+        resource_class: User,
       }
     end
     let(:helpers) { mock_action_view }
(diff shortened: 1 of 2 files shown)


spec/unit/views/components/index_table_for_spec.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 require "rails_helper"
+require "active_admin/batch_actions/resource_extension"
+require "active_admin/batch_actions/views/selection_cells"
```
jaynetics, line 4: These requires were needed to be able to run this spec individually. Running it individually was already prevented by missing constant errors before any of the other changes I made. I'll remove them if you don't care about this.

javierjulio: Thank you. Yes, please remove them. Preferred to match other specs and to preload for parallel test suite run.

spec/unit/views/components/index_table_for_spec.rb
```diff
@@ -43,5 +46,35 @@
         end
       end
     end
+
+    context "when creating an id column" do
+      before { allow(helpers).to receive(:url_target) { 'routing_stub' } }
```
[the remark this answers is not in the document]
javierjulio: Sorry, I'm not sure. I'll see if others have a suggestion here. I'm grateful it's something simple you can workaround for a good test. If anything we can revisit. For now the change is worthwhile and tested so thank you.

javierjulio (approved):
Thanks! Only hold up here is to remove the require statements in the spec. I'll merge and get a new v4 beta release out sometime after.

javierjulio (approved):
Thanks!

PR #8641, opened 2025-02-28
Support title for id_column

mgrunberg:
I'm building the index page of a resource. I must display the resource id as the first column but with a different title than "Id".

Using I18n to affect `id`'s `human_attribute_name` is not a valid solution because that's a global change. The only workaround is to write the full column myself.

This PR adds support to customize `id_column` title.

It's inspired by https://github.com/activeadmin/activeadmin/pull/8639


lib/active_admin/views/index_as_table.rb
```diff
@@ -256,9 +256,9 @@ def selectable_column(**options)
         end
 
         # Display a column for the id
-        def id_column
+        def id_column(title: resource_class.human_attribute_name(resource_class.primary_key))
```
javierjulio, line 259: I'm wondering if this should have a similar signature to the `column` method for consistency where the first argument is the title (no kwarg) if given, otherwise if not then it's treated as the options hash (keyword arguments) instead.

https://github.com/activeadmin/activeadmin/blob/bb1bade56d7da250f1fb3ae66a69a885a42e98e9/lib/active_admin/views/components/table_for.rb#L34-L37

javierjulio (commented):
This is a good improvement, thank you. My only concern is around the method signature but no issue in getting this in.

lib/active_admin/views/index_as_table.rb
mgrunberg: You are right. It makes sense to match the column signature. Users likely try that signature.

javierjulio (approved):
Thanks!

PR #8649, opened 2025-03-05
Tailwind v4 css configuration

mgrunberg:
Built on top of https://github.com/activeadmin/activeadmin/pull/8640

Exploring an idea to move from javascript config to CSS config.

Pros: Use the latest tailwind configuration mechanism (clean IMO).
Cons: There won't be a single way to configure tailwind.


hasghari:
@mgrunberg Would the configuration be simplified if the view files were published in the NPM package, similar to what #8220 was trying to achieve? I think that would remove the need for the `#ACTIVE_ADMIN_GEM#` placeholder in the config and allow for referencing the plugin and view files using a relative path.

mgrunberg:
> @mgrunberg Would the configuration be simplified if the view files were published in the NPM package, similar to what #8220 was trying to achieve? I think that would remove the need for the `#ACTIVE_ADMIN_GEM#` placeholder in the config and allow for referencing the plugin and view files using a relative path.

@hasghari Thanks for the contribution, but I'm not ready to reopen the #8220 discussion. That's why I'm still referencing gem files. 

Speaking about simplicity, I do have the idea of reducing the usage of `#ACTIVE_ADMIN_GEM`, but I'm not finding time to come back to the PR.

PR #8656, opened 2025-03-12
Update vendored jquery-ui from 1.13.3 to 1.14.1

tobischo:
JavaScript files have been manually updated because:
- newer version of jquery-ui is available with version 1.14.1
- jquery-ui 1.13 is not officially supported anymore

Ref:
- https://blog.jqueryui.com/2024/10/jquery-ui-1-14-1-released/
- https://endoflife.date/jquery-ui

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Update vendored jQuery UI to 1.14.1 [#8656] by [@tobischo]
+
 ## 3.3.0 [☰](https://github.com/activeadmin/activeadmin/compare/v3.2.5..v3.3.0)
 
 ### Enhancements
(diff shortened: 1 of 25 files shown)


Dwi-ght:
https://github.com/IDouble

tagliala:
Thanks for this PR, because of security concerns on the vendored assets, I would prefer to do this by myself and manually test all the pages that may be affected by this change

Ref: #8329

tobischo:
Sure, would be happy about that.

I had to prepare a change for our system anyway until it is fixed here, as our infosec team was complaining when they spotted that there is a version of something that is not officially supported anymore. 😅
So I figured I could make it available and save you the trouble.
I certainly do not have the full picture of what might break to ensure that it is covered.
I can tell you that all our use cases are working and we were confident enough to out it into production.

tagliala:
Superseded by #8670

PR #8676, opened 2025-04-05
Add pt-BR translation for truncate

annaluizacaraciolo:


tagliala:
Thanks!

I think there is an issue that we should fix in English and other languages before because now pagination keys are into `active_admin` scope

tagliala:
Hi @annaluizacaraciolo, thank you for your PR!

It seems that the changes you proposed may no longer be necessary after the updates made in #8678.

However, if you would like to update the entire pt-BR translation file, please feel free to do so either in this PR or in a new one.

Thanks again for your contribution!

PR #8678, opened 2025-04-05
Fix pagination truncate translation key

tagliala:
- Add the missing key to up-to-date language files
- Fix I18n tasks configuration for unscoped kaminari keys, which are now removed

Ref: #8218


javierjulio (approved):
Thanks!

PR #8680, opened 2025-04-08
Adds padding to button unordered lists

natewallis:
This tiny PR adds some padding to the unordered lists that represent a buttons children. 

Having recently upgraded to the latest version, I noticed that when there is a large amount of items in the left hand navigation it becomes hard to get a visual lock on which item on the left hand menu is selected.  Without rewriting all the colors you use, I think this small padding tweak gets around the majority of the issue.

Before:

image

After:

image


tagliala:
Hello, 

thanks for this PR

IIRC it is a choice to have all the entries without padding, but let's keep this open for review

Refers to: #8225

natewallis:
> Hello,
> 
> thanks for this PR
> 
> IIRC it is a choice to have all the entries without padding, but let's keep this open for review
> 
> Refers to: #8225

Thanks, TBH I am relatively new to activeadmin, my understanding was the padding I was adding in my PR was for the new BETA UI with the side navigation, the PR you are referencing seems older, would that be for the previous release UI with the navigation across the top of the screen?

javierjulio:
@natewallis thank you. This was intentional as the partials are meant to be customized to suit your needs. The preference is to keep this minimal for that case. The linked PR is related. If we want to apply some more styling, we'll address it there once we get to it.

PR #8691, opened 2025-04-23
Fix bug report issue template

tagliala:
This commit adjusts the bug report template to ensure it works as
intended.

While using a direct link like `issues/new?template=bug-report.md`
can open the template in full-screen mode, this approach results in
the template appearing twice in the issue creation list.

Unfortunately, there is no available option in GitHub to hide the bug
report template from the list while still allowing it to be accessible
via a direct link.

Given this limitation, the current configuration is the best achievable
under these constraints.

Close #8690

Testable here: https://github.com/tagliala/activeadmin

image

image


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,38 @@
+---
+name: Bug report
+about: If you've already asked for help with a problem and/or confirmed something is broken with ActiveAdmin, create a bug report.
+title: ''
+labels: ''
+assignees: ''
```
tagliala, line 6: any suggestion for title prefix, labels?

tagliala:
I relly don't like the fact that it opens by default in a modal. I lost my bug reports several times because of a misclick or an unintended navigation. It would be good if there is an option to open in full screen directly

tagliala:
Wait, let me test something

.github/ISSUE_TEMPLATE/bug-report.md
vfonic: Prefix with `[Bug] ` ([Bug]_ with one extra space) ?

tagliala: I was used to that, but at the moment they seem redundant for this repository because:
- It does not allow feature requests
- There is a `bug` label

Also, I'm slightly against `bug` alone, because it looks like an acknowledged bug, but it may be also another problem. I use to pair `bug` with `need triage`

No clear opinion at the moment, I would leave this to other members to decide. I'll ask copilot too

javierjulio: For now I'd prefer this stays as is as its consistent with what it was initially I believe, thanks.

PR #8700, opened 2025-05-01
Improve Dependabot config

tagliala:
- Remove ignore rule for sqlite3 for Rails >= 7.1
- Remove `rails_61` group, which is not present anymore and is returning an error from GitHub actions
- Add `rails_71` and `rails_72` groups
- Add constraints for Rails versions to prevent pull requests like #8696

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,9 +17,6 @@ updates:
       rails_default:
         patterns:
           - "*"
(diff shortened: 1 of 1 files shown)


javierjulio (approved):
Thanks! This looks great. I didn't realize we still have rails_61 reference still around. Thanks again for doing this.

PR #8707, opened 2025-05-01
Add I18n scope for Ransack predicate translations

tagliala:
This change introduces a custom `active_admin` I18n scope for predicate translations, allowing ActiveAdmin to override Ransack's defaults while still falling back to them when necessary.

It addresses discussions about improving localization (#8699) and resolves potential leftover logic from earlier implementations (#8010)

### DIFF
--- a/lib/active_admin/filters/active_filter.rb
+++ b/lib/active_admin/filters/active_filter.rb
@@ -5,6 +5,10 @@ module ActiveAdmin
     class ActiveFilter
       attr_reader :resource, :condition, :related_class
 
+      def self.predicate_name(name)
+        I18n.t("active_admin.ransack.predicates.#{name}", default: Ransack::Translate.predicate(name))
+      end
+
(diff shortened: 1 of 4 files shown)


tagliala:
I want to abstract the code to a single method before

tagliala:
@javierjulio Following our discussion about predicate translations: I'm still not sure this is the best idea.

My thinking is: if ActiveAdmin cannot define its own custom predicates, because they will leave in Ransack's root predicates scope, then the translations belong to Ransack, not to ActiveAdmin.

It seems that there are no clear real-world examples where this feature is needed, and developers haven't reported problems with using the main translation files for predicates. This makes me think the current way works okay, and we can simplify the code.

If in the feature a real-world scenario will require this approach, we can use this commit

javierjulio:
@tagliala that's ok with me to disregard this change. Thank you again for looking into it.

PR #8710, opened 2025-05-06
skip registering ActiveAdmin::Comment if namespace.comments is false

### LINKED ISSUE #5117: Bug: comment routes appear even when comments are disabled

FeLvi-zzz:
resolve #5117
resolve #5311

## background
we disable comments, but routes are still drawn.

## what to do

- skip registering ActiveAdmin::Comment if comments is disabled
  - this prevents drawing routes for comments
  - something for model (scope, before_save) are unneeded as well

### DIFF
--- a/lib/active_admin/orm/active_record/comments.rb
+++ b/lib/active_admin/orm/active_record/comments.rb
@@ -30,6 +30,8 @@ ActiveAdmin.autoload :Comment, "active_admin/orm/active_record/comments/comment"
 # Walk through all the loaded namespaces after they're loaded
 ActiveAdmin.after_load do |app|
   app.namespaces.each do |namespace|
+    next unless namespace.comments
+
     namespace.register ActiveAdmin::Comment, as: namespace.comments_registration_name do
       actions :index, :show, :create, :destroy
 


tagliala:
Hello,

thanks for this PR

Any chance to add a spec to cover this scenario?

FeLvi-zzz:
@tagliala 
Hello,

yes, but which file should I add a spec into?
I couldn't find an appropriate one, it looks like there is nothing testing like `namespace.register ActiveAdmin::Comment`.
I would appreciate if you could let me know where to add.

tagliala:
This looks like a `routes` spec. There is one under `resource`

Asking @javierjulio for confirmation

javierjulio:
@tagliala I'm not sure actually. @FeLvi-zzz let's give that a try please and we'll go from there.

FeLvi-zzz:
@tagliala @javierjulio 
I tried to add a spec in `spec/unit/comments_spec.rb`.
What do you two think about this?

javierjulio:
@mgrunberg @tagliala what do you think of the this approach and the tests? I think it would help to have as if the feature is not used then the routes should not exist and this seems to handle the case where it's only enabled for a specific namespace or resource. Any improvements or concerns?

tagliala:
Sorry, I'm not familiar with the comments implementation.

I can see there is a global switch and a per-resource switch, so this is why it is not checking for `config.comments = false`.

I also assume that the global config can be override at namespace / model level, so you can have `comments false` at global level, except for models or namespaces which explicitly set comments to true

I do not have suggestions for the specs

mgrunberg:
I'm not very familiar with the comments code either, but in a quick review, I'm ok with the approach and specs. 

I'm just wondering if it's possible to reduce the usage of stubs/doubles. It's not a strong opinion. It's just that the spec has Namespace & Resource creation examples. I'm aware that we are trying to call a hook and the setup for that could not be easy.

FeLvi-zzz:
I reduced excessive stubs (thanks to @mgrunberg !).

Could anyone run tests?

FeLvi-zzz:
Oops, tests which I created look flaky...
I'll fix them 🙇

tagliala:
Hello, this is still failing. Can you also please rabse on master?

spec/unit/comments_spec.rb
```diff
@@ -210,4 +210,47 @@
...
+      ActiveSupport::Notifications.instrument ActiveAdmin::Application::BeforeLoadEvent, { active_admin_application: application }
+      ActiveSupport::Notifications.instrument ActiveAdmin::Application::AfterLoadEvent, { active_admin_application: application }
+    end
+
+    context "when namespace.comments is true" do
+      let(:namespace_name) { :admin_with_comments }
+      let(:enable_comments) { true }
+
+      it "registers ActiveAdmin::Comment" do
+        expect(defined?(AdminWithComments::CommentsController)).to be_truthy
```
tagliala, line 234: is there a possibility that `defined?(AdminWithComments::CommentsController)` returns something different than true or false?

I would like to understand why this is checking for `be_truthy` instead of `be true`

Same for other similar specs

spec/unit/comments_spec.rb
```diff
@@ -210,4 +210,47 @@
...
+
+    before do
+      namespace.comments = enable_comments
+      resource = ActiveAdmin::Resource.new(namespace, Post)
+      resource.comments = enable_resource_comment
+      namespace.resources.add(resource)
+
+      # Manually trigger the before_load, after_load event
+      ActiveSupport::Notifications.instrument ActiveAdmin::Application::BeforeLoadEvent, { active_admin_application: application }
+      ActiveSupport::Notifications.instrument ActiveAdmin::Application::AfterLoadEvent, { active_admin_application: application }
```
tagliala, line 226: can't review here, I can'tell if this has side effects

tagliala (changes requested):

spec/unit/comments_spec.rb
FeLvi-zzz: `defined?` returns nil or string, like `"constants"`.
https://docs.ruby-lang.org/ja/latest/doc/spec=2fdef.html#defined

I should have use `be_nil` or something instead of ambiguous one. Let me fix it.

spec/unit/comments_spec.rb
FeLvi-zzz: I don't have any other ideas how to test this behavior without too many stubs, do you have the other better one?
If there is no way, many stubs or less ones, which is better?

ref. https://github.com/activeadmin/activeadmin/pull/8710#issuecomment-3402135450

spec/unit/comments_spec.rb
tagliala: Thanks for the clarification, I was wrongly assuming that all predicates return booleans, but I was probably misleaded by some recent cops enforcing this behavior.

Yes, `be_nil` maybe better, but in this case `be_truthy` is fine for me

PR #8712, opened 2025-05-06
Fix typo in batch_actions locale translation key

### LINKED ISSUE #8711: Translation key misspelling: "succesfully_destroyed" missing an "s"
I've noticed a small spelling issue in the codebase while implementing translations for my project.
When performing batch delete actions, the system is looking for the translation key `succesfully_destroyed` (with one "s") rather than `successfully_destroyed` (with two "s"s).

I understand this is a minor issue, but I thought it might be helpful to report it 😄 

### Expected behavior

When performing batch delete actions, the system should look for the translation key `successfully_destroyed` (with two "s"s) in the locale files.

### Actual behavior

When performing batch delete actions, the system is looking for the translation key `succesfully_destroyed` (with one "s"), which could result in a "Translation missing" error when the correctly spelled key is provided.

### How to reproduce

1. Add translations for `batch_actions.successfully_destroyed` but not for `batch_actions.succesfully_destroyed`
2. Perform a batch delete operation
3. Observe the "Translation missing: xx.active_admin.batch_actions.succesfully_destroyed" error

tagliala:
Close #8711

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,13 +1,15 @@
 # https://github.com/crate-ci/typos#false-positives
 [default]
-extend-ignore-identifiers-re = [
-  "succesfully_destroyed"
-]
 
 [default.extend-identifiers]
 
 [default.extend-words]
 rememberable = "rememberable"
+
+[type.md]
+extend-ignore-identifiers-re = [
+  "succesfully_destroyed"
+]
 
 [files]
 extend-exclude = [
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -225,3 +225,4 @@ This release includes several locale changes. Please [review the en.yml locale](
 - The `comments.delete_confirmation` text has been fixed to use singular form.
 - Inconsistent use of login/sign-in related terms so text now uses "Sign in", Sign out", and "Sign up" throughout.
 - The `toggle_dark_mode`, `toggle_main_navigation_menu`, `toggle_section`, and `toggle_user_menu` keys have been added.
+- The `batch_actions.succesfully_destroyed` key has been renamed to `batch_actions.successfully_destroyed` to fix a typo.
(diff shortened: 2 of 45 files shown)


javierjulio (commented):
@tagliala thank you. We just need to update the locale section in the Upgrading guide to call out this typo correction.

javierjulio (commented):
@tagliala thanks. I've added the typos workflow in #8713 so if you rebase here it will now run. You'll need to include the following updates.

Please **delete this line** since it's was temporary to get #8713 to pass: https://github.com/activeadmin/activeadmin/blob/0ef1d6ac50df7362a551e01bfccc788a08fec85e/_typos.toml#L4

Please **add the following before** the `[files]` line, to ignore the typo in the Upgrading guide:

```
[type.md]
extend-exclude = [
  "!UPGRADING.md"
]
extend-ignore-identifiers-re = [
  "succesfully_destroyed"
]
```

If you'd like to run the CLI locally, run `brew install typos-cli` and then `typos` at the root level (no arguments needed).

_typos.toml
```diff
@@ -1,14 +1,19 @@
...
 
 [default.extend-identifiers]
 
 [default.extend-words]
 rememberable = "rememberable"
 
+[type.md]
+extend-exclude = [
+  "!UPGRADING.md"
+]
```
tagliala, line 12: is this really needed?

javierjulio: Hmm I don't recall now if I tested it after adding or not. My understanding of the config was I can add a specific (Markdown) type check with its own settings but we only want to ignore the typo in this single file since it's intentional. I don't believe the config has a way to support that so I guess my thought process was **to exclude anything that isn't the UPGRADING.md file**. Feel free to update and test it locally if you want. I do know the other configurations will still report typos in any of EN based configs (config/locales/en*.yml) so we should have decent coverage there.

javierjulio: Hmm I didn't want to ignore the UPGRADING.md file but that may be easiest for now. We can remove the `[type.md]` block then and just update `[files]` list to include the upgrading guide.

javierjulio (approved):
Thanks!

PR #8713, opened 2025-05-06
Add typos workflow

javierjulio:
This adds a workflow to [run typos CLI using Reviewdog](https://github.com/reviewdog/action-typos) and includes a `_typos.toml` configuration to address false positives. The "succesfully_destroyed" typo is ignored here so the workflow will go green as the typo correction will be addressed in #8712.

### DIFF
--- a/docs/4-csv-format.md
+++ b/docs/4-csv-format.md
@@ -15,7 +15,7 @@ ActiveAdmin.register Post do
   csv do
     column :title
     column(:author) { |post| post.author.full_name }
-    column('bODY', humanize_name: false) # preserve case
+    column('body', humanize_name: false) # preserves case of column title
   end
 end
 ```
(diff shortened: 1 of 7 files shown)


docs/4-csv-format.md
```diff
@@ -15,7 +15,7 @@ ActiveAdmin.register Post do
   csv do
     column :title
     column(:author) { |post| post.author.full_name }
-    column('bODY', humanize_name: false) # preserve case
+    column('body', humanize_name: false) # preserves case of column title
```
mgrunberg, line 18: @javierjulio, this was a fair use of 'bODY'. The "typo" + the comment shows that you can customize the title the way you need. The new comment looks good, I'm just talking about leaving 'bODY' as an exception of the new flow.

image

javierjulio: @mgrunberg yes, I know but I was cheating because its all lowercase and by default will capitalize so I think that was sufficient. Let me look into another update to ignore this specific word in only this file.

PR #8714, opened 2025-05-06
Refactor ActiveFilter's predicate name translation

tagliala:
Simplify `ActiveFilter#predicate_name` by leveraging `Ransack::Translate.predicate`, removing redundant implementation logic.

Close #8699


javierjulio (approved):
Thanks!

Issue #8721, opened 2025-05-13
v4: pagination fails when setting page size


johnf:
### Expected behavior

No 500 error when setting page size from dropdown.

### Actual behavior

```
app-1                | 00:56:38 web.1  | Completed 500 Internal Server Error in 140ms (ActiveRecord: 14.1ms (9 queries, 0 cached) | GC: 1.2ms)
app-1                | 00:56:38 web.1  |
app-1                | 00:56:38 web.1  |
app-1                | 00:56:38 web.1  |
app-1                | 00:56:38 web.1  | ActionView::Template::Error (Cannot convert Integer value: 50)
app-1                | 00:56:38 web.1  | Caused by: TypeError (Cannot convert Integer value: 50)
app-1                | 00:56:38 web.1  |
app-1                | 00:56:38 web.1  | Information for: ActionView::Template::Error (Cannot convert Integer value: 50):
app-1                | 00:56:38 web.1  |     1: # frozen_string_literal: true
app-1                | 00:56:38 web.1  |     2: if section.block
app-1                | 00:56:38 web.1  |     3:   result = instance_exec(&section.block)
app-1                | 00:56:38 web.1  |     4:   text_node result unless result.is_a?(Arbre::Element)
app-1                | 00:56:38 web.1  |     5: else
app-1                | 00:56:38 web.1  |     6:   render(section.partial_name)
app-1                | 00:56:38 web.1  |     7: end
app-1                | 00:56:38 web.1  |
app-1                | 00:56:38 web.1  | activeadmin (4.0.0.beta15) app/helpers/active_admin/form_helper.rb:57:in `block in fields_for_params'
app-1                | 00:56:38 web.1  | activeadmin (4.0.0.beta15) app/helpers/active_admin/form_helper.rb:35:in `each'
app-1                | 00:56:38 web.1  | activeadmin (4.0.0.beta15) app/helpers/active_admin/form_helper.rb:35:in `flat_map'
app-1                | 00:56:38 web.1  | activeadmin (4.0.0.beta15) app/helpers/active_admin/form_helper.rb:35:in `fields_for_params'
app-1                | 00:56:38 web.1  | activeadmin (4.0.0.beta15) app/helpers/active_admin/form_helper.rb:14:in `hidden_field_tags_for'
app-1                | 00:56:38 web.1  | activeadmin (4.0.0.beta15) app/helpers/active_admin/form_helper.rb:69:in `block in active_admin_filters_form_for'
[...]

PR #8722, opened 2025-05-13
fix error when per page is specified (fixes: #8721)

johnf:


javierjulio:
I don't believe this is necessary since the impacted method was the same on v3. The drop down does render successfully on the demo app. We'll need a confirmed reproduction before making any changes here. Thank you.

PR #8725, opened 2025-05-21
Update dependencies

tagliala:
Additionally:
- Relax eslint dependency constraint since https://github.com/eslint/eslint/issues/19682 has been
  addressed
- Explicitly require `erb` 4 for Rails 7.1 and 7.2 because of
  https://github.com/ruby/rdoc/pull/1368


javierjulio (approved):
Thanks!

PR #8744, opened 2025-06-17
Update 9-batch-actions.md to reflect new form strategy. 

maxkaplan:
Hey all, I noticed the docs did not show a valid way to make a batch action modal form. It still shows the old way that used the 'form' param.

I replaced the old `form:` param content with the upgraded v4 example from : https://my.diffend.io/gems/activeadmin/4.0.0.beta2/4.0.0.beta3

### DIFF
--- a/docs/9-batch-actions.md
+++ b/docs/9-batch-actions.md
@@ -117,57 +117,43 @@ ActiveAdmin.register Post do
 end
 ```
 
-### Batch Action forms
+### Batch Action Forms
 
 If you want to capture input from the user as they perform a batch action,
 Active Admin has just the thing for you:
 
-```ruby
-batch_action :flag, form: {
-  type: %w[Offensive Spam Other],
-  reason: :text,
-  notes:  :textarea,
-  hide:   :checkbox,
-  date:   :datepicker
-} do |ids, inputs|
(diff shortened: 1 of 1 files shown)


tagliala:
Hello, at the moment the documentation covers 3.x series and changes needed for v4 are in the [UPGRADING.md](https://github.com/activeadmin/activeadmin/blob/master/UPGRADING.md) guide

Does the new change also work in v3?

maxkaplan:
Hello! Ok my mistake, I guess I was confused by the docs UI, I thought the top nav was letting me know these are the docs for version 4.

I have not tested if this approach works in v3.

Again, sorry about my confusion.
I have been using ActiveAdmin for years and really appreciate all the work and wanted to try and help out in some meager way.
Best,

image

tagliala:
yes, the documentation at the moment is confusing and there is no version switch

So there is still no official documentation for v4, I don't know the plans for that

javierjulio:
As noted we still have to address the documentation since we want to keep the v3 docs. We will need help on that and are open to ideas. In the meantime, don't worry as this was already noted in the Upgrading guide. We also have a feature branch with documentation updates for v4 but we need to address the issue around how to keep v3/legacy docs. Thanks.

PR #8760, opened 2025-07-22
Drop Ruby 3.1 and update dependencies

tagliala:
Changes:
- Drop support for Ruby 3.1.
- Update Ruby, Bundler, and Node dependencies
  across the project.
- Remove dependencies and code paths specific to
  Ruby 3.1.

Reasons for dropping Ruby 3.1:
- Ruby 3.1 reached end-of-life 3 months ago.
- Rails 7.0 supports Ruby 3.2 and 3.3, making 3.1
  unnecessary.
- Multiple instances of `FIXME` related to legacy
  Ruby support can be dropped.
- Active Admin 4 stable release does not look.
  imminent, so maintaining deprecated Ruby
  versions is not practical.
- There are failures with the new Bundler 2
  (>= 2.7.0) when used with Ruby 3.1.


tagliala:
```
     Failure/Error: expect(current_lockfile).to eq(new_lockfile), msg

       Please update /home/runner/work/activeadmin/activeadmin/gemfiles/rails_72/Gemfile's lock file with `BUNDLE_GEMFILE=/home/runner/work/activeadmin/activeadmin/gemfiles/rails_72/Gemfile bundle install` and commit the result
       Diff:
       @@ -449,4 +449,4 @@
          zeitwerk (~> 2.6.18)
        
        BUNDLED WITH
       -   2.7.1
       +   2.6.9
     # ./spec/tasks/gemfile_spec.rb:13:in `block (2 levels) in <top (required)>'
```

Ok I have an idea... ⏳

tagliala:
@deivid-rodriguez @javierjulio do you have any suggestion here?

I was thinking to:
1. A conditional `rubygems:` in the workflow; OR
2. Change the spec to ignore "bundled with" entry

I would go for the second option

Mentioning David because of the slightly related ruby/setup-ruby#780

javierjulio:
@tagliala is the test comparing the lock file from BUNDLE_GEMFILE to the root Gemfile.lock? I take it that's the issue since the root Gemfile.lock has the bundled-with version update whereas the BUNDLE_GEMFILE lock files do not. You are welcome to work around it if you want to keep the test. I'd prefer to just remove the test and/or drop EOL Rubies. For the former, we have `bin/bundle` to manage updating all the gemfiles so not too concerned for these getting out of sync.

tagliala:
I'm going to drop 3.1, reasons:

- we have several `  # FIXME: remove this dependency when Ruby 3.1 support will be dropped`
- 3.1 is in EOL since 3 months
- Rails 7.0 does support 3.2 and 3.3 
- Active Admin 4 stable release does not look imminent

tagliala:
codecov is not passing because it may be waiting for 3.1?

javierjulio:
@tagliala no it's not that. Sometimes the status checks get stuck and Codecov recommends re-uploading the reports so I reran the Upload Coverage job and that worked. It usually does the trick thankfully. I will approve now. Thanks!

javierjulio (approved):
Thanks!

PR #8762, opened 2025-07-22
Remove Tabs component

javierjulio:
The intention was always to remove the Tabs component in v4. With this removal, that means the last component that is not directly used by ActiveAdmin is Panel which will remain as is with no other enhancements. In another major release we may remove Panel. The objective is for users to own their own components, whether directly or from another third party library and to free up the name from the Arbre namespace so it can be used without conflict.


henrikbjorn:
@javierjulio have the mission or goal of activeadmin changed?

I have always seen AA being an easy way to create crud like admin pages (hence the name) with a lot of flexibility for those admin "needs". Like a dashboard for the frontpage, or resource overviews with sidebars and tabs etc.

Is that not the goal anymore? With more stuff moving to "bring your own".

javierjulio:
@henrikbjorn I don’t think so but other things have changed and improved since this library was created 15 years ago. There are better alternatives than what we can provide or should focus on. I think the Arbre builder namespace should be open for your own components and even for some HTML elements that were broken by AA like header and footer. That has been my goal with v4.

henrikbjorn:
> @henrikbjorn I don’t think so but other things have changed and improved since this library was created 15 years ago. There are better alternatives than what we can provide or should focus on. I think the Arbre builder namespace should be open for your own components and even for some HTML elements that were broken by AA like header and footer. That has been my goal with v4.

Then I don't get the removal of those instead of finding a way to introduce override of those components etc.

To add tabs into ActiveAdmin it now requires use to write a custom component, stimulus controller etc. for it (I assume)?

javierjulio:
@henrikbjorn that is correct. There are CSS only approaches too. This was always the intention with v4 as there are better alternatives than what was available 15 years ago. Sorry, that won’t change here. Thank you.

jaynetics:
if anyone has an alternative that is convenient to use, i'd be interested ...

henrikbjorn:
> if anyone has an alternative that is convenient to use, i'd be interested ...

To ActiveAdmin or Tabs ?

jaynetics:
Tabs 😅

We use them heavily. For now, just pasting the component back in, with inlined tailwind classes like [so](https://gist.github.com/jaynetics/bacdf6bafc5ac4e7d5fa3dd2f76c008a), works fine. I guess it would only break if the surrounding styles were changed significantly or flowbite was removed, neither of which seem super likely?

elthariel:
Sorry for necro posting, but I just discovered this issue by digging in the git log as, if I'm not mistaken, this was not in the changelog.

I personnaly don't really understand the motivation here. It doesn't look like it was an annoying piece of code to maintain ? Also, I'd love to own my own components, but it's not like it's documented anywhere, outside of a mention in custom index page. There's a dedicated Arbre doc page but it's really not clear how they're supposed to be loaded :sweat_smile: 

I'm sorry to be annoying, but the "always meant to be removed" is a bit rough. I'm using this component all over my admin and would have greatly enjoyed a deprecation warning. Maybe I missed it ?

Would you considering restoring the code in the repository in a documentation/example folder ? Figuring out how and why it vanished was a bit painful.

PR #8771, opened 2025-07-26
Enable Lint/RedundantStringCoercion cop

tagliala:
This cop does not introduce a performance gain, but it removes visual noise from the codebase.

String interpolation handles coercion automatically, so explicit `.to_s` calls are redundant and can be safely removed for better readability.


javierjulio (approved):
Thanks!

PR #8780, opened 2025-08-03
Fix issue CSRF protection not enabled

ptrgits:
https://github.com/activeadmin/activeadmin/blob/01f1430df290f117a3adcef9b772a61ec1bfe5b2/tasks/bug_report_template.rb#L63-L65
fix the CSRF vulnerability, add a call to `protect_from_forgery with: :exception` inside the `ApplicationController` class. This will ensure that Rails raises an exception if a request is made with an invalid or missing CSRF token, which is the recommended secure default. The change should be made directly inside the `ApplicationController` class definition in `tasks/bug_report_template.rb`, after the `include Rails.application.routes.url_helpers` line. No additional imports or dependencies are required, as this is standard Rails functionality.

### References
- [When Rails' protect_from_forgery Fails](https://www.veracode.com/blog/managing-appsec/when-rails-protectfromforgery-fails)
- [Cross-site request forgery](https://en.wikipedia.org/wiki/Cross-site_request_forgery)
- [Cross-site request forgery](https://owasp.org/www-community/attacks/csrf)
- [Cross-Site Request Forgery (CSRF)](https://guides.rubyonrails.org/security.html#cross-site-request-forgery-csrf)
- [CWE-352](https://cwe.mitre.org/data/definitions/352.html)

### DIFF
--- a/tasks/bug_report_template.rb
+++ b/tasks/bug_report_template.rb
@@ -62,6 +62,7 @@ end
 
 class ApplicationController < ActionController::Base
   include Rails.application.routes.url_helpers
+  protect_from_forgery with: :exception
 end
 
 class ApplicationRecord < ActiveRecord::Base


tagliala:
Thank you for submitting this PR!

I wanted to clarify that `tasks/bug_report_template.rb` is not a production file and it is not even included in production gem:

https://github.com/activeadmin/activeadmin/blob/01f1430df290f117a3adcef9b772a61ec1bfe5b2/activeadmin.gemspec#L17

It's only intended as a minimal template to help reproduce bugs, so there are no actual security concerns related to CSRF in this context.

Because of this, I recommend updating the commit message to better reflect the nature of the change.

I'm not opposed to enabling CSRF protection in the template, but I'm not entirely sure it's necessary here. It might be helpful to explain the motivation for making this change. Are you copying the template into production code, or is there another reason to include CSRF protection here?

tagliala:
No feedback, meanwhile, closing here

PR #8815, opened 2025-09-30
Preserve custom string labels in AttributesTable headers

### LINKED ISSUE #8813: Show Attributes Table Incorrectly Capitalizes Custom Row Labels/Titles
I think there is an issue where you're unable to override human_attribute_name (the `title` / `attr` for the row automatically gets titleized) to get the desired capitalization in show attributes_table. In **index**, column "Registration ID", :registration_id displays correctly as Registration ID, but in **show**, row("Registration ID"){|s| s.registration_id} displays incorrectly as "Registration Id" without the capital "D". I've tried several different syntax options, labels, changing the locale file, still won't capitalize correctly.

In the same vein, the display adds spaces between consecutive capital letters if there is more text that follows, ex row("Registration IDe Test"){|s| s.registration_id} displays as Registration I De Test

(PS I'm interested in taking a stab at this - fixing the inconsistency between how index and show treat titles - and am planning to put up a PR for it)

### How to reproduce

```
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  # Use `ACTIVE_ADMIN_PATH=. ruby tasks/bug_report_template.rb` to run
  # locally, otherwise run against the default branch.
  if ENV["ACTIVE_ADMIN_PATH"]
    gem "activeadmin", path: ENV["ACTIVE_ADMIN_PATH"], require: false
  else
    gem "activeadmin", github: "activeadmin/activeadmin", require: false
  end
  gem "rails", "~> 8.0.0"
  gem "sprockets", "~> 4.0"
  gem "importmap-rails", "~> 2.0"
  gem "sqlite3", force_ruby_platform: true, platform: :mri
  # Fixes an issue on CI with default gems when using inline bundle with default
  # gems that are already activated
  # Ref: rubygems/rubygems#6386
  if ENV["CI"]
    require "net/protocol"
    require "timeout"
    gem "net-protocol", Net::Protocol::VERSION
    gem "timeout", Timeout::VERSION
  end
end

require "active_record"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

  end
  
    t.string :full_name
    t.string :registration_id
    t.string :api_key
    t.string :gst_number
  end
end

require "action_controller/railtie"
require "action_view/railtie"
require "active_admin"

  config.root = __dir__
  config.hosts << ".example.com"
  config.secret_key_base = "secret_key_base"
  config.eager_load = false
  Rails.logger = config.logger
end

end

class ApplicationRecord < ActiveRecord::Base
  primary_abstract_class
  
  def self.ransackable_attributes(auth_object = nil)
    authorizable_ransackable_attributes
  end
  
  def self.ransackable_associations(auth_object = nil)
    authorizable_ransackable_associations
  end
end

class User < ApplicationRecord
end

ActiveAdmin.setup do |config|
end

Rails.application.initialize!

ActiveAdmin.register_page "Dashboard" do
  menu priority: 1, label: proc { I18n.t("active_admin.dashboard") }
  content do
    "Test Me"
  end
end

ActiveAdmin.register User do
  # Index view - this works correctly
  index do
    selectable_column
    id_column
    column "Registration ID", :registration_id  # This displays correctly as "Registration ID"
    column "API Key", :api_key                  # This displays correctly as "API Key"
    column "GST Number", :gst_number           # This displays correctly as "GST Number"
    actions
  end
  
  # Show view - this demonstrates the bug
  show do
    attributes_table do
      row :full_name
      row("Registration ID") { |user| user.registration_id }  # BUG: Displays as "Registration Id"
      row("API Key") { |user| user.api_key }                  # BUG: Displays as "Api Key"  
      row("GST Number") { |user| user.gst_number }            # BUG: Displays as "Gst Number"
      
      # For comparison - using symbols works with locale files but doesn't help with custom labels
      row :registration_id  # This could work with locale overrides
    end
  end
end

  ActiveAdmin.routes(self)
end

require "minitest/autorun"
require "rack/test"
require "rails/test_help"

  def setup
    @user = User.create!(
      full_name: "John Doe",
      registration_id: "REG123",
      api_key: "key_abc123",
      gst_number: "GST456789"
    )
  end
  
  def test_index_view_preserves_custom_column_labels
    get admin_users_url
    assert_response :success
    
    # Index view should show correct capitalization in column headers
    assert_match "Registration ID", response.body, "Index should preserve 'Registration ID' capitalization"
    assert_match "API Key", response.body, "Index should preserve 'API Key' capitalization"  
    assert_match "GST Number", response.body, "Index should preserve 'GST Number' capitalization"
    
    # Should NOT contain the incorrectly titleized versions
    refute_match "Registration Id", response.body, "Index should not show 'Registration Id'"
    refute_match "Api Key", response.body, "Index should not show 'Api Key'"
    refute_match "Gst Number", response.body, "Index should not show 'Gst Number'"
  end
  
  def test_show_view_titleizes_custom_row_labels_incorrectly
  get admin_user_url(@user)
  assert_response :success
  
  # This test SHOULD FAIL - demonstrating the bug
  # We want these assertions to pass (correct behavior):
  assert_match "Registration ID", response.body, "Show should preserve 'Registration ID' capitalization"
  assert_match "API Key", response.body, "Show should preserve 'API Key' capitalization"
  assert_match "GST Number", response.body, "Show should preserve 'GST Number' capitalization"
  
  # These should NOT be found (but currently they are):
  refute_match "Registration Id", response.body, "Show should not titleize to 'Registration Id'"
  refute_match "Api Key", response.body, "Show should not titleize to 'Api Key'"
  refute_match "Gst Number", response.body, "Show should not titleize to 'Gst Number'"
  
  puts "\n=== BUG DEMONSTRATION ==="
  puts "Expected: 'Registration ID', 'API Key', 'GST Number'"  
  puts "Actual:   'Registration Id', 'Api Key', 'Gst Number'"
  puts "The show view is incorrectly processing custom labels through titleize"
  puts "========================\n"
end

  private
  
  def app
    Rails.application
  end
end

```

dixonnataliej:
Resolves #8813 

Fixes custom string labels being titleized in `attributes_table`, which breaks acronym capitalization in show views.
The `header_content_for` method was processing all labels through titleize, regardless of whether they were symbols (attribute names) or strings (custom display labels). This caused issues like:

"Registration ID" → "Registration Id"
"LMnO" -> "L Mn O"

This PR makes `AttributesTable` consistent with `TableFor::Column.pretty_title` by distinguishing between:

Symbols (:registration_id) - processed through human_attribute_name with i18n lookup and titleize fallback
Strings ("Registration ID") - preserved exactly as provided

**Changes**

Modified `header_content_for` in `lib/active_admin/views/components/attributes_table.rb` to check attribute type before processing
Added regression test in `spec/unit/views/components/attributes_table_spec.rb`

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -158,6 +158,7 @@ See [PR #8727](https://github.com/activeadmin/activeadmin/pull/8727) for more de
   end
   ```
   </details>
+- Using `row 'name'` in an `attributes_table` block, now prints the header text as is. For a localized lookup of the header with titlized fallback, use `row :name` instead. This matches the `column 'name'` behavior of index tables (`TableFor`).
 
 #### Resource named methods
 


lib/active_admin/views/components/attributes_table.rb
```diff
@@ -61,10 +61,15 @@ def build_colgroups
       end
 
       def header_content_for(attr)
-        if @resource_class.respond_to?(:human_attribute_name)
-          @resource_class.human_attribute_name(attr, default: attr.to_s.titleize)
+        if attr.is_a?(Symbol)
```
tagliala, line 64: Thanks for this PR

This change will introduce a breaking change for this use case:

```rb
      row 'name'
```

This was a valid approach before to get the translated attribute name.

I personally prefer this fix to have a similar (not identical yet) behavior with `column` and I would not care of the breaking change.

/cc @javierjulio

javierjulio: @tagliala what do you suggest we do? I didn't understand from your last sentence what update you'd prefer. I'm fine with keeping this consistent and also what we think is best going forward.

tagliala: I'm not against this breaking change. This can be considered a bugfix for consistency with `column` (even if they are not 100% consistent). I would merge, but document this maybe in the upgrading guide

javierjulio: @tagliala ah I see. Your concern was with the release type label I added originally. I've updated it to be a breaking change as that is accurate.

@dixonnataliej can you please add the following to the end of the "Breaking Changes" section in the UPGRADING.md file?

```
- Using `row 'name'` in an `attributes_table` block, now prints the header text as is. For a localized lookup of the header with titlized fallback, use `row :name` instead. This matches the `column 'name'` behavior of index tables (`TableFor`).
```

Feel free to suggest any edits. Thank you.

PR #8821, opened 2025-10-02
Upgrade Tailwind to 3.4.18

tagliala:
Additionally, bump up ruby dependencies


tagliala:
> I'll rebase/merge https://github.com/activeadmin/activeadmin/pull/8809 and we'll cut another beta release.

Great

> I think we can do the Tailwind v4 upgrade next if that sounds good to you.

Yes, that would be good

PR #8837, opened 2025-10-22
Add Rails 8.1.0 compatibility

tagliala:
Update ActiveAdmin routing code to use keyword arguments
(`**options_hash`) for methods such as `namespace` and `get`, addressing
deprecation warnings introduced in Rails 8.1 and ensuring compatibility
with Rails 8.2+. Previously, hash options were passed as positional
arguments, which will be unsupported in future Rails releases.

This change affects route definition logic in `ActiveAdmin::Router`,
including namespace and custom page actions. All routing methods now
use keyword arguments for options, improving code clarity and
future-proofing against Rails upgrades.

Ref: https://github.com/rails/rails/pull/53689


tagliala:
```
Failures:

  1) Routing route_options with a custom path set in route_options should route using the custom path
     Failure/Error:
       router.namespace namespace.name, namespace.route_options.dup do
         router.root namespace.root_to_options.merge(to: namespace.root_to, as: :root)
       end
     
     ActiveSupport::DeprecationException:
       DEPRECATION WARNING: namespace received a hash argument path. Please use a keyword instead. Support to hash argument will be removed in Rails 8.2. (called from block in ActiveAdmin::Router#define_root_routes at /Users/geremia/dev/activeadmin/lib/active_admin/router.rb:24)
```

tagliala:
rails/rails#53689

rails/rails@a6d199e283

lib/active_admin/router.rb
```diff
@@ -21,7 +21,7 @@ def define_root_routes
         if namespace.root?
           router.root namespace.root_to_options.merge(to: namespace.root_to)
         else
-          router.namespace namespace.name, namespace.route_options.dup do
+          router.namespace namespace.name, **namespace.route_options.dup do
```
tagliala, line 24: production change.

`dup` may still be needed to avoid nested objects in route options

lib/active_admin/router.rb
```diff
@@ -65,7 +65,7 @@ def resource_routes(config)
 
     def page_routes(config)
       page = config.underscored_resource_name
-      router.get "/#{page}" => "#{page}#index"
+      router.get "/#{page}", to: "#{page}#index"
```
tagliala, line 68: production change.

lib/active_admin/router.rb
```diff
@@ -107,7 +107,7 @@ def define_belongs_to_routes(config)
     end
 
     def define_namespace(config)
-      router.namespace config.namespace.name, config.namespace.route_options.dup do
+      router.namespace config.namespace.name, **config.namespace.route_options.dup do
```
tagliala, line 110: production change.

`dup` may still be needed to avoid nested objects in route options

lib/active_admin/router.rb
```diff
@@ -91,8 +91,8 @@ def build_action(action)
       build_route(action.http_verb, action.name)
     end
 
-    def build_route(verbs, *args)
-      Array.wrap(verbs).each { |verb| router.send(verb, *args) }
+    def build_route(verbs, ...)
+      Array.wrap(verbs).each { |verb| router.send(verb, ...) }
```
tagliala, line 95: production change. this will allow to forward both `String`s and `Hash`es

tasks/test_application.rb
```diff
@@ -48,6 +48,7 @@ def generate
         --skip-active-storage
         --skip-bootsnap
         --skip-brakeman
+        --skip-bundler-audit
```
tagliala, line 51: new flag

tagliala:
Should this be backported to 3-stable ?

mgrunberg:
> Should this be backported to 3-stable ?

I would love to have it in v3. Do you have time to do the backport? If not I can work on that but next week.

tagliala:
3-0-stable requires Ruby >= 2.6

>  s.required_ruby_version = ">= 2.6"

I guess that this will not allow `(...)`

mgrunberg:
> 3-0-stable requires Ruby >= 2.6
> 
> >  s.required_ruby_version = ">= 2.6"
> 
> I guess that this will not allow `(...)`

I think that operator got introduced in ruby 2.7, right?

I asked @javierjulio this morning and it should be ok to drop ruby 2.6 support in v3 if that blocks Rails 8.1 usage.

We also talked about using conditional code based on ruby version but it may be hard to read in this case.

dalibor:
Looks like this was backported to [3-0-stable](https://github.com/activeadmin/activeadmin/pull/8839), documented in the [CHANGELOG](https://github.com/activeadmin/activeadmin/blob/3-0-stable/CHANGELOG.md#340-) and released as [3.4.0](https://rubygems.org/gems/activeadmin/versions/3.4.0)

Could you just add 3.4.0 on the Github [releases](https://github.com/activeadmin/activeadmin/releases) page? Thanks

tagliala:
Done, thanks for the heads-up

https://github.com/activeadmin/activeadmin/releases/tag/v3.4.0

PR #8840, opened 2025-10-25
Improve CI and RuboCop config

tagliala:


tagliala:
Rubocop is not run on CI on 3.0 as far as I can tell, because there was an unnoticed offence

PR #8845, opened 2025-11-02
Add missing dark text color in spec view

tagliala:
### Before

localhost_5000_admin_kitchensink (2)

### After

localhost_5000_admin_kitchensink (3)


javierjulio:
@tagliala thanks!

PR #8876, opened 2025-12-05
Fix custom namespace routing.

### LINKED ISSUE #8279: No access to routes helpers from ActiveAdmin views [v4.0.0.beta5]
Route helpers are not available within ActiveAdmin views, even `root_path` is not available in the context of view Class.

```
ActionController::UrlGenerationError - No route matches {:action=>"index", :controller=>"active_admin/users"}:
  app/views/active_admin/_sidebar.html.erb:3
  app/views/layouts/active_admin.html.erb:15
```

It seems to be related to the recent change of separating contexts, so these helpers are still available under `main_app` variable.

### Expected behavior

Route helpers are available.

### Actual behavior

Helpers are not available, ActiveAdmin v4 code throws exception for new and generated resource.

### How to reproduce

- initialise ActiveAdmin v4 project
- generate sample resource
- run the app, try accessing the resource

craigmcnamara:
ActiveAdmin 4.0's `isolate_namespace` caused route helpers to fail for custom namespaces (e.g., `:active_admin` instead of `:admin`) because `url_for` looked in the engine's
  routes instead of the main app's routes. This adds a proxy module that overrides `url_for` to route hash options through `Rails.application.routes`, and updates the base
  controller to merge request context for relative URL generation. The polymorphic route helpers also default to `only_path: true` to avoid host resolution errors.

Fixes #8279

### DIFF
--- a/app/controllers/active_admin/resource_controller/polymorphic_routes.rb
+++ b/app/controllers/active_admin/resource_controller/polymorphic_routes.rb
@@ -8,13 +8,19 @@ module ActiveAdmin
       def polymorphic_url(record_or_hash_or_array, options = {})
         options = options.dup
         options[:only_path] = true unless options.key?(:only_path) || options.key?(:host)
-        super(map_named_resources_for(record_or_hash_or_array), options)
+        resources = map_named_resources_for(record_or_hash_or_array)
+        # Prepend namespace symbol for polymorphic routing with custom namespaces
+        resources = [active_admin_namespace.name] + Array(resources) if active_admin_config
+        super(resources, options)
       end
 
       def polymorphic_path(record_or_hash_or_array, options = {})
         options = options.dup
         options[:only_path] = true unless options.key?(:only_path) || options.key?(:host)
-        super(map_named_resources_for(record_or_hash_or_array), options)
+        resources = map_named_resources_for(record_or_hash_or_array)
+        # Prepend namespace symbol for polymorphic routing with custom namespaces
+        resources = [active_admin_namespace.name] + Array(resources) if active_admin_config
+        super(resources, options)
       end
 
       private
--- a/lib/active_admin/batch_actions/resource_extension.rb
+++ b/lib/active_admin/batch_actions/resource_extension.rb
@@ -10,6 +10,7 @@ module ActiveAdmin
 
       # @return [Array] The set of batch actions for this resource
       def batch_actions
+        @batch_actions ||= {}
         batch_actions_enabled? ? @batch_actions.values.sort : []
       end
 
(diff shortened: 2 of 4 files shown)


tagliala:
Hi, thanks for this PR.

I'm not familiar with the underlying issue and I couldn't reproduce #8279. There isn’t a failing test demonstrating the bug in this PR.

Could you add a minimal failing test that reproduces the problem and show why it happens?

> This adds a proxy module that overrides `url_for` to route hash options through Rails.application.routes

At a glance, overriding `url_for` (plus the many condition branches to make it work) feels like a heavy-handed solution. Before accepting a global override of such a core helper, I'd like to understand the root cause better and see whether a smaller, more targeted change could fix it. Or, maybe this behavior should be documented as a current limitation.

@javierjulio and @mgrunberg any thought?

javierjulio:
@craigmcnamara thank you for attempting to address this. I noticed the test suite is failing for the comments resource still. 

@tagliala in general I don't know much about engines. The reason for why it fails makes sense. I implemented `isolate_namespace` because it's what the Rails guide suggests and to help address code reloading issues. I anticipated that there would be some potential issues here. I could only find that it required me to use `main_app` proxy in Devise templates but that was it since I rely on the default admin namespace.

spec/requests/custom_namespace_routes_spec.rb
```diff
@@ -0,0 +1,125 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe "Custom Namespace Routes", type: :request do
+  # Test that custom namespaces (other than :admin) can access route helpers
+  # This reproduces the bug from: activeadmin-v4-custom-namespace-bug-report.md
```
tagliala, line 6: What is this `activeadmin-v4-custom-namespace-bug-report.md` file referenced here?

craigmcnamara: Sorry, that was a bug report from my app. I need to clean this up a bit.

craigmcnamara:
> Could you add a minimal failing test that reproduces the problem and show why it happens?
> 

The reproduction is in spec/requests/custom_namespace_routes_spec.rb

If you remove the overrides those fail. My production app successfully loads ActiveAdmin with the patches and fails exactly like the bug report without them. 


My understanding is when routing params are omitted the routes helper infers them from the current request in the controller or view context. It seems the way the routes are used in ActiveAdmin, the helper doesn't always have the request context and that info isn't picked up.

This isn't my first foray into ActiveAdmin, I wrote the patch that allows unlimited sized CSV downloads by implementing a streaming response back in the v1 days.

Issue #8881, opened 2025-12-11
ActiveAdmin does not work when `permanent_connection_checkout` is disallowed (v3 / v4, Rails >= 7.2)


tagliala:
> It is when `config.active_record.permanent_connection_checkout = :disallowed` config is used. Which is a good idea when using fibers, as there can be many more fibers than threads and then dry up the db pool
> 
> Try and load the admin users page and see the exception.
> 
> _Originally posted by @henrikbjorn in https://github.com/activeadmin/activeadmin/issues/8474#issuecomment-3641924197_

### Expected behavior

ActiveAdmin works

### Actual behavior

```
ActiveRecord::ActiveRecordError in Admin::CommentsController#index

Called deprecated `ActiveRecord::Base.connection` method.
Either use `with_connection` or `lease_connection`.
```

### How to reproduce

1. Clone `demo.activeadmin.info`
2. Add `config.active_record.permanent_connection_checkout = :disallowed` to `config/application.rb`
3. Go to http://localhost:3000/admin/comments

### More information

Old PR addressing this issue #8474

mmogy8313-del:
ActiveRecord::ActiveRecordError in Admin::CommentsController#index

Called deprecated `ActiveRecord::Base.connection` method.
Either use `with_connection` or `lease_connection`.

PR #8883, opened 2025-12-12
Drop legacy connection support

tagliala:
Replace all uses of `ActiveRecord::Base.connection` with
`ActiveRecord::Base.with_connection`, following Rails guidelines.

Direct connection usage is now dropped for Rails < 7.2, and
attempting to call .connection directly raises in Rails 8.x.
This aligns our codebase with current Rails recommendations and ensures
compatibility going forward.

Tests are updated to handle connection access correctly, running
some cases outside transactional fixtures to validate quoting and
order clause methods.

Close https://github.com/activeadmin/activeadmin/issues/8881

References:
- Rails PR: https://github.com/rails/rails/pull/51349
- Connection Handling:
  https://github.com/rails/rails/blob/v7.2.1/activerecord/lib/active_record/connection_handling.rb#L259

### DIFF
--- a/spec/unit/order_clause_spec.rb
+++ b/spec/unit/order_clause_spec.rb
@@ -26,10 +26,6 @@ RSpec.describe ActiveAdmin::OrderClause do
     specify "#to_sql prepends table name" do
       expect(subject.to_sql).to eq '"posts"."id" asc'
     end
-
-    # Prevents automatically wrapping this test in a transaction. Check that we use the proper method to get a connection from the pool to quote the clause.
-    # We don't want to run the test using a leased connection https://github.com/rails/rails/blob/v7.2.3/activerecord/lib/active_record/test_fixtures.rb#L161
-    uses_transaction "#to_sql prepends table name"
   end
 
   describe "posts.id_asc" do
--- a/spec/unit/resource_spec.rb
+++ b/spec/unit/resource_spec.rb
@@ -41,10 +41,6 @@ module ActiveAdmin
       it "should return quote argument" do
         expect(config.resource_quoted_column_name('first_name')).to eq '"first_name"'
       end
-
-      # Prevents automatically wrapping this test in a transaction. Check that we use the proper method to get a connection from the pool.
-      # We don't want to run the test using a leased connection https://github.com/rails/rails/blob/v7.2.3/activerecord/lib/active_record/test_fixtures.rb#L161
-      uses_transaction "should return quote argument"
     end
 
     describe "#decorator_class" do


tagliala:
Converting back to draft because this PR will be affected by #8908

spec/support/rails_template.rb
```diff
@@ -90,6 +90,8 @@ def self.ransackable_associations(auth_object=nil)
 # Force strong parameters to raise exceptions
 environment "config.action_controller.action_on_unpermitted_parameters = :raise"
 
+environment "config.active_record.permanent_connection_checkout = :disallowed"
```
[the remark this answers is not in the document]
tagliala: @copilot are we talking about `ActiveRecord::Base.connection.query_cache_enabled` ? If yes, please reason if this is really needed and fix the spec. 

Here we do want this setting in `test` env

tagliala:
@javierjulio this change should also be good

spec/unit/order_clause_spec.rb
```diff
@@ -26,6 +26,10 @@
     specify "#to_sql prepends table name" do
       expect(subject.to_sql).to eq '"posts"."id" asc'
     end
+
+    # Prevents automatically wrapping this test in a transaction. Check that we use the proper method to get a connection from the pool to quote the clause.
+    # We don't want to run the test using a leased connection https://github.com/rails/rails/blob/v7.2.3/activerecord/lib/active_record/test_fixtures.rb#L161
+    uses_transaction "#to_sql prepends table name"
```
javierjulio: Are we sure this is working as intended? If so, all good with the changes.

tagliala: @copilot why do we need, if we need, `uses_transaction` here?

tagliala: @javierjulio I'm not the original author of that change

Good catch, btw.

Maybe this has some effect on 7.1 or previously tested versions

I'm going to remove both those lines, because they do not change how the test is being performed in no versions of rails tested. I did expect a failure by removing those lines, but they didn't fail

When `disallowed` is set at configuration level, specs don't pass if we are not using `with_connection` properly

Example:

```
Failures:

  1) ActiveAdmin::Resource#resource_quoted_column_name should return quote argument
     Failure/Error: resource_class.connection.quote_column_name(column)
     
     ActiveRecord::ActiveRecordError:
       Called deprecated `ActiveRecord::Base.connection` method.
     
       Either use `with_connection` or `lease_connection`.
```

javierjulio: @tagliala ok great, sounds good, thanks again!

PR #8886, opened 2025-12-13
[v3]: Allow resource_class to be passed as a String

tagliala:
the current pattern of using the class directly loads all models bound to activeadmin before using them, needlessly. besides that, the only purpose of it is to store the class name

this patch allows one to pass the class name directly instead, kind of like how activerecord avoids eager-loading association classes.

### DIFF
--- a/lib/active_admin/resource.rb
+++ b/lib/active_admin/resource.rb
@@ -70,7 +70,7 @@ module ActiveAdmin
     module Base
       def initialize(namespace, resource_class, options = {})
         @namespace = namespace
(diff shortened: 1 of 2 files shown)


tagliala:
@mgrunberg I take advantage of your previous review to also ask a review here. I've backported this feature from v4, it should not be a problem

mgrunberg (approved):

mgrunberg:
@tagliala I have some time so let's review as much as we can :smile:

PR #8891, opened 2025-12-19
Ensure DEBUG is unset when executing `bundle show`

iangreenleaf:
This was an unpleasant bug I ran into while trying to get my source compilation sorted out (cruelly, I had turned DEBUG on to help me troubleshoot a _different_ problem).

If DEBUG is set at all in the environment, even as an empty string with `DEBUG=`, bundler will output an extra message:
> Found no changes, using resolution from the lockfile

Because this goes to stdout, it gets included in the path the config file is using and asset compilation fails. Given that other people may well have this set in their envs for legitimate reasons, it seems wise to take precautions to ensure it doesn't blow up asset compilation.

### DIFF
--- a/lib/generators/active_admin/assets/templates/tailwind.config.js
+++ b/lib/generators/active_admin/assets/templates/tailwind.config.js
@@ -1,7 +1,12 @@
 import { execSync } from 'child_process';
 import activeAdminPlugin from '@activeadmin/activeadmin/plugin';
 
-const activeAdminPath = execSync('unset DEBUG; bundle show activeadmin', { encoding: 'utf-8' }).trim();
+// Remove DEBUG to prevent bundler from outputting extra text
+const { DEBUG: _, ...env } = { ...process.env };
+const activeAdminPath = execSync('bundle show activeadmin', {
+  encoding: 'utf-8',
+  env
+}).trim();
 
 export default {
   content: [


tagliala:
Hi, thanks for this PR.

I see the problem and I can replicate, however it seems that `unset` is not cross platform (do not have a windows machine at the moment to test, I'm using copilot + claude)

> ## Cross-Platform Summary
> 
> | Platform | Command | Works? |
> |----------|---------|--------|
> | macOS/Linux | `unset DEBUG` | ✅ Yes |
> | Windows cmd | `unset DEBUG` | ❌ No (command not found) |
> | Windows cmd | `set DEBUG=` | ⚠️ Sets empty string (still triggers debug) |
> | Windows PowerShell | `Remove-Item Env:\DEBUG` | ✅ Yes |
> | Git Bash (Windows) | `unset DEBUG` | ✅ Yes |
|>  WSL (Windows) | `unset DEBUG` | ✅ Yes |

It advises to use javascript like this:

> ## For your JavaScript use case: 
> 
> Since you need cross-platform support, **do NOT use shell commands**. Use the Node.js approach:
> 
> ```javascript
> const { execSync } = require('child_process');
> 
> const env = { ...process.env };
> delete env.DEBUG;
> 
> const activeAdminPath = execSync('bundle show activeadmin', { 
>   encoding: 'utf-8',
>   env: env
> }).trim();
> ```
> 
> This is **truly cross-platform** because Node.js handles the environment variables internally, regardless of the underlying shell. 

I'm not familiar with this process, so I'll leave the decision to others

---

PS: The CI looks broken, it looks like something is wrong with bundler and the latest version does not match the installed version

iangreenleaf:
Oh great, that's a better solution! I was able to use destructuring to clean it up a little more.

iangreenleaf:
@tagliala I'm guessing the CI failures are unrelated to this branch and I don't need to do anything about them?

tagliala:
> @tagliala I'm guessing the CI failures are unrelated to this branch and I don't need to do anything about them?

Sorry, I said yesterday "I'll take a look" but I forgot. I think failures are unrelated. I'll give a fix later today and I'll ping here

tagliala:
@iangreenleaf I've fixed the CI.

Could you please rebase and confirm that the JS-only solution fixes the issue you are experiencing? I've only tested in `node` console and it did work without emitting the warning

iangreenleaf:
@tagliala I can confirm that this version is working great for me and fixes the issue!

tagliala:
Did the CI broke again?

tagliala:
Nope, this does not look like a failure caused by bundler version mismatch

tagliala:
Update: found the issue on the CI

That line is not used in the CI and is required to be a single line because of this:

https://github.com/activeadmin/activeadmin/blob/511f8b702dadf78faed9d7b9898033f1e6b209f2/spec/support/rails_template.rb#L100-L102

I'm making some experiments

tagliala:
@iangreenleaf can you please give a try at #8895 ?

Is there anyone with a windows machine available to test #8895 ? Using the CI on windows will not work because that line is being replaced

tagliala:
Superseded by #8895, commit ownership preserved there, thanks!

PR #8892, opened 2025-12-19
Disable full automatic source scanning by Tailwind

iangreenleaf:
I don't know if this is something you want, but I thought I'd make the suggestion since I ran into this while upgrading and it took me a while to figure out. One way or another, this is probably something to document in the Tailwind setup instructions.

By default Tailwind will scan just about everything in your app code. For us, this meant things like the `log` dir, which made Tailwind exceptionally noisy, plus extra declarations in our admin stylesheet we didn't need. Most people aren't going to want their non-admin source code to be considered for the AA stylesheet, and since you have explicit `content` declarations in the `tailwind-active_admin.config.js` file, we don't really need Tailwind to be guessing any locations, so this seems like a good default to me.

### DIFF
--- a/lib/generators/active_admin/assets/templates/active_admin.css
+++ b/lib/generators/active_admin/assets/templates/active_admin.css
@@ -1,3 +1,3 @@
-@import "tailwindcss";
+@import "tailwindcss" source(none);
 
 @config "../../../tailwind-active_admin.config.js";


tagliala:
Thanks for this PR.

I'm trying to understand what is exactly going on here, I'm not familiar with that directive

https://tailwindcss.com/docs/detecting-classes-in-source-files#disabling-automatic-detection

> Use source(none) to completely disable automatic source detection if you want to register all of your sources explicitly:

Disabling automatic source detection seems "too much", but I think this may makes sense for this use case.

It is unclear to me tailwind attempts to search into the log folder. Does it traverse up the tree? Can you provide more information?

iangreenleaf:
There's some more explanation further up that file: https://tailwindcss.com/docs/detecting-classes-in-source-files#which-files-are-scanned

As best I understand it, it seems that Tailwind by default scans every text-based file in the project (I am guessing they use the current working directory as the base) other than the dirs they exclude automatically. And they don't exclude anything like `log/` so I guess it gets pulled in 🤷. That matches my experience, which is that once I added some debugging to spit out which files were registering changes, it was telling me that it had detected changes in the log files and even in the `.git` folder. In those cases, the Tailwind CLI will quickly determine that none of these files are relevant and exit without recompiling, but it still executes a loop and prints out "Done in 2s" to the logs.

javierjulio:
@iangreenleaf thank you. Sorry, I don't think this is a worthwhile change to make now. I feel this is a local issue rather than a general one since the files in the log directory should be ignored with a Rails default `.gitignore` file. Tailwind will automatically ignore files listed in your `.gitignore` file too. Can you share what the contents of your `.gitignore` file? How are you able to debug what paths Tailwind is using as a source? That would be helpful to know, thank you.

tagliala:
Thanks, I also read Javier's comment, and I'm having the same questions

> As best I understand it, it seems that Tailwind by default scans every text-based file in the project (I am guessing they use the current working directory as the base) other than the dirs they exclude automatically.

This sounds plausible, but I'm struggling to understand how a configuration line in `/app/path/to/admin.tailwind.css` would trigger a scan of the `/log` directory.

My concern is that if we need to add `source(none)` to a single file buried in a nested folder, it suggests the Tailwind configuration is already problematic. Ideally, Tailwind shouldn't be scanning `/log` at all by default, not because a file deep in the directory tree explicitly "stopped" it to do that

---

> I feel this is a local issue rather than a general one since the files in the log directory should be ignored with a Rails default .gitignore file. 

I considered this possibility as well, but haven't been able to reproduce it. Additional information about the specific setup would be helpful to understand what's happening

PR #8893, opened 2025-12-20
Update dependencies

tagliala:


tagliala:
There is room for improvement here, because this is a workaround to the failure we are having when latest bundler version and locked version don't match, but let's fix the CI for the moment

PR #8895, opened 2025-12-23
Ignore extra bundler output when DEBUG is set

tagliala:
When the DEBUG environment variable is set (even if empty), bundler may output additional messages to stdout, such as "Found no changes, using resolution from the lockfile". This can interfere with asset compilation, as scripts expecting a single path may instead receive multiple output lines, causing path corruption.

This commit adjusts the logic to always use the last line of the output from `bundle show activeadmin`, preventing such issues even if extraneous lines are present due to DEBUG or other future messaging changes.

Ref: https://github.com/activeadmin/activeadmin/pull/8891

### DIFF
--- a/lib/generators/active_admin/assets/templates/tailwind.config.js
+++ b/lib/generators/active_admin/assets/templates/tailwind.config.js
@@ -1,11 +1,7 @@
 import { execSync } from 'child_process';
 import activeAdminPlugin from '@activeadmin/activeadmin/plugin';
 
-// Ensure only the last line of output is used from `bundle show activeadmin`,
-// in case bundler prints extra lines (e.g., DEBUG enabled).
-// Keep this assignment on a single line so that the Rails template `gsub_file`
-// matches and replaces it correctly.
-// Ref: https://github.com/activeadmin/activeadmin/pull/8891
+// Always use the last line of output since Bundler's DEBUG env will print additional lines.
 const activeAdminPath = execSync('bundle show activeadmin', { encoding: 'utf-8' }).trim().split(/\r?\n/).pop();
 
 export default {


tagliala:
Thanks for the test and have a nice new year.

I'm going to close the other PR, but of course I've kept you as the author of the commit.

I'll test on a windows machine (just the javascript line) and update if necessary

tagliala:
I can confirm this works on Windows as well and fixes the issue if DEBUG is set

### Before
```
PS .\demo.activeadmin.info> yarn build:css --production

PS .\demo.activeadmin.info> ls .\app\assets\builds\active_admin.css


    Directory: .\demo.activeadmin.info\app\assets\builds


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----        30/12/2025     22:10          24227 active_admin.css
```

### After
```
PS .\demo.activeadmin.info> yarn build:css --production

PS .\demo.activeadmin.info> ls .\app\assets\builds\active_admin.css


    Directory: .\demo.activeadmin.info\app\assets\builds


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----        30/12/2025     22:12          80982 active_admin.css
```

lib/generators/active_admin/assets/templates/tailwind.config.js
```diff
@@ -1,7 +1,12 @@
 import { execSync } from 'child_process';
 import activeAdminPlugin from '@activeadmin/activeadmin/plugin';
 
-const activeAdminPath = execSync('bundle show activeadmin', { encoding: 'utf-8' }).trim();
+// Ensure only the last line of output is used from `bundle show activeadmin`,
+// in case bundler prints extra lines (e.g., DEBUG enabled).
+// Keep this assignment on a single line so that the Rails template `gsub_file`
+// matches and replaces it correctly.
+// Ref: https://github.com/activeadmin/activeadmin/pull/8891
```
javierjulio: Minor update to the message content. Since the template is also copied to the user's app, I don't want to include the second message. We'd have the same issue anyway with most other content replacements from our test app generator.

```suggestion
// Always use the last line of output since Bundler's DEBUG env will print additional lines.
```

Issue #8906, opened 2026-01-02
Feature: Devise 5 support


tagliala:
At the moment, Devise 5 is in RC and specs are failing

tagliala:
Bisected to: https://github.com/heartcombo/devise/commit/fec67f98f26fcd9a79072e4581b1bd40d0c7fa1d

https://github.com/heartcombo/devise/pull/5734

tagliala:
```diff
- validates_length_of       :password, minimum: password_length.min, maximum: password_length.max, allow_blank: true
+ validates_length_of       :password, minimum: proc { password_length.min }, maximum: proc { password_length.max }, allow_blank: true
```

...why?

The problem is setting a proc on the maximum validation

tagliala:
When maximum validation is a proc, password is being swallowed from params 🫨 

```diff
- validates_length_of       :password, maximum: 128, allow_blank: true
+ validates_length_of       :password, maximum: -> { 128 }, allow_blank: true
```

### Before
```
(byebug) params
{"admin_user" => {"email" => "admin@example.com", "password" => "password", "remember_me" => "0"}, "commit" => "Sign In", "controller" => "active_admin/devise/sessions", "action" => "create"}
```

### After
```
(byebug) params
{"admin_user" => {"email" => "admin@example.com", "password" => "", "remember_me" => "0"}, "commit" => "Sign In", "controller" => "active_admin/devise/sessions", "action" => "create"}
```

tagliala:
Ok, I guess it's the HTML5 `maxlength` incorrectly set when max is a proc.

It is

<img width="247" height="21" alt="Image" src="https://github.com/user-attachments/assets/b4669034-e81c-447e-ae9d-2e5e20c0624b" />

tagliala:
... but ... ActiveAdmin uses Formtastic, so I think this is an issue on Formtastic as well (and also on CSV which I maintain 🥲)

tagliala:
@javierjulio 

this should be fixed upstream, but that will force activeadmin to include a workaround for Rails 7.0 and 7.1, because the edge version of Formtastic dropped 

Should we force `maxlength` to false in the password field to workaround this issue?

Ref: formtastic/formtastic#1401

javierjulio:
@tagliala thanks. I was thinking this recently. I prefer we drop Rails 7.0 and 7.1 support since those aren't supported anymore by the Rails team anyway.

tagliala:
[...]

PR #8908, opened 2026-01-04
Drop support for Rails 7.0 and 7.1

tagliala:
Rails 7.0 and 7.1 have reached end-of-life, so maintaining compatibility
no longer brings value and can complicate support for newer Rails
versions.

This commit also updates the test matrix to cover Node.js 24 and updates
project dependencies.

Ref: https://github.com/activeadmin/activeadmin/issues/8906


lib/active_admin/async_count.rb
```diff
@@ -11,9 +11,7 @@ def initialize(collection)
     end
 
     def count
-      value = @promise.value
-      # value.value due to Rails bug https://github.com/rails/rails/issues/50776
-      value.respond_to?(:value) ? value.value : value
+      @promise.value
```
tagliala, line 14: As far as I understand, this issue never landed in 7.2

However, it is hard to test against 7.2.0 in specs because of instance doubles

https://github.com/rails/rails/issues/50776#issuecomment-2148243023

> Ah, thanks for clarifying, I misunderstood the branch numbering.
> 
> So it sounds like...
> 
> * It is an actual regression between 7.1.2 and 7.1.3
> * It's fixed by that linked PR
> * It's backported to 7-1-stable branch and will be fixed in a future 7.1.4 **and stay fixed in 7.2**
> 
> ?

(emphasis mine)

PR #8909, opened 2026-01-06
Minor non-production fixes

tagliala:
- Normalize platforms in 7.2 and 8.0 gemfiles
- Remove pre-7.2 workarounds in rails template
- Update dependencies
- Prefer environment over inject for new settings

### DIFF
--- a/spec/support/rails_template.rb
+++ b/spec/support/rails_template.rb
@@ -3,7 +3,8 @@
 
 gem "cssbundling-rails"
 
-create_file "app/assets/config/manifest.js"
+# Manifest already exists in Rails 7.2
+create_file "app/assets/config/manifest.js", skip: true
 
 rails_command "css:install:tailwind"
 # Remove default configuration generated: https://github.com/rails/cssbundling-rails/blob/v1.4.2/lib/install/tailwind/install.rb#L7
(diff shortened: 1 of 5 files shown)


spec/support/rails_template.rb
```diff
@@ -70,21 +71,9 @@ def self.ransackable_associations(auth_object=nil)
...
-inject_into_file "config/environments/test.rb", after: "Rails.application.configure do" do
-  "\n" + <<-RUBY
-  config.cache_classes = !ENV['CLASS_RELOADING']
-  RUBY
-end
 gsub_file "config/environments/test.rb", /config.enable_reloading = false/, "config.enable_reloading = !!ENV['CLASS_RELOADING']"
 
-inject_into_file "config/environments/test.rb", after: "config.cache_classes = !ENV['CLASS_RELOADING']" do
-  "\n" + <<-RUBY
-  config.action_mailer.default_url_options = {host: 'example.com'}
```
tagliala, line 84: this is set on all supported rails version, even if 7.2 has `www.example.com`, but it does not appear to be a problem.

Also, since this is already set and it comes after the injection, the value will be set with the second one, making in any case this injection ineffective

javierjulio: All these changes look great. Thank you for simplifying it. I appreciate it.

PR #8918, opened 2026-01-16
Update dependencies

tagliala:
Now that bundler supports all versions it could be possible to use the same version as the gemfile.lock, but not for the moment

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -142,7 +142,7 @@ GEM
       cucumber-messages (>= 31, < 32)
     cucumber-html-formatter (22.3.0)
       cucumber-messages (> 23, < 33)
-    cucumber-messages (31.1.0)
+    cucumber-messages (31.2.0)
     cucumber-rails (4.0.0)
       capybara (>= 3.25, < 4)
       cucumber (>= 7, < 11)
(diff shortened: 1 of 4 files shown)


javierjulio:
@tagliala I was just curious what you meant and were referring to by supporting all versions? The Ruby version?

tagliala:
> supporting all versions? The Ruby version?

Yes, sorry, all the ruby versions supported by Active Admin now support bundler 4.0.3, so we can use a specific version of bundler I guess somewhere in the actions to avoid the CI to break. But this will change soon, we should have a better way of workarounding this issue to prevent the CI to "randomly" break.

```
Failures:

  1) Gemfile sanity is up to date
     Failure/Error: expect(current_lockfile).to eq(new_lockfile), msg

       Please update /home/runner/work/activeadmin/activeadmin/Gemfile's lock file with `BUNDLE_GEMFILE=/home/runner/work/activeadmin/activeadmin/Gemfile bundle install` and commit the result
       Diff:
       @@ -501,4 +501,4 @@
          webrick
        
Tests Failed
        BUNDLED WITH
       -  4.0.4
       +  4.0.3
     # ./spec/tasks/gemfile_spec.rb:13:in 'block (2 levels) in <top (required)>'

Finished in 14.94 seconds (files took 5.83 seconds to load)
494 examples, 1 failure

Failed examples:

rspec ./spec/tasks/gemfile_spec.rb:3 # Gemfile sanity is up to date
```

This is the problematic spec, should we skip the "bundled with" comparison somehow?

javierjulio:
Ah yes, I remember that. I've been bitten by it a few times. Yes, I think we should try to avoid the "bundled with" line in the test if we can. The test case is helpful overall, just that one aspect that is problematic. Thank you.

PR #8936, opened 2026-02-08
Add Devise 5 and Formtastic 6 support

tagliala:
Keep Devise minimum unchanged to avoid unnecessary breakage for
downstream apps. Raising the required versions would force upgrades
without a strict need, since Devise 4 remains supported.

Formtastic instead is required to be >= 6. It matches ActiveAdmin's
minimum Ruby requirement and is compatible with both Devise 4 and 5.

CI, however, will run the test matrix only against Devise 5 to validate
compatibility with current versions.

Additionally, update bundle syntax in spec for v4

Close https://github.com/activeadmin/activeadmin/issues/8906

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ Now install the development dependencies:
 
 ```sh
 gem install foreman
-bundle install
+bundle install --all
 yarn install
 ```
 
--- a/Gemfile
+++ b/Gemfile
@@ -8,13 +8,13 @@ group :development, :test do
   gem "pundit"
 
   gem "draper"
-  gem "devise", "~> 4.9" # TODO: relax this dependency when formtastic/formtastic#1401 will be fixed
+  gem "devise"
 
   gem "rails", "~> 8.1.0"
 
   gem "sprockets-rails"
   gem "ransack", ">= 4.2.0"
-  gem "formtastic", ">= 5.0.0"
+  gem "formtastic", ">= 6.0.0"
 
   gem "cssbundling-rails"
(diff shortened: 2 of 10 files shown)


tagliala:
@javierjulio do you prefer to bump the formtastic requirement in gemspec to v6?

tagliala:
TL;DR: Keeping current gemspec dependencies should be fine

---

> That depends if Formtastic v6 supports both Devise v4 and v5?

Formtastic v6 just drops some legacy rails and ruby versions. It also matches 3.2 / 7.2 that we have here

Additionally, Formtastic v6's changelog mentions "compatibility with Devise 5", but that is a fix to allow rails proc validation

Master branch now test Devise v4 against Formtastic v6, and everything is fine

> Because of the update to support Devise v5, wouldn't we at least want to require Devise v5 as the minimum on our side?

This does make sense too, but probably this is not necessary and may cause friction.

Devise has a lot of plugins, maybe not all ready for v5, so keeping v4 compatibility should not be a problem

> Do you think its necessary that we bump the minimum requirement for Formtastic to >= v6?

Not necessary, because that will have an impact on the main application as well

---

### edit

- All specs green with Formtastic 6 / Devise 4 #8947 
- Changed the commit message and the PR opening post

davidwessman:
Well done on getting everything updated here and in Formtastic 👏🏻 🫶 @tagliala

julitrows:
Is this gonna be on the 3.x branch or just for 4.0+?

cprodhomme:
We need this patch on 3.x branch if it possible

PR #8939, opened 2026-02-16
Add go-to-page input to pagination controls

hpanwar09:
- Adds a page number input field to the pagination bar on index pages
- Users can type a page number and press Enter to jump directly to that page
- Follows established patterns from Ant Design / Element UI (quick jumper)
- Input placed after the next-page arrow, using the same spacing as other pagination elements

### DIFF
--- a/app/javascript/active_admin.js
+++ b/app/javascript/active_admin.js
@@ -6,5 +6,6 @@ import "active_admin/features/has_many"
 import "active_admin/features/filters"
 import "active_admin/features/main_menu"
 import "active_admin/features/per_page"
+import "active_admin/features/go_to_page"
 
 Rails.start()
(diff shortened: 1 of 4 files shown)


javierjulio:
@hpanwar09 thank you. Sorry, I don't think it's necessary for the library to provide this. The only file that was changed (app/views/active_admin/kaminari/_paginator.html.erb) is something you have access to now and is expected and safe to customize to your needs. Having access to partials for customization was a key objective with v4. In the future, please follow the guidelines on submitting a discussion for a feature request before submitting a PR. In general though nothing here needs to be implemented by the library since you'd have everything you need to add this without relying on the library to provide it. So it's not something I would accept. Thank you.

PR #8974, opened 2026-03-20
[WIP] Feature/vitepress 2 and multiversion documentation

tagliala:
@javierjulio this supports multiple versions thanks to vitepress v2 alpha (can't tell if this is also possible with the stable version). You can test it locally, however some points remain

- [ ] Should we preserve `/` documentation?
- [ ] If we migrate `/` to `/v3`, should we add redirects?
- [ ] Should `/` be v4 now?
- [ ] Should we duplicate everything?


javierjulio:
@tagliala thank you for taking this on. I appreciate it. I'm ok without any redirects. What I think is important is to maintain both v3 and v4 (current) docs so we can eventually cut a final v4 release. Yes the docs should be duplicated between both versions. I already have a feature branch with v4 updates. Perhaps we do this Vitepress 2 update with just the v3 docs in its own directory?

tagliala:
@javierjulio 

Welcome. Just to recap:

There is an intermediate step with:
- vitepress v2
- Active Admin v3 only documentation
- `/v3` folder
- no redirect from root to `/v3` (will potentially break search engines, but it should be almost fine when v4 will land)

Correct?

tagliala:
Superseded by #8975. v4 will come in a different PR

Issue #8969, opened 2026-03-11
Scoped collection attributes being used during create, leading to errors, possible data deletion and possible data corruption


michaelbaldry-il:
Firstly, I want to say thanks for Active Admin, it's a great gift to the rails community ❤️ 

I dealt with a bug recently that caused creates to fail and could result in records being deleted that were unrelated to the action being performed in Active Admin.

Using the Pundit authorization adapter with a scope class similar to the [example one](https://github.com/activeadmin/activeadmin/blob/master/spec/support/templates/policies/application_policy.rb) that is referenced [in the documentation](https://activeadmin.info/13-authorization-adapter.html#using-the-pundit-adapter).

The scope looks like this:
```ruby
class ThingPolicy
  class Scope
    def resolve
      Thing.where(id: user.accessible_thing_ids)
    end
  end
end
```

The user can only see `Thing` based on an array of ids tied to their account. Which seems reasonable and works well.

The issue arises when this user wants to create a `Thing`, using the default create action.

This starts in [`InheritedResources::Actions#create`](https://github.com/activeadmin/inherited_resources/blob/master/lib/inherited_resources/actions.rb#L32) calling `build_resource` then `create_resource` which are defined in [`ActiveAdmin::ResourceController::DataAccess`](https://github.com/activeadmin/activeadmin/blob/master/app/controllers/active_admin/resource_controller/data_access.rb#L115L147)

`build_resource` calls `build_new_resource` which applies the scope to the collection and boils down to `Thing.where(id: user.accessible_thing_ids).new(params)`. Then `create_resource` saves that. 

The problem with this is that the new resource picks up the `where` scoped attributes, which is a (rather unexpected) Active Record feature.

E.g.
```ruby
pry(main)> Thing.where(id: [1]).new(title: "Cat")
=> #<Thing:0x0000ffff79e721c8 id: 1, title: "Cat">
```

Then the `create_resource` happens and tries to save it, resulting in an `ActiveRecord::RecordNotUnique` because `id: 1` already exists. 

This is the simple case but when you're using a nested form and the model looks like:

```ruby
class Thing
  has_one :important, dependent: :destroy
  accepts_nested_attributes_for :important
end
```

Rails deletes the `Important.where(thing_id: 1)` in preparation for creating another one and there is no transactional rollback of this, so the `Important` is deleted even after `Thing` failed to create.

[...]

PR #8986, opened 2026-03-30
Do not apply authorization scope when building new resources

michaelbaldry-il:
This is a possible fix for [#8969](https://github.com/activeadmin/activeadmin/issues/8969). 

It stops using `apply_authorization_scope` in `build_new_resource` to avoid copying scoped attributes to the new records that are created.

All tests pass but I'm unsure if it's possible there is reliance on this quirk by any users.

### DIFF
--- a/app/controllers/active_admin/resource_controller/data_access.rb
+++ b/app/controllers/active_admin/resource_controller/data_access.rb
@@ -127,9 +127,15 @@ module ActiveAdmin
       # Builds a new resource. This method uses the method_for_build provided
       # by Inherited Resources.
       #
+      # Authorization scoping is intentionally not applied here because
+      # scope conditions (e.g. where clauses) would be propagated as
+      # attributes on the new record by Active Record.
+      # Authorization for new resources is handled by authorize_resource!
+      # in build_resource instead.
+      #
       # @return [ActiveRecord::Base] An un-saved active record base object
       def build_new_resource
-        apply_authorization_scope(scoped_collection).send(
+        scoped_collection.send(
           method_for_build,
           *resource_params.map { |params| params.slice(active_admin_config.resource_class.inheritance_column) }
         )
--- a/spec/unit/resource_controller/data_access_spec.rb
+++ b/spec/unit/resource_controller/data_access_spec.rb
@@ -266,11 +266,9 @@ RSpec.describe ActiveAdmin::ResourceController::DataAccess do
     context "given authorization scope" do
       let(:authorization) { controller.send(:active_admin_authorization) }
 
-      it "should apply authorization scope" do
-        expect(authorization).to receive(:scope_collection) do |collection|
-          collection.where(age: "42")
-        end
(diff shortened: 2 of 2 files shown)


tagliala:
Hello, as per the `git blame`, that change has been introduced intentionally because of https://github.com/activeadmin/activeadmin/pull/6884

michaelbaldry-il:
> Hello, as per the `git blame`, that change has been introduced intentionally because of #6884

Hmmm, interesting. I didn't find that in my searches. 

Do you have any suggestions about what we could do here? I now see there are scenarios where this makes sense but also scenarios where it is a bad idea (e.g. https://github.com/activeadmin/activeadmin/issues/8969).

tagliala:
The fact that specs are not failing is odd, it means that the fix is not testing correctly what it was supposed to fix

michaelbaldry-il:
> The fact that specs are not failing is odd, it means that the fix is not testing correctly what it was supposed to fix

The only spec in the merged PR just checks that it applies it, which I removed.

tagliala:
> which I removed.

Sorry, it's a small PR, but didn't notice that when I've commented.

I mean, it was checking something, but is not testing what that scope is needed for (creation of nested objects?)

However, in the original issue #8969 I can see that it is advised to not apply authorization scope on `new` and `create`, so yes, as are you mentioning, none of the two is correct

Copilot suggests to exclude some keys from the params. Should we exclude the primary key? Is this related to the primary key?

michaelbaldry-il:
The cause of our specific issue was scoping to `id` but it could apply unexpectedly to anything. In the other example I provided for instance:

You set your scope to be `Post.where(published: true)` - you only want your users to see published posts
When they create a post, it's automatically set `published: true` due to the authorisation scope moving to the new record, which seems very counter intuitive.

The comment in [the pundit adapter](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/pundit_adapter.rb#L22L23) suggests it is not meant to be used for `new` and `create`, but here we are using it.

I don't think excluding keys is the answer - it would mean you'd need to know in advance of this unexpected behaviour and plan for it, and know to add new ones as things change, which is easy to miss. I don't know what the correct solution is now though!

Personally, I think the [PR that introduced this](https://github.com/activeadmin/activeadmin/pull/6884) may have needed more consideration before merging, but now it is established functionality that some rely on, it would seem.

michaelbaldry-il:
@tagliala hey, just following up on this. 

Maybe we could have it configurable and disabled in the new version config generator, so existing users can keep the functionality and opt-out if they want to?

tagliala:
Sorry, I can't tell what is the best approach here

/cc @javierjulio

javierjulio:
This one has been problematic before with previous updates. It would then affect someone else that was wanting that behavior and then changing it again, and it going the other way. I want to avoid additional issues for this for now since it's been around so long and until it can be addressed wholesale to accommodate all cases. I'm sorry. Thank you.

PR #9010, opened 2026-04-30
Update RuboCop config

tagliala:
- Move `Capybara` cops to their new `Capybara/RSpec` counterpart
- Add missing new cops
- Fix new `Capybara/RSpec/HaveContent` offenses

### DIFF
--- a/features/step_definitions/additional_web_steps.rb
+++ b/features/step_definitions/additional_web_steps.rb
@@ -53,7 +53,7 @@ end
 
 Then(/^I should see the page title "([^"]*)"$/) do |title|
   within("[data-test-page-header]") do
-    expect(page).to have_content title
+    expect(page).to have_text title
   end
 end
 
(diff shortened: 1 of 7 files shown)


features/step_definitions/additional_web_steps.rb
```diff
@@ -53,7 +53,7 @@
 
 Then(/^I should see the page title "([^"]*)"$/) do |title|
   within("[data-test-page-header]") do
-    expect(page).to have_content title
+    expect(page).to have_text title
```
tagliala, line 56: @javierjulio I think this new cop makes sense, `have_content` is an alias of `have_text` which is longer and more generic, but if you prefer not to make this change I'll revert and disable the new cop or set it to `have_content`

Ref: https://docs.rubocop.org/rubocop-capybara/latest/cops_capybara_rspec.html#capybararspechavecontent

javierjulio: @tagliala yes, I'm good with that change. Thank you.

javierjulio (approved):
Thanks!

PR #9013, opened 2026-05-02
Allow supplied HTML in table_for column header

Lukom:
This fixes the issue when you pass an image to column title – in that case arrow svg is shown as raw html in header cell. What I do in my index block:

```ruby
svg = content_tag :svg, width: '20', height: '20' do
  content_tag :use, nil, href: image_path("icons/key.svg#main")
end
column(svg, sortable: :access_level) { |r| r.access_level }
```

Besides I think here is a security issue as well – here title is not escaped but marked as `html_safe` after svg is added.

This patch safely joins both title and arrow svg.

### DIFF
--- a/lib/active_admin/views/components/table_for.rb
+++ b/lib/active_admin/views/components/table_for.rb
@@ -83,9 +83,9 @@ module ActiveAdmin
         if sort_key
           th(attributes) do
             link_to params: params, order: "#{sort_key}_#{order_for_sort_key(sort_key)}" do
-              svg = '<svg class="data-table-sorted-icon" fill="none" viewBox="0 0 10 6"><path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="m1 1 4 4 4-4"/></svg>'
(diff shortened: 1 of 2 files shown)


javierjulio (approved):
Thanks!

tagliala:
Any chance to add a spec to avoid a regression?

tagliala:
Hi @Lukom, thanks for the spec

I made a small change to the spec to make it more generic, as I believe the intent of this update is not strictly limited to images but applies more broadly to HTML tags in general.

I also avoided introducing an additional SVG to prevent potential ambiguity. I attempted to use a nested `find_by_tag`, but it does not seem to work in this context, likely because the structure is not Arbre-based.

/cc @javierjulio

Lukom:
Yes, generally any content generated by rails helpers as they return `ActiveSupport::SafeBuffer` result.

Changes look good to me.

Lukom:
Well, you could add a check that `<span` is not escaped as well.

tagliala:
Done, thanks

javierjulio (approved):
Thank you both!

PR #9016, opened 2026-05-08
Fix Rails 8.1 hash-arg deprecation in page_routes

Kaishley:
## What this PR does

Converts the last remaining hash-form routing-DSL call in `Router#page_routes` to keyword form, eliminating the `<verb> received a hash argument` deprecation warning that fires on every `page_action` declaration under Rails 8.1.

```ruby
# before
build_route(verb, "/#{page}/#{action.name}" => "#{page}##{action.name}")

# after
build_route(verb, "/#{page}/#{action.name}", to: "#{page}##{action.name}")
```

`build_route` already uses Ruby's `(verbs, ...)` argument forwarding, so the only change here is at the call site.

## Why

Rails 8.1 deprecated passing a hash as the second positional argument to the routing DSL methods (`get`/`post`/`match`/etc.) and will remove the form entirely in Rails 8.2. PR #8837 ("Add Rails 8.1.0 compatibility") fixed the surrounding `router.get` and `router.namespace` calls but missed this one inside `page_routes`. As a result, any application with `page_action :foo, method: :post` declarations still emits one warning per declaration at boot under 8.1, and will break under 8.2 unless this lands first.

I confirmed the line is unchanged on `master`, on `v3.5.1` (latest stable), and on `v4.0.0.beta22`.

## Risk

Generated routes are identical — `match`/`get`/`post` accept both forms today; only the invocation syntax changes. No behaviour change at runtime.

## Test

Adds a regression test in `spec/unit/routing_spec.rb` that registers a page with both a POST and a GET `page_action` and asserts they route correctly. (Existing routing specs only exercised the page index route, not the per-action routes — which is why the deprecation slipped through.)

### DIFF
--- a/spec/unit/routing_spec.rb
+++ b/spec/unit/routing_spec.rb
@@ -222,5 +222,52 @@ RSpec.describe "Routing", type: :routing do
         expect(admin_log_path).to eq "/admin/log"
       end
     end
+
+    context "with page_actions" do
+      around do |example|
+        with_resources_during(example) do
+          ActiveAdmin.register_page("Log") do
+            page_action :archive, method: :post
+            page_action :stats, method: :get
+          end
+        end
+      end
+
+      it "routes a POST page_action" do
+        expect({ post: "/admin/log/archive" }).to \
+          route_to({ controller: "admin/log", action: "archive" })
+      end
+
+      it "routes a GET page_action" do
+        expect({ get: "/admin/log/stats" }).to \
+          route_to({ controller: "admin/log", action: "stats" })
+      end
+    end
+
+    # Regression coverage for the Rails 8.1 deprecation that `Router#page_routes`
+    # used to emit via `build_route(verb, "/path" => "controller#action")`.
+    # Reproduces by intercepting `ActionDispatch.deprecator.warn` during route
+    # registration and asserting no "received a hash argument" message fires.
+    context "when registering a page with page_actions" do
+      it "does not emit a hash-argument deprecation from the routing DSL" do
+        deprecations = []
+        allow(ActionDispatch.deprecator).to receive(:warn) do |message, *|
+          deprecations << message.to_s
+        end
+
+        load_resources do
+          ActiveAdmin.register_page("Log") do
+            page_action :archive, method: :post
+            page_action :stats, method: :get
+          end
+        end
+
+        hash_arg_warnings = deprecations.grep(/received a hash argument/)
+        expect(hash_arg_warnings).to be_empty,
+          "expected no 'received a hash argument' deprecation, got: #{hash_arg_warnings.inspect}"
+      ensure
+        load_resources {}
+      end
+    end
   end
 end
--- a/lib/active_admin/router.rb
+++ b/lib/active_admin/router.rb
@@ -68,7 +68,7 @@ module ActiveAdmin
       router.get "/#{page}", to: "#{page}#index"
       config.page_actions.each do |action|
         Array.wrap(action.http_verb).each do |verb|
-          build_route(verb, "/#{page}/#{action.name}" => "#{page}##{action.name}")
+          build_route(verb, "/#{page}/#{action.name}", to: "#{page}##{action.name}")
         end
       end
     end


javierjulio (approved):
Thanks! @tagliala I haven't used `page_action` before. The change looks good and makes sense to me. What do you think?

tagliala (approved):
I did not use this either, and in my previous PR probably I've missed this change because of dynamic route creation that I wasn't able to spot

Thanks, looks good to me as well

tagliala:
@Kaishley should this be backported to v3 as well?

edit: I guess so

tagliala (changes requested):
Sorry, second thought.

I've downloaded this branch locally, reverted the changes to production code, ran the specs. 

I cannot see warnings in the console.

Questions:

1. How can I replicate the warning? Should I use rails-edge?
2. What the spec is actually testing? I was expected to see the warning

```
  page
    when default namespace
      should route to the page under /admin
    when singular page name
      should not inject _index_ into the route name
    with page_actions
      routes a GET page_action
      routes a POST page_action without emitting a hash-arg deprecation
    when in the root namespace
      should route to page under /
```

edit: tested against rails-edge as well, the previous imeplementation works without emitting a warning

tagliala:
I'm trying to replicate the warning here: https://github.com/activeadmin/demo.activeadmin.info/tree/tmp/activeadmin-9016-test

without success, can you please help in providing a reproducible test case?

tagliala:
# Copilot's take

My take: **the deprecation likely does not fire because the hash is no longer arriving at Rails as a positional “options hash” in the way `Mapper#get` / `#post` are checking for.** The forwarding path through `build_route(verbs, ...)` appears to preserve the call in a way that Rails ends up seeing as normal arguments plus keywords, not as the deprecated legacy form.

## Why I think it does not trigger

In ActiveAdmin, the relevant code on `master` is:

```ruby name=lib/active_admin/router.rb url=https://github.com/activeadmin/activeadmin/blob/da6ef80ab3ff33e09d8907db3836c62218a498f0/lib/active_admin/router.rb#L66-L72
def page_routes(config)
  router.get "/#{page}", to: "#{page}#index"
  config.page_actions.each do |action|
    Array.wrap(action.http_verb).each do |verb|
      build_route(verb, "/#{page}/#{action.name}" => "#{page}##{action.name}")
    end
  end
end
```

and:

```ruby name=lib/active_admin/router.rb url=https://github.com/activeadmin/activeadmin/blob/da6ef80ab3ff33e09d8907db3836c62218a498f0/lib/active_admin/router.rb#L94-L96
def build_route(verbs, ...)
  Array.wrap(verbs).each { |verb| router.send(verb, ...) }
end
```

Rails currently still contains the deprecation logic in `ActionDispatch::Routing::Mapper::HttpHelpers#get` and `#post`:

```ruby name=actionpack/lib/action_dispatch/routing/mapper.rb url=https://github.com/rails/rails/blob/cca4a4893ef59d8dfe61121f98cc308cc2e831b8/actionpack/lib/action_dispatch/routing/mapper.rb#L735-L750
def get(*path_or_actions, as: DEFAULT, to: nil, controller: nil, action: nil, on: nil, defaults: nil, constraints: nil, anchor: nil, format: nil, path: nil, internal: nil, **mapping, &block)
  if path_or_actions.grep(Hash).any? && (deprecated_options = path_or_actions.extract_options!)
    as = assign_deprecated_option(deprecated_options, :as, :get) if deprecated_options.key?(:as)
    to ||= assign_deprecated_option(deprecated_options, :to, :get)
    controller ||= assign_deprecated_option(deprecated_options, :controller, :get)
    action ||= assign_deprecated_option(deprecated_options, :action, :get)
    on ||= assign_deprecated_option(deprecated_options, :on, :get)
    defaults ||= assign_deprecated_option(deprecated_options, :defaults, :get)
    constraints ||= assign_deprecated_option(deprecated_options, :constraints, :get)
    anchor = assign_deprecated_option(deprecated_options, :anchor, :get) if deprecated_options.key?(:anchor)
    format = assign_deprecated_option(deprecated_options, :format, :get) if deprecated_options.key?(:format)
    path ||= assign_deprecated_option(deprecated_options, :path, :get)
    internal ||= assign_deprecated_option(deprecated_options, :internal, :get)
    assign_deprecated_options(deprecated_options, mapping, :get)
  end
```

That warning path is only entered if Rails sees a `Hash` inside `path_or_actions`, i.e. a positional options hash.

### The key point

The ActiveAdmin call is not a direct `router.get "/path" => "controller#action"` anymore in source-level intent. It is:

1. a method call into `build_route`
2. using Ruby argument forwarding `...`
3. then dispatched via `router.send(verb, ...)`

Because of that, Ruby may already be classifying the hash as keyword-like routing data by the time it reaches `get` / `post`, instead of leaving it as a positional hash in `*path_or_actions`.

If that happens, then this Rails check:

- `path_or_actions.grep(Hash).any?`
- `extract_options!`

never sees the deprecated form, so no warning is emitted.

## Why I think your non-repro is meaningful

You said you:

- reverted the production change locally
- ran the specs
- tried rails-edge
- still saw no warnings

That is exactly what I would expect if the forwarding path prevents Rails from treating the argument as the deprecated positional hash form.

So my current interpretation is:

- the PR is based on a real Rails deprecation mechanism
- the touched call site is a real leftover from the earlier Rails 8.1 compatibility work
- but the specific ActiveAdmin code path probably does **not** hit the deprecation in practice

spec/unit/routing_spec.rb
```diff
@@ -222,5 +222,52 @@
...
+            page_action :archive, method: :post
+            page_action :stats, method: :get
+          end
+        end
+
+        hash_arg_warnings = deprecations.grep(/received a hash argument/)
+        expect(hash_arg_warnings).to be_empty,
+          "expected no 'received a hash argument' deprecation, got: #{hash_arg_warnings.inspect}"
+      ensure
+        load_resources {}
```
tagliala, line 269: Hello,

the problem here is that the original implementation does not emit a warning as well.

/cc @javierjulio I'm going to close this PR, I'll wait for some human interaction before reopening

tagliala (changes requested):

tagliala:
Closing because of missing developer feedback

Kaishley:
Thanks for the review, @tagliala. After more digging, I realised this isn't observable on master, and I want to share the trace in case it's useful for future routing-DSL deprecation work.

### Why the deprecation doesn't fire on master

In v3.4.0, [`build_route`](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/router.rb#L94) changed from `(verbs, *args)` to `(verbs, ...)`. The Ruby 3+ argument forwarding causes the trailing hash literal at the call site to be absorbed into `**kwargs` before reaching Rails' deprecation gate:

```ruby
# actionpack/lib/action_dispatch/routing/mapper.rb
def post(*path_or_actions, ..., **mapping, &block)
  if path_or_actions.grep(Hash).any? && (deprecated_options = path_or_actions.extract_options!)
    # deprecation fires here — but only when the hash is positional
```

With `(verbs, *args)` (v3.2.x), `build_route(:post, "/path" => "x#y")` keeps the hash in `args`, so `router.send(:post, *args)` passes it positionally → `path_or_actions.grep(Hash).any?` is true → deprecation fires.

With `(verbs, ...)` (v3.4.0+), the same call sends the hash through Ruby's kwargs slot → `path_or_actions` is empty → `grep(Hash).any?` is false → no deprecation.

Confirmed by running both signatures side-by-side against Rails 8.1.3:

| `build_route` signature   | Warning fires? |
| ------------------------- | -------------- |
| `(verbs, *args)` (v3.2.x) | Yes            |
| `(verbs, ...)` (v3.4.0+)  | No             |

Tested across v3.2.3, v3.5.1, and master — the `build_route` call site itself is unchanged across all of them, so what determines whether the warning fires is purely the method signature change in v3.4.0.

### Where the warnings I was chasing actually came from

An app downstream was pinned to AA 3.2.3 and emitted 9 of these warnings per boot — all from `page_action :name, method: :post` declarations going through the legacy `*args` path. Bumping to 3.5.1 eliminated all 9 without needing this PR.

### What about Rails 8.2?

When 8.2 removes hash-form routing, master's kwargs-absorbed path _might_ still work (kwargs aren't deprecated, only positional hashes). Or it might break. Hard to predict from the Rails side until 8.2 lands.

This PR isn't fixing a current bug on master, and future-proofing for 8.2 is speculative. Happy to close and reopen if 8.2's hash-form removal ends up affecting the kwargs-absorbed path.

tagliala:
@Kaishley @javierjulio I've opened #9019 as a cleanup change, with a better commit message

Commit authorship has been preserved

PR #9019, opened 2026-05-11
Use keyword `to:` in `page_routes`

tagliala:
Convert the remaining hash-form route call in
`ActiveAdmin::Router#page_routes` to the keyword `to:` form for uniformity with other route invocations and as a cleanup left over from #8837.

This change is purely stylistic, it does not remove any observed Rails 8.1 deprecation warnings, does not seem to alter Rails 8.2 (edge) compatibility, and has no runtime behavior changes.


javierjulio (approved):
Thanks!

PR #9025, opened 2026-05-13
Add missing Vietnamese translations

### LINKED ISSUE #9024: Can not display the icon toggle dark/light mode and profile icon when set the difference locale
### Expected behavior

Display the same UI with the locale `:en`

Image

### Actual behavior
Can not display the correct icon

What actually happens?
It displays the text instead of an icon. The UI is broken also.

Image

### How to reproduce

change the default locale to `:vi`

[our bug report template script]: https://github.com/activeadmin/activeadmin/blob/master/tasks/bug_report_template.rb

leedrum:
- Fix #9024 
- When missing the translation, the UI will show the text of `aria-label` instead of the icon, and then the UI is broken.


tagliala:
Thanks!

PR #9030, opened 2026-05-23
Fix nested belongs_to scope when resource uses an `as:` alias

Fivell:
When a resource is registered with an `as:` alias different from its model class name, the nested-route action raises NoMethodError. For example:

    class User < ApplicationRecord
      has_many :posts
    end

    ActiveAdmin.register User
    ActiveAdmin.register Post, as: "Story" do
      belongs_to :user, optional: true
    end

    # GET /admin/users/1/stories
    # => NoMethodError: undefined method `stories' for an instance of User

Inherited Resources defaults `method_for_association_chain` to `resource_collection_name`, which Active Admin derives from `controller_name` (and `controller_name` comes from the `as:` alias). That value (`:stories`) does not exist on the parent -- the parent declares `has_many :posts`, named after the model.

Override `method_for_association_chain` so that, when the resource belongs_to a parent, the association name is derived from the resource's model class (`Post.model_name.element.pluralize` => `:posts`). For non-conventional association names (e.g. a parent declaring `has_many :authored_posts, class_name: "Post"`), override the method explicitly in the controller block.

Tests:

* `spec/unit/belongs_to_spec.rb` -- new unit example asserting the derived name comes from the model class, not the `as:` alias.
* `features/belongs_to.feature` -- new scenario hitting `/admin/users/:id/stories`. Uses `as: "Story"` (not `"Article"`) because the test fixture User declares `has_many :articles, class_name: "Post"`, which would mask the bug by coincidence.

### DIFF
--- a/app/controllers/active_admin/resource_controller/data_access.rb
+++ b/app/controllers/active_admin/resource_controller/data_access.rb
@@ -75,6 +75,15 @@ module ActiveAdmin
         end_of_association_chain
       end
 
+      # Derive the parent's `has_many` association name from the
+      # resource's model class so that resources registered with an
+      # `as:` alias still scope through the correct collection.
+      def method_for_association_chain
+        return super unless active_admin_config&.belongs_to?
+
+        active_admin_config.resource_class.model_name.element.pluralize.to_sym
+      end
+
       # Retrieve, memoize and authorize a resource based on params[:id]. The
       # actual work of finding the resource is done in #find_resource.
       #
--- a/features/belongs_to.feature
+++ b/features/belongs_to.feature
@@ -160,3 +160,16 @@ Feature: Belongs To
     When I go to the last author's posts
     And I follow "View"
     Then the "Posts" menu item should be selected
+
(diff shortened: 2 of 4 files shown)


javierjulio (approved):
@Fivell thank you. I don't use the belongs_to functionality in ActiveAdmin so I'm not familiar with it. Overall, this seems fine to me since the override is limited to that functionality. @tagliala do you use this feature and/or have any opinions?

tagliala (approved):
Sorry, I do not use this feature as well, but the change looks good to me!

javierjulio:
@tagliala I was going to merge this after rebasing if all green but saw that you requested a re-review. Let me know if you have anything for this and I'll wait, otherwise I'll merge later. Thanks!

tagliala (approved):
@javierjulio I have some second thoughts (for scoped models, like `Admin::Post`), but this should not break any existing case. Let's approve and see if this break something unexpected

I guess that as far as `Admin::Post` does not have a belongs_to it will be fine

PR #9031, opened 2026-05-23
Redirect to comments index when deleting from own show page

### LINKED ISSUE #7770: wrong redirect url after delete comment from show page of Comment
### Description

When I delete comment ActiveAdmin delete them and redirect to deleted comment that is already deleted.

### Expected behavior
AA delete comment from DB and perform redirect to `/admin/comments`

### Actual behavior
AA delete comment from DB and perform redirect to `/admin/comments/{id}`

### How to reproduce

1. visit /admin/comments/{id}
2. click `Delete comment` link
3. AA redirect to deleted comment and return 404 responce 

Your best chance of getting this bug looked at quickly is to provide a **code
sample** or an **executable test case** demonstrating the expected behavior that is not
occurring. Ideally, use [our bug report template] to create the issue. Simply
copy the content of the appropriate template into a .rb file, make the necessary
changes to demonstrate the issue, and **paste the content into the issue
description**.

[our bug report template]: https://github.com/activeadmin/activeadmin/blob/master/tasks/bug_report_template.rb

Fivell:
Closes #7770.

Previously, deleting a comment from `/admin/comments/:id` redirected back to that same URL (via `redirect_back`), which then 404'd because the comment had just been destroyed.

The existing `redirect_back` is the right behavior when the comment is deleted from the commentable resource's page (e.g. `/admin/posts/5`) -- the user wants to return there. Only the self-referencing case is broken.

Check the referer against the deleted comment's own path and redirect to the comments index when they match; otherwise keep `redirect_back`.

A new cucumber scenario exercises the broken path (delete from the comment show page) and asserts the user lands on the comments index.

### DIFF
--- a/lib/active_admin/orm/active_record/comments.rb
+++ b/lib/active_admin/orm/active_record/comments.rb
@@ -75,12 +75,11 @@ ActiveAdmin.after_load do |app|
         def destroy
           destroy! do |success, failure|
             success.html do
-              # If deleting from the Comments resource page then this will fail, as redirecting back
-              # will be to the comment show page, but comment was deleted. The following can be used
-              # to alleviate that, but then deleting comments on commentable resource pages will
-              # redirect to the comments index which may be undesirable.
-              # redirect_to({ action: :index }, fallback_location: active_admin_root)
-              redirect_back fallback_location: active_admin_root
+              if params[:return_to] == "back"
+                redirect_back fallback_location: active_admin_root
+              else
+                redirect_to action: :index
+              end
             end
             failure.html do
               redirect_back fallback_location: active_admin_root
--- a/app/views/active_admin/shared/_resource_comments.html.erb
+++ b/app/views/active_admin/shared/_resource_comments.html.erb
@@ -33,7 +33,7 @@
           <%= simple_format(comment.body) %>
         </div>
         <% if authorized?(ActiveAdmin::Auth::DESTROY, comment) %>
-          <%= link_to I18n.t("active_admin.comments.delete"), url_for_comments(comment.id), method: :delete, data: { confirm: I18n.t("active_admin.comments.delete_confirmation") } %>
+          <%= link_to I18n.t("active_admin.comments.delete"), url_for_comments(comment.id, return_to: "back"), method: :delete, data: { confirm: I18n.t("active_admin.comments.delete_confirmation") } %>
         <% end %>
       </div>
     <% end %>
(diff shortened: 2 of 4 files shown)


lib/active_admin/orm/active_record/comments.rb
```diff
@@ -75,12 +75,15 @@ def create
...
-              # will be to the comment show page, but comment was deleted. The following can be used
-              # to alleviate that, but then deleting comments on commentable resource pages will
-              # redirect to the comments index which may be undesirable.
-              # redirect_to({ action: :index }, fallback_location: active_admin_root)
-              redirect_back fallback_location: active_admin_root
+              # When deleting from the comment's own show page, redirect
+              # to the comments index instead -- otherwise redirect_back
+              # would return to a now-404 page. Deleting from a
+              # commentable resource still redirects back to that resource.
+              if request.referer && URI(request.referer).path == admin_comment_path(id: params[:id])
```
tagliala: I would personally avoid dealing with the Referer header, even if that means living with the bug for now.

javierjulio: @tagliala yes, I agree. We should avoid relying on that header. @Fivell I've been aware of this bug but didn't seem clear how to reliably resolve it. Since the issue is from the admin/comments page which is implemented internally, could we set a parameter in the request to then change the redirect behavior here only for that page?

Fivell: @javierjulio  @tagliala  thanks for your comments, I'll try alternative implementation this friday and let you know

tagliala (approved):
Thanks, this looks a lot simpler than before and I think this is an excellent solution to the problem

Fivell:
changed the approach so it does not use the Referer header anymore, like @tagliala and @javierjulio asked.
  
  How it works now: #destroy redirects to the comments index by default. The only place that needs to go back — the delete link
  on a commentable resource's page (the _resource_comments partial)-  sends an explicit return_to=back param.

  - Delete from the comment's own show/index page - comments index (no more redirect to a 404 page).
  - Delete from a commentable resource's page -  back to that resource (return_to=back).

Frst idea was to do it in the comments resource by changing the destroy, but action_item has no
  option to change the link's URL or add params - the whole link_to is written inside its block. So to add the param I would need to remove the default :destroy action item and write it again from scratch. The _resource_comments partial
  already belongs to Active Admin, so adding one query param there is much simpler and keeps the default action item as it is.

javierjulio (approved):
Thanks! I agree with @tagliala that this is simpler and looks great so thank you @Fivell for tackling the issue.

PR #9032, opened 2026-05-23
Add `string_input_filters` resource & namespace setting

Fivell:
Allow setting the default filters list for string filters, so you don't have to repeat the per-filter filters option
  everywhere. It can be set at the namespace or resource level.

  Precedence (highest to lowest):
  
  1. per-filter filters: option
  2. resource-level string_input_filters
  3. namespace-level string_input_filters
  4. the class default ActiveAdmin::Inputs::Filters::StringInput.filters

```
  # config/initializers/active_admin.rb
  ActiveAdmin.setup do |config|
  
    # Namespace level:
      admin.string_input_filters = [:eq, :cont]
    end
  end
  ```
  
  ```
  # app/admin/post.rb
  ActiveAdmin.register Post do

    # Resource level:
    string_input_filters [:eq, :cont]
  end
  ```

### DIFF
--- a/lib/active_admin/inputs/filters/base/search_method_select.rb
+++ b/lib/active_admin/inputs/filters/base/search_method_select.rb
@@ -51,7 +51,7 @@ module ActiveAdmin
           end
 
           def filters
-            options[:filters] || self.class.filters
+            options[:filters] || resource_default_filters || namespace_default_filters || self.class.filters
           end
 
           def current_filter
@@ -65,6 +65,18 @@ module ActiveAdmin
             filters.collect do |filter|
               [Ransack::Translate.predicate(filter).capitalize, "#{method}_#{filter}"]
             end
+          end
+
+          private
+
+          def resource_default_filters
+            template.respond_to?(:active_admin_config) &&
+              template.active_admin_config&.string_input_filters
+          end
+
+          def namespace_default_filters
+            template.respond_to?(:active_admin_config) &&
+              template.active_admin_config&.namespace.string_input_filters
           end
 
         end
--- a/docs/3-index-pages.md
+++ b/docs/3-index-pages.md
@@ -111,6 +111,29 @@ To override options for string or numeric filter pass `filters` option.
   filter :title, filters: [:start, :end]
 ```
 
+To set the same options for every string filter at once, instead of passing the
+`filters` option on each one, use `string_input_filters` at the namespace or
+resource level. A resource-level setting overrides the namespace one, and the
+per-filter `filters` option still takes precedence over both.
+
+```ruby
+# config/initializers/active_admin.rb
+ActiveAdmin.setup do |config|
+
+  # Namespace level:
+  config.namespace :admin do |admin|
+    admin.string_input_filters = [:eq, :cont]
+  end
+end
+
+# app/admin/post.rb
+ActiveAdmin.register Post do
+
+  # Resource level:
+  string_input_filters [:eq, :cont]
+end
+```
+
 Also, if you don't need the select with the options 'cont', 'eq', 'start' or
 'end' just add the option to the filter name with an underscore.
 
--- a/docs/3-index-pages.md
+++ b/docs/3-index-pages.md
@@ -111,16 +111,14 @@ To override options for string or numeric filter pass `filters` option.
   filter :title, filters: [:start, :end]
 ```
 
-To set the same options for every string filter at once, instead of passing the
-`filters` option on each one, use `string_input_filters` at the namespace or
+To set the same options for every string filter, instead of passing the `filters`
+option on each, use the `string_input_filters` config at the namespace or
 resource level. A resource-level setting overrides the namespace one, and the
 per-filter `filters` option still takes precedence over both.
 
 ```ruby
 # config/initializers/active_admin.rb
 ActiveAdmin.setup do |config|
-
-  # Namespace level:
   config.namespace :admin do |admin|
     admin.string_input_filters = [:eq, :cont]
   end
@@ -128,8 +126,6 @@ end
 
 # app/admin/post.rb
 ActiveAdmin.register Post do
-
-  # Resource level:
   string_input_filters [:eq, :cont]
 end
 ```
--- a/lib/active_admin/namespace_settings.rb
+++ b/lib/active_admin/namespace_settings.rb
@@ -29,6 +29,11 @@ module ActiveAdmin
     # Whether filters are enabled
     register :filters, true
 
+    # Default `filters` list for string filters
+    # (e.g. `[:cont, :eq, :start, :end]`). When nil, the global
+    # `ActiveAdmin::Inputs::Filters::StringInput.filters` list is used.
+    register :string_input_filters, nil
+
     # The namespace root
     register :root_to, "dashboard#index"
 
--- a/lib/active_admin/namespace_settings.rb
+++ b/lib/active_admin/namespace_settings.rb
@@ -29,9 +29,8 @@ module ActiveAdmin
     # Whether filters are enabled
     register :filters, true
 
-    # Default `filters` list for string filters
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Review by javierjulio — COMMENTED** [review]
I think this is worthwhile to do at some level because the default contains search can be quite slow on large tables. I've run into this problem and can cause a long running database query with cascading effects sometimes. I'm not sure though I want this either or only as a namespace config. Would it be possible to support at a resource level?

**Comment by Fivell** [comment]
@javierjulio will add on resource level as well 👍

**Comment by Fivell** [comment]
@javierjulio @tagliala  resource level setting also supported now

**Review by javierjulio — APPROVED** [review]
Thanks!

**Comment by javierjulio** [comment]
@tagliala any concerns with this change? If all good, then I'll merge soon and include it in a new release.

File: lib/active_admin/inputs/filters/base/search_method_select.rb [hunk 1]
```diff
@@ -51,7 +51,7 @@ def select_html
           end
 
           def filters
-            options[:filters] || self.class.filters
+            options[:filters] || resource_default_filters || namespace_default_filters || self.class.filters
```
tagliala, line 54: **[P2] Keep string defaults out of numeric inputs**

`SearchMethodSelect` is also included by `NumericInput`, so this fallback replaces numeric operators whenever the new string setting is configured. With the documented `string_input_filters = [:eq, :cont]`, an existing `filter :id` now renders `Equals`/`Contains` instead of `Equals`/`Greater than`/`Less than`. This happens with both namespace-level and resource-level settings, removing numeric comparison choices from an unrelated input type. Please scope the new defaults to string inputs and preserve numeric defaults. The two numeric examples in the review body reproduce this through the rendered filter form.

lib/active_admin/inputs/filters/base/search_method_select.rb
```diff
@@ -67,6 +67,18 @@ def filter_options
             end
           end
 
+          private
+
+          def resource_default_filters
+            template.respond_to?(:active_admin_config) &&
+              template.active_admin_config&.string_input_filters
```
tagliala, line 74: **[P2] Handle page configs before reading resource-only settings**

`active_admin_config` is an `ActiveAdmin::Page` on pages created with `register_page`, and `Page` does not implement the new resource-only `string_input_filters` accessor. Consequently, an existing custom-page search form using `ActiveAdmin::Filters::FormBuilder` raises `NoMethodError` when rendering a normal string filter, even if neither new setting is enabled. Checking that the template responds to `active_admin_config` and using `&.` does not guard against this non-nil, non-resource config. Please check the config's capability before reading the resource override so these forms can continue to fall back. The custom-page example in the review body reproduces the failure with a real `ActiveAdmin::Page` instance.

tagliala (changes requested):
**Reviewer: ChatGPT Astra 6 (xhigh).**

Requesting changes for the numeric-filter and custom-page regressions described inline. I also reproduced the nil-config failure already [reported by Copilot](https://github.com/activeadmin/activeadmin/pull/9032#discussion_r3943713118); its regression spec is included below without adding a duplicate inline comment.

**Verification**
Reviewed commit `3cae619dd2d525b59fea873daf3424e70ab814a6` in an isolated worktree. Before running specs, ran `rm -rf tmp/test_apps` and regenerated the application using `bin/rake setup`. Environment: Ruby 4.0.6, Rails 8.1.3.1, Ransack 4.4.1.

- Existing `spec/helpers/filter_form_helper_spec.rb` and `spec/unit/resource_spec.rb`: **164 examples, 0 failures**.
- The same files with the four examples below: **168 examples, 4 failures**. All four failures are the new regression examples.
- RuboCop on the affected helper spec passed before and after adding the examples.
- No production code changes were made. These specs assert the intended behavior and deliberately fail on the current PR.

**Reproduction Specs**
The following block belongs inside the existing top-level `RSpec.describe ActiveAdmin::FormHelper, type: :helper` in `spec/helpers/filter_form_helper_spec.rb`. It uses that file's existing `filter` helper and real `ActiveAdmin::Namespace`, `ActiveAdmin::Resource`, and `ActiveAdmin::Page` configuration objects.

```ruby
describe "PR 9032 regressions" do
  let(:namespace) { ActiveAdmin::Namespace.new(ActiveAdmin::Application.new, :admin) }
  let(:resource_config) { ActiveAdmin::Resource.new(namespace, Post) }

  it "preserves numeric operators with namespace string defaults" do
    namespace.string_input_filters = [:eq, :cont]
    allow(helper).to receive(:active_admin_config).and_return(resource_config)

    expect(filter(:id)).to have_select(options: ["Equals", "Greater than", "Less than"])
  end

  it "preserves numeric operators with resource string defaults" do
    resource_config.string_input_filters = [:eq, :cont]
    allow(helper).to receive(:active_admin_config).and_return(resource_config)

    expect(filter(:id)).to have_select(options: ["Equals", "Greater than", "Less than"])
  end

  it "renders string filters with a custom-page config" do
    page_config = ActiveAdmin::Page.new(namespace, "Report", {})
    allow(helper).to receive(:active_admin_config).and_return(page_config)

    expect(filter(:title)).to have_field("q[title_cont]")
  end

  it "renders string filters with a nil config" do
    allow(helper).to receive(:active_admin_config).and_return(nil)

    expect(filter(:title)).to have_field("q[title_cont]")
  end
end
```

From the PR checkout, regenerate the test app and run the two affected spec files:

```sh
rm -rf tmp/test_apps
bin/rake setup
bin/rspec spec/helpers/filter_form_helper_spec.rb spec/unit/resource_spec.rb --format documentation
```

**Observed Failures**
- Both numeric examples: expected `Equals`, `Greater than`, `Less than`; received `Equals`, `Contains`.
- Custom-page example: `NoMethodError: undefined method 'string_input_filters' for an instance of ActiveAdmin::Page`, originating at `search_method_select.rb:74`.
- Nil-config example: `NoMethodError: undefined method 'string_input_filters' for nil`, originating at `search_method_select.rb:79`.

tagliala:
Hello, I've used some Astra credits. I'm not familiar with this part of the code, the suggestions make sense to me but I'm unable to review if they can represent an issue in a real world scenario

Fivell:
@tagliala hello, thanks, all fixed in new follow-up commit

tagliala (approved):
**Reviewer: ChatGPT Astra 6 (xhigh).**

Re-downloaded the PR description, complete diff, commits, discussion/review threads, and CI results, then re-reviewed `5214b30023bd82165203284617637c229c6eab64` in a fresh worktree.

The previous functional findings are resolved:
- The new fallback now lives in `StringInput`, so numeric filters retain their comparison operators with either namespace-level or resource-level string defaults.
- The resource lookup checks whether the config supports `string_input_filters`, allowing custom-page filter forms to render.
- The namespace fallback now safely handles nil config/namespace values.

**Verification**
Before running specs, ran `rm -rf tmp/test_apps` and regenerated the application with `bin/rake setup`. Re-ran `spec/helpers/filter_form_helper_spec.rb` and `spec/unit/resource_spec.rb`, including all four original reproduction examples from my previous review: **171 examples, 0 failures**. All four previously failing examples now pass. RuboCop passes on all six changed Ruby files, and all 23 GitHub checks pass.

**Nonblocking test suggestion:** The numeric regression context at [`spec/helpers/filter_form_helper_spec.rb:379–390`](https://github.com/activeadmin/activeadmin/blob/5214b30023bd82165203284617637c229c6eab64/spec/helpers/filter_form_helper_spec.rb#L379-L390) sets the namespace value but does not supply `helper.active_admin_config` or configure a resource-level value. Wiring it to a real resource config, as in the two numeric examples from my previous review, would ensure it exercises both configuration paths and catches the original shared-mixin regression. I independently verified those stronger examples pass on this revision.

I found no remaining functional blockers. Approving.

javierjulio (approved):
Thanks!

PR #9039, opened 2026-06-01
Support Non-ActiveRecord resources

Fivell:
ActiveAdmin's filter pipeline assumed an ActiveRecord-backed resource and a Ransack::Search-shaped `@search` in spots that raised for resources backed by ActiveResource, API client wrappers, ActiveModel, or POROs:

* `FormtasticAddons#klass`, `#scope?`
* `Resource::Attributes#default_attributes`
* `Views::TableFor::Column#sortable?`
* `Filters::Active#initialize`

Each now short-circuits when the resource/search doesn't respond to the AR/Ransack-specific method, returning `nil` / `[]` / `{}` / `false` instead of raising. The AR path is unchanged; no rescues, no monkey-patches.

Covered by `features/index/non_ar_filters.feature`: a PORO `ApiPost` resource with a `find_collection` override and a small `Data`-based `@search` hydrated from `params[:q]`, demonstrating server-side pagination and filter narrowing.

Known limitation: the "Active Search" chips sidebar is tightly coupled to `Ransack::Condition` (predicate / attributes / klass / reflect_on_all_associations). It no longer raises but renders "No filters applied" for non-Ransack searches. Hide it with `config.current_filters = false` or provide a Ransack-shaped wrapper.

### DIFF
--- a/features/index/non_activerecord_filters.feature
+++ b/features/index/non_activerecord_filters.feature
@@ -27,8 +27,8 @@ Feature: Filters on non-ActiveRecord resources
         end
 
         controller do
-          # Non-AR resources override `find_collection`; Kaminari.paginate_array
-          # adapts the API response for AA's pagination UI.
+          # Non-ActiveRecord resources override `find_collection`; Kaminari.paginate_array
+          # adapts the API response for ActiveAdmin's pagination UI.
           def find_collection(*)
             q = params[:q] || {}
             @search = ApiPost::Search.new(
--- a/lib/active_admin/filters/active.rb
+++ b/lib/active_admin/filters/active.rb
@@ -14,7 +14,8 @@ module ActiveAdmin
       # @see ActiveAdmin::ResourceController::DataAccess#apply_filtering
       def initialize(resource, search)
         @resource = resource
-        @filters = build_filters(search.conditions)
+        # Ransack::Search uses method_missing for #conditions, so gate on the class.
+        @filters = search.is_a?(::Ransack::Search) ? build_filters(search.conditions) : []
         @scopes = search.instance_variable_get(:@scope_args)
       end
 
--- a/features/index/non_ar_filters.feature
+++ b/features/index/non_ar_filters.feature
@@ -1,0 +1,97 @@
+@filters
+Feature: Filters on non-ActiveRecord resources
+
+  ActiveAdmin should let admins use the `filter` DSL on resources that
+  aren't ActiveRecord-backed, without monkey-patching.
+
+  Known limitation: the "Active Search" sidebar is coupled to Ransack,
+  so for a non-Ransack `@search` it renders empty ("No filters applied")
+  instead of filter chips. The filter form, "Clear Filters" link,
+  pagination, and table are unaffected.
+
+  Background:
+    Given I am logged in
+    And a configuration of:
+    """
+      ActiveAdmin.register ApiPost do
+        actions :index
+
+        filter :title_cont, as: :string, label: "Title contains"
+        filter :status_eq,  as: :select, label: "Status",
+               collection: %w[active inactive]
+
+        index do
+          column :id
+          column :title
+          column :status
+        end
+
+        controller do
+          # Non-AR resources override `find_collection`; Kaminari.paginate_array
+          # adapts the API response for AA's pagination UI.
+          def find_collection(*)
+            q = params[:q] || {}
+            @search = ApiPost::Search.new(
+              title_cont: q[:title_cont],
+              status_eq: q[:status_eq]
+            )
+            page = (params[:page] || 1).to_i
+            per_page = 5
+
+            result = ApiPost.fetch(
+              page: page,
+              per_page: per_page,
+              title_cont: @search.title_cont,
+              status_eq: @search.status_eq
+            )
+
+            Kaminari.paginate_array(
+              result.records,
+              total_count: result.total_count,
+              limit:       result.limit,
+              offset:      result.offset
+            ).page(page).per(per_page)
+          end
+        end
+      end
+    """
+
+  Scenario: Server-side pagination across multiple pages
+    When I am on the index page for api_posts
+    Then I should see "Alpha"
+    And I should see "Epsilon"
+    And I should not see "Zeta"
+    And I should see "Showing 1-5 of 7"
+    And I should see pagination page 2 link
+    And I should see the following filters:
+      | Title contains | string |
+      | Status         | select |
+
+    When I follow "2"
+    Then I should see "Zeta"
+    And I should see "Eta"
+    And I should see "Showing 6-7 of 7"
+    And I should not see "Alpha"
+
+  Scenario: Filter narrows the result set to a single page (no pagination)
+    When I am on the index page for api_posts
+    Then I should see pagination page 2 link
+
+    When I select "inactive" from "Status"
+    And I press "Filter"
+    Then I should see "Beta"
+    And I should see "Epsilon"
+    And I should not see "Alpha"
+    And I should not see "Gamma"
+    And I should see "Showing all 2"
+    And I should not see pagination
+
+  Scenario: Filter narrows the result set to a single record
+    When I am on the index page for api_posts
+    And I fill in "Title contains" with "Bet"
+    And I press "Filter"
+    Then I should see "Beta"
+    And I should not see "Alpha"
+    And I should not see "Epsilon"
+    And I should see "Showing 1 of 1"
+    And I should not see pagination
--- a/lib/active_admin/filters/formtastic_addons.rb
+++ b/lib/active_admin/filters/formtastic_addons.rb
@@ -32,9 +32,9 @@ module ActiveAdmin
       # The below are custom methods that Formtastic does not provide.
       #
 
-      # The resource class, unwrapped from Ransack
+      # The resource class, unwrapped from Ransack. Nil for non-Ransack search objects.
       def klass
-        @object.object.klass
+        @object.object.klass if @object.respond_to?(:object) && @object.object.respond_to?(:klass)
       end
 
       def polymorphic_foreign_type?(method)
@@ -66,6 +66,8 @@ module ActiveAdmin
 
       # Ransack supports exposing selected scopes on your model for advanced searches.
       def scope?
+        return false if klass.nil?
+
         context = Ransack::Context.for klass
         context.respond_to?(:ransackable_scope?) && context.ransackable_scope?(method.to_s, klass)
       end
--- a/lib/active_admin/filters/formtastic_addons.rb
+++ b/lib/active_admin/filters/formtastic_addons.rb
@@ -32,7 +32,7 @@ module ActiveAdmin
       # The below are custom methods that Formtastic does not provide.
       #
 
-      # The resource class, unwrapped from Ransack. Nil for non-Ransack search objects.
+      # The resource class, unwrapped from Ransack or `nil` for non-Ransack search objects.
       def klass
         @object.object.klass if @object.respond_to?(:object) && @object.object.respond_to?(:klass)
       end
--- a/lib/active_admin/resource/attributes.rb
+++ b/lib/active_admin/resource/attributes.rb
@@ -5,6 +5,8 @@ module ActiveAdmin
     module Attributes
 
       def default_attributes
+        return {} unless resource_class.respond_to?(:columns)
+
         resource_class.columns.each_with_object({}) do |c, attrs|
           unless reject_col?(c)
             name = c.name.to_sym
--- a/lib/active_admin/views/components/table_for.rb
+++ b/lib/active_admin/views/components/table_for.rb
@@ -160,7 +160,7 @@ module ActiveAdmin
         def sortable?
           if @options.has_key?(:sortable)
             !!@options[:sortable]
-          elsif @resource_class
+          elsif @resource_class.respond_to?(:column_names)
             @resource_class.column_names.include?(sort_column_name)
           else
             @title.present?
--- a/spec/support/rails_template.rb
+++ b/spec/support/rails_template.rb
@@ -49,6 +49,8 @@ template File.expand_path("templates/migrations/create_taggings.tt", __dir__), "
 
 copy_file File.expand_path("templates/models/tagging.rb", __dir__), "app/models/tagging.rb"
 
+copy_file File.expand_path("templates/models/api_post.rb", __dir__), "app/models/api_post.rb"
+
 copy_file File.expand_path("templates/helpers/time_helper.rb", __dir__), "app/helpers/time_helper.rb"
 
 copy_file File.expand_path("templates/models/company.rb", __dir__), "app/models/company.rb"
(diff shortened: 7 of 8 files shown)


lib/active_admin/filters/active.rb
```diff
@@ -14,7 +14,8 @@ class Active
       # @see ActiveAdmin::ResourceController::DataAccess#apply_filtering
       def initialize(resource, search)
         @resource = resource
-        @filters = build_filters(search.conditions)
+        # Ransack::Search uses method_missing for #conditions, so gate on the class.
+        @filters = search.is_a?(::Ransack::Search) ? build_filters(search.conditions) : []
         @scopes = search.instance_variable_get(:@scope_args)
       end
```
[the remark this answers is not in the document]
tagliala: > Ransack::Search may use method_missing

it does

```
> Screening.ransack.respond_to?(:conditions)
=> false
```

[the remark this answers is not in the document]
Fivell: `respond_to?(:conditions)` would not help here: a real `Ransack::Search` returns `false` for it (`#conditions` is `method_missing`-only, as @tagliala confirmed above), and so does a delegator wrapping one, since `respond_to?` is forwarded to the target. It would only let in foreign objects whose `#conditions` are not `Ransack::Condition`-shaped, which then crash deeper in `ActiveFilter`. Keeping the class check; wrappers that need the chips can subclass `Ransack::Search`, or hide the sidebar via `config.current_filters = false`.

javierjulio (approved):
@Fivell thank you. This all looks good to go. Would you be open to contributing a sample resource using the model for the dev app? It would be a good example for others and us on how it works. I realize some of it can be taken from the feature spec but if it's not much trouble to get working, I'd welcome that in the dev app.

tagliala:
I am GPT-6 Astra. Following the request from Javier Julio for a sample resource, I provided a minimal implementation in the demo repository: https://github.com/activeadmin/demo.activeadmin.info/pull/1001 (draft).

The implementation is AI-generated and has not been reviewed or verified by a human. It adds a read-only, non-ActiveRecord API Posts example with in-memory data, filtering, and pagination, temporarily pinning ActiveAdmin to this PR at `e372fc404d788824c0974894418313849f577a1c`.

All 28 automated tests pass locally, but this is not human verification. The draft is available for human review.

Ccar1989 (approved):

Fivell:
@javierjulio  @tagliala   The demo app currently has nothing beyond admin_users and the dashboard -  no regular AR-backed resource at all. Adding a narrow, read-only non-AR example as the first real resource feels not native: visitors would see an edge case before they see ActiveAdmin's bread-and-butter CRUD/filters/nested resources/associations/etc experience.

javierjulio:
@Fivell that was a misunderstanding as I wrote "dev app" which means the local app generated in development when working on ActiveAdmin in this repository, not the demo app repository. The example could work in both places but I think it's most helpful for now here in the generated dev app. Would you be able to add that here? I wasn't sure how straightforward it would be. If it's not much trouble, it would help us out. Thank you.

Fivell:
image
@javierjulio

tagliala (commented):
I am GPT-6 Astra, posting this review on behalf of @tagliala. I reviewed the existing comments, the resolved Ransack class-check discussion, and the current head (`1db93d960793db27c16cda3c27330354bc1a5e0b`), including the added dev-app example. Per request, I have not run specs or linters; the reproduction below is based on tracing the code, not an observed test run.

### P2 — Checkbox and non-predicate select filters still dereference a nil resource class

This makes part of Copilot's previously suppressed finding concrete: [`klass`](https://github.com/activeadmin/activeadmin/blob/1db93d960793db27c16cda3c27330354bc1a5e0b/lib/active_admin/filters/formtastic_addons.rb#L35-L38) now returns `nil` for the plain search object, but [`searchable_has_many_through?`](https://github.com/activeadmin/activeadmin/blob/1db93d960793db27c16cda3c27330354bc1a5e0b/lib/active_admin/filters/formtastic_addons.rb#L49-L55) still calls `klass.ransackable_associations` unconditionally.

With an explicit collection and a search object exposing the appropriate `status_eq` / `status_in` readers:

| Declaration | Static call-path result |
| --- | --- |
| `filter :status, as: :check_boxes, collection: %w[active inactive]` | `CheckBoxesInput#searchable_method_name` always reaches the nil dereference. |
| `filter :status, as: :select, collection: %w[active inactive]` | `seems_searchable?` is false, so `SelectInput#searchable_method_name` reaches the same nil dereference. |
| The same select with `multiple: true` | Same failure, before it can generate the `_in` input. |
| `filter :status_eq, as: :select, collection: %w[active inactive]` | Bypasses the helper because it already has a predicate; this is the path covered by the PR. |

Could we guard `searchable_has_many_through?` when the class lacks the required API and cover these paths? Supplying the collection and search readers does not avoid this error.

### Regression spec

Add this context inside the existing `RSpec.describe` in `spec/helpers/filter_form_helper_spec.rb`; it reuses that file's `filter` / `render_filter` helpers. The first three examples are expected to fail on this PR with `NoMethodError` for `ransackable_associations` on `nil`; the fourth is a control for the supported predicate-suffixed select. These examples are **unexecuted**.

```ruby
describe "filters with a non-Ransack search object" do
  let(:scope) do
    search_class = stub_const(
      "NonActiveRecordFilterSearch",
      Data.define(:status_eq, :status_in)
    )
    search_class.new(status_eq: "inactive", status_in: ["inactive"])
  end

  it "renders checkboxes and preserves the selected values" do
    body = filter :status, as: :check_boxes,
                           collection: %w[active inactive]

    expect(body).to have_field("q[status_in][]", type: "checkbox",
                                              with: "inactive", checked: true)
  end

  it "renders a select without an explicit predicate" do
    body = filter :status, as: :select,
                           collection: %w[active inactive]

    expect(body).to have_select("q[status_eq]", selected: "inactive")
  end

  it "renders a multi-select without an explicit predicate" do
    body = filter :status, as: :select, multiple: true,
                           collection: %w[active inactive]

    expect(body).to have_select("q[status_in][]", multiple: true,
                                               selected: ["inactive"])
  end

  it "renders a select with an explicit predicate" do
    body = filter :status_eq, as: :select,
                              collection: %w[active inactive]

    expect(body).to have_select("q[status_eq]", selected: "inactive")
  end
end
```

### Other field types checked

I traced all the other built-in filter input implementations. Explicitly typed `:string`, `:numeric`, `:boolean`, `:date_range`, and `:text` do not reach this unguarded association helper. This is a static finding, not a claim that their rendering has been tested. String/numeric predicate dropdowns require readers for the predicates they inspect; date ranges use the `_gteq` / `_lteq` readers.

There is also an inference boundary: without `as:`, [`default_input_type`](https://github.com/activeadmin/activeadmin/blob/1db93d960793db27c16cda3c27330354bc1a5e0b/lib/active_admin/filters/forms.rb#L20-L25) still calls `klass._ransackers` for names outside the `_eq` / `_cont` / `_start` / `_end` shortcut. That raises before reaching the previously mentioned `polymorphic_foreign_type?` helper. Non-AR filter examples should therefore explicitly specify the input type. The checkbox/select issue above occurs even when `as:` is supplied.

tagliala:
I am GPT-6 Astra, confirming the analysis in my review above. I added the regression specs to `spec/helpers/filter_form_helper_spec.rb` and ran them on this branch (head `1db93d960793db27c16cda3c27330354bc1a5e0b`), after force-regenerating the test app with `bin/rake setup:force` so they run against a clean app.

Command:

```bash
bin/rspec spec/helpers/filter_form_helper_spec.rb:583
```

Result: **4 examples, 3 failures** — every one a `NoMethodError: undefined method 'ransackable_associations' for nil` raised at `lib/active_admin/filters/formtastic_addons.rb:50` inside `searchable_has_many_through?`:

- `filter :status, as: :check_boxes, collection: %w[active inactive]` — fails
- `filter :status, as: :select, collection: %w[active inactive]` — fails
- `filter :status, as: :select, multiple: true, collection: %w[active inactive]` — fails
- `filter :status_eq, as: :select, collection: %w[active inactive]` — passes (the predicate-suffixed path handled by this PR)

This confirms these three non-AR filter declarations crash on the PR's head, while only the `_eq`-suffixed select that the PR example happens to use works. Could `searchable_has_many_through?` be guarded on a missing class API, as suggested in my review?

tagliala (changes requested):
Hello, apparently there is another potential issue. You may want to address it if it applies

Fivell:
@tagliala done

tagliala (commented):
## Review reassessment (static, head `46b6168`)

Re-assessed the current diff, prior review threads, and the resolved findings. No code changes or CI runs here.

### Status of previous findings

- **Checkbox / bare-select / multi-select nil-class crash** — addressed. The new nil guard in `searchable_has_many_through?` short-circuits the shared dereference in all three paths, independent of the example's catch-all `method_missing`.
- **`polymorphic_foreign_type?` nil-class crash** — addressed by its guard.
- **`respond_to?(:conditions)` vs. class check in `Active#initialize`** — agree with keeping it resolved; the thread's Ransack introspection reasoning still holds.
- **Dev-app example** — added.
- **Active-filter chips for non-Ransack searches** — accepted and disclosed as a limitation.

The prior changes-requested concern no longer blocks.

### Remaining recommendation before approval

**P3 — Dev-app table offers sorting its controller ignores**

In `spec/support/templates_with_data/admin/api_posts.rb`, all three columns render sortable headers (the updated `Column#sortable?` falls back to the column title when there's no `column_names`), but the example's `find_collection` never consumes `params[:order]` and `ApiPost.fetch` keeps fixtures in their declared order. Clicking a header changes URL and sort indicator without reordering rows. Suggest `sortable: false` on those columns.

**Coverage note (non-blocking)**

The current feature only asserts that a bare select renders. It doesn't carry over the checkbox, multi-select, or selected-value assertions from the earlier failing cases. Adding those would lock in the fix.

spec/support/templates_with_data/admin/api_posts.rb
```diff
@@ -0,0 +1,47 @@
...
+  actions :index
+
+  # Active-filter chips require a Ransack search, unlike the filter form.
+  config.current_filters = false
+
+  filter :title_cont, as: :string, label: "Title contains"
+  filter :status_eq, as: :select, label: "Status", collection: %w[active inactive]
+
+  index title: "PORO Posts", download_links: false do
+    column :id
```
tagliala: P3: all three columns render sortable headers here (updated `Column#sortable?` falls back to the column title when there's no `column_names`), but `find_collection` never consumes `params[:order]` and `ApiPost.fetch` preserves the fixture order. Clicking a header changes the URL and sort indicator without reordering rows. Suggest adding `sortable: false` to these columns.

features/index/non_activerecord_filters.feature
```diff
@@ -0,0 +1,100 @@
...
+    Given I am logged in
+    And a configuration of:
+    """
+      ActiveAdmin.register ApiPost do
+        actions :index
+
+        filter :title_cont, as: :string, label: "Title contains"
+        filter :status_eq,  as: :select, label: "Status",
+               collection: %w[active inactive]
+        filter :title, as: :select, label: "Title (no predicate)",
```
tagliala, line 22: Coverage note (non-blocking): this scenario only proves a bare select renders. To lock in the `searchable_has_many_through?` nil-class fix, add these regression specs to `spec/helpers/filter_form_helper_spec.rb` (they use the file's existing `filter`/`scope` helpers and pass on the current head):

```ruby
describe "filters with a non-Ransack search object" do
  let(:scope) do
    search_class = stub_const(
      "NonActiveRecordFilterSearch",
      Data.define(:status_eq, :status_in)
    )
    search_class.new(status_eq: "inactive", status_in: ["inactive"])
  end

  it "renders checkboxes and preserves the selected values" do
    body = filter :status, as: :check_boxes,
                           collection: %w[active inactive]

    expect(body).to have_checked_field("q[status_in][]", with: "inactive")
  end

  it "renders a select without an explicit predicate" do
    body = filter :status, as: :select,
                           collection: %w[active inactive]

    expect(body).to have_select("q[status_eq]", selected: "inactive")
  end

  it "renders a multi-select without an explicit predicate" do
    body = filter :status, as: :select, multiple: true,
                           collection: %w[active inactive]

    expect(body).to have_select("q[status_in][]", multiple: true,
                                                 selected: ["inactive"])
  end

  it "renders a select with an explicit predicate" do
    body = filter :status_eq, as: :select,
                              collection: %w[active inactive]

    expect(body).to have_select("q[status_eq]", selected: "inactive")
  end
end
```

Note: with the fix in place, the `Data`-based search does not need the catch-all `method_missing` used in the feature's `ApiPost::Search`; these specs cover the supporting paths directly.

tagliala (commented):
Anchoring the two suggestions from the reassessment to the specific lines.

tagliala:
@Fivell this is the real me. It is frustrating that each time you run a review they find something else, but I guess that `sortable: false` and coverage on other fields that can be affected by this change are legit to prevent regressions in the future

features/index/non_activerecord_filters.feature
tagliala, line 22: Regression specs to paste into `spec/helpers/filter_form_helper_spec.rb` (see the full code in this review's body).

tagliala (commented):
Concrete regression specs to add to `spec/helpers/filter_form_helper_spec.rb`, inside the existing `RSpec.describe ActiveAdmin::FormHelper` block (uses the file's `filter`/`scope` helpers):

```ruby
describe "filters with a non-Ransack search object" do
  let(:scope) do
    search_class = stub_const(
      "NonActiveRecordFilterSearch",
      Data.define(:status_eq, :status_in)
    )
    search_class.new(status_eq: "inactive", status_in: ["inactive"])
  end

  it "renders checkboxes and preserves the selected values" do
    body = filter :status, as: :check_boxes,
                           collection: %w[active inactive]

    expect(body).to have_checked_field("q[status_in][]", with: "inactive")
  end

  it "renders a select without an explicit predicate" do
    body = filter :status, as: :select,
                           collection: %w[active inactive]

    expect(body).to have_select("q[status_eq]", selected: "inactive")
  end

  it "renders a multi-select without an explicit predicate" do
    body = filter :status, as: :select, multiple: true,
                           collection: %w[active inactive]

    expect(body).to have_select("q[status_in][]", multiple: true,
                                                 selected: ["inactive"])
  end

  it "renders a select with an explicit predicate" do
    body = filter :status_eq, as: :select,
                              collection: %w[active inactive]

    expect(body).to have_select("q[status_eq]", selected: "inactive")
  end
end
```

Fivell:
@tagliala  added new commit to address review

tagliala (approved):
Approving after final reassessment at head `7bde2b3`.

All prior findings are addressed:
- Checkbox / bare-select / multi-select nil-class crash — guarded in `searchable_has_many_through?` and covered by the new regression specs in `spec/helpers/filter_form_helper_spec.rb`.
- `polymorphic_foreign_type?` nil-class crash — guarded.
- Dev-app example — added, with `sortable: false` on the columns whose sorting the `find_collection` override doesn't honor.
- Non-Ransack active-filter chips — accepted and disclosed limitation.

The previously resolved Ransack class-check thread remains valid. All checks report success.

javierjulio (approved):
Thanks!

PR #9044, opened 2026-06-09
Mitigate supply-chain attacks with bundle cooldown

tagliala:
Introduces the new time-based 'cooldown' filter shipped in Bundler  4.0.13. Most supply-chain attacks exploit a tiny window immediately  after a gem version is published. This feature prevents Bundler from  resolving to any gem version until it has been public for a minimum  number of days, allowing time for community vetting.

Ref: https://blog.rubygems.org/2026/06/03/cooldown-let-new-gems-be-vetted.html

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-source "https://rubygems.org"
+source "https://rubygems.org", cooldown: 5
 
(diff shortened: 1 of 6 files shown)


javierjulio (approved):
Thanks!

PR #9061, opened 2026-07-06
Fix failing tests in local environment

javierjulio:
Often running the full test suite with `bin/rake` or even just RSpec or Cucumber tests separately, would result in hard failures or intermittently failing tests. I couldn't ever figure out why. I had Codex run the test suite and suggested these changes which now solve the test suite failures in development. Worth noting we didn't encounter this in CI although the flaky cucumber test I believe we have. By running `bin/rake` over 200 times, I've yet to encounter a test failure.

### DIFF
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -8,8 +8,8 @@ Given(/^(?:I am logged|log) out$/) do
 end
 
 Given(/^I am logged in$/) do
-  logout(:user)
-  login_as ensure_user_created "admin@example.com"
+  logout(:admin_user)
+  login_as ensure_user_created("admin@example.com"), scope: :admin_user
 end
 
 Given(/^I am logged in with capybara$/) do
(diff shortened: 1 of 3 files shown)


javierjulio:
@tagliala I'm wondering if the collection decorator related commit is the right change since perhaps there is a reason it shouldn't be auto loaded.

javierjulio:
@tagliala nevermind as that collection decorator related fix is unnecessary with another Codex review so I removed the commit. The reason is the app load in the rails_helper seems to resolve all the test suite failures now.

tagliala:
From commit message:

> CI stayed green because its grouping/order happened to load ActiveAdmin through routes or another spec first.

Is this correct? Isn't it because on CI we have `CI=1` env variable which eager loads the application as per

```
$ cat tmp/test_apps/rails_81/config/environments/test.rb | grep CI
  config.eager_load = ENV["CI"].present?
```

?

Can you please try to replicate if you have failures locally with `CI=1 bundle exec bin/rake` ?

javierjulio:
@tagliala yes, I confirmed it's successful. I don't get any failures by running `CI="1" bin/rake` repeatedly in my dev environment.

PR #9064, opened 2026-07-13
Backport PR #7437 to 3-0-stable - run update_resource inside a transaction

mgrunberg:
Backport of https://github.com/activeadmin/activeadmin/pull/7437 (run update_resource inside a transaction) to 3-0-stable.

A copy of https://github.com/activeadmin/activeadmin/pull/6202, to not let it be forgotten.

Fixes https://github.com/activeadmin/activeadmin/issues/5430

When submitting a form with invalid attributes, some associations are autosaved via assign_attributes before calling save_resource. As a consequence, when updating a resource, the user sees a validation error and is not aware that the changes to the associations got persisted.

This PR runs the code inside the update_resource method in an ActiveRecord::Transaction so that everything is rolled back when the model is not valid.

### DIFF
--- a/features/edit_page.feature
+++ b/features/edit_page.feature
@@ -136,3 +136,18 @@ Feature: Edit Page
     And I follow "Edit"
     Then I should see a fieldset titled "Details"
     And the "Name" field should contain "Bugs"
+
+  Scenario: Save resource within a transaction
+    Given a company named "My company" with a store named "First store" exists
+    And a store named "Second store" exists
+    When I am on the index page for companies
+    And I follow "Edit"
+    Then the "Stores" select should have "First store" selected
+    When I fill in "Name" with ""
+    And I select "Second store" from "Stores"
+    And I press "Update Company"
+    Then I should see "can't be blank"
+    When I follow "Cancel"
+    And I follow "View"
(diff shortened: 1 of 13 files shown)


tagliala:
## Cucumber failures: CSS class mismatch in attribute_steps.rb

> _Note: This analysis was generated by an AI agent (opencode). Please verify the findings before applying any changes._

The backport commit changed the CSS selector in `features/step_definitions/attribute_steps.rb` from `.attributes_table` (underscore) to `.attributes-table` (hyphen), but the Arbre component's `default_class_name` at `lib/active_admin/views/components/attributes_table.rb:49` still returns `"attributes_table"` (underscore).

Capybara treats these as different classes, causing all 20 `I should see the attribute` cucumber steps to fail with `attribute missing`.

**Fix**: Revert the selector change in `attribute_steps.rb` back to `.attributes_table` (underscore), since this branch doesn't include the corresponding component change from commit 0e630e675 on main.

mgrunberg:
@tagliala thanks for the help.
@javierjulio this is ready to go. If it's ok, I plan to release a minor 3.x version.

javierjulio (approved):
Thanks!

PR #9090, opened 2026-08-18
Update docs to reflect removed site_title and footer configs

abd0-omar:
These configs (config.footer, config.site_title_link, and
config.site_title_image) were removed in favor of overriding simple
partials in activeadmin/activeadmin#8156. However, the
docs were not updated at the time.

This removes the outdated ruby configurations and instead provides
instructions for overriding the `_site_footer.html.erb` and
`_site_header.html.erb` partials.

### DIFF
--- a/docs/1-general-configuration.md
+++ b/docs/1-general-configuration.md
@@ -38,11 +38,10 @@ If you want, you can customize it.
 
 ```ruby
 config.site_title       = "My Admin Site"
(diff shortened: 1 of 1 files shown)


javierjulio:
@abd0-omar thank you. The current docs are v3. We are working on updating them to account for both v3 and v4 so the necessary content changes around this will be incorporated in that future update.

PR #9093, opened 2026-08-28
feat: Support `:if` option on `member_action` and `collection_action`

Lokideos:
I proposed this in #9050 a while ago, so I've decided to open a PR since it's easier to evaluate the idea with working code.

## Problem

`member_action` and `collection_action` accept an options hash, but silently ignore the `:if` option:

```ruby
member_action :archive, method: :post, if: -> { resource.archivable? } do
  # always reachable; the proc does nothing
end
```

This is surprising because `:if` is honored almost everywhere else in the DSL (action items, sidebars, etc.) via `OptionalDisplay`, so it is easy to assume it works here too. As a result, the custom action stays reachable for anyone who crafts the request, even when the corresponding `action_item` link is hidden.

## Solution

Both methods now accept `:if` as a per-request guard, mirroring the `OptionalDisplay` semantics:

* A `Proc` is evaluated in the controller context on each request, so `resource`, `params` and other controller helpers are available.
* A `Symbol` (or `String`) names a controller method to call, including private ones.

When the condition returns a falsey value, the request raises `ActionController::RoutingError`, which Rails renders as a 404 in production. A guarded-off action therefore renders as if it does not exist.

Implementation-wise:
- `ResourceDSL#action` extracts the `:if` option and installs a `before_action` scoped to that action only. 
- The condition evaluation lives in a new private controller method, `restrict_action_access!`, in `ResourceController::ActionBuilder`, named after the existing [restrict_download_format_access!](https://github.com/activeadmin/activeadmin/blob/0f853ce27e549de320fcb113a247eb782b2c7e7c/app/controllers/active_admin/resource_controller.rb#L92-L100) guard that follows the same pattern.

```ruby
controller do
  before_action(only: [name]) { @page_title = title } if title
  before_action(only: [name]) { restrict_action_access!(condition) } if condition
  define_method(name, &block || Proc.new {})
end
```

The documentation was also updated and gained a **"Conditionally Enabling Actions"** section, describing new behavior.

### DIFF
--- a/app/controllers/active_admin/resource_controller/action_builder.rb
+++ b/app/controllers/active_admin/resource_controller/action_builder.rb
@@ -5,6 +5,20 @@ module ActiveAdmin
     module ActionBuilder
       extend ActiveSupport::Concern
 
+      private
+
+      def restrict_action_access!(condition)
+        allowed =
+          case condition
+          when Symbol, String
+            send condition
+          when Proc
+            instance_exec(&condition)
+          end
+
+        raise ActionController::RoutingError, "Not Found" unless allowed
+      end
+
       module ClassMethods
 
         def clear_member_actions!
--- a/docs/8-custom-actions.md
+++ b/docs/8-custom-actions.md
@@ -70,6 +70,24 @@ member_action :foo, method: [:get, :post] do
 end
 ```
 
+## Conditionally Enabling Actions
+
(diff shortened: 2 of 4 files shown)


beceqada32ps3 (approved):
Approval to 10 Workflows.

PR #9101, opened 2026-09-15
Fix double-escaped <title> for html_safe display

jablan:
## Summary
- `html_head_site_title` combined `page_title`/`display_name` with the site title using plain Ruby string interpolation (`"#{...}"`), which silently strips the `html_safe`/`SafeBuffer` flag from the already-escaped display name.
- Since the resulting string is no longer marked safe, ERB's automatic output escaping in the `<title>` tag escapes it a second time (e.g. `&` becomes `&amp;amp;` instead of `&amp;`), so the browser tab shows literal escaped entities instead of the decoded display name.
- This isn't an XSS (page `<title>` text doesn't execute), just a display/double-escaping correctness bug, but it produces garbled tab titles for ordinary names containing `&`, `<`, `>`, etc.
- Switches `html_head_site_title` to use `safe_join`, which preserves `html_safe` on already-escaped fragments while still escaping any genuinely unsafe input.

## Test plan
- [x] Added a regression test in `spec/helpers/layout_helper_spec.rb` asserting an `html_safe` page title containing `&` is escaped exactly once in the combined title, and is itself `html_safe`.
- [x] Confirmed the new test fails against the unpatched code (`html_safe?` is `false` after interpolation) and passes with the `safe_join` fix.
- [x] `bundle exec rspec spec/helpers/layout_helper_spec.rb spec/helpers/display_helper_spec.rb` — all passing.
- [x] `bundle exec rubocop` on changed files — no offenses.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tagliala:
**Review note from ChatGPT Astra 6 (xhigh) — nonblocking edge case / possible follow-up.**

The `safe_join` change fixes the ordinary escaped-display-name case, and raw untrusted title strings remain escaped. One corner case is an HTML-safe custom heading containing an inline SVG with its own `<title>` element, for example:

```ruby
content_tag(:svg, content_tag(:title, "Report"))
```

This is suitable markup for a visible heading, but inserting it unchanged inside the document's `<title>` causes the SVG's `</title>` to close the document title early. I reproduced this through ERB rendering and HTML5 parsing.

**Important qualification:** this did not produce a useful browser-tab title before the PR either. Previously the tab displayed the SVG markup literally; it never rendered an SVG there. With `site_title == "MyAdmin"`:

| Version | Browser-tab text | Extra body text |
| --- | --- | --- |
| Before this PR | `<svg><title>Report</title></svg> - MyAdmin` | None |
| This PR | `<svg><title>Report` | `- MyAdmin` |

I haven't identified an actual application affected by this customization, so I would not block the double-escaping fix on it. It is a worsening of an already poorly handled edge case, rather than a previously working SVG browser-title feature.

If we want rich headings to produce plain-text tab titles, a small follow-up would be:

```ruby
def html_head_site_title(separator: "-")
  strip_tags(safe_join([@page_title || page_title, site_title], " #{separator} "))
end
```

The order matters: `safe_join` first escapes ordinary strings while preserving already-escaped fragments; `strip_tags` then removes actual HTML markup from HTML-safe fragments. Literal markup or entity spellings in ordinary strings remain literal text. The SVG example becomes `Report - MyAdmin`, while an already-escaped `Smith &amp; Sons` still displays as `Smith & Sons` without double escaping. This only changes the document-title helper, leaving the visible heading's markup available to its own renderer.

**Local verification of that proposed follow-up:** after removing `tmp/test_apps` and regenerating with `bin/rake setup`, the layout/display helper specs plus five rendered-title checks passed: **77 examples, 0 failures**. The checks cover the SVG case, raw closing-title/script text remaining inert, literal markup, literal entity text, and an HTML-formatted site title. The PR's existing ampersand regression also passes. RuboCop passes on both affected files.

tagliala (approved):
**Reviewer: ChatGPT Astra 6 (xhigh).**

Approving the fix for double-escaped display names at `d8199051be7cb64ba29643daa3be22ed6775121b`. All 20 GitHub checks pass. The HTML-rich-title case below is a **nonblocking edge case / possible follow-up**. The ordinary escaped-display-name case is fixed, and raw untrusted title strings remain escaped.

### Nonblocking finding: SVG markup in an HTML-safe custom title

At `app/helpers/active_admin/layout_helper.rb:20`, `safe_join` preserves markup in HTML-safe fragments. A custom heading can contain an inline SVG with its own `<title>` element:

```ruby
content_tag(:svg, content_tag(:title, "Report"))
```

This is suitable markup for a visible heading, but inserting it unchanged inside the document's `<title>` causes the SVG's `</title>` to close the document title early. This was verified through ERB rendering and HTML5 parsing.

**This was already a problem before this PR:** the old implementation displayed the SVG markup literally in the browser tab; it never rendered a useful SVG browser title. The new behavior additionally truncates the title and spills the suffix into the body. With site title `MyAdmin`:

| Version | Browser-tab text | Extra body text |
| --- | --- | --- |
| Before this PR | `<svg><title>Report</title></svg> - MyAdmin` | None |
| This PR | `<svg><title>Report` | `- MyAdmin` |
| Optional plain-text follow-up | `Report - MyAdmin` | None |

I have not identified an actual application affected by this customization, so I would not block the ordinary double-escaping fix on it. It is a worsening of an already poorly handled edge case, rather than a previously working SVG browser-title feature.

### Failing spec for the desired plain-text behavior

This example belongs inside the existing top-level `RSpec.describe ActiveAdmin::LayoutHelper, type: :helper` in `spec/helpers/layout_helper_spec.rb`:

```ruby
it "renders a safe SVG heading as plain title text", aggregate_failures: true do
  allow(helper).to receive(:site_title).and_return("MyAdmin")
  helper.set_page_title(
    helper.content_tag(:svg, helper.content_tag(:title, "Report"))
  )

  html = helper.render(inline: "<title><%= html_head_site_title %></title>")
  document = Nokogiri::HTML5.parse(html)

  expect(document.at_css("head > title").text).to eq "Report - MyAdmin"
  expect(document.at_css("body").text).to be_empty
end
```

On this PR, the actual title text is `"<svg><title>Report"` and the body text is `"- MyAdmin"`. **The plain-text-title assertion also fails on the old implementation**, because it displayed the complete markup literally; only the body-containment assertion passed before. This spec describes the desired follow-up behavior, not a previously working SVG-title feature.

### Optional fix

If we want rich headings to produce plain-text tab titles, a small follow-up would be:

```ruby
def html_head_site_title(separator: "-")
  strip_tags(safe_join([@page_title || page_title, site_title], " #{separator} "))
end
```

The order matters: `safe_join` first escapes ordinary strings while preserving already-escaped fragments; `strip_tags` then removes actual HTML markup from HTML-safe fragments. Literal markup or entity spellings in ordinary strings remain literal text. The SVG example becomes `Report - MyAdmin`, while an already-escaped `Smith &amp; Sons` still displays as `Smith & Sons` without double escaping. This only changes the document-title helper, leaving the visible heading's markup available to its own renderer.

### Verification

Already completed in an isolated worktree after removing `tmp/test_apps` and regenerating with `bin/rake setup`:

- The unmodified PR's layout/display helper specs passed: **72 examples**.
- With the optional fix and five additional rendered-title checks: **77 examples, 0 failures**.
- The additional checks cover the SVG/plain-text case, raw closing-title/script text remaining inert, literal markup, literal entity text, and an HTML-formatted site title.
- The PR's original ampersand regression also passes with the optional fix.
- RuboCop passed on both affected files.

spec/helpers/layout_helper_spec.rb
```diff
@@ -35,6 +35,19 @@
     end
   end
 
+  describe "html_head_site_title escaping" do
+    it "does not double-escape an already html_safe page title" do
+      allow(helper).to receive(:site_title).and_return("MyAdmin")
+      allow(helper).to receive(:page_title).and_return(ERB::Util.html_escape("Smith & Sons"))
```
javierjulio:
```suggestion
      expect(helper).to receive(:site_title).and_return("MyAdmin")
      expect(helper).to receive(:page_title).and_return(ERB::Util.html_escape("Smith & Sons"))
```

javierjulio (approved):
Thanks!

PR #9103, opened 2026-09-18
Explicitly configure every RuboCop cop and sort entries

tagliala:
## Summary

Makes the RuboCop configuration explicit by registering every installed cop, and sorts all entries alphabetically.

- Removes the `Capybara:` and `Performance:` department-level blocks, which silently auto-enable new cops added in future plugin updates.
- Enables every cop with 0 offenses; disables the remaining 268 offending cops, each with a `# TODO:` note.
- Sorts all entries in stable alphabetical order.

## Motivation

This is preliminary work toward synchronizing with modern RuboCop versions. Having every registered cop explicitly pinned ensures that when `rubocop`, `rubocop-capybara`, `rubocop-performance`, or `rubocop-rspec` is updated, newly added cops default to `Enabled: false` instead of silently enabling and breaking CI.


tagliala:
Closing this in favor of a sort-only PR; the full explicit-config sweep is deferred.

